PR #1, opened 2019-02-14
dataset validator

alexfoias:
Initial version of the bids dataset validator. 

The validator only checks the items present in the contrast_dict. This version doesn't check the contents of `/sub-XX/dwi/`


andreanne-lemay:
Since the code changed a lot since this PR was created, this PR can be closed (the problem is not applicable anymore).

PR #70, opened 2019-10-24
Change == to >= for torch requirement

### LINKED ISSUE #67: Better to use >= rather than == in requirements
"==" are mostly reserved for releases (not relevant yet). Having ">=" prevents from missing critical library updates.

it is the responsibility of the developer to be up to date with the requirements,

implementing unit tests facilitates the discovery of version-related inconsistencies between local installation code.

### LINKED ISSUE #69: Pytorch 1.3.0 cuda not detected
To reproduce :

1. Create venv and install dependencies
2. `python -c 'import torch;print(torch.cuda.is_available())'` gives `False`

~Weird since pytorch doc requires CUDA 9 or above and we have CUDA 10~

Possibly a driver version and not CUDA version issue? `print(torch.cuda.get_device_name(0))
` gives :

> AssertionError: 
The NVIDIA driver on your system is too old (found version 10000).
Please update your GPU driver by downloading and installing a new
version from the URL: http://www.nvidia.com/Download/index.aspx
Alternatively, go to: https://pytorch.org to install
a PyTorch version that has been compiled with your version
of the CUDA driver.

Looks related https://github.com/pytorch/pytorch/issues/27738

### LINKED ISSUE #210: Upgrade torchvision
I thought we should put this question back on the table: Should we upgrade our torchvision?

@andreanne-lemay & @olix86 --> What are your thoughts?

Small new argument I can across today in favour of an upgrade --> This [new functio](https://pytorch.org/docs/stable/nn.functional.html#instance-norm)n `instance_norm` is not available in our current `torchvision` version but is exactly what we need for: [NormalizeInstance2D & 3D](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/transforms.py#L439).

olix86:
Fix #67, fix #69, fix #210


olix86:
Looks like this PR is back on the menu 😄 

I've ran pytest manually on `rosenberg` and everything passes with pytorch 1.5, and torchvision 0.6.0. Also they've fixed a bug which means we don't have to import `future`, and we don't have to pin the `pillow` version anymore. So this closes the same number of issues as the number of modified lines 😛

PR #112, opened 2019-12-04
(WIP) Multi-GPU : use threads instead of processes

olix86:
This PR aims to fix the fact that the current version of dev/automate_training.py is not compatible with a non-zero `num_workers` which became a problem after what @charleygros  discovered in #111 

This  [error](https://stackoverflow.com/questions/51485212/multiprocessing-gives-assertionerror-daemonic-processes-are-not-allowed-to-have) was previously [discussed ](https://github.com/neuropoly/ivado-medical-imaging/issues/75#issuecomment-547513590)

Using shared queue to assign an available GPU to each thread


jcohenadad:
@olix86 can you please add a justification when closing a PR without merging it?

olix86:
It fixes a really old bug with automate training that hasn't occured in > 1 year

PR #122, opened 2019-12-11
Add macOS and python 3.7

olix86:
Windows compatibility will take more work since there is no pypi package for pytorch but our project is already compatible with macOS so this PR adds it to the test cases. Same for python 3.7.


AnBucquet:
Thx @olix86!

PR #123, opened 2019-12-13
Adapt loader and transforms to MedicalTorch modifications + Modifications on HeMIS implementation

AnBucquet:
Related to MedicalTorch [PR-9](https://github.com/neuropoly/medicaltorch/pull/9). This PR aims to adapt the current code to MedicalTorch modifications for multi-inputs and HeMIS U-net training. 

**Done**:
- Adapt loader.py to new MedicalTorch dataset (cf [PR-9](https://github.com/neuropoly/medicaltorch/pull/9))
- Adapt transforms.py, tests as well
- add test for Hemis U-net

**TODO**
- Adapt normalize_metadata to multi-inputs . 
- Add tests for multiple inputs

Both of them will be implemented in the next PR.


config/config_tumorSeg.json
```diff
@@ -71,6 +71,7 @@
         "size": [256, 256]
       },
       "ToTensor": {},
+      "StackTensors": {},
```
AnBucquet: @andreanne-lemay I've added this transformation in [MedicalTorch](https://github.com/neuropoly/medicaltorch/blob/ab/generalization_dataset/medicaltorch/transforms.py#L118) to stack the input tensors so the new loader should not impact your work. Yet I haven't try to train a network, but the [`tests_unet()` ](https://github.com/neuropoly/ivado-medical-imaging/blob/ab/Missing_Modality/testing/test_training_time.py#L35) works. Let me know if you've any troubles!

ivadomed/utils.py
```diff
@@ -154,3 +153,14 @@ def threshold_predictions(predictions, thr=0.5):
...
+    """
+    This function sends input_var to GPU.
+    :param input_var: either a tensor or a list of tensors
+    :return: same as input_var
+    """
+
+    if isinstance(input_var, list):
+        return [t.cuda() for t in input_var]
+    else:
+        return input_var.cuda()
```
AnBucquet: Need to add this function otherwise `.cuda()` returns an error when `input_samples` is a list of tensors.

charleygros:
`test_undoTrans` is failling because the CenterCrop2D is no longer working.
I have added some matplotlib lines to help the debug.
See current result: the undo SC (right image) is no longer centered.
test_undo_err_806

charleygros:
We now have a fail with `RoiCrop2D` --> something possibly breaked it in this PR.

AnBucquet:
> We now have a fail with `RoiCrop2D` --> something possibly broke it in this PR.

Fixed by bd46981.

ivadomed/transforms.py
```diff
@@ -142,23 +143,25 @@ def __call__(self, sample):
...
+            w, h = input_data[i].size
+            th, tw = self.size
+            th_half, tw_half = int(round(th / 2.)), int(round(tw / 2.))
 
-        params = (fh, fw, h, w)
-        self.propagate_params(sample, params, 0)
+            # compute top left corner of the crop area
+            fh = y_roi - th_half
+            fw = x_roi - tw_half
+            params = (fh, fw, w, h)
```
charleygros: Error comes from here! w and h have been swapped compared to `master`

ivadomed/transforms.py
```diff
@@ -152,23 +143,24 @@ def __call__(self, sample):
...
         # compute center of mass of the ROI
         x_roi, y_roi = center_of_mass(np.array(roi_data).astype(np.int))
         x_roi, y_roi = int(round(x_roi)), int(round(y_roi))
 
-        # compute top left corner of the crop area
-        fh = y_roi - th_half
-        fw = x_roi - tw_half
+        for i in range(len(input_data)):
+            w, h = input_data[i].size
+            tw, th = self.size
```
charleygros: Can be out of the loop. (I am in a fancy mood ;) )

ivadomed/transforms.py
```diff
@@ -152,23 +143,24 @@ def __call__(self, sample):
...
+            th_half, tw_half = int(round(th / 2.)), int(round(tw / 2.))
+
+            # compute top left corner of the crop area
+            fh = y_roi - th_half
+            fw = x_roi - tw_half
```
charleygros: Can go out of the loop (I'm in a fancy mood ;) ).

ivadomed/transforms.py
```diff
@@ -152,23 +143,24 @@ def __call__(self, sample):
         input_data = sample['input']
         roi_data = sample['roi']
 
-        w, h = input_data.size
-        tw, th = self.size
-        th_half, tw_half = int(round(th / 2.)), int(round(tw / 2.))
-
         # compute center of mass of the ROI
         x_roi, y_roi = center_of_mass(np.array(roi_data).astype(np.int))
```
charleygros: To keep in mind --> We assume here that the ROI, and all the modalities share the same image size! (which is not the case in our current dataset... maybe the case in challenges datasets?)

ivadomed/transforms.py
AnBucquet: Wow, good catch! Thx @charleygros

ivadomed/transforms.py
AnBucquet: Like [this](https://github.com/neuropoly/ivado-medical-imaging/blob/ab/Missing_Modality/ivadomed/transforms.py#L152-L162)? 😃

PR #126, opened 2019-12-18
Implement evaluation metrics at 3D data scale

### LINKED ISSUE #45: Revisit performance scores
The Dice is highly sensitive to the size of the object (e.g., small lesion will highly impact Dice, and potentially be noisy), and the specificity includes voxels of class=0 (to confirm), which brings it to close to 1, which is not informative.

We might want to introduce more representative metrics, such as Hausdorff distance, and only include voxels=1 in the computation of the specificity. Object-based metrics (see [Gros et al 2018](https://paperpile.com/view/a3911a98-1d16-0bac-a8f4-26844d869998)) could also be relevant.

charleygros:
Evaluation on the testing dataset is currently performed on a slice-wise basis, the average value of the metrics is displayed in the terminal.

Overall, this PR aims at computing the evaluation metrics per 3D data, saving the results for each image, which opens the door to a finer analyses of the results, eg across center? across tumour type? while being more meaningful in practice than slice-wise approach.

### Implementation
1. `class Evaluation3DMetrics` in `ivadomed/utils.py` which compute the different metrics from `fname_pred` and `fname_gt`. Possible to request only one metric at a time (e.g. `self.get_dice()`) or all at the same time via `self.get_all_metrics()` which output a dictionary.
2. `config` --> You can now specify `eval` for the `command` parameter.
3. `cmd_eval(context)` in `ivadomed/main.py`.

### How to run
1. Modify the `config` file with `eval` for the `command` parameter.
2. Run `ivadomed [config.json]`
```
Working on cpu.
Evaluation: 100%|████████████████████████████████████████████████████████████████████████████████████████████| 266/266 [01:02<00:00,  4.27it/s]
```
3. Grab your results in: `log_directory/results_eval/evaluation_3Dmetrics.csv`, which look like something like this (ignore the poor results, obtained with model trained with 2 epochs, charlie impatient):
```
                                                avd      dice  precision      recall         rvd  specificity       vol_gt     vol_pred
image_id                                                                                                                               
sub-nyuShepherd102_acq-sag_T2w                  NaN  0.000000   0.000000         NaN         NaN    99.970678     0.000000   711.187899
sub-milanFilippi063_acq-sag_T2w                 NaN  0.000000   0.000000         NaN         NaN    99.968749     0.000000   646.016861
sub-montpellierLesion012_acq-sup_T2star  162.336585  0.118762  44.067797  100.000000 -162.336585    99.997392    44.049699   115.558475
sub-nihReich013_T2star                    79.286144  1.320280  44.910645   92.038217  -79.286144    99.982746  1273.158445  2282.596680
sub-bwh041_acq-sagstir_T2w                      NaN  0.000000   0.000000         NaN         NaN    99.989306     0.000000   265.386313
```
Note: the `NaN` occurs when division by zero or empty GT or...etc. see `class Evaluation3DMetrics` for further details.

### Metrics:
All metrics are in percentage, except the last ones.
For all metrics in percentage, the best value is `100.` while it is `0.` for `avd`, `rvd` and `lfdr`.
- Dice score, of course
- Relative Volume Difference (`rvd`), Absolute Volume Difference (`avd`)
- Voxel-wise recall, precision, specificity.
- Structure-wise True Postive Rate (`ltpr`) and False Detection Rate (`lfdr`).
- Volume GT (`vol_gt`) and prediction (`vol_pred`), in mm3.
- Number of lesions (`n_gt` and `n_pred`).

Besides the opportunity to feed tables of results, this csv file could be used to generate violin plots (e.g. `seaborn`), or correlation plots `vol_gt` vs. `vol_pred` such as:
Screenshot 2019-12-18 17:21:46

Fixes #45

### DIFF
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -49,6 +49,8 @@ class Evaluation3DMetrics(object):
 
         self.px, self.py, self.pz = self.get_pixdim(self.fname_pred)
 
+        # Note: Some papers suggested to remove all lesion with less than 3 voxels: Todo?
+
         # 18-connected components
         self.data_pred_label, self.n_pred = label(self.data_pred,
                                                     connectivity=3,
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -78,43 +78,15 @@ class Evaluation3DMetrics(object):
         """Absolute volume difference."""
         return abs(self.get_rvd())
 
-    def get_dice(self):
-        return dice_score(self.data_gt, self.data_pred) * 100.
-
-    def get_precision(self):
-        """Positive predictive values, precision."""
-        FP, FN, TP, TN = mt_metrics.numeric_score(self.data_pred, self.data_gt)
-        if (TP + FP) <= 0.0:
-            return np.nan
-
-        precision = np.divide(TP, TP + FP)
-        return precision * 100.0
-
-    def get_recall(self):
-        """Recal, TPR, sensitivity."""
-        FP, FN, TP, TN = mt_metrics.numeric_score(self.data_pred, self.data_gt)
-        if (TP + FN) <= 0.0:
-            return np.nan
-
-        TPR = np.divide(TP, TP + FN)
-        return TPR * 100.0
-
-    def get_specificity(self):
-        """Specificity, TNR."""
-        FP, FN, TP, TN = mt_metrics.numeric_score(self.data_pred, self.data_gt)
-        if (TN + FP) <= 0.0:
-            return 0.0
-        TNR = np.divide(TN, TN + FP)
-        return TNR * 100.0
-
     def get_all_metrics(self):
         dct = {}
         dct['vol_pred'] = self.get_vol(self.data_pred)
         dct['vol_gt'] = self.get_vol(self.data_gt)
         dct['rvd'], dct['avd'] = self.get_rvd(), self.get_avd()
-        dct['dice'] = self.get_dice()
-        dct['recall'], dct['precision'] = self.get_recall(), self.get_precision()
-        dct['specificity'] = self.get_specificity()
+        dct['dice'] = dice_score(self.data_gt, self.data_pred) * 100.
+        dct['recall'] = mt_metrics.recall_score(self.data_pred, self.data_gt, err_value=np.nan)
+        dct['precision'] = mt_metrics.precision_score(self.data_pred, self.data_gt, err_value=np.nan)
+        dct['specificity'] = mt_metrics.specificity_score(self.data_pred, self.data_gt, err_value=np.nan)
 
         return dct
 
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -34,6 +34,48 @@ class IvadoMetricManager(mt_metrics.MetricManager):
         return res_dict
 
 
+class Evaluation3DMetrics(object):
+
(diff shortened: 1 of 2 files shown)


ivadomed/utils.py
```diff
@@ -34,6 +34,91 @@ def get_results(self):
...
+        return rvd
+
+    def get_avd(self):
+        """Absolute volume difference."""
+        return abs(self.get_rvd())
+
+    def get_dice(self):
+        return dice_score(self.data_gt, self.data_pred) * 100.
+
+    def get_precision(self):
```
andreanne-lemay: This is duplicate from medicaltorch metrics.py, the methods get_precision, gt_recall, get_specificity could be removed and the methods from medical torch could be called directly in get_all_metrics

charleygros: The reason I created a method `get_dice` is to allow the user to get only dice score as metric from this class (compared to `get_all_metrics`), sort of way to be more modular. But it could be modified. What do you think?

`get_precision`, `gt_recall`, `get_specificity` are not exactly the same than in `medicaltorch` --> return `np.nan` instead of `0` in case of null denominator. Does this make sense? Similar to `dice_score` issue.

andreanne-lemay: It could be an option, but if we want to do so, we should allow the option between only dice or all metrics in the config file no? But I am not sure of the necessity, if all metrics are returned which include dice, is it necessary to give the option? The only reason I see is if there is a significant difference in computing time to measure all metrics vs only dice (I didn't test this)

For nan vs 0, could this differentiation be beneficial for 2Dmetrics as well? If so metrics.py could be modified and we could avoid method duplication.

charleygros: Initially, it was one for the reason I created the `param` attribute. The class could also be called independently, as a debug function.
But I see what you mean, and agree, I will modify this.

`nan` vs `0` --> when we had this debate for the `dice_score` we thought that it was quite specific and we did not wanted to impose this to `medicaltorch`. I mean: in GM segmentation (Christian's project) all your samples are positive (ie 1 true voxel at least) --> so you do not meet the issue that we have where both pred and gt can be empty, and we do not necessary want to include these values.
Alternatively, I can modify the medical torch functions by adding a parameter `err_value` (or something like this) which returns 0 or nan depending on your case. Cool, thank you for feedback.

andreanne-lemay: Oh I see! Fair point!

charleygros:
Comments about the structure-wise metrics:
- both gt and pred data are first labelled, ie each connected component (18-connectivity neighbourhood) is labelled with a value from 1 to `n_gt` or `n_pred`. For instance, data with 3 objects: one object is filled with ones, the second with 2s and the last with 3s.
- I used these labelled data to compute the TP, FN, FP.
- Debate #1: some papers considered that a lesion_pred is considered as TP if at least one voxel overlap with a lesion_gt. While other papers asked for 3 voxels. While other papers required 25% of the lesion_gt. In brief --> there is no consensus. I added a parameter to deal with this --> `overlap_vox`, fixed at 3 voxels by default.
- Debate #2: some papers remove all the lesion with less than 3 voxels prior to this evaluation (see below, from [here](https://arxiv.org/pdf/1808.01200.pdf)), which has not been implemented yet --> to discuss.
> To obtain lesion-level statistics from voxel segmentations, ground truth le-sions smaller than 3 voxels were removed, as per clinical protocol

Issue #127, opened 2019-12-20
Travis Integration


AnBucquet:
As mentioned during the last meeting, it has been decided to use [Travis CI](https://travis-ci.org/) instead of Github actions. 

I guess that the integration of [ReadTheDocs](https://readthedocs.org/) can be done at the same time?

jcohenadad:
> I guess that the integration of ReadTheDocs can be done at the same time?

i would do 2 separate PRs

PR #128, opened 2020-01-07
Travis intregation

AnBucquet:
Related to #127 

**Update:**

**Done**
- add the repository to Travis
- add .travis.yml
- work on Linux
- python 3.5/3.6/3.7 
- OSX (10.12/10.13/10.14)

**TODO**
- Integration test
- Transfer Github actions to Travis

**N.B.**
As the new version of **Pillow** (7.0.0) is not compatible with our version of **Torchvision** (because **_Pillow_version_** has been removed), I.ve pinned the version of **Pillow** to 6.1 to fix this bug.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,41 @@
 # testing file for Travis
 # https://github.com/neuropoly/ivado-medical-imaging
 language: python
+# ===== Linux ======
+dist:
+  - xenial
+  - bionic
+  - trusty
+
 python:
-  - "3.6"
-  - "3.7"
+  - 3.5
+  - 3.6
+  - 3.7
 matrix:
   include:
-    - os: linux  # list of Linux env: https://docs.travis-ci.com/user/reference/overview/
-      name: "Ubuntu 18.04 (Bionic Beaver)"
-      dist: bionic
-    - os: linux
-      name: "Ubuntu 16.04 (Xenial)"
-      dist: xenial
-    - os: linux
-      name: "Ubuntu 14.04 (Trusty)"
-      dist: trusty
-    - os: osx
-      name: "OSX 10.13 (High Sierra)"
-      osx_image: xcode11.3
-      language: shell
+    - name: "Python 3.6.5 on macOS 10.13"
+      os: osx
+      osx_image: xcode9.4  # Python 3.6.5 running on macOS 10.13
+      language: shell       # 'language: python' is an error on Travis CI macOS
+      before_install:
+        - python3 --version
+        - pip3 install -U pip
+        - pip3 install -U pytest
+      install: pip3 install -e .
+      script:
+        -
+          echo *** UNIT TESTS ***
+          travis_wait 30 pytest
+      after_success: python 3 -m codecov
 
 before_install:
-  - echo *** Python Version ***
   - python --version
   - pip install -U pip
   - pip install -U pytest
 
-install:
-  - pip install -e .
+install: pip install -e .
 
 script:
-  - |
+  -
     echo *** UNIT TESTS ***
-    travis_wait 30 pytest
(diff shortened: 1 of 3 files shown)


AnBucquet:
**Update:**

**Done**
- add the repository to Travis
- add .travis.yml
- work on Linux with python 3.6/3.7
- Fix Pillow bug (_Pillow_version_ has been removed in Pillow 7.0.0)

**WIP**

- Transfer Github actions to Travis
- Fix problems with OSX

**TODO**
- Integration test
- Coverage

.travis.yml
```diff
@@ -0,0 +1,35 @@
+# testing file for Travis
```
jcohenadad: @AnBucquet could you update the PR body (it says "WIP: add .travis.yml-- which seems done)

requirements.txt
```diff
@@ -10,5 +10,6 @@ torch==1.2.0
 torchvision==0.2.1
 tqdm>=4.30
 future
+pillow==6.1.0
```
jcohenadad: is that related to this PR?

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+    #  language: shell
+
+before_install:
+  - python3 --version
+  - pip3 install -U pip
+  - pip3 install -U pytest
+  - pip3 install codecov
+
+install:
+  - pip3 install -e .
```
jcohenadad: the doc asks to use pip (not pip3)-- i would replicate what is in the doc (the purpose of this testing is to make sure that what a user is instructed to do will work in the long term)

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+    - os: linux  # list of Linux env: https://docs.travis-ci.com/user/reference/overview/
+      name: "Ubuntu 18.04 (Bionic Beaver)"
+      dist: bionic
+    - os: linux
+      name: "Ubuntu 16.04 (Xenial)"
+      dist: xenial
+    - os: linux
+      name: "Ubuntu 14.04 (Trusty)"
+      dist: trusty
+    #- os: osx
```
jcohenadad: any reason to remove osx?

AnBucquet: Seems that most **OSX** distributions on Travis don't have python 3.X by default (waiting for Catalina...). For some unknown reasons, _xcode9.4_ works with python3 but tests return a timeout error while **Linux** tests pass.

Trying to fix this. Any help is welcomed

.travis.yml
AnBucquet: Done

requirements.txt
AnBucquet: It fixed a bug with a new version of **Pillow** (which leads tests to fail). I can move it to another PR if you prefer.

.travis.yml
jcohenadad: > tests return a timeout error
i would start looking at where is the lag (which test?) and then try to reduce it. Travis has a maximum running time for processes.

requirements.txt
jcohenadad: > I can move it to another PR if you prefer.

it's fine, but just mention it in the body of the PR so reviewers understand this is not related to the main purpose of this PR

.travis.yml
AnBucquet: **pip** works on Linux but not on **OSX/ Windows** as it points to **Python 2.X**.
Any advice?

jcohenadad: did you look into these posts?
- https://github.com/travis-ci/travis-ci/issues/9744#issuecomment-426894873
- https://travis-ci.community/t/osx-default-environment-with-python-doesnt-work/6116/2

another easy solution would be to install python locally via distribs like minconda

PR #132, opened 2020-01-20
Adaptative loader

AnBucquet:
Implementation of the **adaptative loader** mentioned in [129](https://github.com/neuropoly/ivado-medical-imaging/issues/129)

- **Done**
- [x] Create a _Dataframe_ class
- [x] transform _Bids dataset_ into **HDF5 format**
- [x] Create test_adaptative.py
- [x] implement _load_into_ram()_ & _partially_load_into_ram()_ options in new dataset and add to test

- [x] Create new dataset that uses that _dataframe_ & _HDF5_
  - [x] __get_item()__ method needs to return slices not volume in 2D
  - [x] add it to test_adaptative.py
  - [x] Transform numpy array into PIL images to be compatible with **MedicalTorch**
- [x] implement HeMIS strategy and adapt model to it.
- **WIP**
- [ ] Test Loader on CSF segmentation (cf. [2277](https://github.com/neuropoly/spinalcordtoolbox/issues/2277))
- **TODO**
- Add other strategies (Curriculum learning, Active learning, ...) and play with it.

Sister PR on medicaltorch:
https://github.com/neuropoly/medicaltorch/pull/24


AnBucquet:
Update: the implementation of `HDF5_to_Bids.py` seems a bit tricky as we do not necessarily save all slices in the HDF5 file. Indeed a filter is applied to keep only the slices with lesions, ...
[`save_nii`](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/utils.py#L168) could be used but it would imply to replace all missing slices by `0`s. 

@charleygros do you agree? see any way to get around?

jcohenadad:
> it would imply to replace all missing slices by 0s

this is what we have in GT slices without lesion masks (i.e. per definition: non-zero values), so that shouldn't be a problem, right?

AnBucquet:
Ok! I wasn't sure it was what you were looking for. I'll add this feature to the PR!

ivadomed/adaptative.py
```diff
@@ -517,18 +515,19 @@ def __getitem__(self, index):
...
 
         # Inputs
-        for ct in self.cst_lst:
+        for i, ct in enumerate(self.cst_lst):
             if self.status[ct]:
-                input_tensor = line[ct]
+                input_tensor = line[ct] * missing_modalities[i]
             else:
-                input_tensor = self.hdf5_file[line[ct]][line['Slices']]
+                input_tensor = self.hdf5_file[line[ct]][line['Slices']] * missing_modalities[i]
```
AnBucquet: I've added the **"cst_matrix"** (stands for contrast matrix) in the **Dataset variables.** The latter is basically a matrix full of "ones" except in the **HeMIS strategy** where some _"ones"_ are replaced by _"zeros"_ with a certain probability. That probability increase with time, which means that more and more modalities are missing over time. 

By using the method [**update(p)**](https://github.com/neuropoly/ivado-medical-imaging/blob/ab/adaptative_learning/ivadomed/adaptative.py#L576), we can change which modalities are missing. 

If the modality of a slice is missing, the slice is replaced by an array of "zeros". Besides the index of that modality is given to **HeMIS-Unet** to make sure that we do not take it into account when computing both **Mean** and **Var**. (see **HeMIS-Unet** architecture for more details)

charleygros: ! (--> this is a sign of appreciation / excitation)

And you are filling your encoders with zeros when missing modalities to ensure a constant batch_size, is it correct?

AnBucquet: Yes exactly!

testing/test_adaptative.py
```diff
@@ -0,0 +1,157 @@
...
+            if i > 0:
+                tot_gen = time.time() - start_gen
+                gen_lst.append(tot_gen)
+            start_load = time.time()
+            input_samples, gt_samples = batch["input"], batch["gt"]
+            print("len input = {}".format(len(input_samples)))
+            print("Batch = {}, {}".format(input_samples[0][0].shape, gt_samples.shape))
+            break
+        break
+    print("Congrats your dataloader works! You can go Home now and get a beer.")
```
charleygros: Amen

testing/test_adaptative.py
```diff
@@ -0,0 +1,157 @@
...
+    optimizer = optim.Adam(model.parameters(), lr=INIT_LR)
+    scheduler = optim.lr_scheduler.CosineAnnealingLR(optimizer, N_EPOCHS)
+
+    load_lst, pred_lst, opt_lst, schedul_lst, init_lst, gen_lst = [], [], [], [], [], []
+    for epoch in tqdm(range(1, N_EPOCHS + 1), desc="Training"):
+        start_time = time.time()
+
+        start_init = time.time()
+        lr = scheduler.get_lr()[0]
+        model.train()
```
charleygros: Do you plan to feed the model in this testing function?

AnBucquet: Finally, I decided to test the training of HeMIS in `test_HeMIS.py` and forgot to remove those lines of code. Thx

ivadomed/models.py
```diff
@@ -284,34 +289,41 @@ def __init__(self, modalities, depth=3, drop_rate=0.4, bn_momentum=0.1):
...
                 features_mod[j].append(features[j].unsqueeze(0))
 
         # Abstraction
         for j in range(self.depth + 1):
-            features_mod[j] = torch.cat([torch.cat(features_mod[j], 0).mean(0).unsqueeze(0), \
-                                         torch.cat(features_mod[j], 0).var(0).unsqueeze(0)], 0)
+            features_cat = torch.cat(features_mod[j], 0).transpose(0, 1)
+
+            features_mod[j] = torch.cat([torch.cat([features_cat[i][indexes_mod[i].nonzero()].squeeze(1).mean(0), 
+                                                    features_cat[i][indexes_mod[i].nonzero()].squeeze(1).var(0)], 0).unsqueeze(0) for i in range(len(indexes_mod))],0)
```
ieee8023: Here

config/config_spineGeHemis.json
```diff
@@ -0,0 +1,82 @@
+{
+	"command": "train",
+	"gpu": 6,
+	"hdf5_path": "../duke/projects/csf_deepseg/spineGeneric/hdf5/SpineGe.hdf5",
+	"csv_path": "../duke/projects/csf_deepseg/spineGeneric/hdf5/SpineGe.csv",
```
charleygros, line 5: super

ivadomed/adaptative.py
```diff
@@ -0,0 +1,695 @@
...
+
+class Dataframe:
+    """
+    This class aims to create a dataset using an HDF5 file, which can be used by an adapative loader
+    to perform curriculum learning, Active Learning or any other strategy that needs to load samples in a specific way.
+    It works on RAM or on the fly and can be saved for later.
+    """
+
+    def __init__(self, hdf5, contrasts, path, target_suffix=None, roi_suffix=None,
+                 slices=False, dim=2):
```
charleygros, line 22: SOrry to be picky, could we rename `slices` by `slice_filter_fn`? I was not sure what it meant.

ivadomed/adaptative.py
```diff
@@ -0,0 +1,695 @@
...
+
+        # Initialize the data frame
+        col_names = [col for col in empty_line.keys()]
+        col_names.append('Subjects')
+        df = pd.DataFrame(columns=col_names)
+        print(hdf5.attrs['patients_id'])
+        # Filling the data frame
+        for subject in hdf5.attrs['patients_id']:
+            # Getting the Group the corresponding patient
+            print(subject)
```
charleygros, line 92: rm?

ivadomed/adaptative.py
```diff
@@ -0,0 +1,695 @@
...
+            inputs = grp['inputs']
+            for c in inputs.attrs['contrast']:
+                if c in col_names:
+                    line[c] = '{}/inputs/{}'.format(subject, c)
+                else:
+                    continue
+            # GT
+            assert 'gt' in grp.keys()
+            inputs = grp['gt']
+            print("input = ", inputs.attrs['contrast'])
```
charleygros, line 108: or in a debug mode?

ivadomed/adaptative.py
charleygros, line 22: Not sure to understand in which case we would have `dim=2` and `slices=False`!?

ivadomed/loader.py
```diff
@@ -386,6 +389,22 @@ def load_dataset(data_list, data_transform, context):
...
+                                         csv_name=context["csv_path"],
+                                         target_suffix=context["target_suffix"],
+                                         contrast_lst=context["contrast_train_validation"],
+                                         ram=context['ram'],
+                                         contrast_balance=context["contrast_balance"],
+                                         slice_axis=AXIS_DCT[context["slice_axis"]],
+                                         transform=data_transform,
+                                         metadata_choice=context["metadata"],
+                                         slice_filter_fn=utils.SliceFilter(**context["slice_filter"]),
+                                         roi_suffix=context["roi_suffix"],
```
charleygros, line 405: rm?

ivadomed/main.py
```diff
@@ -129,16 +135,17 @@ def cmd_train(context):
         print(
             f"Loaded {len(ds_train)} volumes of size {context['length_3D']} for the training set.")
 
-    if context['balance_samples']:
+    if context['balance_samples'] and not HeMIS:
```
charleygros, line 138: great

ivadomed/main.py
```diff
@@ -388,9 +398,8 @@ def cmd_train(context):
 
             num_steps += 1
 
-            # Only write sample at the first step
             if i == 0:
-                utils.save_tensorboard_img(writer, epoch, "Train", input_samples, gt_samples, preds, unet_3D)
+                imed_utils.save_tensorboard_img(writer, epoch, "Train", input_samples, gt_samples, preds, unet_3D)
```
charleygros, line 402: I should have communicated with you about this changes :/ duplication. Not a worry tho

ivadomed/main.py
```diff
@@ -401,6 +410,15 @@ def cmd_train(context):
             dice_train_loss_total_avg = dice_train_loss_total / num_steps
             tqdm.write(f"\tDice training loss: {dice_train_loss_total_avg:.4f}.")
 
+        # In case of curriculum Learning we need to update the loader
+        if HeMIS:
+            p = p ** (5 / 6)
```
charleygros, line 415: Should we move the `5 / 6` in the config file? dunno

config/config_spineGeHemis.json
```diff
@@ -0,0 +1,82 @@
...
+	"film_layers": [0, 0, 0, 0, 0, 0, 0, 0],
+	"mixup_bool": false,
+	"mixup_alpha": 2,
+	"metadata": "constrast",
+	"multichannel": false,
+	"HeMIS": true,
+	"ram": true,
+	"missing_probability": 0.0001,
+	"target_lst": ["T2w_seg-manual"],
+	"roi_lst": null,
```
charleygros, line 34: I am not sure to understand how `roi_lst` is used.

testing/test_HeMIS.py
```diff
@@ -1,89 +1,111 @@
...
+                                     ram=False,
+                                     contrast_balance={},
+                                     slice_axis=2,
+                                     transform=train_transform,
+                                     metadata_choice=False,
+                                     dim=2,
+                                     slice_filter_fn=utils.SliceFilter(filter_empty_input=True, filter_empty_mask=True),
+                                     canonical=True,
+                                     roi_suffix="_seg-manual",
+                                     target_lst=['T2w'],
```
charleygros, line 65: Why sometimes it is `T2w_seg-manual`, while sometimes it is `T2w`?

config/config_spineGeHemis.json
jcohenadad: could you please move it to ivado/config/

ivadomed/main.py
jcohenadad, line 415: what do these numbers correspond to? a comment would be welcome

testing/test_dilateGT.py
```diff
@@ -15,7 +15,7 @@
 
 cudnn.benchmark = True
 
-GPU_NUMBER = 5
+GPU_NUMBER = 1
```
jcohenadad, line 18: [here](https://github.com/neuropoly/ivado-medical-imaging/pull/132/files?file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-64f10db8dac07e1cb7f592cb33a24b7cR21) GPU_NUMBER is set to 0. What is the reasoning behind setting those to 0 or 1?

andreanne-lemay: I changed it to 0. The tests were created with different GPU numbers, but I believe there was no particular reasoning for that.

ivadomed/main.py
```diff
@@ -44,18 +45,22 @@ def cmd_train(context):
...
     attention = context["attention_unet"]
-    HeMIS = context['missing_modality']
     if film_bool:
         context["multichannel"] = False
         HeMIS = False
     elif context["multichannel"]:
         HeMIS = False
+    if HeMIS:
+        # Initializing the missing probability for HeMIS.
+        # The lower the probability the higher, the more modalities are missing.
```
charleygros, line 62: "The lower the probability, the more modalities are missing." ?

ivadomed/main.py
```diff
@@ -401,6 +410,16 @@ def cmd_train(context):
             dice_train_loss_total_avg = dice_train_loss_total / num_steps
             tqdm.write(f"\tDice training loss: {dice_train_loss_total_avg:.4f}.")
 
+        # In case of curriculum Learning we need to update the loader
+        if HeMIS:
+            # Increase the probability of a missing modality
+            p = p ** (context["missing_probability_growth"])
```
charleygros, line 416: great!

ivadomed/main.py
andreanne-lemay: I agree the comment is not clear, but I believe it means Higher probability, means more modalities missing. I'll reprhase it ;)

ivadomed/adaptative.py
andreanne-lemay: Since I never really read those print I simply removed them :)!

ivadomed/adaptative.py
andreanne-lemay: The parameter `slices` is misleading, it represents if the slices are filter or not (slice_filter_fn). So in 2D it is possible to filter or not the slices if gt or inputs are empty. I'll change the param name to slices_filter instead

config/config_spineGeHemis.json
andreanne-lemay: target_lst and roi_lst indicates the modality you want to use for the ground truth or for the roi. For example, i'd like to use the roi from T1w modality (sub000_T1w_seg-manual and not sub000_T2w_seg-manual). Target_lst is a list to have the option to have more than one ground truth. Roi_lst is a list since some places deal with roi and gt the same way.

testing/test_HeMIS.py
andreanne-lemay: Both work since it is `if "T2w" in "string"`, but it should be only the modality i'll standardize it :)

PR #133, opened 2020-01-21
Integrate 3D UNet + Loader

andreanne-lemay:
- [x] Addition of a modified 3D Unet model. (from https://arxiv.org/abs/1802.10508) This implementation of 3D Unet worked better than 3D Unet from medicaltorch.
- [x] Added 3D transforms.

To do:
- [x] Add tests
- [x] one metric is not working (hausdorff score)


andreanne-lemay:
Related PR: https://github.com/neuropoly/medicaltorch/pull/15

ivadomed/main.py
```diff
@@ -59,8 +59,11 @@ def cmd_train(context):
 
     if bool(sum(context["film_layers"])) and not (metadata_bool):
         print('\tWarning FiLM disabled since metadata is disabled')
-    print('\nArchitecture: {} with a depth of {}.\n' \
-          .format('FiLMedUnet' if film_bool else 'HeMIS-Unet' if HeMIS else 'Unet', context['depth']))
+    if context['unet_3D']:
+        print('\nArchitecture: 3D Unet')
+    else:
+        print('\nArchitecture: {} with a depth of {}.\n' \
+              .format('FiLMedUnet' if film_bool else 'HeMIS-Unet' if HeMIS else 'Unet', context['depth']))
```
charleygros: Could we combine all this & give the info of depth in 3D Unet case too?

ivadomed/main.py
```diff
@@ -116,17 +121,33 @@ def cmd_train(context):
...
+                                      target_suffix=context["target_suffix"],
+                                      roi_suffix=context["roi_suffix"],
+                                      contrast_lst=context["contrast_train_validation"],
+                                      metadata_choice=context["metadata"],
+                                      contrast_balance=context["contrast_balance"],
+                                      slice_axis=axis_dct[context["slice_axis"]],
+                                      transform=train_transform,
+                                      multichannel=True if context['multichannel'] else False,
+                                      slice_filter_fn=SliceFilter(**context["slice_filter"]))
+
```
charleygros: Almost all parameter are the same between the two loaders, could we have only one class for both and an attribute that could differentiate the two? Or would it be too much work?

ivadomed/main.py
```diff
@@ -158,17 +179,33 @@ def cmd_train(context):
...
+                                    padding=context["padding_3D"])
+    else:
+        ds_val = loader.BidsDataset(context["bids_path"],
+                                    subject_lst=valid_lst,
+                                    target_suffix=context["target_suffix"],
+                                    roi_suffix=context["roi_suffix"],
+                                    contrast_lst=context["contrast_train_validation"],
+                                    metadata_choice=context["metadata"],
+                                    contrast_balance=context["contrast_balance"],
+                                    slice_axis=axis_dct[context["slice_axis"]],
```
charleygros: Same comment as above.

ivadomed/main.py
```diff
@@ -205,7 +244,9 @@ def cmd_train(context):
     if context['multichannel']:
         in_channel = len(context['multichannel'])
 
-    if context['retrain_model'] is None:
+    if context['unet_3D']:
+        model = models.Modified3DUNet(in_channels=in_channel, n_classes=1)
```
charleygros: The 2D-Unet is also not the original Unet --> what about calling it `Unet3D` or something like this and explain in the docstring the difference with the original 3D Unet?

charleygros: Would it be too much work to input the depth?

ivadomed/main.py
andreanne-lemay: I put them separate because BidsDataset and Bids3DDataset don't have the same parent class. Do you know a good way to change the parent according to a parameter?

ivadomed/main.py
andreanne-lemay: Yes, will do!

andreanne-lemay: For the depth, I will investigate. It is probably possible!

charleygros: Reading briefly the article, the main differences are:
- instance normalisation
- "segmentation" layers on the decoder part of the Unet
- leaky ReLU
Is there something else?

ivadomed/main.py
charleygros: Oh! I see..
We could do something like suggested herebelow but not sure if it would be cleaner:
- in `main.py` --> call a function (eg `load_dataset`)with a parameter which controls 2D or 3D
- in `loader.py` --> the function `load_dataset` initialise the BidsDataset with either `Bids3DDataset`, either `BidsDataset`.

I am mainly annoying with that because I am conscious that the `main.py` is getting very long.

ivadomed/main.py
```diff
@@ -413,6 +471,8 @@ def cmd_train(context):
                       mt_metrics.specificity_score,
                       mt_metrics.intersection_over_union,
                       mt_metrics.accuracy_score]
+        if context["unet_3D"]:
+            metric_fns.remove(mt_metrics.hausdorff_score)
```
charleygros: Just realise --> this block could certainly be moved out of the training loop I think.

ivadomed/main.py
```diff
@@ -701,18 +797,25 @@ def cmd_test(context):
...
-                # re-init pred_stack_lst
-                pred_tmp_lst, z_tmp_lst = [], []
-
-            # add new sample to pred_tmp_lst
-            pred_tmp_lst.append(np.array(rdict_undo['gt']))
-            z_tmp_lst.append(int(rdict_undo['input_metadata']['slice_index']))
-            fname_tmp = fname_ref
+                # Choose only one modality
+                save_nii(rdict_undo['gt'][0, :, :, :].transpose((1, 2, 0)), [], fname_ref, fname_pred,
+                         axis_dct[context['slice_axis']], unet_3D=True)
```
charleygros: That s with the case where you have each patch = one full image. We will have to remember to change this section if one day we do smaller patches.

ivadomed/transforms.py
```diff
@@ -308,3 +312,190 @@ def __call__(self, sample):
...
+class ElasticTransform3D(mt_transforms.ElasticTransform):
+    @staticmethod
+    def elastic_transform(image, alpha, sigma):
+        shape = image.shape
+        dx = gaussian_filter((np.random.rand(*shape) * 2 - 1),
+                             sigma, mode="constant", cval=0) * alpha
+        dy = gaussian_filter((np.random.rand(*shape) * 2 - 1),
+                             sigma, mode="constant", cval=0) * alpha
+        dz = gaussian_filter((np.random.rand(*shape) * 2 - 1),
+                             sigma, mode="constant", cval=0) * alpha
```
charleygros: If not too much work --> can we combine it with the 2D transform? I think there is not too much difference between the two implementations.
(it can also be done later, after the MIDL deadline, just opening an issue as reminder :) )

ivadomed/utils.py
```diff
@@ -214,7 +219,8 @@ def save_nii(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False):
     arr_pred_ref_space = nib.orientations.apply_orientation(arr_ras, trans_orient)
 
     # create nii
-    nib_pred = nib.Nifti1Image(arr_pred_ref_space, nib_ref.affine)
+    nib_pred = nib.Nifti1Image(np.where(arr_pred_ref_space > 0.5, 1.0, 0.0), nib_ref.affine)
```
charleygros: I would recommend to move this thresholding out of this function or at least to add a parameter for this as this is a key post processing step (and to use the function already implemented in utils?).

ivadomed/main.py
andreanne-lemay: I think this sums it up :)!

ivadomed/loader.py
```diff
@@ -323,4 +346,34 @@ def __iter__(self):
             self.weights, self.nb_samples, replacement=True))
 
     def __len__(self):
-        return self.num_samples
\ No newline at end of file
+        return self.num_samples
+
+
+def load_dataset(data_list, data_transform, context):
```
charleygros: :+1:

ivadomed/main.py
```diff
@@ -119,21 +125,10 @@ def cmd_train(context):
...
-                                  roi_suffix=context["roi_suffix"],
-                                  contrast_lst=context["contrast_train_validation"],
-                                  metadata_choice=context["metadata"],
-                                  contrast_balance=context["contrast_balance"],
-                                  slice_axis=axis_dct[context["slice_axis"]],
-                                  transform=train_transform,
-                                  multichannel=True if context['multichannel'] else False,
-                                  missing_modality=HeMIS,
-                                  slice_filter_fn=SliceFilter(**context["slice_filter"]))
+    ds_train = loader.load_dataset(train_lst, train_transform, context)
```
charleygros: :100:

ivadomed/main.py
```diff
@@ -211,14 +203,16 @@ def cmd_train(context):
     in_channel = 1
 
     if context['multichannel']:
-        in_channel = len(context['multichannel'])
+        in_channel = len(context['contrast_train_validation'])
```
charleygros: When `contrast_train_validation` was introduced in the config files, it was used in the following context: Train a Unet on  `contrast_train_validation` and inference on `contrast_testing`.
What was the reason to get ride of `multichannel` param?

ivadomed/main.py
```diff
@@ -319,6 +313,16 @@ def cmd_train(context):
     epsilon = context["early_stopping_epsilon"]
     val_losses = []
 
+    metric_fns = [dice_score,  # from ivadomed/utils.py
```
charleygros: Yep!

ivadomed/main.py
```diff
@@ -319,6 +313,16 @@ def cmd_train(context):
...
+                  mt_metrics.hausdorff_score,
+                  mt_metrics.precision_score,
+                  mt_metrics.recall_score,
+                  mt_metrics.specificity_score,
+                  mt_metrics.intersection_over_union,
+                  mt_metrics.accuracy_score]
+    if unet_3D:
+        metric_fns.remove(mt_metrics.hausdorff_score)
```
charleygros: If we want to have hausdorff for 3D models --> hausdorff_score is implemented for 2D slices --> we could loop across the slices of the 3D patch, then return the mean. As you prefer, your call, only a suggestion.

ivadomed/transforms.py
```diff
@@ -296,3 +301,126 @@ def __call__(self, sample):
             sample.update(rdict)
 
         return sample
+
+
+class StackTensors(mt_transforms.StackTensors):
+    """This class extends mt_transforms.NormalizeInstance"""
+    def undo_transform(self, sample):
+        rdict = {}
```
charleygros: No update of the Tensor? I am not familiar with this function, trust you ;-)

ivadomed/transforms.py
```diff
@@ -296,3 +301,126 @@ def __call__(self, sample):
...
+
+        return sample
+
+
+# 3D Transforms
+class CenterCrop3D(mt_transforms.MTTransform):
```
charleygros: :1st_place_medal:

ivadomed/utils.py
```diff
@@ -211,11 +216,12 @@ def save_nii(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False):
     # Return the orientation that transforms from ras to ref_orientation
     trans_orient = nib.orientations.ornt_transform(ras_orientation, ref_orientation)
     # apply transformation
-    arr_pred_ref_space = nib.orientations.apply_orientation(arr_ras, trans_orient)
+    arr_pred_ref_space = threshold_predictions(nib.orientations.apply_orientation(arr_ras, trans_orient))
```
charleygros: WOuld it be possible to add a param which controls the thresholding? Either a Bool either the threshold value --> config files. For some applications (eg uncertainty computations), we may want to keep non-binarised values.

ivadomed/main.py
andreanne-lemay: Yes good idea, I submitted the changes :)!

ivadomed/transforms.py
andreanne-lemay: Under undo transforms don't support, undo on lists of tensors so keeping it concatenate is more useful. For now, I see no benefit in resplitting the input in lists of tensors.

ivadomed/utils.py
andreanne-lemay: Updated :)!

ivadomed/utils.py
```diff
@@ -244,6 +244,19 @@ def dice_score(im1, im2):
     return (2. * intersection.sum()) / im_sum
 
 
+def hausdorff_score(prediction, groundtruth):
```
charleygros: :+1:

ivadomed/utils.py
```diff
@@ -216,7 +216,9 @@ def save_nii(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False, une
     # Return the orientation that transforms from ras to ref_orientation
     trans_orient = nib.orientations.ornt_transform(ras_orientation, ref_orientation)
     # apply transformation
-    arr_pred_ref_space = threshold_predictions(nib.orientations.apply_orientation(arr_ras, trans_orient))
+    arr_pred_ref_space = nib.orientations.apply_orientation(arr_ras, trans_orient)
+    if binarize:
```
charleygros: :+1:

PR #140, opened 2020-02-13
in_channel is set twice

olix86:


AnBucquet:
Thx for  the PEP8 corrections 😍

PR #141, opened 2020-02-14
Fix errors in test command on 3D and multichannel

andreanne-lemay:
When running the test command, some errors occurred on 3D models or multichannel.

### DIFF
--- a/ivadomed/main.py
+++ b/ivadomed/main.py
@@ -741,7 +741,7 @@ def cmd_test(context):
             rdict['gt'] = preds.cpu()
             batch.update(rdict)
 
-            if batch["input"].shape[1] > 1:
+            if batch["input"].shape[1] > 1 and not i_monteCarlo:
                 batch["input_metadata"] = batch["input_metadata"][0]  # Take only second channel
 
             # reconstruct 3D image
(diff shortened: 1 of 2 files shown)


ivadomed/main.py
```diff
@@ -799,36 +803,37 @@ def cmd_test(context):
...
-
-        fname_unc_vox = os.path.join(path_3Dpred, subj_acq+'_unc-vox.nii.gz')
-        fname_unc_struct = os.path.join(path_3Dpred, subj_acq+'_unc.nii.gz')
-        if not os.path.isfile(fname_unc_vox) or not os.path.isfile(fname_unc_struct):
-           # compute voxel-wise uncertainty map
-           voxelWise_uncertainty(fname_pred_lst, fname_unc_vox)
-
-           # compute structure-wise uncertainty
-           structureWise_uncertainty(fname_pred_lst, fname_pred, fname_unc_vox, fname_unc_struct)
+    if context['uncertainty']['epistemic'] or context['uncertainty']['aleatoric']:
```
andreanne-lemay: Without this condition, fname_pred_list is empty which triggers an error in combine_predictions.

charleygros: Indeed :-) Thanks!

ivadomed/main.py
```diff
@@ -741,7 +741,7 @@ def cmd_test(context):
             rdict['gt'] = preds.cpu()
             batch.update(rdict)
 
-            if batch["input"].shape[1] > 1:
+            if batch["input"].shape[1] > 1 and not i_monteCarlo:
```
charleygros: Not sure to understand here: why the first iteration of MC is not taking only second channel?

andreanne-lemay: Because it is the same batch across the for loop and the operation needs to be done only once :)

charleygros: oooo! makes sense!

Issue #142, opened 2020-02-14
prepare_data.sh not compatible with current version of sct.


AnBucquet:
In `prepare_data.sh`, we're using an old version of sct (cf line [72](
https://github.com/neuropoly/ivado-medical-imaging/blob/master/prepare_data/prepare_data.sh#L72)):

`sct_crop_image -i ${file_t1w_mts}.nii.gz -o ${file_t1w_mts}_crop.nii.gz -start 3 -end -3 -dim 2` 

**returns** 

`sct_crop_image: error: unrecognized arguments: -start 3 -end -3 -dim 2`

AnBucquet:
Fixed by #143.

PR #143, opened 2020-02-17
Update prepare_data.sh

AnBucquet:
Related to #142.


prepare_data/README.md
```diff
@@ -13,6 +13,8 @@ In its current state, this pipeline uses [SCT development version](https://githu
 See [README](../README.md)
 ~~~
 source PATH_TO_YOUR_VENV/venv-ivadomed/bin/activate
+# install prepdata
+cd .. && python setup.py install && cd -
```
AnBucquet: It seems that [prepdata](https://github.com/neuropoly/ivado-medical-imaging/tree/master/prepare_data/prepdata) is not installed with ivadomed, which leads to the failure of `./run_process.sh parameters.sh final_qc.sh`.

prepare_data/README.md
```diff
@@ -100,5 +102,5 @@ Copy final files to anat/, copy json sidecars, move segmentations to derivatives
 Once QC and manual correction is done, remove tmp/ folder:
 
 ~~~
-./run_process.sh delete_tmp_files.sh
+./run_process.sh parameters.sh delete_tmp_files.sh
```
jcohenadad: actually, we can now use `sct_run_batch`.

AnBucquet:
Tested on **ub-balgrist02**, seems to work.
(cf. `/Duke/projects/csf_deepseg/SpineGe_debbug/result/sub-balgrist02`)

prepare_data/README.md
jcohenadad: duplication with https://github.com/neuropoly/ivado-medical-imaging/blob/master/README.md#installing

jcohenadad: ah yes you're right... it's not the most intuitive way of organizing a software structure (i'm the one to blame), and the only dependent python software is actually `prepdata remove-slice` (called by `final_qc.sh`). So i guess until we find a better structure/organization (feel free to suggest), we should indeed keep that install documentation.

Issue #135, opened 2020-02-04
Post-processing


charleygros:
A key feature that we are currently missing is post-processing pipeline.

Some tools could be implemented and integrated in our testing pipeline:
- basic image processing to remove false positive ?
- basic image processing to fill holes ?
- lesion filtering based on uncertainty measures ?
- segmentation adjustment with CRF postprocessing fed with uncertainty measures and soft segmentations

jcohenadad:
Excellent point, which we should probably discuss during our meeting today

PR #146, opened 2020-02-23
Paint prediction TP / FP / FN

charleygros:
Implement few tools to analyse results:
- paint prediction TP / FP / FN --> ie TP lesions are labeled 1, FP 2, FN 3 --> easy visualisation of the results when opening the nifti file in fsleyes --> DONE
- remove objects of less than 3 voxels (common practice in litterature) --> DONE
- compute object-wise metrics per lesion size (small, medium, large size) --> DONE

More details herebelow.


charleygros:
Remove objects made of less than 3 voxels:
-  as per clinical protocol in MS lesions [eg REF Nair 2019, Carass 2017]
- currently done within the class `Evaluation3DMetrics` before computing the 3D metrics. But could be eventually moved in a "post-processing" pipeline, see #135.

charleygros:
Painting tool:
- called when running `Evaluation3DMetrics.run_eval()`
- saved in `pred_masks/*_pred_painted.nii.gz`
- TP: 1 ; FP: 2 ; FN: 3

Example:
```
fsleyes sub-bwh055_acq-ax_T2w.nii.gz sub-bwh055_acq-ax_T2w_pred_painted.nii.gz -cm hsv
```
Screenshot 2020-02-24 14:29:14

The use of `hsv` color map leads to: TP: green ; FP: blue ; FN: red.

charleygros:
Adding dev / utils script to compute the target size distribution of each individual object (ie 1 lesion or 1 tumour or..).

Command:
```
python dev/target_size.py -c config/config_sctTesting_lesionAx.json
```
It uses the config file to extract `bids_path`, `contrast_test`, `target_suffix`.

Terminal output:
```
Target distribution in vox:
	Mean: 118.49115417017691
	Median: 49.0
	Inter-quartile range: [21.0, 117.75]
	Save as: _lesion-manual_vox.png

Target distribution in mm3:
	Mean: 93.58197132028437
	Median: 43.89957133959979
	Inter-quartile range: [19.78110589838496, 101.89889602643234]
	Save as: _lesion-manual_mm3.png
```
 
Plots:
_lesion-manual_mm3

charleygros:
After looking at the distribution & discussing with Anne, I am thinking of the following thresholds for the lesion size:
```
small lesions (3-20 vox), medium (21-100 vox), large (101+)
```

Note: Nair et al. used the following on their brain dataset:
```
small (3-10 vox), medium (11-50 vox), and large (51+) lesion.
```

--> Any thoughts?

jcohenadad:
> After looking at the distribution & discussing with Anne, I am thinking of the following thresholds for the lesion size (in vox):
small lesions (3-20 vox), medium (21-100 vox), large (101+)
Note: Nair et al. used the following on their brain dataset:
small (3-10 vox), medium (11-50 vox), and large (51+) lesion.

why not hard-coding the thresholds in physical dimensions? e.g., if in 2 years someone changes the output resolution (i.e. voxel size), the effective threshold size will also change (without the user being necessarily aware of it)

charleygros:
I am adding these parameters in the config files to avoid hard coding:
```
        "eval_params": {
            "targetSize": {"unit": "vox", "thr": [20, 100]},
            "removeSmall": {"unit": "vox", "thr": 3},
            "overlap": {"unit": "vox", "thr": 3}
            },
```

These parameters are detailed in the below comments.

charleygros:
### Overlap

Overlap controls the overlap threshold used to compute object-wise TP, FP, FN. It can be expressed in:
- voxels: if lesion_gt overlaps with at least 3 vox with lesion_pred, then TP++
- mm3: if lesion_gt overlaps with at least 3 mm3 with lesion_pred, then TP++
- percentage of the GT. if lesion_pred overlaps with at least 50% of lesion_gt, then TP++

charleygros:
### RemoveSmall

RemoveSmall is used to remove small blobs in pred prior to metrics' computation.
It can be expressed either in voxels, either in mm3.
If not specified, then nothing removed.

charleygros:
### TargetSize

TargetSize is used to compute metrics for different size of object
It can be expressed either in voxels, either in mm3.

If not specified, then metrics are computed on all objects (ie 1 output dataframe).
If specified, then metrics are computed on all objects + on objects grouped by size
--> example: ```"targetSize": {"unit": "vox", "thr": [20, 100]}``` leads to 1 + 3 output dataframes: (1) all objects, (ii) objects between 0 and 20 vox, (iii) objects between 21 and 100 vox, (iv) objects for more than 101 vox.
Output fname examples: ```['log_lesionAx_retrain/results_eval/evaluation_3Dmetrics_3-20vox.csv', 'log_lesionAx_retrain/results_eval/evaluation_3Dmetrics_21-100vox.csv', 'log_lesionAx_retrain/results_eval/evaluation_3Dmetrics_101-INFvox.csv']```.

Could also be of interest for @andreanne-lemay.

charleygros:
The current implementation of analysis across size does not work. What I did so far: filter the pred and gt for a range of size, then compute the metrics on the resulting mask.
It does not work because the GT "small lesion" are not necessarily the pred "small lesion" etc.
What I should do instead: label the GT mask in terms of size, then refer to this labeled mask to run the analysis across size.

charleygros:
New version:
- analysis per size is only done for the metrics LTPR and LFDR.
- Only one csv saved, containing the following new metrics:
- `n_3-20vox`, `n_21-100vox` etc. --> lesion count per size
- `ltpr_3-20vox` etc & `lfdr_3-20vox` etc. --> object-wise metrics per size

charleygros:
Results expressed as median:
```
lfdr = 63.1578947368
lfdr_3-20vox = 0
lfdr_21-100vox = 0
lfdr_101-INFvox = 0

ltpr = 100
ltpr_3-20vox = 83.3333333333
ltpr_21-100vox = 100
ltpr_101-INFvox = 100

n_gt = 2
n_3-20vox = 2
n_21-100vox = 2
n_101-INFvox = 1
n_pred = 7
```

Conclusions:
- LFDR results: incoherent, need to fix bug
- LTPR: coherent results, less good results with smaller lesions
- median n_pred >> n_gt --> lot of FP

charleygros:
LFDR size analysis now run on pred mask labeled per size.

charleygros:
New LFDR results are:
```
lfdr_all = 20.5907452426
lfdr_3-20vox = 13.1113423517 
lfdr_21-100vox = 31.0090317486
lfdr_101-INFvox = 22.5953472913
```

Which makes much more sense.

ivadomed/utils.py
```diff
@@ -69,6 +114,64 @@ def get_pixdim(self, fname):
         px, py, pz = nib_im.header['pixdim'][1:4]
         return px, py, pz
 
+    def remove_small_objects(self, data):
```
andreanne-lemay: I feel like the pred_mask should be overwritten with the new post-processing step (save the new nii.gz file)

dev/target_size.py
```diff
@@ -0,0 +1,101 @@
...
+                 kde_kws={'linewidth': 2},
+                 rug_kws={'color': 'black'})
+
+    plt.xlabel(label)
+    plt.xlim(xlim)
+    plt.ylabel('Density')
+    fig.savefig(fname_out)
+    print('\tSave as: '+fname_out)
+
+def run_main(args):
```
andreanne-lemay: Very nice!

ivadomed/utils.py
charleygros: Good Idea!
I could move this function to a `post_processing` module, will do in #151 !

ivadomed/utils.py
```diff
@@ -38,27 +41,69 @@ def get_results(self):
...
+
+            self.data_gt_per_size = self.label_per_size(self.data_gt)
+            label_gt_size_lst = list(set(self.data_gt_per_size[np.nonzero(self.data_gt_per_size)]))
+            self.data_pred_per_size = self.label_per_size(self.data_pred)
+            label_pred_size_lst = list(set(self.data_pred_per_size[np.nonzero(self.data_pred_per_size)]))
+            self.label_size_lst = [label_gt_size_lst + label_pred_size_lst,
+                                    ['gt']*len(label_gt_size_lst) + ['pred']*len(label_pred_size_lst)]
+
+        else:
+            self.label_gt_size_lst, self.label_gt_size_lst = []
```
andreanne-lemay: here it should probably be: self.label_gt_size_lst, self.label_size_lst = [], []

ivadomed/utils.py
```diff
@@ -161,9 +293,21 @@ def get_all_metrics(self):
         dct['precision'] = mt_metrics.precision_score(self.data_pred, self.data_gt, err_value=np.nan)
         dct['specificity'] = mt_metrics.specificity_score(self.data_pred, self.data_gt, err_value=np.nan)
         dct['n_pred'], dct['n_gt'] = self.n_pred, self.n_gt
-        dct['ltpr'] = self.get_ltpr()
+        dct['ltpr'], _ = self.get_ltpr()
         dct['lfdr'] = self.get_lfdr()
 
+        for lb_size, gt_pred in zip(self.label_size_lst[0], self.label_size_lst[1]):
```
andreanne-lemay: if target_size is not in eval_params these values are not defined !

PR #147, opened 2020-02-27
Fix errors in prepare_data.sh

AnBucquet:
- Following the PR #143, there was still a problem  with the use of the parameter **zmax** in _sct_crop_image_. As we want to remove the 3 lastest slices we need to set **zmax** to -4.
- We also added a registration of the manual seg onto the image voxel space in case, for some reasons, the manual seg is not in the same space (e.g. this could happen if the cropping values change across versions of this pipeline)


jcohenadad:
> As the manual segmention have been created with the old version of sct_crop_image, they are bigger than the images which leads sct_create_mask to fail. For that reason, we register the manual segmentations in the space of the images.

the way this PR is written is misleading because the previous point (changing -zmax from 3 to 4) fixes the "bigger segmentation" problem. I would rephrase as: We also added a registration of the manual seg onto the image voxel space in case, for some reasons, the manual seg is not in the same space (e.g. this could happen if the cropping values change across versions of this pipeline)

AnBucquet:
Thx for the review! I've updated the PR body.

PR #153, opened 2020-03-07
Reproducibility

olix86:
Done 
- Introduce flag `-n / --n-iterations` to run multiple times each config with `dev/automate_training.py`
- Compute average, standard dev for each config
- Compute student T-test to get p-values between each pair of configs 
- Test on random data
- Rename automate_training.py to compare_models.py
- Random seed change
- Add option to run `cmd_eval` and compare test scores
- Test with real configs (train / validation)
- Test with real configs (eval / test)


dev/automate_training.py
```diff
@@ -21,6 +22,7 @@
 from ivadomed import main as ivado
 from ivadomed import loader
 from itertools import product
+from scipy.stats import ttest_ind_from_stats
```
charleygros: Great!

dev/automate_training.py
```diff
@@ -29,48 +31,56 @@
...
     # Save config file in log directory
-    config_copy = open(config["log_directory"] + "/config.json","w")
+    config_copy = open(config["log_directory"] + "/config.json", "w")
     json.dump(config, config_copy, indent=4)
 
     return config["log_directory"], best_training_dice, best_training_loss, best_validation_dice, best_validation_loss
 
 
+def test_worker(config):
+    return config["log_directory"], np.random.normal(0, 1), np.random.normal(0, 1), np.random.normal(0, 1), np.random.normal(0, 1)
```
charleygros: Could you please add a docstring about the purpose of this function? Is it to init the dataframe with values for `best_training_dice` etc?

charleygros:
This tool is really great! Actually I find it awesome, thank you.

General comment: This p-value computation is to assess the reproducibility of training across different hyperparameters.
For instance, it would help to answer the question: "Does CyclicLR scheduler lead to a significant  better training than CosineAnnealingLR scheduler when comparing the validation_dice?" --> @olix86 : Is it correct?

Once hyperparameter optimisation performed: To answer the reproducibility of our results with a specific model (as mentioned during our past abstract submissions), we will need to implement a similar tool but running `cmd_train` then `cmd_eval` (or `cmd_test`) across `n_iterations` (instead of `cmd_train` only.

dev/automate_training.py
olix86: It was just to generate fake values to avoid running `ivadomed` when testing the script 😛

olix86:
> For instance, it would help to answer the question: "Does CyclicLR scheduler lead to a significant better training than CosineAnnealingLR scheduler (when comparing the validation_dice)?"

Yep exactly

> Once hyperparameter optimisation performed: To assess the reproducibility of our results with a specific model (as mentioned during our past abstract submissions) or to compare two approches (i.e. FiLM vs no-FiLM), we will need to implement a similar tool but running cmd_train then cmd_eval (or cmd_test) across n_iterations (instead of cmd_train only).

The script was designed for broad hyperparam optimization but it's possible to run it for only 1 or 2 configs to assess the reproducibility of  a specific model or for FiLM vs no-FiLM. We could add a flag to call the test command when `cmd_train` has completed to cover this use case? What do you think?

charleygros:
> The script was designed for broad hyperparam optimization but it's possible to run it for only 1 or 2 configs to assess the reproducibility of a specific model or for FiLM vs no-FiLM.

Yes exactly, that is what I was thinking of too :-)

> We could add a flag to call the test command when `cmd_train` has completed to cover this use case

As the script is called `automate_training` --> we could either rename the file, something like `compare_models`, either have a different script dedicated to "testing comparison" (my preference). If we go with the second option, we could share some code between the two scripts as a lot of duplication.

> and compare test dice instead of validation dice
Yep absolutely --> if not too complicated, we could run the `cmd_eval` instead of the `cmd_test` since `cmd_eval` calls `cmd_test` if `pred_masks` folder does not exit yet then run the 3D metrics evaluation (so we will have more info than the dice) and output a csv file --> which should be friendly with your Panda dataframe approach.

Also, if not implemented yet, this user case would need to change the `random seed` for each iteration (but same seed for the 2 models during one iteration).

olix86:
Alright I've added the following : 
-  `--run-test` to run `cmd_eval` to get metrics on test dataset (p-value of test dice)
- `--fixed-split` which keeps the dataset split constant between iterations (default is same seed for all model in a given iteration, seed varies between iterations)

Question : 
- Currently using a two-sided T-test -> which answers  is X different than baseline U-Net?
-In our case the problem is mostly one-sided  i.e. is X better than baseline U-Net? So the outputted p-value should be divided by 2 to answer this question . Should I output p/2 directly?

charleygros:
> Alright I've added the following :
> 
>     * `--run-test` to run `cmd_eval` to get metrics on test dataset (p-value on test dice)
> 
>     * `--fixed-split` which keeps the dataset split constant between iterations (default is same seed for all model in a given iteration, seed varies between iterations)

Really really great!

> Question :
> 
>     * Currently using a two-sided T-test -> which answers  is model X different than baseline U-Net?
> 
>     * In our case the problem is mostly one-sided  i.e. is model X better than baseline U-Net? So the outputed p-value should be divided by 2 to answer this question . Should I output p/2 directly?

Not sure to understand the question --> You wrote the same question for the "currently" and "our case".

(On another note: I remember vaguely attending a course called GBM6125 --> given the sample size, should we consider a non-parametric test?)

jcohenadad:
> In our case the problem is mostly one-sided i.e. is model X better than baseline U-Net? So the outputed p-value should be divided by 2 to answer this question . Should I output p/2 directly?

i would just output p with a two-sided test and argue in the report "the two cases are statistically different"

dev/compare_models.py
```diff
@@ -0,0 +1,307 @@
...
+    keep = list(param_dict.keys())
+    keep.append("log_directory")
+    config_df = config_df[keep]
+
+    results_df = config_df.set_index('log_directory').join(results_df.set_index('log_directory'))
+    results_df = results_df.sort_values(by=['best_validation_loss'])
+
+    results_df.to_pickle("output_df.pkl")
+
+    print("Detailled results")
```
jcohenadad: detailed

dev/compare_models.py
```diff
@@ -0,0 +1,307 @@
...
+    # Merge config and results in a df
+    config_df = pd.DataFrame.from_dict(config_list)
+    keep = list(param_dict.keys())
+    keep.append("log_directory")
+    config_df = config_df[keep]
+
+    results_df = config_df.set_index('log_directory').join(results_df.set_index('log_directory'))
+    results_df = results_df.sort_values(by=['best_validation_loss'])
+
+    results_df.to_pickle("output_df.pkl")
```
jcohenadad: can we avoid pkl output and prefer csv files which are human readable? (and could be opened by many more software)

dev/compare_models.py
```diff
@@ -0,0 +1,307 @@
+##############################################################
+#
+# This script enable training and comparison of models on multiple GPUs
```
jcohenadad: enables

dev/compare_models.py
```diff
@@ -0,0 +1,307 @@
...
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-c", "--config", required=True, help="Base config file path.")
+    parser.add_argument("-n", "--n-iterations", dest="n_iterations",
+                        type=int, help="Number of times to run each config .")
+    parser.add_argument("--all-combin", dest='all_combin', action='store_true',
+                        help="To run all combinations of config")
+    parser.add_argument("--run-test", dest='run_test', action='store_true',
+                        help="Run cmd_eval to get metrics on test dataset")
```
jcohenadad: not sure i understand this argument

dev/compare_models.py
```diff
@@ -0,0 +1,307 @@
...
+                test_df.set_index('log_directory'))
+            combined_df = combined_df.reset_index()
+
+        else:
+            combined_df = val_df
+
+        results_df = pd.concat([results_df, combined_df])
+
+    # Compute avg, std, p-values
+    if(n_iterations > 1):
```
jcohenadad: do you want to separate statistics calculation in another script? e.g. i can imaging that running all the combinations will take a lot of time, so in case you want to play with the statistics (e.g. add another test, try other threshold values, etc.) you would not have to re-run all the training but you would just input your pandas DF in your compute_statistics module. just an idea.

dev/compare_models.py
charleygros: What about:
```
Evaluate the trained model on the testing sub-set.
```

dev/compare_models.py
olix86: Yep good idea, will look into it

olix86:
Sample output
```
[5 rows x 12 columns]
Detailed results
            log_directory  batch_size  initial_lr  ...  best_validation_dice  best_validation_loss  test_dice
0     log_sc-batch_size=8           8       0.001  ...             -0.871973             -0.871973   0.711156
1     log_sc-batch_size=8           8       0.001  ...             -0.851089             -0.851089   0.741970
2  log_sc-initial_lr=0.01          18       0.010  ...             -0.827729             -0.827729   0.509227
3  log_sc-initial_lr=0.01          18       0.010  ...             -0.800095             -0.800095   0.525435

[4 rows x 8 columns]
Average dataframe
                        batch_size  initial_lr  best_training_dice  ...  best_validation_dice  best_validation_loss  test_dice
log_directory                                                       ...                                                       
log_sc-batch_size=8              8       0.001           -0.472531  ...             -0.861531             -0.861531   0.726563
log_sc-initial_lr=0.01          18       0.010           -0.439524  ...             -0.813912             -0.813912   0.517331

[2 rows x 7 columns]
Standard deviation dataframe
                        batch_size  initial_lr  best_training_dice  ...  best_validation_dice  best_validation_loss  test_dice
log_directory                                                       ...                                                       
log_sc-batch_size=8            0.0         0.0            0.005822  ...              0.014767              0.014767   0.021789
log_sc-initial_lr=0.01         0.0         0.0            0.028881  ...              0.019540              0.019540   0.011461

[2 rows x 7 columns]
P-values array
                        log_sc-batch_size=8  log_sc-initial_lr=0.01
log_sc-batch_size=8                1.000000                0.006851
log_sc-initial_lr=0.01             0.006851                1.000000

```

dev/compare_models.py
```diff
@@ -305,7 +302,7 @@ def compute_statistics(results_df, n_iterations):
     results_df = results_df.reset_index()
     results_df = results_df.sort_values(by=['best_validation_loss'])
 
-    results_df.to_csv("output_df.csv")
+    results_df.to_csv("detailled_results.csv")
```
charleygros: Detailed?

dev/automate_training.py
```diff
@@ -29,48 +34,80 @@
...
+
+    # Use GPU i from the array specified in the config file
+    config["gpu"] = config["gpu"][ID]
+
+    # Call ivado cmd_eval
+    try:
+        # Save best test score
+
+        config["command"] = "eval"
+        test_dict, eval_df = ivado.cmd_eval(config)
```
charleygros: What about using the 3D dice from `eval_df`?

dev/compare_models.py
olix86: I named `detailled_results` because it has the results for all iterations, I thought `output_df` was too generic since avg, std and p-values are also outputs, open to suggestions :)

dev/automate_training.py
olix86: Sorry if it's stupid question, but wouldn't it give the same result?

olix86:
Note : added a wiki page for this PR 
https://github.com/neuropoly/ivado-medical-imaging/wiki/Automate-training-and-model-comparison

dev/compare_models.py
jcohenadad: i think Charley's suggestion was about the typo: detailled --> detailed

jcohenadad:
> Note : added a wiki page for this PR
> https://github.com/neuropoly/ivado-medical-imaging/wiki/Automate-training-and-model-comparison

great! i just slightly changed the name for the more explicit [Pipeline for training and model comparison](https://github.com/neuropoly/ivado-medical-imaging/wiki/pipeline-training-and-model-comparison) (this is PR is de facto about a pipeline, which happens to be automated)

dev/compare_models.py
olix86: Oups yep thanks!

dev/automate_training.py
charleygros: Not necessary, but probably close indeed --> computed by batch vs computed by subjects.

dev/automate_training.py
```diff
@@ -88,6 +87,10 @@ def test_worker(config):
         config["command"] = "eval"
         test_dict, eval_df = ivado.cmd_eval(config)
         test_dice = test_dict['dice_score']
+
+        # Uncomment to use 3D dice
+        # test_dice = eval_df["dice"].mean()
```
charleygros: Great

PR #160, opened 2020-03-16
Multiclass segmentation

andreanne-lemay:
This PR adapts the code to allow multiclass segmentation.
- [x] Modification to loader (`target_suffix`  and label data become lists)
- [x] Modify model to specify number of classes
- [x] Adapt transformations to loader
- [x] Implement and test different multi class losses
- [x] Modify metrics to compute dice score for each label
- [x] Reconstruct prediction during testing
- [x] Adapt eval to multiclass
- [x] Test for 2D models
- [x] Add test
 
Related medical torch PR : https://github.com/neuropoly/medicaltorch/pull/17


ivadomed/utils.py
```diff
@@ -745,3 +809,95 @@ def save_feature_map(batch, layer_name, context, model, test_input, slice_axis):
...
+    n_classes = len(pil_list)
+    w, h = pil_list[0].size
+    numpy_array = np.zeros((w, h, n_classes))
+    for idx, pil_img in enumerate(pil_list):
+        numpy_array[..., idx] = np.array(pil_img)
+    return numpy_array
+
+
+def convert_labels_to_RGB(grid_img):
+    # Keep always the same color labels
```
andreanne-lemay: image
This is necessary to have a decent view if the number of classes is not 1 or 3.

ivadomed/utils.py
```diff
@@ -745,3 +809,95 @@ def save_feature_map(batch, layer_name, context, model, test_input, slice_axis):
         nib_pred = nib.Nifti1Image(attention_map, nib_ref.affine)
 
         nib.save(nib_pred, save_directory)
+
+
+def save_color_labels(gt_data, binarize, gt_filename, output_filename, slice_axis):
+    rdict = {}
```
andreanne-lemay: Saves .nii.gz with color labels which helps identify the different labels compared to grayscales (especially if there is a lot of labels)
image

ivadomed/main.py
```diff
@@ -180,14 +179,16 @@ def cmd_train(context):
     if context['retrain_model'] is None:
         if HeMIS:
             model = models.HeMISUnet(modalities=context['contrast_train_validation'],
+                                     out_channel=len(context["target_suffix"]) + 1,
```
olix86: Why +1?

config/config.json
```diff
@@ -45,6 +45,53 @@
         "Resample": {
             "wspace": 0.75,
             "hspace": 0.75
+<<<<<<< HEAD
```
olix86: Merge conflict

ivadomed/utils.py
olix86: Really cool! :smiley:

ivadomed/main.py
andreanne-lemay: It is for the background class since I use softmax I need this extra output I'll leave a comment in the code :)

config/config.json
andreanne-lemay: Oupsie Forgot this one thanks!

ivadomed/loader.py
```diff
@@ -98,15 +97,18 @@ def __init__(self, root_dir, subject_lst, target_suffix, contrast_lst, contrast_
...
                 derivatives = subject.get_derivatives("labels")
-                target_filename, roi_filename = None, None
+                target_filename, roi_filename = [None] * len(target_suffix), None
 
                 for deriv in derivatives:
-                    if deriv.endswith(subject.record["modality"] + target_suffix + ".nii.gz"):
-                        target_filename = deriv
-                    if not (roi_suffix is None) and deriv.endswith(subject.record["modality"] + roi_suffix + ".nii.gz"):
-                        roi_filename = deriv
+                    for idx, suffix in enumerate(target_suffix):
```
charleygros: super

ivadomed/losses.py
```diff
@@ -1,16 +1,54 @@
...
+            intersection = (iflat * tflat).sum()
+            dice_per_class += intersection / (iflat.sum() + tflat.sum() + smooth)
+        return -(2.0 * dice_per_class) / n_classes
+
+
+class CombinedDiceLoss(nn.Module):
+    """
+    :param classes_of_interest: list containing the index of a class which dice will be added to the loss.
+    :param multiclass_dice: boolean representing whether or not multiclass dice is included in the current loss.
+    :param dice: boolean representing whether or not dice is included in the current loss.
```
charleygros: My understanding is that these three params are mutually exclusive, is it correct? eg [XX], False, False OR [], True, False OR [], False, True.
Could we send only one param instead?eg  [0] for monoclass, [0, 2] for class 1 and 2, [0, 1, 2] for all classes.

ivadomed/losses.py
```diff
@@ -1,16 +1,54 @@
...
+        self.multi_class_dice = multi_class_dice
+        self.dice = dice
+
+    def forward(self, prediction, target):
+        loss = torch.zeros(1).cuda()
+        if len(self.class_of_interest):
+            for idx in self.class_of_interest:
+                loss += - dice_loss(prediction[:, idx, ], target[:, idx, ])
+        if self.multi_class_dice:
+            loss += MultiClassDiceLoss()(prediction, target)
```
charleygros: Between these two cases: I can see only one difference: the `/ n_classes` in `MultiClassDiceLoss` --> is it correct?
Just wondering if we could combine all cases in only one loop (cf above comment).
--> This class would be the one to call for both mono and multi label tasks using the Dice Loss

ivadomed/losses.py
```diff
@@ -1,16 +1,54 @@
...
+        return loss / (len(self.class_of_interest) + self.dice + self.multi_class_dice)
 
 
 def dice_loss(input, target):
     smooth = 1.0
 
-    iflat = input.view(-1)
-    tflat = target.view(-1)
+    iflat = input.reshape(-1)
+    tflat = target.reshape(-1)
```
charleygros: I was just reading: https://stackoverflow.com/questions/49643225/whats-the-difference-between-reshape-and-view-in-pytorch
Could you comment why you changed? I am not saying this is wrong, just curious :)

ivadomed/models.py
```diff
@@ -147,6 +147,8 @@ def forward(self, features, context=None, w_film=None):
             x, w_film = self.last_film(x, context, w_film)
         if self.out_channel > 1:
             preds = F.softmax(x, dim=1)
+            # Remove background class
+            preds = preds[:, 1:, ]
```
charleygros: super!

ivadomed/transforms.py
```diff
@@ -122,11 +123,15 @@ def __call__(self, sample):
...
+            rdict['gt'] = []
+            for gt in gt_data:
+                rdict['gt'].append(self.resample_bin(gt, wshape_new, hshape_new))
 
         if sample['roi'] is not None:
             roi_data = sample['roi']
-            rdict['roi'] = self.resample_bin(roi_data, wshape_new, hshape_new, 0.0)
+            rdict['roi'] = []
+            for roi in roi_data:
+                rdict['roi'] = self.resample_bin(roi, wshape_new, hshape_new, 0.0)
```
charleygros: I thought `roi_data` was only one element. Different from `gt_data` which is now a list.

ivadomed/transforms.py
```diff
@@ -423,6 +433,19 @@ def undo_transform(sample):
         return sample
 
 
+class BackgroundClass(mt_transforms.MTTransform):
```
charleygros: Great!
Question --> do you exclude this class when calling `CombinedDiceLoss` or rather just after the softmax?

ivadomed/transforms.py
```diff
@@ -447,11 +470,13 @@ def __call__(self, sample):
...
-                pil_img = Image.fromarray(gt, mode='F')
-                gt_vol[idx, :, :] = np.array(self.sample_augment(pil_img, params))
-            ret_gt = gt_vol.astype('float32')
+            ret_gt = []
+            for labels in gt_data:
+                gt_vol = np.zeros(labels.shape)
+                for idx, gt in enumerate(labels):
+                    pil_img = Image.fromarray(gt, mode='F')
+                    gt_vol[idx, :, :] = np.array(self.sample_augment(pil_img, params))
+                ret_gt.append(gt_vol.astype('float32'))
```
charleygros: Do you need `float32`?

ivadomed/utils.py
```diff
@@ -61,7 +62,16 @@ def __init__(self, fname_pred, fname_gt, params={}):
...
+
+        h, w, d = self.data_pred.shape[:3]
+        self.n_classes = len(self.fname_gt)
+        self.data_gt = np.zeros((h, w, d, self.n_classes))
+
+        for idx, file in enumerate(fname_gt):
+            if os.path.exists(file):
+                self.data_gt[..., idx] = self.get_data(file)
+            else:
+                self.data_gt[..., idx] = np.zeros((h, w, d), dtype='u1')
```
charleygros: In which case(s) are we falling here?

ivadomed/utils.py
```diff
@@ -122,7 +141,7 @@ def __init__(self, fname_pred, fname_gt, params={}):
 
     def get_data(self, fname):
         nib_im = nib.load(fname)
-        return nib_im.get_data()
+        return nib_im.get_data().astype('u1')
```
charleygros: super!

ivadomed/utils.py
```diff
@@ -562,6 +586,40 @@ def structureWise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
     nib.save(nib_avgUnc, fname_avgUnc)
 
 
+def multiclass_dice_score(im1, im2):
+    im1 = np.asarray(im1[0, ]).astype(np.bool)
+    im2 = np.asarray(im2[0, ]).astype(np.bool)
```
charleygros: What are we extracting here? not sure to follow you re the first index.

ivadomed/utils.py
```diff
@@ -562,6 +586,40 @@ def structureWise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+
+    dice_per_class = 0
+    n_classes = im1.shape[0]
```
charleygros: I thought the "class dimension" was the last dimension --> are you doing some reshaping at some point?

ivadomed/utils.py
```diff
@@ -562,6 +586,40 @@ def structureWise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    n = 0
+    for i in range(n_classes):
+        intersection = np.logical_and(im1[i, ], im2[i, ])
+
+        # Verify that sum is not null
```
charleygros: super

ivadomed/utils.py
```diff
@@ -562,6 +586,40 @@ def structureWise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+        if im1[i, ].sum() + im2[i, ].sum():
+            dice_per_class += intersection.sum() / (im1[i, ].sum() + im2[i, ].sum())
+            n += 1
+    if not n:
+        n = 1
+    return (2.0 * dice_per_class) / n
```
charleygros: same annoying question than above, sorry: why not looping and calling `dice_score`.

ivadomed/utils.py
```diff
@@ -562,6 +586,40 @@ def structureWise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+
+
+def tumor_dice_score(im1, im2):
```
charleygros: could you call `dice_score` by sending `im1[0,]` instead of `im1`, same for `im2`?

ivadomed/utils.py
```diff
@@ -562,6 +586,40 @@ def structureWise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    n = 0
+    for i in range(n_classes):
+        intersection = np.logical_and(im1[i, ], im2[i, ])
+
+        # Verify that sum is not null
+        if im1[i, ].sum() + im2[i, ].sum():
+            dice_per_class += intersection.sum() / (im1[i, ].sum() + im2[i, ].sum())
+            n += 1
+    if not n:
+        n = 1
```
charleygros: Not sure, could you explain please?

ivadomed/utils.py
```diff
@@ -581,6 +639,10 @@ def dice_score(im1, im2):
 
 
 def hausdorff_score(prediction, groundtruth):
+    if len(prediction.shape) == 4:
+        l, h, d, w = prediction.shape
+        prediction = prediction.reshape((h, l * d, w))
+        groundtruth = groundtruth.reshape((h, l * d, w))
```
charleygros: Not sure here, could you please explain?

ivadomed/main.py
```diff
@@ -244,7 +252,8 @@ def cmd_train(context):
     var_contrast_list = []
 
     # Loss
-    if context["loss"]["name"] in ["dice", "cross_entropy", "focal", "gdl", "focal_dice"]:
+    if context["loss"]["name"] in ["dice", "cross_entropy", "focal", "gdl", "focal_dice", "multi_class_dice",
+                                   "combined_dice"]:
```
charleygros: cf comment about dice/multiclass/combined

ivadomed/main.py
```diff
@@ -286,7 +303,9 @@ def cmd_train(context):
     epsilon = context["early_stopping_epsilon"]
     val_losses = []
 
-    metric_fns = [dice_score,  # from ivadomed/utils.py
+    metric_fns = [multiclass_dice_score,
+                  tumor_dice_score,
+                  dice_score,  # from ivadomed/utils.py
```
charleygros: cf below comment

ivadomed/main.py
```diff
@@ -364,41 +384,7 @@ def cmd_train(context):
...
-                    grid_img = vutils.make_grid(preds.data.cpu(),
-                                                normalize=True,
-                                                scale_each=True)
-                    writer.add_image('Train/Predictions', grid_img, epoch)
-
-                    grid_img = vutils.make_grid(gt_samples,
-                                                normalize=True,
-                                                scale_each=True)
-                    writer.add_image('Train/Ground Truth', grid_img, epoch)
+                save_tensorboard_img(writer, epoch, "Train", input_samples, gt_samples, preds, unet_3D)
```
charleygros: youhou
refactoring en masse

ivadomed/main.py
```diff
@@ -738,62 +694,77 @@ def cmd_test(context):
...
+                                                 slice_axis=AXIS_DCT[context['slice_axis']],
+                                                 kernel_dim='2d',
+                                                 bin_thr=0.5 if context["binarize_prediction"] else -1)
+
+                        if output_nii.get_data().shape[-1] > 1:
+                            save_color_labels(output_nii.get_data(),
+                                              context["binarize_prediction"],
+                                              fname_tmp,
+                                              fname_pred.split(".nii.gz")[0] + '_color.nii.gz',
+                                              AXIS_DCT[context['slice_axis']])
```
charleygros: In the case of multiclass --> is the prediction saved as one or multiple files? If one file --> it will be seen with different color when using fsleyes & cie anyway, no? same for multiple files. I may miss something :(

ivadomed/losses.py
andreanne-lemay: You can have all three at the same time: [xx], True, True -> yes, one param would be best I'll change that!

andreanne-lemay: I changed it for a list of the params, I think it might affect clarity? Tell me what you think!

ivadomed/losses.py
andreanne-lemay: I think I understand what you mean! I optimized MultiClassDiceLoss and CombinedDiceLoss -> tell me if you had something else in mind

ivadomed/losses.py
andreanne-lemay: view only supports contiguous data, so I had errors when using MultiClassDiceLoss because the way I used it, data was not always contiguous. I also had the error on the HeMIS branch. Since then I made changes and the data is always contiguous, but I see no disadvantage to use reshape since it supports contiguous and discontiguous data :). Should I add comment in the code as well?

ivadomed/transforms.py
andreanne-lemay: Yes you are totally right, but the thing is in medicaltorch a this line https://github.com/neuropoly/medicaltorch/blob/master/medicaltorch/datasets.py#L257, we treat roi_filename the same as gt_filename (line 260), so roi_filename need to be a list since I modified SegmentationPair2D to deal with list instead of a single file.

ivadomed/transforms.py
andreanne-lemay: I don't use it anymore since I exclude the background class in the models. I let it in this PR, because I thought it might be useful in the future if we want to include the background as a class and measure different metrics on it. But since I don't use it currently, it might pollute the transform files, opinion?

ivadomed/utils.py
andreanne-lemay: In case, the path doesn't exist. For instance, not all my subject has edema and cavity. If does paths don't exist for a specific user, the gt should be an empty array

ivadomed/utils.py
andreanne-lemay: Modified it :)!

ivadomed/utils.py
andreanne-lemay: This shouldn't be here -> I modified it

ivadomed/utils.py
andreanne-lemay: The shape format is batch_size, n_classes, height, depth, width. There used to be a squeeze before metrics which made the class number be in first place if the batch_size was 1, but I removed it. n_classes should be in im1.shape[1]! Thanks for that!

ivadomed/utils.py
andreanne-lemay: I removed this score, I used it only for debugging and forgot to take it off :-1:

ivadomed/utils.py
andreanne-lemay: Did a refactor :) !

ivadomed/utils.py
andreanne-lemay: I added a comment to explain! Tell me if you need more explanation

ivadomed/main.py
andreanne-lemay: Ok made the change, but it requires to add another parameter `dice_params` in the config files. Let me know if that's what you had in mind :)

ivadomed/main.py
andreanne-lemay: The labels are saved as one file. If the number of class is different than 3 (which can be seen as an RGB image only if there is 3 channels in last dimension), the file cannot be seen (to the best of my knowledge) as RGB. For instance, I tried with 2 labels only (tumor and edema), the last dimension is 2 -> I cannot use the RGB view option.I can only assume the same if there were 4 or more labels since the viewer cannot interprete what color you want on an image not encoded in RGB. For multiple files, a few manual manipulations are needed to see each labels in different color. I find it more convenient to see my result as one file :). Does it make sense?

ivadomed/main.py
```diff
@@ -836,8 +795,10 @@ def cmd_eval(context):
...
         fname_pred = os.path.join(path_pred, subj_acq + '_pred.nii.gz')
-        fname_gt = os.path.join(context['bids_path'], 'derivatives', 'labels', subj, 'anat',
-                                subj_acq + context['target_suffix'] + '.nii.gz')
+        fname_gt = []
+        for suffix in context['target_suffix']:
+            fname_gt.append(os.path.join(context['bids_path'], 'derivatives', 'labels', subj, 'anat',
+                                         subj_acq + suffix + '.nii.gz'))
 
         # 3D evaluation
         eval = Evaluation3DMetrics(fname_pred=fname_pred,
```
andreanne-lemay: in cmd_test why return metrics_dict? The variable is not always assigned a this point causing an error

ivadomed/main.py
charleygros: Sorry I am missing something: why for?
I just had in mind to call dice --> loop across a list of one or more elements.

ivadomed/main.py
charleygros: I am not sure to know which RGB view option you are referring to.
I was thinking of something like:
```
cd spinalcordtoolbox/data/PAM50/template/
fsleyes PAM50_levels.nii.gz -cm subcortical
```
or another colormap.

ivadomed/main.py
charleygros: I agree!
I think it has been added for this application: https://github.com/neuropoly/ivado-medical-imaging/blob/83a7b58d8458393b04dbe2f9c876221d95471db6/dev/automate_training.py#L88

ivadomed/losses.py
```diff
@@ -1,16 +1,59 @@
...
 
 
-def dice_loss(input, target):
+# Inspired from https://arxiv.org/pdf/1802.10508.pdf
+def multi_class_loss(prediction, target, classes_of_interest=None):
+    n_classes = prediction.shape[1]
+    if classes_of_interest is None:
+        classes_of_interest = range(n_classes)
+    elif not len(classes_of_interest):
+        return 0.0
```
charleygros: Should we print a warning message here?
Why are you returning zero?

ivadomed/losses.py
charleygros: Yes sure :-)
Many thanks

ivadomed/losses.py
```diff
@@ -1,16 +1,59 @@
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
 
 
-def dice_loss(input, target):
+# Inspired from https://arxiv.org/pdf/1802.10508.pdf
+def multi_class_loss(prediction, target, classes_of_interest=None):
```
charleygros: `multi_class_diceLoss`?

ivadomed/losses.py
```diff
@@ -1,16 +1,59 @@
...
+    :return: loss
+    """
+    if params is None:
+        classes_of_interest, multi_class_dice, dice = [], False, True
+    else:
+        classes_of_interest, multi_class_dice, dice = params
+
+    loss = multi_class_loss(prediction, target, classes_of_interest) + \
+           multi_class_dice * multi_class_loss(prediction, target) + \
+           dice * dice_function(prediction, target)
```
charleygros: What about: use only `multi_class_loss`?
dice_function --> [0, 1]
multi_class_loss --> [0, 1, 2, 3]
selected class --> [0, 2, 3]

ivadomed/losses.py
```diff
@@ -1,16 +1,59 @@
...
+    :param prediction: torch tensor generated as the model's prediction
+    :param target: torch tensor representing the ground truth
+    :param params: list of the losses considered for this loss
+    first item in list: represents classes of interest, list containing the index of a class which dice will be added to
+                        the loss.
+    second item in list: represents multi_class_dice, boolean representing whether or not multiclass dice is included in
+                         the current loss.
+    third item in list: represents dice, boolean representing whether or not dice is included in the current loss.
+
+    e.i: [[0, 1], True, False]
```
charleygros: If we send `[0, 1]` --> will it compute the dice of the background class?

ivadomed/transforms.py
charleygros: Mmmmm so true! Perfect

ivadomed/transforms.py
charleygros: Yes, I would agree! There is no planet B.

ivadomed/losses.py
charleygros: Oooo --> I have just realised [0, 1] refers to the two first classes as you remove the background after the softmax, correct?

ivadomed/transforms.py
charleygros: I thought that Anthime modified the loader / transforms so that the gt are not float32. Good to know!
Yep, could do this during medicaltorch-ivadomed fusion :-)

ivadomed/main.py
charleygros: But a very good application for the way you implemented it --> is to be able to manipulate it directly in Python / matplotlib etc which could be very handy. Same idea with [data_painted](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/utils.py#L265) from eval.

ivadomed/losses.py
```diff
@@ -3,50 +3,28 @@
...
-
-    e.i: [[0, 1], True, False]
-    :return: loss
+    :param classes_of_interest:  list containing the index of a class which dice will be added to the loss.
     """
-    if params is None:
-        classes_of_interest, multi_class_dice, dice = [], False, True
-    else:
-        classes_of_interest, multi_class_dice, dice = params
+    def __init__(self, classes_of_interest):
```
charleygros, line 10: Great!

ivadomed/transforms.py
```diff
@@ -129,7 +134,7 @@ def __call__(self, sample):
             roi_data = sample['roi']
             rdict['roi'] = []
             for roi in roi_data:
-                rdict['roi'] = self.resample_bin(roi, wshape_new, hshape_new, 0.0)
+                rdict['roi'].append(self.resample_bin(roi, wshape_new, hshape_new, 0.0))
```
charleygros, line 137: The `thr` of `resample_bin` is not used anymore.
We may want to binarise during the "do transform" to keep a binary GT, but not mandatory, to discuss and check with the loss functions.
We do not want to binarise during the "undo transform".

ivadomed/transforms.py
```diff
@@ -474,7 +486,7 @@ def __call__(self, sample):
                 for idx, gt in enumerate(labels):
                     pil_img = Image.fromarray(gt, mode='F')
                     gt_vol[idx, :, :] = np.array(self.sample_augment(pil_img, params))
-                ret_gt.append(gt_vol.astype('float32'))
```
charleygros: You may have corrected this in a latter commit --> but we want to keep float32 when using this transform because we are modifying the GT: GT_voxels=1 and new surrounding voxels € ]0, 1[.

ivadomed/losses.py
```diff
@@ -21,7 +21,7 @@ def forward(self, prediction, target):
         for i in self.classes_of_interest:
             dice_per_class += dice_loss(prediction[:, i, ], target[:, i, ])
 
-        return dice_per_class / len(self.classes_of_interest)
+        return - dice_per_class / len(self.classes_of_interest)
```
charleygros, line 24: Would it be okay to keep it positive in the return?
Can't remember exactly why but we decided this with @olix86 last Autumn --> we can re-disucss of course if needed :)
I think it was a "convention" across the loss, so that they can be combined knowing they are positive.

andreanne-lemay: This would require that I add a `if` in the code (if multi_class_dice), because the other loss functions don't require a negative

ivadomed/transforms.py
andreanne-lemay: Yes, it was corrected :)

ivadomed/losses.py
```diff
@@ -2,14 +2,36 @@
...
+
+    def forward(self, prediction, target):
+        dice_per_class = 0
+        n_classes = prediction.shape[1]
+
+        if self.classes_of_interest is None:
+            self.classes_of_interest = range(n_classes)
+
+        for i in self.classes_of_interest:
+            dice_per_class += dice_loss(prediction[:, i, ], target[:, i, ])
```
lrouhier, line 22: Does `prediction[:, i, ]` really works ? (i am a bit surprised about the empty third index) :)

ivadomed/main.py
```diff
@@ -274,12 +281,15 @@ def cmd_train(context):
...
+        elif context["loss"]["name"] == "multi_class_dice":
+            loss_fct = losses.MultiClassDiceLoss(classes_of_interest=context["loss"]["params"]["classes_of_interest"])
 
         if not context["loss"]["name"].startswith("focal"):
             print("\nLoss function: {}.\n".format(context["loss"]["name"]))
 
     else:
-        print("Unknown Loss function, please choose between 'dice', 'focal', 'focal_dice', 'gdl' or 'cross_entropy'")
+        print("Unknown Loss function, please choose between 'dice', 'focal', 'focal_dice', 'gdl', 'cross_entropy',"
+              "'multi_class_dice' or 'combined_dice'")
```
lrouhier, line 292: 'combined dice' does not appear in the option line 260, and I don't see it in the previous `if...elif` statement. Unless this is implemented in another loss function (multiclass) ?

andreanne-lemay: Yes forgot to remove it, good catch thanks :)

PR #162, opened 2020-03-19
Merge medicaltorch into ivadomed 1 of 3

olix86:
We only import from `medicaltorch` the following : `datasets, transforms, filters, metrics` so no need for  `losses, models`

Plan :

- PR 1 (this PR)
  - `filters`  -> contains only `SliceFilter` funtions, moved into `utils.py`
  - `metrics`  -> We have only reimplemented dice and hausdorff in `utils.py`  so moved in `metrics.py` and use the `medicaltorch` version for the rest
- PR 2
  - `transforms` -> combine with our transforms, so no need to override
- PR 3
  - `datasets` -> idem, combine with our datasets MRI2D / 3D in `loader.py`

Done in this PR : 
- Remove lots of unused imports
- Incorporate `filters` and `metrics`
- remove wildcard imports (`from ivadomed.utils import *`) since they make it impossible to know which function comes from where

Todo in next PRs: 
- incorporate `datasets` and `transforms`

### DIFF
--- a/dev/class_balance.py
+++ b/dev/class_balance.py
@@ -10,7 +10,6 @@
 #
 ##############################################################
 
-import os
 import json
 import argparse
 import numpy as np
--- a/testing/test_inference.py
+++ b/testing/test_inference.py
@@ -91,7 +91,7 @@ def test_inference(film_bool=False):
     model.eval()
 
     metric_fns = [metrics.dice_score,  # from ivadomed/utils.py
-                  metrics.hausdorff_score,
+                  metrics.hausdorff_2D_score,
                   metrics.precision_score,
                   metrics.recall_score,
                   metrics.specificity_score,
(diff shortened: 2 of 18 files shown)


dev/class_balance.py
```diff
@@ -10,7 +10,6 @@
 #
 ##############################################################
 
-import os
```
charleygros: Not used?

olix86: Indeed 👍

testing/test_inference.py
```diff
@@ -91,7 +91,7 @@ def test_inference(film_bool=False):
     model.eval()
 
     metric_fns = [metrics.dice_score,  # from ivadomed/utils.py
-                  metrics.hausdorff_score,
+                  metrics.hausdorff_2D_score,
```
charleygros: 3D? or 2D okay here?

olix86: I was discussing that with @andreanne-lemay in theory 3D should work everywhere since it fallbacks to 2D when it received a not 3D input

PR #169, opened 2020-03-22
segment_volume(): folder_model instead of (fname_model, fname_model_metadata).

### LINKED ISSUE #166: segment_volume(): couldn't fname_model and fname_model_metadata be merged?
Training of a model does require a json metadata file. Which means: one model necessarily has a json file associated with it.

The function `utils.segment_volume()` does as for two separate arguments: `fname_model` and `fname_model_metadata`, which is source of confusion and error. Why not packaging a model with its associated json file within a single folder, so callers can simply point to the folder instead of the two separate files?

charleygros:
Instead of `fname_model, fname_model_metadata`, `segment_volume()` needs `folder_model` as argument which contains both the model and the related metadata, `folder_model/model.pt` and `folder_model/model_metadata.json` resp.

Fixes #166

To test:
```
from ivadomed.utils import segment_volume
n = segment_volume(folder_model='../duke/projects/ivado-medical-imaging/github_PR155/model_large_t2star', fname_image='../duke/projects/ivado-medical-imaging/spineGeneric_202002200220/result/sub-queensland04/anat/sub-queensland04_T2star.nii.gz')
```

### DIFF
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -643,8 +643,7 @@ def threshold_predictions(predictions, thr=0.5):
 def segment_volume(folder_model, fname_image, fname_roi=None):
     """Segment an image.
 
-    Segment an image (fname_image) using a already trained model given its
-    training parameters (both in folder_model). If provided, a RegionOfInterest (fname_roi)
+    Segment an image (fname_image) using a pre-trained model (folder_model). If provided, a region of interest (fname_roi)
     is used to crop the image prior to segment it.
 
     Args:
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -640,17 +640,18 @@ def threshold_predictions(predictions, thr=0.5):
     return thresholded_preds
 
 
-def segment_volume(fname_model, fname_model_metadata, fname_image, fname_roi=None):
+def segment_volume(folder_model, fname_image, fname_roi=None):
     """Segment an image.
 
-    Segment an image (fname_image) using a already trained model (fname_model) given its
-    training parameters (fname_model_metadata). If provided, a RegionOfInterest (fname_roi)
+    Segment an image (fname_image) using a already trained model given its
+    training parameters (both in folder_model). If provided, a RegionOfInterest (fname_roi)
     is used to crop the image prior to segment it.
 
     Args:
-        fname_model (string): model filename (.pt) to use.
-        fname_model_metadata (string): Configuration file filename (.json) used for the training,
-            see https://github.com/neuropoly/ivado-medical-imaging/wiki/configuration-file
+        folder_model (string): foldername which contains
+            (1) the model ('model.pt') to use
+            (2) its configuration file ('model_metadata.json') used for the training,
+                see https://github.com/neuropoly/ivado-medical-imaging/wiki/configuration-file
         fname_image (string): image filename (e.g. .nii.gz) to segment.
         roi_fname (string): Binary image filename (e.g. .nii.gz) defining a region of interest,
             e.g. spinal cord centerline, used to crop the image prior to segment it if provided.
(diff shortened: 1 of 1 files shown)


ivadomed/utils.py
```diff
@@ -640,17 +640,18 @@ def threshold_predictions(predictions, thr=0.5):
     return thresholded_preds
 
 
-def segment_volume(fname_model, fname_model_metadata, fname_image, fname_roi=None):
+def segment_volume(folder_model, fname_image, fname_roi=None):
     """Segment an image.
 
-    Segment an image (fname_image) using a already trained model (fname_model) given its
-    training parameters (fname_model_metadata). If provided, a RegionOfInterest (fname_roi)
+    Segment an image (fname_image) using a already trained model given its
```
jcohenadad:
```suggestion
    Segment an image (fname_image) using a pre-trained model given its
```

ivadomed/utils.py
```diff
@@ -640,17 +640,18 @@ def threshold_predictions(predictions, thr=0.5):
...
+    training parameters (both in folder_model). If provided, a RegionOfInterest (fname_roi)
```
jcohenadad:
```suggestion
    Segment an image (fname_image) using a pre-trained model (folder_model). If provided, a region of interest (fname_roi)
```

ivadomed/utils.py
```diff
@@ -640,17 +640,18 @@ def threshold_predictions(predictions, thr=0.5):
...
     is used to crop the image prior to segment it.
 
     Args:
-        fname_model (string): model filename (.pt) to use.
-        fname_model_metadata (string): Configuration file filename (.json) used for the training,
-            see https://github.com/neuropoly/ivado-medical-imaging/wiki/configuration-file
+        folder_model (string): foldername which contains
+            (1) the model ('model.pt') to use
+            (2) its configuration file ('model_metadata.json') used for the training,
```
jcohenadad: I suggest to keep the same basename for the .json and .pt files (as done e.g. with the BIDS convention between the NIfTI binaries and the json metadata).
```suggestion
            (2) its configuration file ('model.json') used for the training,
```

ivadomed/utils.py
```diff
@@ -661,6 +662,21 @@ def segment_volume(fname_model, fname_model_metadata, fname_image, fname_roi=Non
     # Define device
     device = torch.device("cpu")
 
+    # Check if model folder exists
+    if os.path.isdir(folder_model):
+        # Check if model and model metadata exist
+        fname_model = os.path.join(folder_model, 'model.pt')
```
jcohenadad: is it an official convention in IVADOMED to have all models called "model.pt"? If it is called model_bob.pt, this line will fail. 

- If it is an official convention, i would remove the explicit file name here and move it to a separate class, e.g. `ivadomed.metadata.get_model()`, as done [in SCT](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/metadata.py#L191). All the appropriate i/o checks could be done in this dedicated module. We can address this refactoring in a separate PR though (and leave the current code as is for this PR).
- If it not an official convention, I would maybe suggest to name the model with the same basename as the folder: `model_bob/model_bob.pt`. This might prove useful in the future, if we manually copy models during debugging/testing, to avoid confusion (e.g. if all models are called "model.pt, robert might think he plays with the model from bob, but in fact it is the model from ariane. calling it model_bob.pt clarifies it). Again, inspired by the BIDS convention.

charleygros: It may be a good time to discuss the official convention in ivadomed :-)

I do like the Bob solution! So should we have `model_bob/model_bob.json` and `model_bob/model_bob.pt` for the metadata and the model, resp?

PR #178, opened 2020-03-31
Missing imports

### LINKED ISSUE #179: Imports or reference missing
In main:
-AXIS_DCT (should be utils.AXIS_DCT) 
(https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L7270)
(https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L764)
(https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L789)
-torch.nn (https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L211)
-pred_to_nib (should be utils.pred_to_nib)
(https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L760)
(https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L785)

In testing:
Conflict in `test_HeMIS` when `from ivadomed.utils import SliceFilter` with the import in loader (https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/loader.py#L3)
image (1)

-Need to `import torch` in the following: `test_inference`, `test_HeMIS`, `test_training`, `undo_transform`, `test_sampler`
-In test `test_inference` incorrect call of: pred_to_nib, IvadoMetricManager, threshold_predictions
-`import matplotlib.pyplot as plt` in test `test_undoTrans`

olix86:
Fix #179

### DIFF
--- a/ivadomed/loader.py
+++ b/ivadomed/loader.py
@@ -1,6 +1,7 @@
 from bids_neuropoly import bids
 from medicaltorch import datasets as mt_datasets
-from ivadomed.utils import SliceFilter
+from ivadomed import utils
+#from ivadomed.utils import SliceFilter
 from ivadomed import __path__
 
 from sklearn.preprocessing import OneHotEncoder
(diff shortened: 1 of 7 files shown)


andreanne-lemay:
In `test_inference` there is still the one_hot_encoder that is not defined, but it is not used -> I am not sure the best way to deal with this (why aren't we testing inference with FiLM?). If we don't want to test inference with FiLM we could just delete the parts with FiLM.

PR #180, opened 2020-03-31
Combine metric manager

olix86:
No need to override MetricManager with IvadoMetricManager anymore

### DIFF
--- a/ivadomed/main.py
+++ b/ivadomed/main.py
@@ -420,7 +420,7 @@ def cmd_train(context):
         val_loss_total, dice_val_loss_total = 0.0, 0.0
         num_steps = 0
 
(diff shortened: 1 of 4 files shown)


ivadomed/main.py
```diff
@@ -420,7 +420,7 @@ def cmd_train(context):
         val_loss_total, dice_val_loss_total = 0.0, 0.0
         num_steps = 0
 
-        metric_mgr = utils.IvadoMetricManager(metric_fns)
+        metric_mgr = metrics.MetricManager(metric_fns)
```
charleygros, line 423: imed_metrics?

charleygros:
@olix86 --> could you please add a import nibabel as nib in main? thx

PR #186, opened 2020-04-07
Fix travis pytest

### LINKED ISSUE #175: Travis tests not running
Unless I'm missing something huge, it looks like Travis tests are not working since +2 months. For some reason the `pytest` command ends up on the same like as the `echo` on the line before, so `pytest` is simply echoed and not executed.

Current situation :
image

Latest commit @AnBucquet on the travis config shows the same thing : [link](https://travis-ci.org/github/neuropoly/ivado-medical-imaging/builds/637506140)

olix86:
Fix #175
- Pytest is now failing but at least it's getting executed
- Added `travis_wait` to mitigate timeout


kousu:
Nice. That's a good hole to plug.

Some failures are due to using f-strings which aren't py3.5 compatible. But the default python on the current Ubuntu LTS is 3.6, and our installer sets python to 3.6, so I think we should just drop 3.5 support.

olix86:
Agreed about 3.5, however it still fails on 3.6/3.7 presumably because the tests use to much ressources?
```
ERROR testing/test_training_time.py - RuntimeError: [enforce fail at CPUAlloc...
ERROR testing/test_undoTrans.py - OSError: [Errno 12] Cannot allocate memory
```

kousu:
Travis probably has tight memory limits, which is fair. I don't know if there's a way to limit that. It looks like we're not the only ones running into it: https://github.com/pytorch/pytorch/issues/18626

Issue #183, opened 2020-04-02
Inconsistency between: Dice score and Dice loss


charleygros:
We often observe inconsistency between the Validation Dice loss during training and the Dice score on the testing set.
This inconsistency is mainly visible on the MS lesion dataset.

A very probable reason is the different way these two functions are dealing with empty GT and empty pred.
--> Dice loss returns 1 (thanks of the `smooth` param): https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/losses.py#L12
--> Dice score returns `np.nan`: https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/metrics.py#L64

A naive solution would be to call `dice_score` in `dice_loss` --> but the optimiser would not appreciate to receive `np.nan` (this is the reason why the two functions have been separated until now).

Any thoughts? :-)

charleygros:
Closed with #188

PR #184, opened 2020-04-06
Basic post-processing tools

charleygros:
Note: Please update your neuropoly/medicaltorch before testing this PR.

Implement basic post-processing tools.
Fixes #135.

New file `ivadomed/postprocessing.py`:
- `threshold_predictions()`: moved from `ivadomed.utils.py`. Note that we now return a data with int as type, instead of the input data type
- `keep_largest_object()`: works with 2 or 3d binary data
- `keep_largest_object_per_slice()`: loop across slices and call above function
- `fill_holes()`: works with 2 or 3d binary data
- `mask_predictions()`: eg to mask soft pred

Example:
```

from ivadomed.utils import segment_volume
import nibabel as nib
from ivadomed import postprocessing as imed_postPro

nib_soft=segment_volume('/home/GRAMES.POLYMTL.CA/chgroc/ivado-medical-imaging/models_uq_mice/model-sc_uq_mice', '/home/GRAMES.POLYMTL.CA/chgroc/duke/projects/ivado-medical-imaging/data_mice_abdullah-UQ/437/anat/437_t1.nii.gz')

data_soft = nib_soft.get_data()
nib.save(nib_soft, 'seg_soft.nii.gz')

data_thr = imed_postPro.threshold_predictions(data_soft, 0.2)
nib.save(nib.Nifti1Image(data_thr, nib_soft.affine), 'seg_hard.nii.gz')

data_rmBlobs = imed_postPro.keep_largest_object_per_slice(data_thr, axis=2)
nib.save(nib.Nifti1Image(data_rmBlobs, nib_soft.affine), 'seg_hard_rmBlobs.nii.gz')

data_soft_masked = imed_postPro.mask_predictions(data_soft, data_rmBlobs)
nib.save(nib.Nifti1Image(data_soft_masked, nib_soft.affine), 'seg_soft_masked.nii.gz')

data_holes_filled = imed_postPro.fill_holes(data_rmBlobs, (3, 3, 1))
nib.save(nib.Nifti1Image(data_soft_masked, nib_soft.affine), 'seg_hard_holesFilled.nii.gz')
```

PR #188, opened 2020-04-08
Soft dice score

charleygros:
This PR replaces the hard `dice_score` by a soft version of it.
This implementation accepts both hard and soft input masks.
This partially addresses the issue #183.

Also, a new parameter `empty_score` --> returned value when both input array are empty. The default value is `np.nan`.

Example:
```
>>> import numpy as np
>>> import nibabel as nib
>>> from ivadomed.metrics import dice_score
>>> 
>>> soft_1 = nib.load('seg_soft.nii.gz').get_data()
>>> soft_2 = nib.load('seg_soft_masked.nii.gz').get_data()
>>> dice_score(soft_1, soft_2)
0.983514115962626
>>> 
>>> hard_1 = np.zeros(soft_1.shape)
>>> hard_1[soft_1 > 1e-3] = 1
>>> hard_1[soft_1 <= 1e-3] = 0
>>> dice_score(hard_1, soft_1)
0.9595146937789737
>>> dice_score(hard_1, soft_2)
0.9504741207898071
>>> 
>>> empty = np.zeros(soft_1.shape)
>>> dice_score(empty, empty)
nan
>>> dice_score(empty, empty, empty_score=1.0)
1.0
```


jcohenadad:
About the proposition of setting a Dice score of 1 if both GT and predictions are empty: After googling around I found other groups using that approach:
- https://arxiv.org/pdf/1707.03237.pdf (see definition of DL2)
- https://www.kaggle.com/c/understanding_cloud_organization/discussion/114093
- https://stackoverflow.com/questions/47084179/how-to-calculate-multi-class-dice-coefficient-for-multiclass-image-segmentation

So i find it reasonable to give the option to set the output value to 1 (or nan or whatever) depending on the application.

charleygros:
Add `import nibabel as nib` in `ivadomed/main.py`.

jcohenadad:
> Add `import nibabel as nib` in `ivadomed/main.py`.

Because main.py is currently untouched in this PR, to prevent unnecessary conflicts, I suggest you rather add this line in this PR https://github.com/neuropoly/ivado-medical-imaging/pull/184

charleygros:
> > Add `import nibabel as nib` in `ivadomed/main.py`.
> 
> Because main.py is currently untouched in this PR, to prevent unnecessary conflicts, I suggest you rather add this line in this PR #184

Better practice indeed. Will do a new PR as I dont know when #184 will be merged.

PR #200, opened 2020-04-14
(WIP) Import nibabel in main.py

charleygros:
Import nibabel in main.py


charleygros:
Close because this line has been added in #132 .

Issue #203, opened 2020-04-15
Move prepdata to dev/ folder


jcohenadad:
In relation to https://github.com/neuropoly/ivado-medical-imaging/issues/195

Issue #195, opened 2020-04-13
Missing file when installing ivadomed via pip


jcohenadad:
When attempting to install ivadomed pointing to a specific branch in the [requirements.txt](https://github.com/neuropoly/spinalcordtoolbox/pull/2639/commits/4c4a5f6d8a0866614809c64b24d1f882a2501910), I bumped into the following error message:
~~~
error: could not create 'build/bdist.macosx-10.9-x86_64/wheel/ivadomed/metrics.py': No such file or directory
~~~

<details><summary>Full output</summary>

~~~
pip install -r requirements.txt
Requirement already satisfied: numpy in ./python/envs/venv_sct/lib/python3.6/site-packages (from -r requirements.txt (line 1)) (1.18.2)
Requirement already satisfied: cryptography in ./python/envs/venv_sct/lib/python3.6/site-packages (from -r requirements.txt (line 2)) (2.8)
Requirement already satisfied: dipy in ./python/envs/venv_sct/lib/python3.6/site-packages (from -r requirements.txt (line 3)) (1.1.1)
Requirement already satisfied: futures in ./python/envs/venv_sct/lib/python3.6/site-packages (from -r requirements.txt (line 4)) (3.1.1)
Requirement already satisfied: h5py in ./python/envs/venv_sct/lib/python3.6/site-packages (from -r requirements.txt (line 5)) (2.10.0)
Collecting ivadomed@ git+https://github.com/neuropoly/ivado-medical-imaging@cg/postPro-basics
  Cloning https://github.com/neuropoly/ivado-medical-imaging (to revision cg/postPro-basics) to /private/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/pip-install-31boj_sy/ivadomed
  Running command git clone -q https://github.com/neuropoly/ivado-medical-imaging /private/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/pip-install-31boj_sy/ivadomed
^[f^[b  Running command git checkout -b cg/postPro-basics --track origin/cg/postPro-basics
  Switched to a new branch 'cg/postPro-basics'
  Branch 'cg/postPro-basics' set up to track remote branch 'cg/postPro-basics' from 'origin'.
Requirement already satisfied: Keras==2.1.5 in ./python/envs/venv_sct/lib/python3.6/site-packages (from -r requirements.txt (line 7)) (2.1.5)
[...]

PR #204, opened 2020-04-16
prep_datadata => dev/prepare_data

kousu:
Addresses #203.

Besides just moving prepare_data, this has a second commit that adds a dev/bin/ folder to contain all the scripts (or symlinks to them). The problem in #195, as I understand it, was that we wanted a `prepdata` command on $PATH and the straightforward way to do that was to add a python `console_scripts` line; that works because virtualenv makes sure to put the special python `bin/` folder on $PATH, but caused unnecessary complication. This circumvents virtualenv by implementing only that one piece of its functionality.

Please consider this a suggestion. I can revert the second commit easily enough if this is too disruptive. There's more work to do before it's ready: I am unsure about the dependency on bash this adds, and if we were to really do this we should probably add symlinks for all the tools we use to dev/bin/, and make sure they all behave themselves when called from arbitrary working directories.


kousu:
I've added most of the dev scripts to this virtual environment, so now you do:

```
. dev/activate
```

Then you can run

```
(sct_venv) $ automate_training -h
usage: automate_training [-h] -c CONFIG [-n N_ITERATIONS] [--all-combin]
                         [--run-test] [--fixed-split]

optional arguments:
  -c CONFIG, --config CONFIG
                        Base config file path.
  -n N_ITERATIONS, --n-iterations N_ITERATIONS
                        Number of times to run each config .
  --all-combin          To run all combinations of config
  --run-test            Evaluate the trained model on the testing sub-set.
  --fixed-split         Keep a constant dataset split for all configs and
                        iterations
(sct_venv) $
(sct_venv) $ class_balance -h
usage: class_balance [-h] [-c C]

optional arguments:
  -h, --help  show this help message and exit
  -c C        Config file path.
(sct_venv) [kousu@requiem ivado-medical-imaging]$ comp
compare         compare_models  compgen         compile_et      complete        compopt         composite      
(sct_venv) $ 
(sct_venv) $ compare_models -h
usage: compare_models [-h] -df DATAFRAME -n N_ITERATIONS [--test-set]

optional arguments:
  -df DATAFRAME, --dataframe DATAFRAME
                        Path to saved dataframe (csv file).
  -n N_ITERATIONS, --n-iterations N_ITERATIONS
                        Number of times each config was run .
  --test-set            Evaluate the trained model on the testing sub-set
                        instead of validation.
```

etc.

Issue #205, opened 2020-04-17
Binarize image prior to applying morphomath operations


jcohenadad:
Some postprocessing methods, such as keep_large_objects or fill_holes only work on binary images. So if people want soft output while still benefiting from these algorithms, one solution is to call the thresholding with low value (eg 10e-3), apply morphomath operation, and then use the mask_prediction function to remove the small objects from the soft pred based on the binary output of the morphomath process.

charleygros:
Fixed with #206

PR #206, opened 2020-04-17
Make morphomath postprocessing compatible with soft predictions

charleygros:
Following #205, we implement here a solution to make morphomath postprocessing compatible with soft predictions.

To do so, we call the thresholding with low thr (1e-3), run the morphomath on the binary array, mask the soft input with the binary array output by the morphomath operation.

It also implement tests specific to soft inputs.


testing/test_postprocessing.py
```diff
@@ -64,45 +64,54 @@ def test_threshold(nii_seg):
...
     # Make sure it works with nibabel input
     nii_seg_proc = postproc.keep_largest_object(nii_seg)
     assert isinstance(nii_seg_proc, nib.nifti1.Nifti1Image)
     assert nii_seg_proc.dataobj[coord] == 0
 
 
-@pytest.mark.parametrize('nii_seg', [nii_dummy_seg()])
+@pytest.mark.parametrize('nii_seg', [nii_dummy_seg(), nii_dummy_seg(softseg=True)])
 def test_keep_largest_object_per_slice(nii_seg):
+    nii_seg_copy = np.copy(nii_seg)
```
jcohenadad, line 87: this copy doesn't seem to be used later on

testing/test_postprocessing.py
```diff
@@ -64,45 +64,54 @@ def test_threshold(nii_seg):
...
     # Set a voxel to 1 at the corner to make sure it is set to 0 by the function
     coord = (1, 1, 1)
     nii_seg.dataobj[coord] = 1
     # Test function with array input
     arr_seg_proc = postproc.keep_largest_object_per_slice(np.copy(np.asanyarray(nii_seg.dataobj)), axis=2)
     assert isinstance(arr_seg_proc, np.ndarray)
     assert arr_seg_proc[coord] == 0
+    # Make sure it equals the input data, in particular: still binary / soft if the input was binary / soft
+    assert np.array_equal(nii_seg.dataobj, arr_seg_proc)
```
jcohenadad, line 96: not sure i understand this assert: the purpose of preprocessing is to *modify* the data, so we should not expect the input to equal the output. Or am i missing something?  Maybe you wanted to test if a soft input produces a a soft output (ditto for binary input/output)

ivadomed/postprocessing.py
```diff
@@ -41,37 +41,47 @@ def threshold_predictions(predictions, thr=0.5):
...
-        predictions (array or nibabel object): Input binary segmentation. Image could be 2D or 3D.
+        predictions (array or nibabel object): Input segmentation.
+            Image could be 2D or 3D, soft or binary.
     Returns:
         Array or nibabel (same object as the input).
     """
-    assert np.array_equal(predictions, predictions.astype(bool))
     predictions_proc = np.copy(predictions)
+    # If input is not binary, then make it binary by thresholding it
+    if not np.array_equal(predictions, predictions.astype(bool)):
```
jcohenadad, line 56: I believe you can use the same trick as for the @nifti_capable to minimize code duplication. See my suggestions for `keep_largest_object()`, which you can apply to the rest of the functions if you agree with the suggestions: 
https://github.com/neuropoly/ivado-medical-imaging/compare/cg/postpro-soft-friendly...jca/postpro-soft-friendly

testing/test_postprocessing.py
charleygros: My understanding of these test functions is:
1. Alter the input dummy seg on a particular coordinate
2. Run postpro
3. Check if postprocessing did want we expect from it at this particular coordinate
--> I would like, with this test I added, to check if the input (ie before alteration) is not modified.

ivadomed/postprocessing.py
charleygros: (Oups... I have just seen this comment)
Very elegant solution! Thank you very much, I am merging these

ivadomed/postprocessing.py
```diff
@@ -38,40 +47,36 @@ def threshold_predictions(predictions, thr=0.5):
...
 
 @nifti_capable
+@binarize_with_low_threshold
 def keep_largest_object(predictions):
     """
     Keep the largest connected object from the input array (2D or 3D).
-    Note: This function only works for binary segmentation.
 
     Args:
-        predictions (array or nibabel object): Input binary segmentation. Image could be 2D or 3D.
```
jcohenadad, line 47: Technically-speaking this function does *only* accept binary inputs. The decorator (which is not part of the function) makes sure that input is indeed binary. However this is a blurry zone as when called outside, we could input non-binary (thanks to the decorator)... @kousu what is the proper way of documenting this docstring?

testing/test_postprocessing.py
```diff
@@ -64,49 +85,56 @@ def test_threshold(nii_seg):
...
+    assert check_bin_vs_soft(nii_seg.dataobj, nii_seg_proc.dataobj)
     assert nii_seg_proc.dataobj[coord] == 0
 
 
-@pytest.mark.parametrize('nii_seg', [nii_dummy_seg()])
+@pytest.mark.parametrize('nii_seg', [nii_dummy_seg(), nii_dummy_seg(softseg=True)])
 def test_fill_holes(nii_seg):
     # Set a voxel to 0 in the middle of the segmentation to make sure it is set to 1 by the function
     coord = (7, 7, 4)
+    value_coord_copy = nii_seg.dataobj[coord]
```
jcohenadad, line 126: are you sure that this equality will not simply make `value_coord_copy` act as a pointer to the `nii_seg.dataobj` object, hence if the latter object is modified (which we want to assert), then `value_coord_copy` will be also modified , making the test below not sensitive to potential alteration of nibabel input object?

ivadomed/postprocessing.py
charleygros: Mmm I see. If we go this way, we should change `array or nibabel object` --> `array`.
But I thought it would be informative to make clear in the docstring that we could input a soft nibabel.

jcohenadad: > Mmm I see. If we go this way, we should change array or nibabel object --> array.

indeed, good point. 

> But I thought it would be informative to make clear in the docstring that we could input a soft nibabel.

yup, agreed. Let's wait for @kousu 's feedback.

charleygros:
I am removing the soft input possibility for `fill_holes` as we can't "recover" the voxel soft value from the input (as done with `keep_largest`) as ... there is a hole. More complex strategies could be implemented in the future.

charleygros:
```
chgroc@rosenberg:~/ivado-medical-imaging$ pytest -q testing/test_postprocessing.py 
.......                                                                  [100%]
7 passed in 0.28s
```

ivadomed/postprocessing.py
charleygros: (Opening this issue #207, as I am about to merge this PR)

Issue #212, opened 2020-05-01
Folder imports


charleygros:
In order to reduce the length / complexity of the imports, we could replace:
```python
from ivadomed.loader import utils as imed_loader_utils, adaptative as imed_adaptative, film as imed_film
```
by
```python
from ivadomed import loader as imed_loader
```

To allow this, we would need to add in the [__init__.py](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/loader/__init__.py) these following lines:
```python
from loader import *
from adaptative import *
from utils import *
from film import *
```

jcohenadad:
that sounds great!

my suggestion however would be, instead of renaming functions/classes with "imed_" prefix, why not doing:
```python
import ivadomed as imed
import ivadomed.loader

# in the code
imed.loader
```

it takes the same amount of characters, it is more explicit and it allows to use automatic completion within IDE.

PR #209, opened 2020-04-22
MedicalTorch integration: transforms

charleygros:
### Done:
- `ToTensor`
- `ToPIL`
- `IMEDTransform`
- `UndoTransform`
- `StackTensors`
- `Crop`
- `CenterCrop2D`
- `Normalize`
- `NormalizeInstance`
- `NormalizeInstance3D`
- `RandomRotation`
- `RandomRotation3D`
- `RandomFlip3D`
- `RandomAffine`
- `RandomTensorChannelShift`
- `ElasticTransform`
- `Resample`
- `AdditiveGaussianNoise`
- `Clahe`
- `Clipping`
- Correct imports across `ivadomed`
- Run tests

PR #213, opened 2020-05-01
Refactor transforms

charleygros:
Following #209, some refactoring in the transforms:
- im and seg transforms run separetely, see [comment here](https://github.com/neuropoly/ivado-medical-imaging/issues/211#issuecomment-621686783).
- Only use `numpy` operations (i.e. no `PIL`)
- Be robust to both "sample" and "list of sample"
- Split file between different small files, using #212
- Implement test functions


charleygros:
PS: I will do another PR once #209 has been merged, so that the "old commit" will not be included in this one.

charleygros:
In order to be compatible for both "sample" and "list of sample" I implemented the following decorator, `@list_capable`:
```
def list_capable(wrapped):
    @functools.wraps(wrapped)
    def wrapper(self, sample, metadata):
        if isinstance(sample, list):
            list_data, list_metadata = [], []
            for s_cur, m_cur in zip(sample, metadata):
                # Run function for each sample of the list
                data_cur, metadata_cur = wrapped(self, s_cur, m_cur)
                list_data.append(data_cur)
                list_metadata.append(metadata_cur)
            return list_data, list_metadata
        return wrapped(self, sample, metadata)
    return wrapper
```

The idea is:
- the transformation function (and undo) can only receive a sample
- if the input is a list, then the decorator will iterate through the sample and call the tranformation for each of them, and return a list of "transformed sample"
- same idea for the metadata

charleygros:
For instance, the class `HistogramClipping`:

### Before:
```python
class HistogramClipping(IMEDTransform):

    def __init__(self, min_percentile=5.0, max_percentile=95.0):
        self.min_percentile = min_percentile
        self.max_percentile = max_percentile

    def do_clipping(self, data):
        data = np.copy(data)
        # Ensure that data is a numpy array
        data = np.array(data)
        # Run clipping
        percentile1 = np.percentile(data, self.min_percentile)
        percentile2 = np.percentile(data, self.max_percentile)
        data[data <= percentile1] = percentile1
        data[data >= percentile2] = percentile2
        return data

    def __call__(self, sample):
        input_data = sample['input']

        # TODO: Decorator?
        if isinstance(input_data, list):
            output_data = [self.do_clipping(data) for data in input_data]
        else:
            output_data = self.do_clipping(input_data)

        # Update
        rdict = {'input': output_data}
        sample.update(rdict)
        return sample
```


### After:
```python
class HistogramClipping(IMEDTransform):

    def __init__(self, min_percentile=5.0, max_percentile=95.0):
        self.min_percentile = min_percentile
        self.max_percentile = max_percentile

    @list_capable
    def __call__(self, sample, metadata={}):
        data = np.copy(sample)
        # Run clipping
        percentile1 = np.percentile(sample, self.min_percentile)
        percentile2 = np.percentile(sample, self.max_percentile)
        data[sample <= percentile1] = percentile1
        data[sample >= percentile2] = percentile2
        return data, metadata
```

Note: it also simplifies the code regarding the labeled data.

charleygros:
Another example with `RandomShiftIntensity`:

### Before:
```python
class RandomTensorChannelShift(IMEDTransform):

    def __init__(self, shift_range):
        self.shift_range = shift_range

    @staticmethod
    def get_params(shift_range):
        sampled_value = np.random.uniform(shift_range[0],
                                          shift_range[1])
        return sampled_value

    @staticmethod
    def sample_augment(input_data, params):
        np_input_data = np.array(input_data)
        np_input_data += params
        input_data = Image.fromarray(np_input_data, mode='F')
        return input_data

    def __call__(self, sample):
        input_data = sample['input']
        params = self.get_params(self.shift_range)

        if isinstance(input_data, list):
            ret_input = [self.sample_augment(item, params) for item in input_data]
        else:
            ret_input = self.sample_augment(input_data, params)

        rdict = {'input': ret_input}

        sample.update(rdict)
        return sample
```

### After
```python
class RandomShiftIntensity(IMEDTransform):

    def __init__(self, shift_range):
        self.shift_range = shift_range

    @list_capable
    def __call__(self, sample, metadata={}):
        # Get random offset
        offset = np.random.uniform(self.shift_range[0], self.shift_range[1])
        # Update metadata
        metadata['offset'] = offset
        # Shift intensity
        data = sample + offset
        return data, metadata

    @list_capable
    def undo_transform(self, sample, metadata={}):
        assert 'offset' in metadata
        # Get offset
        offset = metadata['offset']
        # Substract offset
        data = sample - offset
        return data, metadata
```

charleygros:
For testing purposes, I implemented a function to create dummy data with labels:
```python
def create_test_image_2d(width, height, num_modalities, noise_max=10.0, num_objs=1, rad_max=30, num_seg_classes=1):
    """Create test image.

    Create test image and its segmentation with a given number of objects, classes, and maximum radius.

    Args:
        width (int): width image
        height (int): height image
        num_modalities (int): number of modalities
        noise_max (float): noise from the uniform distribution [0,noise_max)
        num_objs (int): number of objects
        rad_max (int): maximum radius of objects
        num_seg_classes (int): number of classes
    Return:
        list, list: image and segmentation, list of num_modalities elements of shape (width, height).

    Adapted from: https://github.com/Project-MONAI/MONAI/blob/master/monai/data/synthetic.py#L17
    """
```

charleygros:
We can know run tests such as:
```python
@pytest.mark.parametrize('im_seg', (create_test_image_2d(100, 100, 1),
                                    create_test_image_2d(100, 100, 3)))
def test_RandomShiftIntensity(im_seg):
    im, _ = im_seg
    # Transform
    transform = RandomShiftIntensity(shift_range=[0., 10.])

    # Apply Do Transform
    metadata_in = [{} for _ in im] if isinstance(im, list) else {}
    do_im, do_metadata = transform(sample=im, metadata=metadata_in)
    # Check result has the same number of modalities
    assert len(do_im) == len(im)
    # Check metadata update
    assert all('offset' in m for m in do_metadata)
    # Check shifting
    for idx, i in enumerate(im):
        assert isclose(np.max(do_im[idx]-i), do_metadata[idx]['offset'], rel_tol=1e-02)

    # Apply Undo Transform
    undo_im, undo_metadata = transform.undo_transform(sample=do_im, metadata=do_metadata)
    # Check result has the same number of modalities
    assert len(undo_im) == len(im)
    # Check undo
    for idx, i in enumerate(im):
        assert np.allclose(undo_im[idx], i, rtol=1e-02)
```

charleygros:
The old `ToPIL`, `ToTensor` and `ToTensor3D` are now replaced by only one class: `NumpyToTensor`.

Todo:
- check if difference between `F.to_tensor` and `torch.from_numpy`
- Do we still need to take into account [these lines](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/transforms.py#L275)?

charleygros:
`Resample` is now PIL-free.

Please note the `interpolation_mode` attribute --> set to 0 for label data, to 2 for images.

```python
class Resample(IMEDTransform):
    """
    Resample image to a given resolution.

    Args:
        wspace (float): Resolution along the first axis, in mm.
        hspace (float): Resolution along the second axis, in mm.
        interpolation_order (int): Order of spline interpolation. Set to 0 for label data. Default=2.
    """
```

TODO:
- finish `test_Resample`

andreanne-lemay:
> The old `ToPIL`, `ToTensor` and `ToTensor3D` are now replaced by only one class: `NumpyToTensor`.
> 
> Also implemented a pytest testing function for it.
> 
> Todo:
> 
> * check if difference between `F.to_tensor` and `torch.from_numpy`
> * Do we still need to take into account [these lines](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/transforms.py#L275)?

The torch.cat was used to convert the list to a torch tensor, so we don't need this part anymore. If the numpy array has the following shape:
2D: (num_channel, height, width) (maybe its (width, height), not sure)
3D: (num_channel, height, depth, width)

We don't need to unsqueeze either, but if the shape doesn't include the number of channels in dimension 0, we would need to add it in dimension 0 (`unsqueeze`).

ivadomed/transforms.py
```diff
@@ -1,26 +1,44 @@
...
+                # Run function for each sample of the list
+                data_cur, metadata_cur = wrapped(self, s_cur, m_cur)
+                list_data.append(data_cur)
+                list_metadata.append(metadata_cur)
+            return list_data, list_metadata
+        return wrapped(self, sample, metadata)
+    return wrapper
 
 
 class IMEDTransform(object):
```
jcohenadad, line 36: Maybe use the camel convention? 'ImedTransform'

charleygros:
To assess data consistency during transforms, I have implemented a small function under `ivadomed/metrics.py` to compute the MSE between two images:
```python
def mse(im1, im2):
    """ Compute the Mean Squared Error.

    Compute the Mean Squared Error between the two images, i.e. sum of the squared difference.

    Args:
        im1 (array): First array.
        im2 (array): Second array.
    Returns:
        float: Mean Squared Error.

    """
```

charleygros:
Utils function implemented in `testing/test_transforms.py`, useful for debugging:
```python
def plot_sample(before, after):
    """Utils tool to plot sample before and after transform, for debugging.

    Args:
        before: sample before transform.
        after: sample after transform.
    """
```

Example of output:
Figure_1

Question: Should we move it in a new file (e.g. `testing/utils.py`) so that could be used by other testing functions? or keep it here for now? Same for `create_test_image_2d`.

charleygros:
An idea would be to add a param in transform spec' in the config file, e.g. `applied_to`, which indicates to which data the transform is applied to.

For instance:
```json
    "transformation_training": {
        "Resample": {
            "wspace": 0.75,
            "hspace": 0.75,
            "applied_to": ["im", "seg", "roi"]
        },
        "RandomShiftIntensity": {
            "shift_range": [0.2, 0.8],
            "applied_to": ["im"]
        },
        "Crop2D": {
            "size": [100, 100],
            "applied_to": ["im", "seg"]
        },
        "NumpyToTensor": {
            "applied_to": ["im", "seg"]
        }
    },
```

- `Crop2D` and `NumpyToTensor` will be applied to both the image and the segmentation
- `Resample` will be applied to the image, the ROI and the segmentation
- `RandomShiftIntensity` will be applied to the image only

That way we would have a better control of what is applied to what data. And it simplifies the code in `ivadomed/transforms.py`.

Any thoughts?

charleygros:
Within the loader, the only changes will be on [these lines](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/loader/loader.py#L347) --> replaces by:
```
        # Run transforms on images
        list_input, metadata_input = self.transform['im'](list_input, seg_pair_slice['input_metadata'])
        # Run transforms on ROI
        list_roi, metadata_roi = self.transform['roi'](list_roi, roi_pair_slice['roi_metadata'])
        # Run transforms on GT
        list_gt, metadata_gt = self.transform['gt'](list_gt, seg_pair_slice['gt_metadata'])

        data_dict = {
            'input': list_input,
            'gt': list_gt,
            'roi': list_roi,
            'input_metadata': seg_pair_slice['input_metadata'],
            'gt_metadata': seg_pair_slice['gt_metadata'],
            'roi_metadata': roi_pair_slice['gt_metadata']
        }
```

Where `self.transforms` is a dict of `Compose` objects.

charleygros:
The param `applied_to` is used in the constructor of `Compose`, implemented in `ivadomed/transforms.py`:
```
class Compose(object):
    """Composes transforms together.

    Composes transforms together and split between images, GT and ROI.

    self.transform is a dict:
        - keys: "im", "gt" and "roi"
        - values torchvision_transform.Compose objects.

    Attributes:
        dict_transforms (dictionary): Dictionary where the keys are the transform names
            and the value their parameters.
        requires_undo (bool): If True, does not include transforms which do not have an undo_transform
            implemented yet.
    """
```

charleygros:
The only change in the main is to replace[ this line](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L90) by:
```python
    train_transform = imed_transforms.Compose(context["transformation_training"])
```

ivadomed/transforms.py
charleygros: Done :)

andreanne-lemay:
> ```python
> plot_sample
> ```

I think a testing/utils.py is a good idea!

charleygros:
ANNOUNCMENT: `NormalizeInstance` is now compatible with both numpy array or torch tensor --> it could hence be used before `NumpyToTensor`.

```python
class NormalizeInstance(ImedTransform):
    """Normalize a tensor or an array image with mean and standard deviation estimated
    from the sample itself.
    """

    def __call__(self, sample, metadata={}):
        if isinstance(sample, np.ndarray):
            data_out = (sample - np.mean(sample)) / np.std(sample)
        elif torch.is_tensor(sample):
            data_out = F.instance_norm(sample)
        else:
            raise NotImplementedError("Invalid data type.")

        return data_out, metadata
```

charleygros:
> > ```python
> > plot_sample
> > ```
> 
> I think a testing/utils.py is a good idea!

Done in `testing/utils.py`.
Also added this to `testing/test_transforms.py`:
```python
DEBUGGING = False
if DEBUGGING:
    from testing.utils import plot_transformed_sample
```

charleygros:
> ANNOUNCMENT: `NormalizeInstance` is now compatible with both numpy array and torch tensor --> it could hence be used before `NumpyToTensor`.

@andreanne-lemay & @olix86 --> I wonder if we should restrict numpy arrays as inputs, which will imply that `NumpyToTensor` has to be the last transform listed in the config files. What do you think?

charleygros:
Adding even more randomness in `RandomShiftIntensity`:
```python
        if np.random.random() < self.prob:
            # Get random offset
            offset = np.random.uniform(self.shift_range[0], self.shift_range[1])
        else:
            offset = 0.0
```

andreanne-lemay:
> An idea would be to add a param in transform spec' in the config file, e.g. `applied_to`, which indicates to which data the transform is applied to.
> 
> For instance:
> 
> ```json
>     "transformation_training": {
>         "Resample": {
>             "wspace": 0.75,
>             "hspace": 0.75,
>             "applied_to": ["im", "gt", "roi"]
>         },
>         "RandomShiftIntensity": {
>             "shift_range": [0.2, 0.8],
>             "applied_to": ["im"]
>         },
>         "Crop2D": {
>             "size": [100, 100],
>             "applied_to": ["im", "gt"]
>         },
>         "NumpyToTensor": {
>             "applied_to": ["im", "gt"]
>         }
>     },
> ```
> 
> * `Crop2D` and `NumpyToTensor` will be applied to both the image and the GT
> * `Resample` will be applied to the image, the ROI and the GT
> * `RandomShiftIntensity` will be applied to the image only
> 
> That way we would have a better control of what is applied to what data. And it simplifies the code in `ivadomed/transforms.py`.
> 
> Any thoughts?

Yes it could be useful for transforms we only want to apply to images (e.i. HistogramClipping). By default, a transform could be applied to all images type (roi, gt and input) and only if the key "applied_to" exists, then the transform is applied to the specified image types. This way we don't have to change every transform in the config files + would avoid extra lines and repetition.

charleygros:
> By default, a transform could be applied to all images type (roi, gt and input) and only if the key "applied_to" exists, then the transform is applied to the specified image types. This way we don't have to change every transform in the config files + would avoid extra lines and repetition.

Love this idea! I have modified the code accordingly:
```python
            # Get list of data type
            if "applied_to" in parameters:
                list_applied_to = parameters["applied_to"]
                del parameters['applied_to']
            else:
                list_applied_to = ["im", "gt", "roi"]
```

charleygros:
TODO: Use [torch.from_numpy](https://pytorch.org/docs/stable/torch.html#torch.from_numpy)

charleygros:
New cropping function (intense diet!) here below. Just to say: @andreanne-lemay --> I am assuming here the convention we discussed for np arrays, ie C x H x W x D
```python
    @list_capable
    def __call__(self, sample, metadata={}):
        # Crop parameters
        th, tw = self.size
        h, w = sample.shape
        fh = int(round((h - th) / 2.))
        fw = int(round((w - tw) / 2.))
        params = (fh, fw, w, h)
        metadata['crop_params'] = params

        # Crop data
        data_out = sample[fh:fh+th, fw:fw+tw]

        return data_out, metadata
```

andreanne-lemay:
> New cropping function (intense diet!) here below. Just to say: @andreanne-lemay --> I am assuming here the convention we discussed for np arrays, ie C x H x W x D
> 
> ```python
>     @list_capable
>     def __call__(self, sample, metadata={}):
>         # Crop parameters
>         th, tw = self.size
>         h, w = sample.shape
>         fh = int(round((h - th) / 2.))
>         fw = int(round((w - tw) / 2.))
>         params = (fh, fw, w, h)
>         metadata['crop_params'] = params
> 
>         # Crop data
>         data_out = sample[fh:fh+th, fw:fw+tw]
> 
>         return data_out, metadata
> ```

Yes -> #222 here it is ;)

charleygros:
Almost there :-)

charleygros:
Another debate we had was: the distinction between 2D and 3D transforms. Should we keep distinct transforms or try to merge them?

Well, I tried to merge `Crop2D` and `Crop3D` and... I feel quite satisfied with the result but I would appreciate your feedback :)
The main "trick" is the introduction of a decorator (`@two_dim_compatible`) that will take care of each function where the extra dimension of 3D was limiting the merge between 2D and 3D. It simply expand the dimension of a 2D array, process it as if it was a 3D array, then squeeze the last dimension before returning the final result:
```python
def two_dim_compatible(wrapped):
    @functools.wraps(wrapped)
    def wrapper(self, sample, metadata):
        # Check if sample is 2D
        if len(sample.shape) == 2:
            # Add one dimension
            sample = np.expand_dims(sample, axis=-1)
            # Run transform
            result_sample, result_metadata = wrapped(self, sample, metadata)
            # Remove last dimension
            return np.squeeze(result_sample, axis=-1), result_metadata
        else:
            return wrapped(self, sample, metadata)
    return wrapper
```


(besides some intense code refactoring in the way to handle it --> especially

charleygros:
Other refactoring in the Cropping:
- `CenterCrop` and `ROICrop` compute the cropping params but the cropping is actually done in `Crop` --> via calling `super().__call__`
- `undo_transform` is now common between `CenterCrop` and `ROICrop`
- It takes a great care of the here below issue, especially with `CroppableArray` and `_adjust_padding` tools
Screenshot 2020-05-13 21:12:35

andreanne-lemay:
> Another debate we had was: the distinction between 2D and 3D transforms. Should we keep distinct transforms or try to merge them?
> 
> Well, I tried to merge `Crop2D` and `Crop3D` and... I feel quite satisfied with the result but I would appreciate your feedback :)
> The main "trick" is the introduction of a decorator (`@two_dim_compatible`) that will take care of each function where the extra dimension of 3D was limiting the merge between 2D and 3D. It simply expand the dimension of a 2D array, process it as if it was a 3D array, then squeeze the last dimension before returning the final result:
> 
> ```python
> def two_dim_compatible(wrapped):
>     @functools.wraps(wrapped)
>     def wrapper(self, sample, metadata):
>         # Check if sample is 2D
>         if len(sample.shape) == 2:
>             # Add one dimension
>             sample = np.expand_dims(sample, axis=-1)
>             # Run transform
>             result_sample, result_metadata = wrapped(self, sample, metadata)
>             # Remove last dimension
>             return np.squeeze(result_sample, axis=-1), result_metadata
>         else:
>             return wrapped(self, sample, metadata)
>     return wrapper
> ```
> 
> --> the Crop would be then performed by `CenterCrop` or `ROICrop` for both 2D and 3D.
> (validated by unit tests :) )

Great, I love it! Making all transforms 2D and 3D compatible will highly reduce the number of methods and lines, I think it is the way to go!

ivadomed/transforms.py
```diff
@@ -42,148 +138,76 @@ def __call__(self, sample):
...
-        rdict['gt'] = self.do_resample(list_data=sample["gt"],
-                                       new_shape=(wshape, hshape),
-                                       interpolation_mode=self.interpolation)
+        hshape, wshape = metadata['data_shape']
 
-        # Update
-        sample.update(rdict)
-        return sample
+        # Undo resampling
+        data_out = resize(sample,
```
andreanne-lemay, line 176: For GT, the `resize` function returns a soft output, should we binarize after resampling or using a soft output as GT is preferable?

charleygros: Good point
Few thoughts:
- I will try order = 0 for GT
- There is other interpolations after resampling (elastic transform, rotation...etc.) --> I feel that if we binarise after each transform we may accumulate a fair amount of error along the way (versus binarize at the very end of the transform pipeline)

charleygros:
We now have a common function to create dummy data for both 2D and 3D :-)

```python
def create_test_image(width, height, depth=0, num_modalities=1, noise_max=10.0, num_objs=1, rad_max=30, num_seg_classes=1):
    """Create test image.

    Create test image and its segmentation with a given number of objects, classes, and maximum radius.
    Compatible with both 2D (depth=0) and 3D images.

    Args:
        height (int): height image
        width (int): width image
        depth (int): depth image, if 0 then 2D images are returned
        num_modalities (int): number of modalities
        noise_max (float): noise from the uniform distribution [0,noise_max)
        num_objs (int): number of objects
        rad_max (int): maximum radius of objects
        num_seg_classes (int): number of classes
    Return:
        list, list: image and segmentation, list of num_modalities elements of shape (height, width, depth).
```

charleygros:
I am removing `StackTensor` as we will have tensors with shape batch_nb, nb_channel, H, W (, D) after our `NumpyToTensor`.

charleygros:
I am also removing `NormalizeInstance3D` as it could be achieve through `NormalizeInstance` --> Note that it needs numpy as input --> to use before `NumpyToTensor`.

charleygros:
We do not apply `@list_capable` to `AddBackgroundClass` as we need all the channels (ie different labels in a multi-class case) to determine the Background.

charleygros:
For all functions were random elements are defined, the new approach requires us to update the GT metadata with the metadata IM so that the same transform is applied to both.

In the transform, we check if the params exist in the metadata. If so, we use them. Otherwise we get random ones and save them in metadata. Cf example in `RandomRotation`:
```python
        # If angle and metadata have been already defined for this sample, then use them
        if 'angle' in metadata and 'rotation' in metadata:
            angle, axes = metadata['rotation'][0]
        # Otherwise, get random ones
        else:
            # Get the random angle
            angle = np.random.uniform(self.degrees[0], self.degrees[1])
            # Get the two axes that define the plane of rotation
            axes = tuple(random.sample(range(3 if sample.shape[2] > 1 else 2), 2))
            # Save params
            metadata['rotation'] = [angle, axes]
```

charleygros:
For some mysterious reasons, the RandomRotation.undo works well for seg but not for images... Can't see why at this stage.
@olix86 & @andreanne-lemay --> if you have time to have a look, that would be great.
To reproduce --> turn `DEBUGGING` to `True` in `testing/test_transforms.py` and run `pytest -q testing/test_transforms.py -s -k test_RandomRotation`

charleygros:
I am introducing two new functions in `ivadomed/loader/utils.py`:

1. `clean_metadata` --> Remove the transform parameters from the previous iteration so that the transform are not the same at each iteration eheh
```python
def clean_metadata(metadata_lst):
    metadata_out = []
    for metadata_cur in metadata_lst:
        for key_ in metadata_cur:
            if key_ in TRANSFORM_PARAMS:
                del metadata_cur[key_]
            metadata_out.append(metadata_cur)
    return metadata_out
```

2. `update_metadata` is used to spread the transform param from ROI to IM then from IM to GT, in order to apply the same transform between them
```python
def update_metadata(metadata_src_lst, metadata_dest_lst):
    metadata_out_lst = []
    for idx in range(len(metadata_src_lst)):
        metadata_updated = dict(list(metadata_src_lst[idx].items()) + list(metadata_dest_lst[idx].items()))
        metadata_out_lst.append(metadata_updated)
    return metadata_out_lst

**Comment by charleygros** [comment]
What remains to do:
- Update config files
- Test function with nibabel object (primarly for debugging experimentation purposes)
- Update test functions that use transforms
- Update 3D loader

A couple of Transforms have not been converted yet --> I was thinking of doing this in another PR, since this one is quite obese which makes its review quite tedious.

**Comment by andreanne-lemay** [comment]
> For some mysterious reasons, the RandomRotation.undo works well for seg but not for images... Can't see why at this stage.
> @olix86 & @andreanne-lemay --> if you have time to have a look, that would be great.
> To reproduce --> turn `DEBUGGING` to `True` in `testing/test_transforms.py` and run `pytest -q testing/test_transforms.py -s -k test_RandomRotation`

I tried to debug it, and frankly this is weird! I can't see why the behavior is different for gt and img, but I found this: https://stackoverflow.com/questions/46657423/rotated-image-coordinates-after-scipy-ndimage-interpolation-rotate . I think the problem is related to this. Basically, rotation in X then in -X won't give you the same image according to this post. Maybe another library might have a more intuitive rotate tool?

File: ivadomed/transforms.py [hunk 3]
```diff
@@ -42,148 +142,86 @@ def __call__(self, sample):
...
-        wshape_new = int(round(wshape * wfactor))
-
-        # Input data
-        rdict['input'] = self.do_resample(list_data=sample["input"],
-                                          new_shape=(wshape_new, hshape_new),
-                                          interpolation_mode=self.interpolation)
+        zooms = metadata["zooms"]
+        if len(zooms) == 2:
+            zooms += tuple([1.0])
+        hfactor = zooms[0] / self.hspace
```
andreanne-lemay, line 204: This branch is not merged with master, so I'm getting the wrong zooms for sagittal images (so images don't respect de hwd format). The same problem applies to CenterCrop. Let me know if you need help to merge with master :) .

ivadomed/transforms.py
```diff
@@ -1101,166 +838,88 @@ def __call__(self, sample):
...
+class ElasticTransform(ImedTransform):
+    """Elastic transform for 2D and 3D inputs."""
 
-    def __init__(self, alpha_range, sigma_range,
-                 p=0.5, labeled=True):
+    def __init__(self, alpha_range, sigma_range):
         self.alpha_range = alpha_range
         self.sigma_range = sigma_range
-        self.labeled = labeled
-        self.p = p
```
andreanne-lemay, line 1145: Why remove p from ElasticTransform?

ivadomed/transforms.py
```diff
@@ -1101,166 +838,88 @@ def __call__(self, sample):
...
-        self.is3D = False  # Set as default, it will be re-assesed when called
 
-    @staticmethod
-    def get_params(alpha, sigma):
-        alpha = np.random.uniform(alpha[0], alpha[1])
-        sigma = np.random.uniform(sigma[0], sigma[1])
-        return alpha, sigma
+    @multichannel_capable
```
andreanne-lemay, line 878: Awesome!

ivadomed/transforms.py
```diff
@@ -599,180 +566,30 @@ def __call__(self, sample):
         return sample
 
 
-class StackTensors(IMEDTransform):
-    """Stack all modalities, as a list, in a single tensor."""
+# TODO: unit test
+class AddBackgroundClass(ImedTransform):
```
andreanne-lemay, line 570: Delete?

ivadomed/transforms.py
```diff
@@ -42,148 +142,86 @@ def __call__(self, sample):
...
+    def undo_transform(self, sample, metadata):
+        assert "resample" in metadata
 
-        # Input data
-        rdict['input'] = self.do_resample(list_data=sample["input"],
-                                          new_shape=(wshape, hshape),
-                                          interpolation_mode=self.interpolation)
+        # Get params
+        params_do = metadata["resample"]
+        params_undo = [1 / x for x in params_do]
```
andreanne-lemay, line 184: Smooth!

ivadomed/config/config_sctTesting.json
```diff
@@ -90,7 +90,7 @@
         "ROICrop2D": {
             "size": [48, 48]
         },
-        "ToTensor": {},
+        "NumpyToTensor": {},
         "NormalizeInstance": {},
```
andreanne-lemay, line 94: Should probably add applie_to: "img" as well

testing/test_transforms.py
```diff
@@ -0,0 +1,353 @@
...
+            if DEBUGGING:
+                plot_transformed_sample(seg[idx], undo_seg[idx], ['raw', 'undo'])
+                plot_transformed_sample(i, undo_im[idx], ['raw', 'undo'])
+        else:
+            assert np.array_equal(i[fh:fh+th, fw:fw+tw, fd:fd+td], undo_im[idx][fh:fh+th, fw:fw+tw, fd:fd+td])
+            assert np.array_equal(seg[idx][fh:fh+th, fw:fw+tw, fd:fd+td], undo_seg[idx][fh:fh+th, fw:fw+tw, fd:fd+td])
+
+
+@pytest.mark.parametrize('im_seg', [create_test_image(100, 100, 0, 2)])
+@pytest.mark.parametrize('crop_transform', [CenterCrop((80, 60)),
```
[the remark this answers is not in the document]
charleygros: The test were passing though, but happy to change that

ivadomed/transforms.py
charleygros: ok

ivadomed/config/config_sctTesting.json
charleygros: What do you mean? I thought you suggested that when "applied_to" is not mentioned then the transform is applied to all kind of images. Have I missed something? :)
Also: this config file is not updated yet (cf comment TODO "Update config files" --> only config_small is ready, cf Slack message). Sorry for wasting your time, I should not have asked you to review it before all ready.

ivadomed/transforms.py
charleygros: I was annoying me when I was optimising alpha & sigma eheh Forgot to put it back! Many thanks :) Good spotting

testing/test_transforms.py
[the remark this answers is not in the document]
andreanne-lemay: weird, I had an error, that's why I put the comment, but tuples are not mutable so it makes sense to put a list since if it's 2D we modify  the crop dimensions :)

ivadomed/config/config_sctTesting.json
andreanne-lemay: Ah yes you are right I see it in your TODO list now :) I just meant to add "applied_to":"img"  as parameter to "NormalizeInstance" in all config files since we never want to normalize the binarized mask. But totally my bad, I missed this element on your todo :). Not a waste of time at all! As said here:

> Wouhou awesome refactoring :)! Definitely worth the time. I was able to run everything, but since the orientation were not correct for sagittal couldn't fully test CenterCrop and Resample. Runs well in multichannel and multilabel!

Everything works very well! Once it's merged with master I'll test those transforms :rocket:

charleygros:
> A couple of Transforms have not been converted yet --> I was thinking of doing this in another PR, since this one is quite obese which makes its review quite tedious.

I changed my mind --> I refactored all the transforms in this PR.

charleygros:
What remains to do:
* Check data type init loader
* Test function with nibabel object (primarly for debugging experimentation purposes)
* Finish to update cmd_test

ivadomed/config/config_sctTesting.json
charleygros: done :)

ivadomed/transforms.py
charleygros: should be better now :)

testing/test_transforms.py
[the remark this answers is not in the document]
charleygros: done

charleygros:
`segment_volume()` now compatible with the new implementation.
To test:
```shell
import nibabel as nib
from ivadomed.utils import segment_volume
m = 'model_t2s'
i = 'testing_data/sub-test001/anat/sub-test001_T2star.nii.gz'
p = segment_volume(m, i)
```

Note: The config file in the testing_data has been updated with the new transform names.

ivadomed/utils.py
```diff
@@ -688,28 +690,17 @@ def segment_volume(folder_model, fname_image, fname_roi=None):
...
-        for i_slice in range(len(batch['gt'])):
-            # Undo transformations
-            rdict = {}
-            # Import transformations parameters
-            for k in batch.keys():
-                if len(batch[k]):
-                    rdict[k] = batch[k][i_slice]
-            rdict_undo = undo_transforms(rdict)
+        for i_slice in range(len(preds)):
+            # undo transformations
```
jcohenadad, line 695: stupid question here: why do we need to undo transformations during inference? also, how come data_type='gt' (we don't have ground truth during normal use-case inference)

andreanne-lemay: Transformations are undone so that the prediction mask match the original image. For instance, if I crop the image 48x48 during inference, but my original image is 100x100, I want the prediction to superpose the 100x100 image. 

For the data_type='gt', the transforms are organized in a dictionary with the keys 'gt' (segmentation), 'img' (input image) and 'roi' (roi to help crop the image around a mask in 2D) (e.i. `transforms = {'gt': transforms_applied_to_gt, 'img':  transforms_applied_to_img, 'roi': transforms_applied_to_roi}`). For example. we don't want to normalize the gt (or any segmentation), so NormalizeInstance will only be in the key 'img'. This was added to remove all the [if calls](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/transforms.py#L1055) in the transforms (if self.labeled: do thing for labels) and the code duplication to apply the transform to the image then to the gt. So here 'gt' refers to the type of input that is given -> we only want to apply the 'gt' transforms since predictions are processed the same way as 'gt'. The name 'gt' may not be the more adapted, we could change it to 'seg' or 'segmentation' or 'mask'?

jcohenadad: cool! thank you for the explanation. 

> we could change it to 'seg'

👍  i would also suggest the more general 'label' (which can include segmentations, single points, etc.)

charleygros: The reason for not choosing `seg` is that the `roi` is also a segmentation.
In the config files we opted for `target`.
But I agree that, at the end, this data is the segmentation (target or predicted) --> so `seg` could make more sense. To discuss :)

ivadomed/loader/utils.py
```diff
@@ -229,6 +229,9 @@ def clean_metadata(metadata_lst):
 
 def update_metadata(metadata_src_lst, metadata_dest_lst):
     if metadata_src_lst and metadata_dest_lst:
+        if len(metadata_src_lst) > len(metadata_dest_lst):
+            metadata_dest_lst = metadata_dest_lst + [SampleMetadata({})] * \
+                                (len(metadata_src_lst) - len(metadata_dest_lst))
```
charleygros, line 234: Great!

ivadomed/utils.py
```diff
@@ -946,3 +946,10 @@ def __call__(self, sample):
                 return False
 
         return True
+
+
+def unstack_tensors(sample):
+    list_tensor = []
+    for i in range(sample.shape[1]):
+        list_tensor.append(sample[:, i, ].unsqueeze(1))
+    return list_tensor
```
charleygros, line 961: To move to `transforms.py`?

ivadomed/utils.py
```diff
@@ -674,7 +674,9 @@ def segment_volume(folder_model, fname_image, fname_roi=None):
         ds = imed_loaded_utils.filter_roi(ds, nb_nonzero_thr=context["slice_filter_roi"])
 
     if not context['unet_3D']:
-        print("\nLoaded {} {} slices..".format(len(ds), context['slice_axis']))
+        print(f"\nLoaded {len(ds)} {context['slice_axis']} slices..")
+    else:
+        print(f"\nLoaded {len(ds)} {context['slice_axis']} volumes of shape {context['length_3D']}")
```
charleygros: I would like to learn: why? :-)

ivadomed/utils.py
jcohenadad: so why is there a 'roi' and a 'gt' (or 'seg') case? are there treatment exceptions between the two? (i guess there is, otherwise you would have called it the same thing)? or to rephrase, should there be different treatments?

charleygros: Yes `roi` and `gt` have not the same treatment --> we use the roi only for the crop --> passed this transform we do not use it anymore.
The requirements are not the same: Only need 1 ROI, but we may need several GT in case of multi label.
Also, ROI is optional, while GT is obviously mandatory.
So in brief: there are different treatments between the two.

jcohenadad: ah, makes sense. Is this documented somewhere?

charleygros:
@andreanne-lemay & @olix86 : Just so that you are aware: `test_RandomAffine` and `test_RandomRotation` may "fail" from time to time because of the problem illustrated here below, even if the do and undo transforms are technically correct.
For now, we could just rerun the test. If it happens too often and becomes annoying, we could think of an alternative.
Does it sound okay?

IMG_20200521_114729

ivadomed/utils.py
charleygros: The difference between ROI and GT is explained in Wiki / config file (cf `*_suffix`): https://github.com/neuropoly/ivado-medical-imaging/wiki/configuration-file

ivadomed/transforms.py
```diff
@@ -302,7 +304,6 @@ def __getitem__(self, item):
 class Crop(ImedTransform):
     def __init__(self, size):
         self.size = size if len(size) == 3 else size + [0]
-        self.is_2D = True if len(size) == 2 else False
```
charleygros: The reason why I check the 2D versus 3D here instead of via sample --> was to be robut to 3D data with shape[2] == 1 (yes I know.. not common). While by checking with the crop_size --> we are certain that the user wants a 2D transform.
What do you think @andreanne-lemay ?

ivadomed/transforms.py
```diff
@@ -181,11 +181,15 @@ def __init__(self, hspace, wspace, dspace=1., interpolation_order=2):
...
-        assert "resample" in metadata
+        assert "data_shape" in metadata
+        is_2d = sample.shape[-1] == 1
 
         # Get params
-        params_do = metadata["resample"]
-        params_undo = [1. / x for x in params_do]
+        original_shape = metadata["data_shape"]
+        current_shape = sample.shape
+        params_undo = [x / y for x, y in zip(original_shape, current_shape)]
```
charleygros: Works too. Did you have issues with the other version? curious

ivadomed/transforms.py
```diff
@@ -211,10 +216,7 @@ def __call__(self, sample, metadata):
         hfactor = zooms[0] / self.hspace
         wfactor = zooms[1] / self.wspace
         dfactor = zooms[2] / self.dspace
-        params_resample = (hfactor, wfactor, dfactor)
-
-        # Save params
-        metadata['resample'] = params_resample
+        params_resample = (hfactor, wfactor, dfactor) if not is_2d else (hfactor, wfactor, 1.0)
```
charleygros, line 216: Cool

ivadomed/transforms.py
```diff
@@ -20,45 +20,45 @@
 
 def multichannel_capable(wrapped):
     @functools.wraps(wrapped)
-    def wrapper(self, sample, metadata):
+    def wrapper(self, sample, metadata, data_type):
```
charleygros: Oh.. what I meant in Slack was to add data_type in the metadata --> Is there advantages in doing the way you did? as it is only used for the resampling

ivadomed/utils.py
andreanne-lemay: I put unstack_tensors as transform at first, but because of the collate (to create batches) I wasn't able to get the right format. I got a list of length batch_size, but what I actually want is a list of length of the number of modality with the shape: batch_size, 1, H, W. Since it is not a transform, I would not put it in transforms, but there might be a better place to put it than utils.py

ivadomed/transforms.py
```diff
@@ -23,9 +23,9 @@ def multichannel_capable(wrapped):
     def wrapper(self, sample, metadata, data_type):
         if isinstance(sample, list):
             list_data, list_metadata = [], []
-            for s_cur, m_cur, d_cur in zip(sample, metadata, data_type):
```
charleygros: Same comment

ivadomed/utils.py
andreanne-lemay: Hehe no reason, I think it's just what was on master? I merged quickly the other line is fine as well!

ivadomed/loader/utils.py
```diff
@@ -21,6 +18,8 @@
     'uint8': torch.ByteTensor,
 }
 
+TRANSFORM_PARAMS = ['resample', 'elastic', 'rotation', 'offset', 'crop_params', 'reverse', 'affine', 'gaussian_noise']
```
charleygros, line 21: TODO charley: `resample` can be removed.

ivadomed/transforms.py
andreanne-lemay: Yes I wanted to discuss that as well! The way it was before worked fine, why I changed it is so the transform automatically adapts to 2D or 3D output without having to change the tranforms. For instance, in the test_orientation.py I can use the same transforms for 2D and 3D. Also, I found it useful to not have to change multiple parameters when I want to train in 2D or 3D (now I only have to change the unet_3D bool).

ivadomed/transforms.py
andreanne-lemay: I also wanted to comment on this! I changed it because using the inverse was not enough precise so I didn't have the same input and output shape after doing and undoing transforms. (e.i 51 instead of 52). When using the dimensions, I get a zoom of example 2.01343 instead of 2, which always give me the right output size.

ivadomed/transforms.py
charleygros: It is convenient indeed. Ok, we can go with that.

ivadomed/transforms.py
charleygros: Excellent :)

ivadomed/transforms.py
andreanne-lemay: Changed it :)!

andreanne-lemay:
On my end, everything runs well, I'm running my usual pipeline to check the performance -> looks good for now. I didn't test or modify the files in dev (we probably should check if there are all still functional?).

I didn't change 'gt' wasn't sure if we wanted `label` or `seg`?

As requested [here](https://github.com/neuropoly/ivado-medical-imaging/pull/220#discussion_r428299843) I removed the explicits calls on all tests.

testing/test_transforms.py
```diff
@@ -155,7 +155,9 @@ def test_NumpyToTensor(im_seg):
 def _test_Resample(im_seg, resample_transform, native_resolution, is_2D=False):
     im, seg = im_seg
     metadata_ = {'zooms': native_resolution,
-                 'data_shape': im[0].shape if len(im[0].shape) == 3 else list(im[0].shape) + [1]}
+                 'data_shape': im[0].shape if len(im[0].shape) == 3 else list(im[0].shape) + [1],
+                 'data_type': 'im'
```
charleygros, line 159: TODO: charley or.. : make another metadata_ for seg with data_type = 'gt'

andreanne-lemay:
> @andreanne-lemay & @olix86 : Just so that you are aware: `test_RandomAffine` and `test_RandomRotation` may "fail" from time to time because of the problem illustrated here below, even if the do and undo transforms are technically correct.
> For now, we could just (i) reduce the translation / rotation params (e.g. test only rotation of 5°) (ii) add a specific warning (eg "the test may have fail because the segmentation lost 80% of its coverage after the transform") and (iii) rerun the test. If it happens too often and becomes annoying, we could think of an alternative. Does it sound okay?
> 
> IMG_20200521_114729

In the long run, we might want to change. If we test a small shift or rotation, will this problem disappear? Or for these we might want to always use the same image (as opposed to randomly generated)

PR #217, opened 2020-05-06
pypi package

olix86:
So for some unknown reason, Travis runs out of memory with 8GB of RAM while github runs fine with 7GB of RAM when running `pytest`. Testing on `Rosenberg` suggests that we don't use near 8GB so I don't see why travis fails. Therefore this PR does the following : 
- revert to GH actions for CI
- add a GH actions script for creating `pypi` package and uploading to pypi.org
- The package is built and uploaded when a release is created
- add GH README to pypi package

Available at : https://pypi.org/project/ivadomed/

via
`pip install ivadomed`


README.md
```diff
@@ -49,21 +48,29 @@ This data augmentation is motivated by the large inter-rater variability that we
 
 ## Installing
 
-This project requires Python 3.6 and PyTorch >= 1.2.0. We recommend you work under a virtual environment:
+
```
andreanne-lemay, line 51: On PyPi some images from the Readme doesn't display properly: 
image
We should change the paths of these images.

ivadomed/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.1"
+__version__ = "0.14"
```
andreanne-lemay, line 1: What are our conventions for versioning :)? Why 0.14?

andreanne-lemay:
Currently the version on PyPi is not the same as this PR (I get version 0.1 with the dependency to pytorch 1.2.0). Can you publish the new version of the package on PyPi or do we need to pass this PR first? Also, on github actions I see 2 actions related to publishing the package: `Python Package` and `Publish Package`. What is Python Package?

ivadomed/__init__.py
andreanne-lemay: Oh I see now! The versioning goes with PyPi's release version number! I see on PyPi that the current version is PyPi 0.13 but I was only able to install 0.1. Also, I came across this error when installing, then uninstalling the package: 
image

README.md
olix86: weird that the badges were absolute but the images were relative paths, now fixed :)

olix86:
>Currently the version on PyPi is not the same as this PR (I get version 0.1 with the dependency to pytorch 1.2.0). Can you publish the new version of the package on PyPi or do we need to pass this PR first? 

I've reuploaded it to be up to date with this PR, after merging the package will be published automatically every time we create a release, but for now it's manual. I think we should do a 1.0 release once this is merged.

> Also, on github actions I see 2 actions related to publishing the package:  and Publish Package. What is Python Package?

It was originally named Python Package, I've renamed it to Publish Package but I don't see a way to delete the old one :(
https://github.community/t5/GitHub-Actions/Delete-old-workflow-results/td-p/30589

ivadomed/__init__.py
olix86: > What are our conventions for versioning :)? Why 0.14?
No particular reason, it started from `0.1` and I had to increment a few times during my tests since pypi doesn't allow to reupload / overwrite using the same version number. 

> Also, I came across this error when installing, then uninstalling the package:
Was that in an existing venv with medicaltorch still installed or an older torchvision/pytorch? It should be fine with a new venv

jcohenadad:
> It was originally named Python Package, I've renamed it to Publish Package but I don't see a way to delete the old one :(

maybe @kousu knows?

PR #220, opened 2020-05-11
Al/segment volume 3 d

andreanne-lemay:
Code adaptation to segment images from 3D models.
Main modifications:
- Minor additions to `segment_volume`
- Modification to `Resample()` transformation adapted for 3D images and all slice axes
- Refactoring + bug fix of transformation `CenterCrop3D`
- Modification to transforms to respect the HWD (height, width, depth) convention
- Addition of configuration parameter to modify the number of base filters for 3D unet (e.i. `n_filters`, `dropout`).
- Minor refactor of `SegmentationPair`: move some code of `get_pair_slice()` in a new method `get_pair_metadata` needed for the 3D loader.
- Refactor of `MRI3DSubVolumeSegmentationDataset`: Apply transforms before extracting subvolumes.


andreanne-lemay:
Resampling with skimage.transform.resize works well in 2D, but not in 3D (I tried order 0, 1 and 2)
Original image:
image

Image after resample and undo resample with 3D volume:
image

Resampling with scipy.ndimage.zoom with order 0 or 1 works well for 2D and 3D (passes test_orientation in 2D and 3D)  
image

The following resampling was applied: {hspace:2, wspace: 3, dspace:1}

andreanne-lemay:
> Looks good to me, my only question is (not directly related to your PR but segment volume in general) that it looks like we don't have a test for `segment_volume`, do you think it could be useful?

Yes, it would be useful! I was sure we already had one, but I doubled-checked and you're right. I'll add one, thanks :)!

jcohenadad:
> Looks good to me, my only question is (not directly related to your PR but segment volume in general) that it looks like we don't have a test for `segment_volume`, do you think it could be useful?

https://github.com/neuropoly/ivado-medical-imaging/issues/187

ivadomed/loader/loader.py
```diff
@@ -437,30 +445,44 @@ def __getitem__(self, index):
...
+            'gt': torch.zeros(data_dict['input'].shape[0], shape_x, shape_y, shape_z),
+            'input_metadata': seg_pair_slice['input_metadata'],
+            'gt_metadata': seg_pair_slice['gt_metadata']
+        }
+
+        for idx in range(len(data_dict['input'])):
+            subvolumes['input'] = data_dict['input'][:,
+                                  coord['x_min']:coord['x_max'],
+                                  coord['y_min']:coord['y_max'],
+                                  coord['z_min']:coord['z_max']]
```
charleygros, line 477: I only had a quick look: Is `idx` not used?

ivadomed/loader/loader.py
```diff
@@ -437,30 +445,44 @@ def __getitem__(self, index):
...
+        shape_x = coord["x_max"] - coord["x_min"]
+        shape_y = coord["y_max"] - coord["y_min"]
+        shape_z = coord["z_max"] - coord["z_min"]
+
+        subvolumes = {
+            'input': torch.zeros(data_dict['input'].shape[0], shape_x, shape_y, shape_z),
+            'gt': torch.zeros(data_dict['input'].shape[0], shape_x, shape_y, shape_z),
+            'input_metadata': seg_pair_slice['input_metadata'],
+            'gt_metadata': seg_pair_slice['gt_metadata']
+        }
```
charleygros, line 471: Nice change

ivadomed/utils.py
```diff
@@ -652,23 +652,27 @@ def segment_volume(folder_model, fname_image, fname_roi=None):
         context["slice_filter"]["filter_empty_mask"] = False
 
     # Load data
-    filename_pairs = [([fname_image], None, fname_roi, [{}])]
+    filename_pairs = [([fname_image], None, [fname_roi], [{}])]
```
charleygros, line 655: ~~

ivadomed/loader/loader.py
andreanne-lemay: Updated :) thanks!

testing/test_segment_volume.py
```diff
@@ -0,0 +1,105 @@
...
+                             out_channel=1,
+                             depth=2,
+                             drop_rate=DROPOUT,
+                             bn_momentum=BN)
+
+    # temporary folder that will be deleted at the end of the test
+    if not os.path.exists(PATH_MODEL):
+        os.mkdir(PATH_MODEL)
+
+    torch.save(model, os.path.join(PATH_MODEL, "model_test.pt"))
```
charleygros, line 33: Good idea :)

charleygros: In the future we could consider download models from OSF

testing/test_segment_volume.py
```diff
@@ -0,0 +1,105 @@
...
+        "slice_filter_roi": 10,
+        "slice_axis": "axial",
+        "unet_3D": False,
+    }
+    PATH_CONFIG = os.path.join(PATH_MODEL, 'model_test.json')
+    with open(PATH_CONFIG, 'w') as fp:
+        json.dump(config, fp)
+
+    nib_img = imed_utils.segment_volume(PATH_MODEL, IMAGE_PATH, ROI_PATH)
+    assert nib_img.shape == nib.load(IMAGE_PATH).shape
```
charleygros, line 56: Add data type test? plus range of values (not sure if it will work with init_Unet)?

testing/test_segment_volume.py
```diff
@@ -0,0 +1,105 @@
...
+    with open(PATH_CONFIG, 'w') as fp:
+        json.dump(config, fp)
+
+    nib_img = imed_utils.segment_volume(PATH_MODEL, IMAGE_PATH)
+    assert nib_img.get_fdata()[..., 0].shape == nib.load(IMAGE_PATH).shape
+
+    shutil.rmtree(PATH_MODEL)
+
+
+print("Test segment volume with 2D model.")
```
charleygros, line 101: Really like it! Many thanks @andreanne-lemay

testing/test_segment_volume.py
andreanne-lemay: Good idea! I'll add it :)

andreanne-lemay: For the data type, segment_volume returned float64, which is not necessary. I forced the output to be float32 in pred_to_nib since it's the datatype used in the whole pipeline (never float64). Also, float32 rather than intXX because we still want the option to have a soft output (our value range is between 0 and 1, so casting in int returns a hard mask).

testing/test_segment_volume.py
```diff
@@ -0,0 +1,110 @@
...
+    nib_img = imed_utils.segment_volume(PATH_MODEL, IMAGE_PATH)
+    assert nib_img.get_fdata()[..., 0].shape == nib.load(IMAGE_PATH).shape
+    assert (nib_img.dataobj.max() < 1.0) and (nib_img.dataobj.max() > 0.0)
+    assert nib_img.dataobj.dtype == 'float32'
+
+    shutil.rmtree(PATH_MODEL)
+
+
+print("Test segment volume with 2D model.")
+test_segment_volume_2d()
```
jcohenadad, line 107: i'm coming a bit late (PR already merged) but why calling this function here? With pytest (which [seems to be what we are using](https://github.com/neuropoly/ivado-medical-imaging/blob/master/requirements.txt#L8)), all tests are run if they are defined with `def test_my_function` (which already the case), without the need to call them explicitly.

andreanne-lemay: It's indeed not required, it was this way on our other tests + it is easier to debug with an IDE (simply run file), but it is not very elegant. I'll change the tests in the next PR to remove those explicit calls.

jcohenadad: yes, i suggest to remove them as it is confusing (at least it confused me)-- plus: you can run pytest within most IDE (e.g. Pycharm's debug mode allows you to run nose, pytest, etc.): Run > Edit configuration... > template > Python tests > pytest

PR #222, opened 2020-05-12
Modify image orientation

andreanne-lemay:
Images are loaded, then are oriented in RAS. According to slice axis (axial, coronal or sagittal), this creates images with varying dimensions relatively to height, width and depth. For instance, to extract a 2D slice, this is needed:
```python
        for data_object in input_dataobj:
            if slice_axis == 2:
                input_slices.append(np.asarray(data_object[..., slice_index],
                                               dtype=np.float32))
            elif slice_axis == 1:
                input_slices.append(np.asarray(data_object[:, slice_index, ...],
                                               dtype=np.float32))
            elif slice_axis == 0:
                input_slices.append(np.asarray(data_object[slice_index, ...],
                                               dtype=np.float32))
```
Currently, the `Resample` transformation only works for axial slices since the zooms are extracted from the 2 first dimensions:
```python
            input_meta_dict.append(imed_loader_utils.SampleMetadata({
                "zooms": handle.header.get_zooms()[:2],
                "data_shape": handle.header.get_data_shape()[:2],
            }))
```
In ToTensor transformation, F.to_tensor() method used converts numpy arrays into torch tensors, but also apply a transpose, changing the dimensions.

To resolve this issues, all the images are reoriented with the following dimensions: (batch number, channel number, height, width, depth). This is coherent with the refactor transform PR #213 that will use only numpy. Reorienting according to dimensions will be needed for this PR if we want to get rid of image list.

### DIFF
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -349,11 +349,16 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False, 
         # create data and stack on depth dimension
         arr = np.stack(tmp_lst, axis=-1)
 
+        # Reorient data
+        arr_pred_ref_space = reorient_image(arr, slice_axis, nib_ref, nib_ref_can)
+
     else:
         arr = data_lst
-
-    # Reorient data
-    arr_pred_ref_space = reorient_image(arr, slice_axis, nib_ref, nib_ref_can)
+        n_channel = arr.shape[0]
+        oriented_volumes = []
+        for i in range(n_channel):
+            oriented_volumes.append(reorient_image(arr[i, ], slice_axis, nib_ref, nib_ref_can))
+        arr_pred_ref_space = np.asarray(oriented_volumes).transpose((1, 2, 3, 0))
 
     if bin_thr >= 0:
         arr_pred_ref_space = imed_postpro.threshold_predictions(arr_pred_ref_space, thr=bin_thr)
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -308,7 +308,8 @@ class Evaluation3DMetrics(object):
         return dct, self.data_painted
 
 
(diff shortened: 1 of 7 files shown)


ivadomed/utils.py
```diff
@@ -345,18 +346,24 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
...
+        oriented_volumes = []
+        for i in range(n_channel):
+            oriented_volumes.append(reorient_image(arr[i, ], slice_axis, nib_ref, nib_ref_can))
+        # transpose to locate the channel dimension at the end to properly see image on viewer
+        arr_pred_ref_space = np.asarray(oriented_volumes).transpose((1, 2, 3, 0))
 
     if bin_thr >= 0:
         arr_pred_ref_space = imed_postpro.threshold_predictions(arr_pred_ref_space, thr=bin_thr)
-    else:  # discard noise
+    elif discard_noise:  # discard noise
```
charleygros, line 366: Ok fair enough :)

charleygros:
The only thing that I am scared of is that a data flip (eg L-R) may happen somewhere. Would you like me to check this? or Have you check this already?

andreanne-lemay:
> The only thing that I am scared of is that a data flip (eg L-R) may happen somewhere. Would you like me to check this? or Have you check this already?

I checked the orientation throughout the training and the slices were always oriented the same way. I also did a quick training in 2D and in 3D and my predictions are good. If there were a problem with orientations, I believe the model would have poor predictions. Also I removed all the unecessary axis swaps + transpose, so I don't see were these flips would come from.

jcohenadad:
> > The only thing that I am scared of is that a data flip (eg L-R) may happen somewhere. Would you like me to check this? or Have you check this already?
> 
> I checked the orientation throughout the training and the slices were always oriented the same way. I also did a quick training in 2D and in 3D and my predictions are good. If there were a problem with orientations, I believe the model would have poor predictions. Also I removed all the unecessary axis swaps + transpose, so I don't see were these flips would come from.

sounds like a good candidate for a specific unit test?

PR #224, opened 2020-05-14
Lr/script manual labeling

lrouhier:
This PR aims at adding a script to label manually the intervertebral disc in sct_testing large. 
This uses the sct_label_utils function in a shell script. 

Changes includes: 
- Addition of script files 
- Addition of SOP
- Addition of the list of subjects to process for this task.


jcohenadad:
@lrouhier after giving it some thoughts (and discussing with Charley), we thought it might be better to have such scripts under neuropoly/internal-tools repos, given that this is internal to the lab (and given the dependency with SCT). What do you think?

lrouhier:
@jcohenadad Yes I think This could be useful. Does the repo already exist because I didn't see it?

jcohenadad:
@lrouhier just added you

lrouhier:
moved to https://github.com/sct-pipeline/manual-labeling

Issue #218, opened 2020-05-07
Integration of object detection model


andreanne-lemay:
Currently, SCT has different machine learning and deep learning models. These models don't currently have a framework to easily retrain them on custom data (or the models are in Tensorflow). Ivado-medical-imaging offers a framework to train models, so integrating an object detection model could solve this issue. This way, we could have a flexible model to detect custom ROI (e.i. the spinal cord). The output of object detection models is a bounding box, so a second model could then be applied to the ROI contained in the bounding box.

Popular object detection models:
- YOLOv3 
- Faster RCNN
- SSD 

These models are 2D object detection models which can be an issue. Recombining the 2D boxes in one 3D bounding box is not always trivial. A model directly generating a 3D bounding box could be more fitted for medical images.

jcohenadad:
awesome suggestion! to list some concrete examples of how this could be used:
- detection of ponto-medullary junction (currently `sct_detect_pmj` @charleygros )
- detection of C2-C3 disc (@lrouhier)
- detection of spinal cord on sagittal data, to be followed by a cropping and size-adjusted CNN for tumor detection (@andreanne-lemay)

I'm sure we can think of other applications

charleygros:
> awesome suggestion! to list some concrete examples of how this could be used:
> 
>     * detection of ponto-medullary junction (currently `sct_detect_pmj` @charleygros )
> 
>     * detection of C2-C3 disc (@lrouhier)
> 
>     * detection of spinal cord on sagittal data, to be followed by a cropping and size-adjusted CNN for tumor detection (@andreanne-lemay)
> 
> 
> I'm sure we can think of other applications

Could add [Figure 3](https://arxiv.org/pdf/1805.06349.pdf).

charleygros:
@jcohenadad --> would it be relevant to integrate [OptiC training pipeline](https://github.com/neuropoly/spinalcordtoolbox/blob/master/dev/detect_c2c3/train.py)?
The framework does exit.

charleygros:
A crucial point would be to allow **end-to-end training** if this module is used for initialising a segmentation tool.

charleygros:
We could start implementing a simple DenseNet for Detection baseline?
It would also imply to create a module of patch generation that will be plug to the current loader :)

andreanne-lemay:
> A crucial point would be to allow **end-to-end training** if this module is used for initialising a segmentation tool.

[...]

PR #240, opened 2020-05-26
Object detection features to find ROI

andreanne-lemay:
Integration of the elements to crop images according to bounding boxes (#218 ). 
Changes:
- Addition of a new folder for object detection. For now there is only a utils.py file, eventually we could add one or different models (models.py) and maybe the custom parts of the bounding box loader (loader.py).
- Function for generation of a json file containing  bounding box coordinates in the following format:
```
{"path/to/img.nii.gz": [height_max, height_min, width_max, width_min, depth_max, depth_min]}
```
The bounding boxes are generated this way: 
1) Train UNet model to segment the ROI (e.i. spinal cord). 
2) Predict ROI segmentation 
3) Generate bounding boxes from segmentation
- Addition of new transform BoundingBoxCrop
- Support to resample bounding boxes and generate dimensions divisible by 16 (requirement for our current version of 3D unet)
- Add GPU option on `segment_volume`. -> To generate the ROI segmentation, the code from `segment_volume` is reused. Using a GPU accelerates the bounding box generation.

To do: 
- [x] Add 2D support
- [x] Add test
- [x] Update all config files
- [x] Refactor of some classes in loader to pass context variable instead of multiple parameters

In future PR:
- Integrate object detection unet for 3D images (e.i. Retina Unet)


ivadomed/loader/loader.py
```diff
@@ -504,7 +538,7 @@ def __init__(self, root_dir, subject_lst, target_suffix, contrast_lst, contrast_
 class BidsDataset(MRI2DSegmentationDataset):
```
andreanne-lemay, line 538: Here I was thinking to reduce the number of input parameters by passing directly the `context` variable. The instantiation of these classes is [massive](https://github.com/neuropoly/ivado-medical-imaging/pull/240/files#diff-edd3e7e75e956aeb056f2c88401ec9eaR520) . @charleygros do you see any down sides to this approach?

charleygros:
>     * Function for generation of a json file containing  bounding box coordinates in the following format:
> 
> 
> ```
> {"path/to/img.nii.gz": [height_max, height_min, width_max, width_min, depth_max, depth_min]}
> ```

Great idea! --> could be further used as metadata in the segmentation Module.

For the implementation / format --> we may want to keep in mind the "multiple object case" as in the future we may want to use this avenue for the MS lesions for instance.

ivadomed/loader/loader.py
charleygros: I felt the same while doing the refactoring of `main.py`.
At the same time, I do not want to fall in the other extrem, eg `loader(context)` --> with `context` being an obscure dictionary with all the parameters --> makes it "back-box".
What I am suggesting for the `main.py` (open to discussion) would be a middle ground between the two: params grouped by "application". For instance, I have:
```
def train(model_params, dataset_train, dataset_val, training_params, log_directory, cuda_available=True, metric_fns=None, debugging=False):
```
Where `model_params` and `training_params` are dictionary containing parameters related to model and training_strategy, resp.

andreanne-lemay:
> > ```
> > * Function for generation of a json file containing  bounding box coordinates in the following format:
> > ```
> > 
> > 
> > ```
> > {"path/to/img.nii.gz": [height_max, height_min, width_max, width_min, depth_max, depth_min]}
> > ```
> 
> Great idea! --> could be further used as metadata in the segmentation Module.
> 
> For the implementation / format --> we may want to keep in mind the "multiple object case" as in the future we may want to use this avenue for the MS lesions for instance.

Yes good point! How about a list of coordinate list -> could then convert into a numpy array after the loading. What do you think?

ivadomed/loader/loader.py
andreanne-lemay: Okay great! I won't modify it since you already made the changes in the main refactor PR :)!

charleygros:
> Yes good point! How about a list of coordinate list -> could then convert into a numpy array after the loading. What do you think?

Yep great :-)
Do we want to also consider multi-label detection? in that case we would have something like:
```json
{"path/to/img.nii.gz": {
    "label_1": [[height_max, height_min, width_max, width_min, depth_max, depth_min]],
    "label_2: [[height_max, height_min, width_max, width_min, depth_max, depth_min]]
    }
}
```

ivadomed/config/config_spineGeHemis.json
```diff
@@ -3,6 +3,7 @@
     "gpu": 0,
     "log_directory": "HeMIS",
     "debugging": false,
+    "object_detection_path": null,
```
charleygros, line 6: What about having directly in the config file / loader_params a dict `object_detection_params` with `path`. Then in main, add to this dict "gpu" and "log_directory" etc.?

ivadomed/transforms.py
```diff
@@ -544,6 +546,22 @@ def __call__(self, sample, metadata=None):
             return sample, metadata
 
 
+class BoundingBoxCrop(Crop):
```
charleygros, line 549: Great!
Were you considering adding a undo_transform for testing time purposes? or maybe thats done somewhere else? or for another PR..

ivadomed/loader/loader.py
```diff
@@ -297,8 +307,18 @@ def _load_filenames(self):
...
 
+            resample = False
+            for idx, transfo in enumerate(self.transform.transform["im"].transforms):
+                if "Resample" in str(type(transfo)):
+                    resample = True
+                    resample_param = (transfo.hspace, transfo.wspace, transfo.dspace)
+
             for idx_pair_slice in range(input_data_shape[-1]):
                 slice_seg_pair = seg_pair.get_pair_slice(idx_pair_slice)
+                self.bounding_box = 'bounding_box' in slice_seg_pair['input_metadata'][0]
```
charleygros, line 318: I am not sure how to fix this but.. here we assume that all images of the dataset has bounding_box in its metadata. If not, it could (if unlucky) crash later.

ivadomed/object_detection/utils.py
```diff
@@ -0,0 +1,152 @@
...
+        dimensions = []
+        for i in range(len(coords)):
+            dimensions.append(int(coords[i].min()))
+            dimensions.append(int(coords[i].max()))
+        bounding_boxes.append(dimensions)
+
+    return bounding_boxes
+
+
+def adjust_bb_size(bounding_box, factor, multiple_16, resample=False):
```
charleygros, line 37: I know you already explained to me via Slack the purpose of this function, multiple_16 etc. but could you please help me to remember? via a docstring? :-)

ivadomed/loader/loader.py
jcohenadad: maybe @Drulex has some ideas?

Drulex: Some general comments without digging too deep:
1. `bounding_box` -> `has_bounding_box` or something more representative since `bounding_box` could mean an actual box/coordinates of a box (this is what I thought at first).
2. If `bounding_box` is local to this function/iteration it should not be a class attribute. I am confused that `self.bounding_box` is set in each iteration of this loop, but later referenced in `__get_item()__`. Is it supposed to be a global flag to ensure that all `slice_seg_pair` have a `bounding_box`? In that case it is better to do something like this:

```
# in __init__() have self.has_bounding_box = True
# [...]
for idx_pair_slice in range(input_data_shape[-1]):
    slice_has_bounding_box = 'bounding_box' in slice_seg_pair['input_metadata'][0]
    self.has_bounding_box &= slice_has_bounding_box # notice bitwise operator here

# [...] later when we want to know if all images have the bounding box
if not self.has_bounding_box:
    # log appropriate message
    # raise ValueError("Some images are missing bounding_box in their metadata!") or something more useful
```
^ this assumes that if we are ok with a potential `IndexError` if `slice_seg_pair['input_metadata'][0]` does not exist, otherwise it's better to sanitize the data somewhere in the initialization.

ivadomed/transforms.py
andreanne-lemay: The undo is accessible to its parent Crop :)!

ivadomed/object_detection/utils.py
andreanne-lemay: For Unet 3D I need patch dimensions of 16, with the training by patch (as discussed) I'll be able to get rid of this :)!

ivadomed/config/config_spineGeHemis.json
andreanne-lemay: Fixed :)

ivadomed/loader/loader.py
andreanne-lemay: Fixed it with imed_obj_detect.verify_metadata -> verifies all input_metadata, gt_metadata and indices have the 'bounding_box' key

ivadomed/loader/loader.py
Drulex: Another potential solution to keep the code clean and avoid repeating arguments is to pass keyword arguments.

For example:
```
def func_that_takes_a_lot_of_params(some_required_param, **ctx):
    """
    Info that includes what is expected to be in context
    """
    self.some_required_param = some_required_param

    # perform any param checking early and raise if something is wrong
    if "something_important" not in ctx:
        raise ValueError("Something important is missing!")

    # store context 
    self.context = dict()

    for k,v in ctx.items():
        # perform any param-specific initialization here
        if k == "param_that_needs_special_handling":
            do_something_special(k)

        # keep a reference to the context params in self.context
        self.context[k] = v
```

This allows caller to use the function in 2 ways:

```
context = dict(param1=1, param2=2, param3=3)
func_that_takes_a_lot_of_params(some_param, **context)
```
or
```
func_that_takes_a_lot_of_params(some_param, param1=1, param2=2, param3=3)
```

charleygros: Many thanks for this tip @Drulex !! great great
I have opened an issue [here](https://github.com/neuropoly/ivado-medical-imaging/issues/275).
I would be happy to implement it in a dedicated PR

andreanne-lemay:
Following changes were made:
- 3D Loader:
  - Training and testing by patch
  - Change padding (overlap) to stride (amount of movement) for the patches
  - Addition of 'coord' parameter in metadata for volume reconstruction
  
- Tranforms:
  - 'crop_params' is now a dictionary where the key is the class name. This is needed in cases where there are 2 types of cropping (ex: Bounding box (Resample) CenterCrop). If there is only one 'crop_params' then the transforms are not undone properly (since they have different crop_params)
  - Addition of 'preresample_shape' param. When undoing  the transform Resample, the original data shape was used, but this supposes that Resample is the first parameter applied.

- segment_volume:
  - GPU support of segment_volume.py if available. To generate the spinal cord tumor mask I use the function segment_volume.py, but with cpu only the first detection step is very long.

The BoundingBoxCrop is applied after Resample, because the other way around required adding 2 transforms automatically rather than one (BoundingBoxCrop, Resample, CenterCrop).

Glad to discuss/modify these topics :)

charleygros:
This is a significant addition to the repo @andreanne-lemay Many Many thanks!

charleygros:
Before reviewing the PR, I am not exactly sure of what would be the typical succession of the different events, eg detection module, BBox data generation, segment_volume for mask generation, patch extraction, resample, BBoxCrop, CenterCrop...
Could you please provide us an overview of your pipeline? That would really help me. Thanks in advance.

andreanne-lemay:
> Before reviewing the PR, I am not exactly sure of what would be the typical succession of the different events, eg detection module, BBox data generation, segment_volume for mask generation, patch extraction, resample, BBoxCrop, CenterCrop...
> Could you please provide us an overview of your pipeline? That would really help me. Thanks in advance.

Gladly :) 
1. Load images with Bids loader
    - Load bounding box file (with function `load_bounding_boxes`) if the path exists or create a json file if the param 'object_detection_path' is not None.
    - Add 'bounding_box' param in metadata that is unique to every image
2. Prepare indices (load_filenames)
    - The preprocessing transforms are applied to seg_pair. If the 'bounding_box' param exists throughout all metadata -> BoundingBoxCrop is automatically added after the Resample transform (if it exists) with function `adjust_transforms`.  For the 3D loader, since the training is by patch now, I modify the bounding box so that it is at least the size of the patch and so that there are a whole number of strides (avoids mismatch dimension errors during training). 
3. Training as usual
4. Testing: During testing transformations are undone. Since all images don't have the same bounding box params (different size) the transform is adapted to the specific metadata of the image with function `adjust_undo_transforms`.
   - Volume reconstruction in 3D: the transformed whole image is initialize to zero. For every coordinates of a given image, prediction are added to this image and a weight matrix is updated (number of predictions for a given voxel). At the last patch of the image, the image adjusted with the weight matrix.

Note: For 2D training, since it works by slice rather than by patch, a cropping method (CenterCrop) needs to be added to the transforms of the config file. This way, all the cropped images by the bounding box are the same size. We might also want to consider implementing patch wise training in 2D one day.

jcohenadad:
> We might also want to consider implementing patch wise training in 2D one day.

ha! @mariehbourget will definitely be interested in that

ivadomed/config/config.json
```diff
@@ -3,6 +3,9 @@
     "gpu": 7,
     "log_directory": "spineGeneric",
     "debugging": false,
+    "object_detection_params": {
+        "object_detection_path": null
```
jcohenadad, line 7: in another PR (https://github.com/neuropoly/ivado-medical-imaging/pull/274) we discussed the use of "null" in json files, vs. empty string. We should probably settle on a convention.
@Drulex @kousu do you have any suggestions about what to use / not to use?

charleygros: Simple note: in #274, empty string have been used instead of `null`.

ivadomed/transforms.py
```diff
@@ -545,6 +548,22 @@ def __call__(self, sample, metadata=None):
...
+    """
+    Crops image according to given bounding box
+    """
+    @multichannel_capable
+    @two_dim_compatible
+    def __call__(self, sample, metadata):
```
charleygros, line 557: great!

ivadomed/transforms.py
```diff
@@ -872,6 +891,11 @@ def get_subdatasets_transforms(transform_params):
 
 
 def get_preprocessing_transforms(transforms):
+    """
+    Processes all the transforms and returns only the ones labeled for preprocessing
```
charleygros, line 895: What about
```
Checks the transformations parameters and selects the transformations which are done during preprocessing only.
```

ivadomed/transforms.py
```diff
@@ -884,21 +908,35 @@ def get_preprocessing_transforms(transforms):
...
-    stack_gt, metadata_gt = transforms(sample=seg_pair["gt"],
-                                       metadata=seg_pair['gt_metadata'],
-                                       data_type="gt")
+    if seg_pair['gt_metadata'][0] is not None:
+        metadata_gt = imed_loader_utils.update_metadata(metadata_input, seg_pair['gt_metadata'])
+        stack_gt, metadata_gt = transforms(sample=seg_pair["gt"],
+                                           metadata=metadata_gt,
+                                           data_type="gt")
+    else:
+        stack_gt, metadata_gt = seg_pair["gt"], []
```
charleygros, line 938: Have you try using this option?
https://github.com/neuropoly/ivadomed/blob/master/ivadomed/transforms.py#L120

ivadomed/transforms.py
```diff
@@ -914,3 +952,21 @@ def apply_preprocessing_transforms(transforms, seg_pair, roi_pair=None):
...
+    """
+    This function seperates the preprocessing transforms from the others and generates the undo transforms related.
+    :param transform_dict (dict): Dictionary containing the transforms and there parameters
+    :param requires_undo (bool): Boolean indicating if transforms can be undone
+    :return: transform lst containing the preprocessing transforms and regular transforms, UndoCompose object containing
+    the transform to undo
+    """
+    training_undo_transform = None
+    if requires_undo:
+        training_undo_transform = UndoCompose(Compose(transform_dict.copy()))
```
charleygros, line 967: For the undo, we do not need to split prepro vs non_prepro because thery are all done only once, correct?

ivadomed/utils.py
```diff
@@ -433,8 +433,9 @@ def segment_volume(folder_model, fname_image, fname_roi=None):
     preds_list, slice_idx_list = [], []
     for i_batch, batch in enumerate(data_loader):
         with torch.no_grad():
-            img = batch['input']
+            img = batch['input'].cuda() if torch.cuda.is_available() else batch['input']
```
charleygros, line 436: What about using:
https://github.com/neuropoly/ivadomed/blob/master/ivadomed/utils.py#L469

testing/test_bounding_box.py
```diff
@@ -0,0 +1,97 @@
...
+
+    handler = ds.handlers if "UNet3D" in config else ds.indexes
+    for index in handler:
+        seg_pair, _ = index
+        if "UNet3D" in config:
+            assert seg_pair['input'][0].shape[-3:] == (mx2 - mx1, my2 - my1, mz2 - mz1)
+        else:
+            assert seg_pair['input'][0].shape[-2:] == (mx2 - mx1, my2 - my1)
+
+    shutil.rmtree(LOG_DIR)
```
charleygros, line 97: Gucci

ivadomed/main.py
```diff
@@ -152,15 +156,16 @@ def run_main():
             transformation_dict = transform_test_params
         else:
             transformation_dict = transform_valid_params
+
+        # UNDO TRANSFORMS
+        _, undo_transforms = imed_transforms.preprare_transforms(transformation_dict.copy())
+
```
charleygros, line 162: When looking inside `prepare_transforms`: I don't see why we are calling this function here:
```python
undo_transforms = imed_transforms.UndoCompose(imed_transforms.Compose(transformation_dict))
```
as before is enough I think.
Either way I am happy. But I am missing something?

ivadomed/testing.py
```diff
@@ -179,36 +188,63 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
-                pred_tmp_lst.append(preds_idx_arr)
+                    # add new sample to pred_tmp_lst, of size n_label X h X w ...
+                    pred_tmp_lst.append(preds_idx_arr)
 
-                # TODO: slice_index should be stored in gt_metadata as well
-                z_tmp_lst.append(int(batch['input_metadata'][smp_idx][0]['slice_index']))
-                fname_tmp = fname_ref
+                    # TODO: slice_index should be stored in gt_metadata as well
+                    z_tmp_lst.append(int(batch['input_metadata'][smp_idx][0]['slice_index']))
+                    fname_tmp = fname_ref
```
charleygros, line 196: I think this does not work: if `pred_tmp_lst` is not fed, then we never go into the "NEW COMPLETE VOLUME" section --> Niftis aren't saved.
I tried to undo this change locally and it seems to fix the issue.
What are your thoughts?

ivadomed/config/config_tumorSeg.json
```diff
@@ -3,11 +3,15 @@
     "gpu": 0,
     "log_directory": "tumor_segmentation",
     "debugging": true,
+    "object_detection_params": {
+        "object_detection_path": "/home/andreanne/Documents/models/sc_model_30mm",
+        "safety_factor": [1.1, 1.1, 1.0]
+    },
```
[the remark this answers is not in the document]
andreanne-lemay: Done :)

ivadomed/loader/utils.py
```diff
@@ -227,7 +227,7 @@ def items(self):
 
     def _update(self, ref, list_keys):
         for k in list_keys:
-            if k not in self.metadata.keys() and k in ref.metadata.keys():
+            if (k not in self.metadata.keys() or not bool(self.metadata[k])) and k in ref.metadata.keys():
```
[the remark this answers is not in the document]
andreanne-lemay: Now that 'crop_params' value is a key-value and is initialize to {}, the parameter is always present in metadata. bool(self.metadata[k]) verifies if the value is empty ({}) or not ({ROICrop: ...}).

ivadomed/transforms.py
```diff
@@ -206,6 +208,7 @@ def __call__(self, sample, metadata=None):
         # Get params
         # Voxel dimension in mm
         is_2d = sample.shape[-1] == 1
+        metadata['preresample_shape'] = sample.shape
```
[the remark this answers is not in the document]
andreanne-lemay: `data_shape` holds the initial dimensions of the image. If Resample is the first transform, `preresample_shape` and `data_shape` have the same value. If Resample is not the first transform, for example if I crop the image to a size of 100x100 before resampling, I want the size of the undone resampled image to be 100x100 and not the initial shape of the image. Using `data_shape` implies Resample will always be used before any cropping, but this way is more versatile and doesn't assume the transform order.

ivadomed/transforms.py
```diff
@@ -335,7 +338,7 @@ def __call__(self, sample, metadata):
         # Get params
         is_2d = sample.shape[-1] == 1
         th, tw, td = self.size
-        fh, fw, fd, h, w, d = metadata['crop_params']
+        fh, fw, fd, h, w, d = metadata['crop_params'][self.__class__.__name__]
```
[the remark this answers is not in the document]
andreanne-lemay: Exactly :), currently for 2D images with object detection, we need to apply a CenterCrop after the BoundingBoxCrop to ensure all batches have the same size (since we don't have patch-wise training in 2D). So I added this to accommodate this need.

ivadomed/transforms.py
```diff
@@ -403,7 +406,7 @@ class ROICrop(Crop):
     def __call__(self, sample, metadata=None):
         # If crop_params are not in metadata,
         # then we are here dealing with ROI data to determine crop params
-        if 'crop_params' not in metadata:
+        if self.__class__.__name__ not in metadata['crop_params']:
```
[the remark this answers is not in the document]
andreanne-lemay: Yes right here: https://github.com/neuropoly/ivadomed/pull/240/files#diff-edd3e7e75e956aeb056f2c88401ec9eaR232 (L232 in loader.py)

ivadomed/transforms.py
andreanne-lemay: Love it! Changed it

ivadomed/transforms.py
andreanne-lemay: I am not sure I understand :). Are you asking if it respects the conditions in the line you cited? if so, yes it falls under these conditions since `len(metadata) == 0`

ivadomed/transforms.py
andreanne-lemay: Exactly :) we still want to undo all the transforms

ivadomed/utils.py
andreanne-lemay: Changed it!

ivadomed/testing.py
andreanne-lemay: Yes absolutely! While modifying the code I didn't indent these lines correctly and forgot to test for 2D models! Did the modifications :) everything should be good now

ivadomed/transforms.py
charleygros: As it falls under these conditions: why do we need "if seg_pair['gt_metadata'][0] is not None: // else"?

docs/source/configuration_file.md
```diff
@@ -127,6 +127,13 @@ Uncertainty computation is performed if `n_it>0` and at least `epistemic` or `al
 - `aleatoric`: Bool. Image-based uncertainty with [test-time augmentation](https://doi.org/10.1016/j.neucom.2019.01.103).
 - `n_it`: Integer. Number of Monte Carlo iterations. Set to 0 for no uncertainty computation.
 
+## Cascaded Architecture Features
+
+### object_detection_params (Optional)
+- `object_detection_path`: String. Path to object detection model. The model's prediction will be used to generate bounding boxes.
```
charleygros, line 133: Is it the path to the model? I thought it was the path to the predictions of the first step (.json).

docs/source/configuration_file.md
```diff
@@ -127,6 +127,13 @@ Uncertainty computation is performed if `n_it>0` and at least `epistemic` or `al
...
+- `safety_factor`: List. List of length 3 containing the factors to multiply each dimension of the bounding box.
```
charleygros, line 134: Not sure: does it mean: it takes the coordinates of the prediction, compute the bounding box, then expand this bounding box with this safety factor? :-)

ivadomed/testing.py
charleygros: I ve just got this when re running it:
```
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivado-medical-imaging/ivadomed/metrics.py", line 71, in dice_score
    raise ValueError("Shape mismatch: im1 and im2 must have the same shape.")
ValueError: Shape mismatch: im1 and im2 must have the same shape.
```
But it may be because of me.. looking deeper at it.

ivadomed/testing.py
```diff
@@ -187,28 +196,55 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
+                    output_nii = imed_utils.pred_to_nib(data_lst=[pred_undo],
+                                                        z_lst=[],
+                                                        fname_ref=fname_ref,
+                                                        fname_out=fname_pred,
+                                                        slice_axis=imed_utils.AXIS_DCT[testing_params['slice_axis']],
+                                                        kernel_dim='3d',
+                                                        bin_thr=0.5 if testing_params["binarize_prediction"] else -1)
+                    preds_npy_list.append(output_nii.get_fdata().transpose(3, 0, 1, 2))
+                    gt_lst = []
+                    for gt in metadata[0]['gt_filenames']:
```
charleygros, line 239: Here there is a mismatch between fname_pred and fname_gt:
`metadata[0]['gt_filenames']` refers to the next image, which is currently being filled.

ivadomed/testing.py
```diff
@@ -172,11 +172,10 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
                                                         kernel_dim='2d',
                                                         bin_thr=0.9 if testing_params["binarize_prediction"] else -1)
-                    preds_npy_list.append(output_nii.get_fdata().transpose(3, 0, 1, 2))
-                    gt_lst = []
-                    for gt in metadata_idx[0]['gt_filenames']:
-                        gt_lst.append(nib.load(gt).get_fdata())
-                    gt_npy_list.append(np.array(gt_lst))
+                    # TODO: Adapt to multilabel
+                    preds_npy_list.append(output_nii.get_fdata()[:, :, :, 0])
+                    gt_npy_list.append(nib.load(fname_tmp).get_fdata())
```
charleygros, line 177: These lines fix [this](https://github.com/neuropoly/ivadomed/pull/240#discussion_r440092679).
This is however not robust to multi label task --> the entire section would actually benefit from some work (my bad.. !), which could be done in a future PR I think :-)

docs/source/configuration_file.md
andreanne-lemay: No it's the path to the detection model to apply segmen_volume. The .json has always the same path: log_directory/bounding_box.json

docs/source/configuration_file.md
andreanne-lemay: Yes that's pretty much it: If the original bounding box has coordinates of 10x20x30, you can apply a safety factor of for instance 1.5, so you'll get a final bounding box size of 15x30x45.

ivadomed/object_detection/utils.py
```diff
@@ -0,0 +1,268 @@
...
+from ivadomed import postprocessing as imed_postpro
+from ivadomed import transforms as imed_transforms
+from ivadomed import utils as imed_utils
+from ivadomed.loader import utils as imed_loader_utils
+
+
+def get_bounding_boxes(mask):
+    """
+    Generates a 3D bounding box around a given mask.
+    :param mask: numpy array with the mask of the ROI
```
jcohenadad, line 18: can you please use the [Google style docstrings](https://github.com/neuropoly/ivadomed/blob/master/CONTRIBUTING.md#documentation-and-docstrings)? thx

ivadomed/object_detection/utils.py
```diff
@@ -0,0 +1,268 @@
...
+            dimensions.append(int(coords[i].max()))
+        bounding_boxes.append(dimensions)
+
+    return bounding_boxes
+
+
+def adjust_bb_size(bounding_box, factor, resample=False):
+    """
+    Modifies the bounding box dimensions according to a given factor
+    :param bounding_box: coordinates of bounding box (h_min, h_max, w_min, w_max, d_min, d_max)
```
jcohenadad, line 40: i find it confusing to alternate between h/w/d and x/y/z. I suggest to stick to one convention (x/y/z seems the most common approach in our current codebase)

ivadomed/object_detection/utils.py
```diff
@@ -0,0 +1,268 @@
...
+        padding = (m - (dim_len - l) % m) if (m - (dim_len - l) % m) != m else 0
+        new_dim.append(dim_len + padding)
+    return new_dim
+
+
+def generate_bounding_box_file(subject_list, model_path, log_dir, gpu_number=0, slice_axis=0, contrast_lst=None,
+                               keep_largest_only=True, safety_factor=None):
+    """
+    Creates json file containing the bounding box dimension for each images. The file has the following format:
+    {"path/to/img.nii.gz": [[h1, h2, w1, w2, d1, d2], [h1, h2, w1, w2, d1, d2]]}
```
jcohenadad, line 81: same comment about h/w/d

ivadomed/object_detection/utils.py
```diff
@@ -0,0 +1,268 @@
...
+            if "BoundingBoxCrop" == transfo.__class__.__name__:
+                transforms.transform[img_type].transforms.pop(idx)
+            if "Resample" == transfo.__class__.__name__:
+                resample_idx[i] = idx
+
+    resample_bounding_box(seg_pair, transforms)
+    index_shape = []
+    for i, img_type in enumerate(transforms.transform):
+        h_min, h_max, w_min, w_max, d_min, d_max = seg_pair['input_metadata'][0]['bounding_box']
+        size = [h_max - h_min, w_max - w_min, d_max - d_min]
```
jcohenadad, line 157: same comment about h/w/d

ivadomed/object_detection/utils.py
```diff
@@ -0,0 +1,268 @@
...
+    bounding_box_dict = {}
+    if object_detection_params is None or object_detection_params['object_detection_path'] is None:
+        return bounding_box_dict
+    bounding_box_path = os.path.join(object_detection_params['log_directory'], 'bounding_boxes.json')
+    if os.path.exists(bounding_box_path):
+        with open(bounding_box_path, 'r') as fp:
+            bounding_box_dict = json.load(fp)
+    elif object_detection_params['object_detection_path'] is not None and \
+            os.path.exists(object_detection_params['object_detection_path']):
+        print("Generating bounding boxes...")
```
jcohenadad, line 217: we should avoid using `print()` in API functions, but i guess this is out-of-scope https://github.com/neuropoly/ivadomed/issues/10

docs/source/configuration_file.md
andreanne-lemay: I added some precision in the doc :)

ivadomed/transforms.py
andreanne-lemay: I removed the `if ... else` condition, but I had to add this https://github.com/neuropoly/ivadomed/pull/240/files#diff-05cfa4695be522c97dc494a15bf08959R280 in the clean metadata function. It is slightly cleaner. In a future PR, we might want to clean this. Instead of using None or [None] we could use [] for empty GTs this way, we can remove the `if is not None` condition since the the for loops the empty list will simply be ignored

Issue #251, opened 2020-05-31
Some pytest tests fail


andreanne-lemay:
Tests in test_transforms.py use randomly generated inputs and ground truths. This causes some tests to fail by time to time ([related comment](https://github.com/neuropoly/ivado-medical-imaging/pull/213#issuecomment-631864898)). These tests should be modified so that they systematically pass. List of tests failing:
- test_NormalizeInstance
- test_DilateGT
- test_RandomRotation
- test_RandomAffine

charleygros:
Duplication: https://github.com/neuropoly/ivado-medical-imaging/issues/236
I am closing #236

charleygros:
Adding another one that is in trouble: `test_RandomShiftIntensity`.

jcohenadad:
Is it fixed by #254?

PR #254, opened 2020-06-01
Fix testing/test_transforms.py fails

charleygros:
This PR is motivated by #251.

DONE:
- `RandomAffine` and `RandomRotation` were sometime failing as illustrated on this [beautiful drawing](https://github.com/neuropoly/ivado-medical-imaging/pull/213#issuecomment-631864898) --> `create_test_image` has now a parameter `random_position` --> set to false by default which implies that the object is located at the center of the image. Plus, the testing parameters were chosen so that the object would not go out of the image boundaries.
- Minor fix for `test_NormalizeInstance`
- I am experimenting unexpected behaviour of `RandomRotation` which would be investigated in another PR (reported in the issue XX) --> I created `RandomTranslation` so that DA with translation can still be used while we are fixing `RandomRotation` as rotation and translation were interweaved within `RandomAffine`.
- I think `DilateGT` could be further improved: the holes are not quite realistic, they should only be located on the periphery of the dilated area (--> it will be also further explored in a future PR, reported in issue XX). In this issue, the behaviour of `DilateGT` has been simplified (ie remove the holes generation) so that the tests run smoothly. The tests failed because some holes were located within the true_lesion (ie non dilated area).

### DIFF
--- a/ivadomed/config/config_sctTesting_lesionAx.json
+++ b/ivadomed/config/config_sctTesting_lesionAx.json
@@ -88,8 +88,8 @@
         "ROICrop": {
             "size": [48, 48]
         },
-        "RandomRotation": {
-            "degrees": 30,
+        "RandomTranslation": {
+            "translate": [0.1, 0.3],
             "applied_to": ["im", "gt"],
         },
         "NumpyToTensor": {"applied_to": ["im", "gt"]},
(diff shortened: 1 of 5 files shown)


charleygros:
New candidate: `RandomShiftIntensity` --> investigating

charleygros:
The issues were:
- `test_NormalizeInstance`: modify test: we do a NumpyToTensor prior to the test --> we need to assess the mean and std of the entire "batch" instead of each sample.
- `test_RandomShiftIntensity`: implement more specific test to assess intensity change
- `test_DilateGT`: as explained in the first post: simplified function as it needs further dev #255. The "muted" behaviour was the one responsible for the fail.

PR #256, opened 2020-06-01
Add Tversky and Focal Tversky Losses

charleygros:
#### Original papers:
- [Tversky Loss 2017](https://arxiv.org/pdf/1706.05721.pdf)
- [Tversky Focal Loss 2018](https://arxiv.org/pdf/1810.07842.pdf)

#### Motivation:
Have been recommended for high class imbalance issues.

#### Limitation:
Some hyperparameters to optimise.

#### Background:
- Tversky Index is a asymmetric similarity measure
- If alpha=beta=0.5 --> Tversky index is equivalent to the Dice Score
- If alpha=beta=1 --> Tversky index is equivalent to the Jaccard Index
- alpha > beta --> penalise false negatives
- Focal Tversky Loss --> Same idea than Focal Loss: non-linear nature of the loss gives control over how the loss behaves at different values of the Tversky index obtained.
- For gamma > 1: the loss is higher for samples where Tversky < 0.5 --> the model focus on harder samples.

### DIFF
--- a/ivadomed/losses.py
+++ b/ivadomed/losses.py
@@ -210,7 +210,7 @@ class FocalTverskyLoss(TverskyLoss):
             # Compute Tversky index
             tversky_index = self.tversky.tversky_index(y_pred, y_true)
             # Compute Focal Tversky loss, Equation 4 in the original paper
-            focal_tversky_sum += torch.pow(1-tversky_index, exponent=1/self.gamma)
+            focal_tversky_sum += torch.pow(1 - tversky_index, exponent=1 / self.gamma)
 
         # TODO: Take the opposite?
         return focal_tversky_sum / n_classes
--- a/ivadomed/losses.py
+++ b/ivadomed/losses.py
@@ -124,7 +124,7 @@ class TverskyLoss(nn.Module):
 
     """
 
-    def __init__(self, alpha=0.3, beta=0.7, smooth=1.0):
+    def __init__(self, alpha=0.7, beta=0.3, smooth=1.0):
         """
         Args:
             alpha (float): weight of false positive voxels
@@ -158,7 +158,7 @@ class TverskyLoss(nn.Module):
         fp = torch.sum((1 - y_true) * y_pred)
         # Compute Tversky for the current class, see Equation 3 of the original paper
         numerator = tp + self.smooth
-        denominator = tp + self.alpha * fn + self.beta * fp + self.smooth
+        denominator = tp + self.alpha * fp + self.beta * fn + self.smooth
         tversky_label = numerator / denominator
         return tversky_label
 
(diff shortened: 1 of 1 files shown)


ivadomed/losses.py
```diff
@@ -114,3 +114,103 @@ def forward(self, input, target):
...
+        """
+        # Compute TP
+        tp = torch.sum(y_true * y_pred)
+        # Compute FN
+        fn = torch.sum(y_true * (1 - y_pred))
+        # Compute FP
+        fp = torch.sum((1 - y_true) * y_pred)
+        # Compute Tversky for the current class, see Equation 3 of the original paper
+        numerator = tp + self.smooth
+        denominator = tp + self.alpha * fn + self.beta * fp + self.smooth
```
olix86, line 161: It's a detail, but I think the alpha and beta are switched up, from the paper : 
> where α and β control the magnitude of penalties for FPs and FNs, respectively.

olix86: The confusion probably comes from `eq3` since they define :
> the `p0i` is the probability of voxel `i` be a lesion and `p1i` is the probability of voxel `i` be a non-lesion

Which is kind of the opposite of what I would have expected

charleygros: Mmm this is confusing I agree. I ve just swapped!

ivadomed/losses.py
```diff
@@ -114,3 +114,103 @@ def forward(self, input, target):
...
+        # TODO: Add class_of_interest?
+        for i_label in range(n_classes):
+            # Get samples for a given class
+            y_pred, y_true = input[:, i_label, ], target[:, i_label, ]
+            # Compute Tversky index
+            tversky_index = self.tversky.tversky_index(y_pred, y_true)
+            # Compute Focal Tversky loss, Equation 4 in the original paper
+            focal_tversky_sum += torch.pow(1 - tversky_index, exponent=1 / self.gamma)
+
+        # TODO: Take the opposite?
```
andreanne-lemay, line 215: Here what do you mean with this TODO :)?

charleygros: We have the convention to minimize our loss during training, ie -1=good, 0=bad.
I just want to make sure that I have not made a mess of the signs --> I will carefully check this via #257 and make sure all losses have the same trend.

ivadomed/losses.py
```diff
@@ -114,3 +114,103 @@ def forward(self, input, target):
...
+        denominator = tp + self.alpha * fp + self.beta * fn + self.smooth
+        tversky_label = numerator / denominator
+        return tversky_label
+
+    def forward(self, input, target):
+        n_classes = input.shape[1]
+        tversky_sum = 0.
+
+        # TODO: Add class_of_interest?
+        for i_label in range(n_classes):
```
andreanne-lemay, line 170: I am wondering here: why do we have to loop? The operations in tversky_index (substraction, addition and division) can be done on multichannel images, no?

charleygros: Because in the FocalTverskyLoss --> we apply the gamma to (1-tversky_class_i) --> we need to compute them per label --> that's why I created this loop.
I could re-write this with array manipulation if you think it would be better.

Issue #230, opened 2020-05-21
Refactor main


charleygros:
The last step of the refactoring that we planned is the refactoring of `main.py`, which goes along with the config files, e.g. #193 & #167.

### Main
Suggestion:
1. `main.py`:
- keep the `run_main` ([here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L885))
- config file / param parsing
- eventually deal with [data_loader](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L151) since there is a lot of duplication between the training, validation and testing loader

2. `train.py`, `test.py`: basically what is currently in `cmd_train` and `cmd_test` but modularised in order to improve clarity.
For instance:
- `get_loss_function` instead of [these lines](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L265)
- `get_lr_scheduler` instead of [these lines](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L235)
- `get_model` instead of t[hese lines](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/main.py#L198)
`train.py` will be in charge to generate a `log.json`, including ivadomed version and dataset info see [here](https://github.com/neuropoly/ivado-medical-imaging/issues/182).
Another objective would be to use the loop from `cmd_test` in `segment_volume` since these lines are mostly duplicated

3. `eval.py`: basically what is currently in `cmd_eval`. Also maybe also pull [Evaluation3DMetrics](https://github.com/neuropoly/ivado-medical-imaging/blob/06c1d6ce7132946541d0951f128d1a8ffe0a9c80/ivadomed/utils.py#L29) in this file.

### Config files

To achieve [this](https://github.com/neuropoly/ivado-medical-imaging/issues/193#issuecomment-631145765), I can see 2 main avenues:
1. we keep our (long) config files and `main.py` will be in charge to split the info between inference, and training and save it into the log_directory
2. we split our config files (config_ms_lesion/training.json, inference.json, evaluation.json) and the main will simply copy/paste these files in the log_directory.
If we go with (2), we will need to run `ivadomed model_ms_lesion/training.json` to trigger the training etc. (which is maybe more handy than opening the json file, changing `cmd_train` by `cmd_test`)

Any thoughts? We could start to draft a PR in parallel

andreanne-lemay:
[...]

PR #257, opened 2020-06-01
Tests for Loss functions

charleygros:
This PR aims at introducing tests for our loss functions.

Note: this PR is aligned with #256.

DONE:
- test_multiclassdiceloss
- test_generaliseddiceloss
- Change GDL sign so that it is consistent with the other losses, ie we want to minimize the loss: best=-1, worst=0
- test_tverskyloss
- test_focaltverskyloss
- test_diceloss (cf #230 )
- test_bce


charleygros:
The beauty of GDL is:

> weighting,  the  contribution  of  each  label  iscorrected by the inverse of its volume, thus reducing the well known correlationbetween region size and Dice score

... --> our implementation is useless!
Indeed, from our implementation perspective we have only one class (one channel) --> only one weight is computed (the one of the SC/tumor/lesion) --> our GDL is basically the DiceLoss.

I have added a parameter (`include_background`) which essentially add a channel to the input tensor reprensenting the background class (ie 1 for background, 0 for foreground). Thus, the weighting between classes is correctly done.

```python
        if self.include_background:
            # init
            input_background = torch.zeros(input.size(), dtype=input.dtype)
            target_background = torch.zeros(target.size(), dtype=target.dtype)
            # fill with opposite
            input_background[input == 0] = 1
            target_background[target == 0] = 1
            # Concat
            input = torch.cat([input, input_background], dim=1)
            target = torch.cat([target, target_background], dim=1)
```

charleygros:
All tests receive tuples organised as follows: input tensor, target tensor, expected_value, loss function.

ivadomed/losses.py
```diff
@@ -112,27 +112,44 @@ class GeneralizedDiceLoss(nn.Module):
...
+            input_background[input.sum(1).expand_as(input_background) == 0] = 1
+            target_background[target.sum(1).expand_as(input_background) == 0] = 1
+            # Concat
+            input = torch.cat([input, input_background], dim=1)
+            target = torch.cat([target, target_background], dim=1)
+
+        # Compute class weights
         target = target.float()
-        target_sum = target.sum(-1)
+        target_sum = target.sum(-1).sum(-1)
```
andreanne-lemay, line 143: Here im guessing in 3D we'd want .sum(-1).sum(-1).sum(-1)?

testing/test_losses.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+# pytest unit tests for ivadomed.losses
+
+import torch
```
andreanne-lemay, line 5: Great test :100:

PR #271, opened 2020-06-04
Add deep learning vertebral labeling code

lrouhier:
This PR aims at adding code related to the deep learning vertebral labeling projects. 

DONE:
* Adding Countception model in models.py 
* Adding L2 Loss and Adaptive Wing loss in losses.py
* Added 'prepare_dataset.py' This python script create nifti image from bids path. This generates a nifti image with a single slice. Heatmap are generated and stored under derivatives, while mid nifti a stored inside subject folder
* uses the bids loader with modality 'T2_mid' --> generated by prepare data and suffix 'heatmap'
* Modify config file/main for new losses and model 
* Loss_combination. This loss take a list of several loss as input and return the sum of these on (input, target). Params need to be listed in a separate list.
* Added config file used for training.
* Documentation for implemented function and model
* Added soft_input boolean for training with soft input (no thresholding)
* Fixed small glitch in ivadomed.utils
* Update doc 

Idea:
The idea behind the prepare_dataset file is to generate a bids dataset with the heatmap and the sagittal slice that we want to train on (average of middle slice +- 3 slices along the RL axis)


ivadomed/losses.py
```diff
@@ -233,3 +234,60 @@ def forward(self, input, target):
...
+        A = w * (1 / (1 + torch.pow(theta / e, alpha - target))) * (alpha - target) * torch.pow(theta / e,
+                                                                                              alpha - target - 1) * (1 / e)
+        C = (theta * A - w * torch.log(1 + torch.pow(theta / e, alpha - target)))
+
+        batch_size = target.size()[0]
+        hm_num = target.size()[1]
+
+        mask = torch.zeros_like(target)
+        # W=10
+        kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (2, 2))
```
charleygros, line 270: Would it be possible to use scipy instead of cv2?
eg https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.morphology.generate_binary_structure.html#scipy.ndimage.morphology.generate_binary_structure

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
+import nibabel as nib
```
charleygros, line 1: Would it be possible to rename the file `preprocessing.py` to match `postprocessing.py`?
If too specific to a project, we could also move some functions to the dev folder.

ivadomed/losses.py
lrouhier: yes probably I'll add that on the todo list !

ivadomed/prepare_dataset.py
lrouhier: I think it would be more fitting to put it in the dev folder it's not really 'preprocessing' it's mode the creation of the dataset :)

lrouhier:
Yes That's some refactoring i need to do (it is called inplanes and out planes but are the same thing)

ivadomed/prepare_dataset.py
charleygros: I was thinking eg some image normalisation tools or multivariate gaussians that could find a spot in `transforms.py` or `postprocessing.py`. Only a suggestion, your call :-)

ivadomed/losses.py
lrouhier: Done. It runs now i need to verify performance and all.

ivadomed/prepare_dataset.py
lrouhier: yes perhaps! Multi variate gaussian mostly might be useful for detection and not hidden here. I'll think about it.

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+import PIL
+import skimage
+import os
+import sys
+from utils import reorient_image
+
+# normalize Image
+def normalize(arr):
```
jcohenadad, line 11: we probably have that code somewhere else

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+    ma = arr.max()
+    mi = arr.min()
+    return ((arr - mi) / (ma - mi))
+
+
+# Useful function to generate a Gaussian Function on given coordinates. Used to generate groudtruth.
```
jcohenadad, line 17:
```suggestion
# Useful function to generate a Gaussian Function on given coordinates. Used to generate ground truth.
```

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+def label2maskmap_gt(data, shape, c_dx=0, c_dy=0, radius=10, normalize=False):
```
jcohenadad, line 18: maybe move to `ivadomed/utils.py`?

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    Generate a Mask map from the coordenates
```
jcohenadad, line 20:
```suggestion
    Generate a heatmap resulting from the convolution between a 2D Gaussian kernel and a pair of 2D coordinates.
```

ivadomed/prepare_dataset.py
jcohenadad, line 18: maybe it would be more useful for other purposes to have a function that takes as input a n-dim binary array (ones where labels are present), and that would output an array of the same dimension, that would be the convolution of the input array by a Gaussian kernel.  There are probably already a one-liner call to scipy/scikit-learn/numpy that does that.

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+
+    maskMap.append(Z)
+
+    if len(maskMap) == 1:
+        maskMap = maskMap[0]
+
+    return np.asarray(maskMap)
+
+
+def multivariate_gaussian(pos, mu, Sigma):
```
jcohenadad, line 73: same as comment above, that could probably be imported from existing Python packages

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+    for i in range(len(img_list)):
+        img = img_list[i]
+        img_tmp = np.zeros((x_val, y_val, 1), dtype=np.float64)
+        img_tmp[0:img.shape[0], 0:img.shape[1], 0] = img
+        img_zero_padding_list.append(img_tmp)
+
+    return img_zero_padding_list
+
+
+def mask2label(path_label, aim='full'):
```
jcohenadad, line 113: - move to utils
- if not used anymore (based on [previous suggestion](https://github.com/neuropoly/ivado-medical-imaging/pull/271/files#r436086717)), can remove

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+            if arr[x, y, z] <30 and arr[x, y, z] != 1:
+                list_label_image.append([x, y, z, arr[x, y, z]])
+        elif aim == 'c2':
+            if arr[x, y, z] == 3:
+                list_label_image.append([x, y, z, arr[x, y, z]])
+    list_label_image.sort(key=lambda x : x[3])
+    return list_label_image
+
+
+def get_midnifti(path_im, ind):
```
jcohenadad, line 137: i would rename: `get_midslice_average()`

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,215 @@
...
+    final = np.zeros(shape_tmp)
+    for x in list_coord_label:
+        train_lbs_tmp_mask = label2maskmap_gt(x, shape_tmp)
+        for w in range(shape_im[0]):
+            for h in range(shape_im[1]):
+                final[0, w, h] = max(final[0, w, h], train_lbs_tmp_mask[w, h])
+    return final
+
+
+def transform_dataset(bids_path, suffix, aim):
```
jcohenadad, line 183:
```suggestion
def transform_dataset(bids_path, suffix, aim):
   """This function takes as input a path to a dataset and generate two sets of images: (i) mid-sagittal image and (ii) heatmap of disc labels associated with the mid-sagittal image. 
```

jcohenadad, line 183: need to be more specific
```suggestion
def extract_mid_slice_and_convert_coordinates_to_heatmaps(bids_path, suffix, aim):
```

ivadomed/losses.py
```diff
@@ -233,3 +234,60 @@ def forward(self, input, target):
 
         # TODO: Take the opposite?
         return focal_tversky_sum / n_classes
+
+
+class L2loss(nn.Module):
```
jcohenadad, line 239: don't forget to update the documentation that lists the losses: https://github.com/neuropoly/ivado-medical-imaging/blob/master/docs/source/api_ref.rst#modivadomedlosses

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,270 @@
+import nibabel as nib
+import numpy as np
+import matplotlib.pyplot as plt
+import PIL
+import skimage
+import os
+import sys
+from utils import reorient_image
```
charleygros, line 8: I would suggest: `from ivadomed import utils as imed_utils`

ivadomed/utils.py
```diff
@@ -466,6 +466,34 @@ def segment_volume(folder_model, fname_image, fname_roi=None):
     return pred_nib
 
 
+def detect_image(image,folder_model):
```
jcohenadad, line 469: add space after , (PEP8)

jcohenadad, line 469: why not generalizing `segment_volume()` to accept 2d inputs?

ivadomed/training.py
```diff
@@ -314,7 +314,8 @@ def get_loss_function(params):
 
     # Check if implemented
     loss_function_available = ["DiceLoss", "FocalLoss", "GeneralizedDiceLoss", "FocalDiceLoss", "MultiClassDiceLoss",
-                               "BinaryCrossEntropyLoss", "TverskyLoss", "FocalTverskyLoss"]
+                               "BinaryCrossEntropyLoss", "TverskyLoss", "FocalTverskyLoss","AdapWingLoss","L2loss",
```
jcohenadad, line 317:
```suggestion
                               "BinaryCrossEntropyLoss", "TverskyLoss", "FocalTverskyLoss", "AdapWingLoss", "L2loss",
```

ivadomed/loader/loader.py
```diff
@@ -297,7 +304,7 @@ class MRI2DSegmentationDataset(Dataset):
     """This is a generic class for 2D (slice-wise) segmentation datasets."""
 
     def __init__(self, filename_pairs, slice_axis=2, cache=True, transform=None, slice_filter_fn=None,
-                 task="segmentation"):
+                 task="segmentation",soft_input=False):
```
jcohenadad, line 307:
```suggestion
                 task="segmentation", soft_input=False):
```

ivadomed/loader/loader.py
```diff
@@ -585,10 +594,10 @@ def __init__(self, root_dir, subject_lst, target_suffix, model_params, contrast_
 class BidsDataset(MRI2DSegmentationDataset):
     def __init__(self, root_dir, subject_lst, target_suffix, contrast_params, slice_axis=2,
                  cache=True, transform=None, metadata_choice=False, slice_filter_fn=None, roi_suffix=None,
-                 multichannel=False, task="segmentation"):
+                 multichannel=False, task="segmentation",soft_input=False):
```
jcohenadad, line 597:
```suggestion
                 multichannel=False, task="segmentation", soft_input=False):
```

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+        img = img_list[i] - np.mean(img_list[i])  # zero-center
+        if std:
+            img_std = np.std(img)  # normalize
+            epsilon = 1e-100
+            img = img / (img_std + epsilon)  # epsilon is used in order to avoid by zero division
+        img_norm_list.append(img)
+    return img_norm_list
+
+
+def extract_mid_slice_and_convert_coordinates_to_heatmaps(bids_path, suffix, aim, ap_pad=128, is_pad=320):
```
jcohenadad, line 147: make it a transfo when using the loader as discussed (Lucas/Julien/Charley)

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+            heatmap = heatmap_generation(label_array[:, :], 10)
+            if heatmap.shape[1] > is_pad:
+                heatmap = heatmap[:, heatmap.shape[1] - is_pad:]
+            elif heatmap.shape[1] < is_pad:
+                heatmap = add_zero_padding(heatmap, heatmap.shape[0], is_pad)[0]
+            if heatmap.shape[0] > ap_pad:
+                heatmap = heatmap[heatmap.shape[0] - ap_pad:, :]
+            elif heatmap.shape[0] < ap_pad:
+                heatmap = add_zero_padding(heatmap, ap_pad, heatmap.shape[1])[0]
+
```
jcohenadad, line 207: make it a transfo when using the loader as discussed (Lucas/Julien/Charley)

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
+import nibabel as nib
+import numpy as np
```
jcohenadad, line 2: maybe move it into the new `scripts/`

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+# normalize Image
+def normalize(arr):
+    ma = arr.max()
+    mi = arr.min()
+    return ((arr - mi) / (ma - mi))
+
+
+def gkern(kernlen=10):
```
jcohenadad, line 19: maybe call it `gaussian_kernel()`

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+    Returns:
+        array: 2D array heatmap matching the label.
+
+    """
+    kernel = gkern(kernel_size)
+    map = scipy.signal.convolve(image, kernel, mode='same')
+    return normalize(map)
+
+
+def add_zero_padding(img_list, x_val=512, y_val=512):
```
jcohenadad, line 54: to remove if padding is moved to a loader transfo

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+        elif aim == 'c2':
+            if arr[x, y, z] == 3:
+                list_label_image.append([x, y, z, arr[x, y, z]])
+    list_label_image.sort(key=lambda x: x[3])
+    return list_label_image
+
+
+def get_midslice_average(path_im, ind):
+    """
+    Retrieve the input images for the network. This images are generated by
```
jcohenadad, line 110:
```suggestion
    Retrieve the input image for the network. This image is generated by
```

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+    averaging the 7 slices in the middle of the volume
+    Args:
+        path_im(string): path to image
+        ind(int): index of the slice around which we will average
+
+    Returns:
+        array: an array containing the average image oriented according to RAS+ convention.
+
+    """
+    image = nib.load(path_im)
```
jcohenadad, line 120: add check for out of bound error

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+    image = nib.as_closest_canonical(image)
+    arr = np.array(image.dataobj)
+    numb_of_slice = 3
+    if ind + 3 > arr.shape[0]:
```
jcohenadad, line 124: add comment about out of bound

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+        numb_of_slice = arr.shape[0] - ind
+    if ind - numb_of_slice < 0:
+        numb_of_slice = ind
+
+    return np.mean(arr[ind - numb_of_slice:ind + numb_of_slice, :, :], 0)
+
+
+def images_normalization(img_list, std=True):
```
jcohenadad, line 132: will be removed

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+            img_std = np.std(img)  # normalize
+            epsilon = 1e-100
+            img = img / (img_std + epsilon)  # epsilon is used in order to avoid by zero division
+        img_norm_list.append(img)
+    return img_norm_list
+
+
+def extract_mid_slice_and_convert_coordinates_to_heatmaps(bids_path, suffix, aim, ap_pad=128, is_pad=320):
+    """
+     This function takes as input a path to a dataset and generate two sets of images:
```
jcohenadad, line 149:
```suggestion
     This function takes as input a path to a dataset and generates two sets of images:
```

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+   (i) mid-sagittal image of common size (1,ap_pad,is_pad) and
+   (ii) heatmap of disc labels associated with the mid-sagittal image.
+
+    Args:
+        bids_path (string): path to BIDS dataset form which images will be generated
+        suffix (string): suffix of image that will be processed (e.g., T2w)
+        aim(string): 'full' or 'c2'. If 'c2' retrieves only c2 label (value = 3) else create heatmap with all label.
+        ap_pad(int): desired output size of AP axis which will be
```
jcohenadad, line 157: try to avoid using MRI-specific orientation terms (here: AP), anticipating accommodation to histology

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+                    achieved from padding small images and cropping bigger ones
+        is_pad(int): Desired output size of IS axis.
```
jcohenadad, line 159: same here about IS axis. maybe use '"3rd dimension", or whatever is relevant

ivadomed/prepare_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+    Returns:
+        None. Image are saved in Bids folder
```
jcohenadad, line 163:
```suggestion
        None. Images are saved in BIDS folder
```

ivadomed/loader/loader.py
```diff
@@ -208,10 +213,14 @@ def get_pair_data(self):
         if self.gt_handle is None:
             gt_data = None
         for gt in self.gt_handle:
-            if gt is not None:
+            if gt is not None and self.soft_input==False:
```
andreanne-lemay, line 216: PEP8: `and not self.soft_input` instead of `== False`

ivadomed/loader/loader.py
```diff
@@ -208,10 +213,14 @@ def get_pair_data(self):
...
                 hwd_oriented = imed_loader_utils.orient_img_hwd(gt.get_fdata(cache_mode, dtype=np.float32),
                                                                 self.slice_axis)
                 gt_data.append(hwd_oriented.astype(np.uint8))
+            elif gt is not None and self.soft_input:
```
andreanne-lemay, line 220: instead of adding a `elif here` we could add a line like this: `data_type = np.float if self.soft_input else np.uint8`, `gt_data.append(hwd_oriented.astype(data_type))`. Currently this line is duplicated, `hwd_oriented = imed_loader_utils.orient_img_hwd(gt.get_fdata(cache_mode, dtype=np.float32),                                                             self.slice_axis)`, so this suggestion will prevent this duplication

ivadomed/loader/loader.py
```diff
@@ -408,7 +419,7 @@ def __getitem__(self, index):
                                                    metadata=metadata_gt,
                                                    data_type="gt")
             # Make sure stack_gt is binarized
-            if stack_gt is not None:
+            if stack_gt is not None and self.soft_input==False:
```
andreanne-lemay, line 422: same comment

ivadomed/losses.py
```diff
@@ -252,3 +254,102 @@ def forward(self, input, target):
...
+    adaptive Wing loss : https://arxiv.org/abs/1904.07399
+    Used for heatmap ground truth.
+
+    """
+
+    def __init__(self):
+        super(AdapWingLoss, self).__init__()
+
+    def forward(self,input,target):
+        theta = 0.5
```
andreanne-lemay, line 282: Do we want to put these as parameters that we could change (theta, alpha, w, e)? Also for clarity, if it's possible, can you use more explicit variable names for w, e, A and C?

ivadomed/losses.py
```diff
@@ -252,3 +254,102 @@ def forward(self, input, target):
...
+        sum_loss = torch.sum(AWingLoss)
+        all_pixel = torch.sum(mask)
+        mean_loss = sum_loss  # / all_pixel
+
+        return mean_loss
+
+
+class Loss_Combination(nn.Module):
+    """
+    Loss that sums different other implemented loss
```
andreanne-lemay, line 323: :100:

ivadomed/losses.py
```diff
@@ -252,3 +254,102 @@ def forward(self, input, target):
...
+    """
+
+    def __init__(self, losses_list,params_list=None):
+        """
+
+        Args:
+            losses_list: list of losses that will be summed
```
andreanne-lemay, line 330: Can you add the variable type in parenthesis (`losses_list (list): ...`) to all your docstrings https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

ivadomed/main.py
```diff
@@ -16,7 +16,9 @@
 cudnn.benchmark = True
 
 # List of not-default available models i.e. different from Unet
-MODEL_LIST = ['UNet3D', 'HeMISUnet', 'FiLMedUnet', 'NAME_CLASSIFIER_1']
+
+MODEL_LIST = ['UNet3D', 'HeMISUnet', 'FiLMedUnet', 'NAME_CLASSIFIER_1', 'Countception']
```
andreanne-lemay, line 20: :tada:

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
         return outputs
 
 
+class ConvBlock(nn.Module):
+    def __init__(self, in_chan, out_chan, ksize=3, stride=1, pad=0, activation=nn.LeakyReLU()):
+        """
+        perform convolution, activation and  batch normalization and
+        Args:
+            in_chan: number of channels on input
```
andreanne-lemay, line 803: Same comment for docstring

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
...
+    def forward(self, x):
+        return self.activation(self.batch_norm(self.conv1(x)))
+
+
+class SimpleBlock(nn.Module):
+    def __init__(self, in_chan, out_chan_1x1, out_chan_3x3, activation=nn.LeakyReLU()):
+        """
+        Inception module with 3 convolutions that are then concatenated. Max pooling performed on concatenation
+        Args:
+            in_chan: number of channel of input
```
andreanne-lemay, line 824: Data type in Docstring

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
     assert predictions.shape == mask_binary.shape
     assert np.array_equal(mask_binary, mask_binary.astype(bool))
     return predictions * mask_binary
+
+
+def coordinate_from_heatmap(nifti_image, thresh=0.3, slice_axis=0):
+    """
+    Retrieve coordinate from a soft output in a 2D image
+    Args:
+        image(array): A 2D array representing the soft output
```
andreanne-lemay, line 147: array -> nd.array

ivadomed/utils.py
```diff
@@ -61,6 +62,11 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
             for arr in tmp_lst:
                 print("Shape element lst {}".format(arr.shape))
 
+        # single slice can be mistaken with channel we add an empty slice in the volume
+        if len(tmp_lst) == 1:
```
andreanne-lemay, line 66: 1 channel should not be mistaken with 1 slice dimensions of the input image should be  (channel x height x width x depth). Our approach should be robust to 1 slice without conditions. As I mentioned on slack these lines shoul be removed :
 ```   
 # If only one channel then return 3D arr
    if arr_pred_ref_space.shape[0] == 1:
        arr_pred_ref_space = arr_pred_ref_space[0, :]
```

File: ivadomed/loader/loader.py (hunk 31 above)
Comment (lrouhier) [reply ivadomed/loader/loader.py:220]: You are right ! Thanks !

File: ivadomed/losses.py (hunk 33 above)
Comment (lrouhier) [reply ivadomed/losses.py:282]: Yep will do !

File: ivadomed/loader/loader.py [hunk 41]
```diff
@@ -20,7 +20,8 @@
 def load_dataset(data_list, bids_path, transforms_params, model_params, target_suffix, roi_params,
                  contrast_params, slice_filter_params, slice_axis, multichannel,
                  dataset_type="training", requires_undo=False, metadata_type=None,
-                 object_detection_params=None, **kwargs):
+                 object_detection_params=None, soft_input=False,**kwargs):
```
jcohenadad, line 23: cannot soft_input  go under kwargs?
also: space after "," PEP8

ivadomed/loader/loader.py
```diff
@@ -37,6 +38,7 @@ def load_dataset(data_list, bids_path, transforms_params, model_params, target_s
         metadata_type (string): None if no metadata
         dataset_type (string): training, validation or testing
         requires_undo (Bool): If True, the transformations without undo_transform will be discarded
+        soft_input(Bool): if True ground truth will be float32 non binarized images
```
jcohenadad, line 41:
```suggestion
        soft_input (Bool): if True ground truth will be float32 non binarized images
```
or remove if in kwargs

ivadomed/losses.py
```diff
@@ -79,9 +80,10 @@ def forward(self, input, target):
 class FocalDiceLoss(nn.Module):
     """
     Motivated by https://arxiv.org/pdf/1809.00076.pdf
-    :param beta: to bring the dice and focal losses at similar scale.
-    :param gamma: gamma value used in the focal loss.
-    :param alpha: alpha value used in the focal loss.
+    Args:
+            beta: to bring the dice and focal losses at similar scale.
```
jcohenadad, line 84: follow number of space convention for tabs (4x)

jcohenadad, line 84:
```suggestion
        beta: to bring the dice and focal losses at similar scale.
```

ivadomed/losses.py
```diff
@@ -252,3 +254,101 @@ def forward(self, input, target):
...
+    def __init__(self):
+        super(L2loss, self).__init__()
+
+    def forward(self, input, target):
+        return torch.sum((input - target) ** 2) / 2
+
+
+class AdapWingLoss(nn.Module):
+    """
+    adaptive Wing loss : https://arxiv.org/abs/1904.07399
```
jcohenadad, line 273:
```suggestion
    adaptive Wing loss as in: https://arxiv.org/abs/1904.07399
```

ivadomed/losses.py
```diff
@@ -252,3 +254,101 @@ def forward(self, input, target):
 
         # TODO: Take the opposite?
         return focal_tversky_sum / n_classes
+
+
+class L2loss(nn.Module):
+    """
+    L2 loss between two images : inputs and target
```
jcohenadad, line 261:
```suggestion
    L2 loss between inputs and target
```

ivadomed/losses.py
```diff
@@ -252,3 +254,101 @@ def forward(self, input, target):
...
+    def __init__(self):
+        super(AdapWingLoss, self).__init__()
+
+    def forward(self,input,target):
+        theta = 0.5
+        alpha = 2.1
+        w = 14
+        e = 1
+        A = w * (1 / (1 + torch.pow(theta / e, alpha - target))) * (alpha - target) * torch.pow(theta / e,
+                                                                                              alpha - target - 1) * (1 / e)
```
jcohenadad, line 287: PEP8

ivadomed/losses.py
```diff
@@ -252,3 +254,101 @@ def forward(self, input, target):
...
+        sum_loss = torch.sum(AWingLoss)
+        all_pixel = torch.sum(mask)
+        mean_loss = sum_loss  # / all_pixel
+
+        return mean_loss
+
+
+class Loss_Combination(nn.Module):
+    """
+    Loss that sums different other implemented loss
```
jcohenadad, line 322:
```suggestion
    Loss that sums other implemented losses.
```

ivadomed/losses.py
```diff
@@ -252,3 +254,101 @@ def forward(self, input, target):
...
+    """
+
+    def __init__(self, losses_list,params_list=None):
+        """
+
+        Args:
+            losses_list(list): list of losses that will be summed. Elements should be string.
```
jcohenadad, line 329:
```suggestion
            losses_list (list): list of losses that will be summed. Elements should be string.
```
pls: fix everywhere

ivadomed/losses.py
```diff
@@ -252,3 +254,101 @@ def forward(self, input, target):
...
+            (e.g., losses_list = ["L2loss","DiceLoss"]
+                   params_list = [None,{"param1:0.5"}])
+        returns:
+            tensor: sum of losses computed on (input,target) with the params
+        """
+        super(Loss_Combination, self).__init__()
+        self.losses_list = losses_list
+        self.params_list = params_list
+
+    def forward(self, input, target):
```
jcohenadad, line 341: add docstrings

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
         return outputs
 
 
+class ConvBlock(nn.Module):
+    def __init__(self, in_chan, out_chan, ksize=3, stride=1, pad=0, activation=nn.LeakyReLU()):
+        """
+        perform convolution, activation and  batch normalization and
```
jcohenadad, line 801:
```suggestion
        perform convolution, activation and batch normalization.
```

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
...
+        conv1_out = self.conv1(x)
+        conv2_out = self.conv2(x)
+        conv3_out = self.conv3(x)
+        output = torch.cat([conv1_out, conv2_out, conv3_out], 1)
+        output = self.MP(output)
+        return output
+
+
+class Countception(nn.Module):
+    """based on : https://arxiv.org/abs/1703.08710
```
jcohenadad, line 845:
```suggestion
    """Countception model, based on: https://arxiv.org/abs/1703.08710
```

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
...
+    modified from github https://github.com/roggirg/count-ception_mbm/
```
jcohenadad, line 846:
```suggestion
    modified from: https://github.com/roggirg/count-ception_mbm/
```

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
...
+    """
+
+    def __init__(self, in_channel=3, out_channel=1, use_logits=False, logits_per_output=12, name='CC'):
+        """
+
+        Args:
+            in_channel(int): number of channel on input image
```
jcohenadad, line 853:
```suggestion
            in_channel(int): number of channels on input image
```

ivadomed/models.py
```diff
@@ -795,6 +795,127 @@ def forward(self, inputs):
...
+            out_channel(int): number of channels on output image
+            use_logits(bool): boolean to change output
+            logits_per_output(int): number of outputs of final convolution which will mulitplied by the number of chanel
```
jcohenadad, line 856:
```suggestion
            logits_per_output(int): number of outputs of final convolution which will be multiplied by the number of channels
```

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
     assert np.array_equal(mask_binary, mask_binary.astype(bool))
     return predictions * mask_binary
+
+
+def coordinate_from_heatmap(nifti_image, thresh=0.3, slice_axis=0):
+    """
+    Retrieve coordinate from a soft output in a 2D image
+    Args:
+        nifti_image(nibabel object): Single slice nifti image of the soft output.
+        thresh(float):Relative threshold for local maxima.
```
jcohenadad, line 148:
```suggestion
        thresh(float): Relative threshold for local maxima, i.e., after normalizing the min and max between 0 and 1, respectively. 
```

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
     assert predictions.shape == mask_binary.shape
     assert np.array_equal(mask_binary, mask_binary.astype(bool))
     return predictions * mask_binary
+
+
+def coordinate_from_heatmap(nifti_image, thresh=0.3, slice_axis=0):
+    """
+    Retrieve coordinate from a soft output in a 2D image
```
jcohenadad, line 145:
```suggestion
    Retrieve coordinates of local maxima in a soft segmentation.
```

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
+    Args:
+        nifti_image(nibabel object): Single slice nifti image of the soft output.
```
jcohenadad, line 147:
```suggestion
        nifti_image(nibabel object): Single slice nifti image. Typically, a soft segmentation.
```

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
+        thresh(float):Relative threshold for local maxima.
+
+    Returns:
+        list: A list of computed coordinates found by local maximum in RAS convention
```
jcohenadad, line 151: why in RAS convention? if user inputs a slice in LPI, i would expect my list of coordinates to be in LPI (otherwise it might break things down the line)

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
+    slc = [slice(None)] * len(arr_can.shape)
+    slc[slice_axis] = 0
+    image = arr_can[tuple(slc)]
+    coordinates_tmp = peak_local_max(image[:, :], min_distance=5, threshold_rel=thresh)
+    return coordinates_tmp
+
+
+def label_file_from_coordinates(fname_image, coord_list, slice_axis):
+    """
+    Creates a nifti object with single voxel label. All labels have a value of 1. The nifti object as the same
```
jcohenadad, line 165:
```suggestion
    Creates a nifti object with single-voxel labels. Each label has a value of 1. The nifti object as the same
```

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
+    orientation as the input
+    Args:
+        fname_image(str): Path to the image that matches the labels.
```
jcohenadad, line 168:
```suggestion
        fname_image(str): Path to the image which affine matrix will be used to generate a new image with labels.
```

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
+        coord_list(list): list of coordinates in RAS oriented space. Each element is [x, y].
```
jcohenadad, line 169: is that for 2D or not? please clarify.

jcohenadad, line 169: why in RAS? it should be orientation-agnostic

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,52 @@ def mask_predictions(predictions, mask_binary):
...
+        slice_axis(int): axis of the original image where the label should be.
+
+    Returns:
+        nib: A single slice nifti object containing the singe-voxel label of value 1 in fname_image ref space.
```
jcohenadad, line 173: "single slice" → not clear. see my comment above

ivadomed/preprocessing.py
```diff
@@ -0,0 +1,38 @@
+# Transformation to perform on image before network processing
+import nibabel as nib
+import numpy as np
+import ivadomed.utils as imed_utils
+
+
+def get_midslice_average(path_im, ind, slice_axis=0):
+    """
+    Retrieve the input image for the network. This image is generated by
```
jcohenadad, line 9:
```suggestion
    Extract an average 2D slice out of a 3D volume. This image is generated by
```

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,144 @@
...
+import ivadomed.utils as imed_utils
+import ivadomed.preprocessing as imed_preprocessing
+import nibabel as nib
+import numpy as np
+import os
+import scipy
+import skimage
+
+
+# normalize between 0 and 1.
```
jcohenadad, line 11: a few of these functions could go in ivadomed package and be exposed as APIs, e.g. normalize, gaussian_kernel, etc.

maybe an `ivadomed.math`? `ivadomed.labels`, etc.

jcohenadad: could be done in another PR--

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,144 @@
...
+        elif aim == 'c2':
+            if arr[x, y, z] == 3:
+                list_label_image.append([x, y, z, arr[x, y, z]])
+    list_label_image.sort(key=lambda x: x[3])
+    return list_label_image
+
+
+def extract_mid_slice_and_convert_coordinates_to_heatmaps(bids_path, suffix, aim="full"):
+    """
+     This function takes as input a path to a dataset  and generates two sets of images:
```
jcohenadad, line 86:
```suggestion
    This function takes as input a path to a dataset and generates a set of images:
```

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,144 @@
...
+   (i) mid-sagittal image of common size (1,ap_pad,is_pad) and
```
jcohenadad, line 87: update

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,144 @@
...
+            label_array = np.zeros(imsh[1:])
+
+            for j in range(len(list_points)):
+                label_array[list_points[j][1], list_points[j][2]] = 1
+
+            heatmap = heatmap_generation(label_array[:, :], 10)
+            arr_pred_ref_space = imed_utils.reorient_image(np.expand_dims(heatmap[:, :], axis=0), 2, lab, nib_ref_can)
+            nib_pred = nib.Nifti1Image(arr_pred_ref_space, lab.affine)
+            nib.save(nib_pred,
+                     bids_path + 'derivatives/labels/' + t[i] + '/anat/' + t[i] + suffix + '_mid_heatmap.nii.gz')
```
jcohenadad, line 123: will break in windows. use `os.path.join()`

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            pass
+
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-p", "--path", dest="path", required=True, type=str, help="Path to bids folder with "/" at the end.")
```
jcohenadad, line 130: OS-specific. Use filesep

ivadomed/config/config_vertebral_labeling.json
```diff
@@ -0,0 +1,103 @@
...
+            "testing": [ "T1w_straight_mid"],
+            "balance": {}
+        },
+        "slice_filter_params": {
+            "filter_empty_mask": false,
+            "filter_empty_input": true
+        },
+        "slice_axis": "sagittal",
+        "multichannel": false,
+	"soft_input": true
```
charleygros, line 24: Before merging, may I ask you to add this param in the other config files? many thanks

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,47 @@ def mask_predictions(predictions, mask_binary):
...
+        list: A list of computed coordinates found by local maximum. each element will be a list composed of
+        [x, y, z]
+    """
+    image_ref = nib.load(nifti_image)
+    image = np.array(image_ref.dataobj)
+    coordinates_tmp = peak_local_max(image, min_distance=5, threshold_rel=thresh)
+    return coordinates_tmp
+
+
+def label_file_from_coordinates(fname_image, coord_list):
```
charleygros, line 161: Just for consistency purposes, what about input a nifti object as the other functions?

ivadomed/scripts/prepare_dataset.py
charleygros: This function already exists in `ivadomed.transforms.py`: `rescale_values(arr, minv, maxv)`

ivadomed/utils.py
```diff
@@ -387,8 +399,10 @@ def segment_volume(folder_model, fname_image, fname_roi=None, gpu_number=0):
     if fname_roi is None and 'filter_empty_mask' in loader_params["slice_filter_params"]:
         print("\nWARNING: fname_roi has not been specified, then the entire volume is processed.")
         loader_params["slice_filter_params"]["filter_empty_mask"] = False
+        filename_pairs = [([fname_image], None, None, [{}])]
+    else:
+        filename_pairs = [([fname_image], None, [fname_roi], [{}])]
 
-    filename_pairs = [([fname_image], [None], fname_roi, [{}])]
```
charleygros, line 391: Reminder to myself: Want to come back to this

ivadomed/losses.py
```diff
@@ -252,3 +256,115 @@ def forward(self, input, target):
...
+    """
+    Adaptive Wing loss as in: https://arxiv.org/abs/1904.07399
+    Used for heatmap ground truth.
+
+    """
+
+    def __init__(self, theta=0.5, alpha=2.1, omega=14, epsilon=1):
+        """
+
+        Args:
```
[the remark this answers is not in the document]
lrouhier: Will change

ivadomed/loader/loader.py
```diff
@@ -37,6 +37,7 @@ def load_dataset(data_list, bids_path, transforms_params, model_params, target_s
         metadata_type (string): None if no metadata
         dataset_type (string): training, validation or testing
         requires_undo (Bool): If True, the transformations without undo_transform will be discarded
+        soft_input (Bool): if True ground truth will be float32 non binarized images
```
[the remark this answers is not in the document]
lrouhier: Yep you are right! It might be clearer

ivadomed/postprocessing.py
lrouhier: The thing is, It kind of goes with the previous function. Therefore there is a second option: compact them both into a function that has an argument that determine if the output is coordinate list or a file with single-voxel label what do you think is the best ?

ivadomed/losses.py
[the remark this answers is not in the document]
lrouhier: done in 3f3eacbc

ivadomed/utils.py
lrouhier: Well depends on what you call height, width, and depth since we orient it to its original shape just before...I think it needs more extensive testing..

ivadomed/postprocessing.py
```diff
@@ -137,3 +138,47 @@ def mask_predictions(predictions, mask_binary):
...
+    Args:
+        nifti_image (nibabel object): nifti image of the soft output.
+        thresh (float): Relative threshold for local maxima, i.e., after normalizing
+        the min and max between 0 and 1, respectively.
+
+    Returns:
+        list: A list of computed coordinates found by local maximum. each element will be a list composed of
+        [x, y, z]
+    """
+    image_ref = nib.load(nifti_image)
```
charleygros, line 155: Is nifti_image a nibabel object or a fname? I am confused by the nib.load

ivadomed/postprocessing.py
charleygros: In order to reduce the number of I/O --> what about loading the image outside of these functions and only deal with nibabel objects for the I/O of these functions.
But I may miss something: couldyou please help me?

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-p", "--path", dest="path", required=True, type=str,
+                        help="Path to bids folder with " / " at the end.")
+    parser.add_argument("-s", "--suffix", dest="suffix", required=True,
+                        type=str, help="suffix of image file")
+    parser.add_argument("-a", "--aim", dest="aim", default="full", type=str,
+                        help="If set to 'c2' only points with value 3 will be converted to heatmap")
```
charleygros, line 127: What about indicating the choices? ie full or c2?
Also, what about being more general: input the label of interest --> 3 for C2/C3, 5 for C4/C5, -1 for all?

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,146 @@
+import argparse
```
charleygros, line 1: I would recommend to rename this script as this is very specific to vert labelling task, something like prepare_dataset_vertebral_labelling?

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,146 @@
...
+            nib.save(nib_pred, os.path.join(bids_path, 'derivatives', 'labels', t[i], 'anat', t[i] + suffix +
+                                            '_mid_heatmap.nii.gz'))
+        else:
+            pass
+
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-p", "--path", dest="path", required=True, type=str,
+                        help="Path to bids folder with " / " at the end.")
```
charleygros, line 123: what about using os.path.join or similar?

ivadomed/scripts/prepare_dataset.py
```diff
@@ -0,0 +1,146 @@
...
+    parser.add_argument("-s", "--suffix", dest="suffix", required=True,
+                        type=str, help="suffix of image file")
```
charleygros, line 125: image = label file ?

ivadomed/scripts/prepare_dataset.py
lrouhier: Yeah I was thinking about it afterward! I'll see what I can do about it !

ivadomed/scripts/prepare_dataset.py
lrouhier: I'll add an example but the idea is sub-xxx_SUFFIX.

ivadomed/postprocessing.py
lrouhier: It' a fname so i'll update this.

lrouhier:
>In order to reduce the number of I/O --> what about loading the image outside of these functions and only deal with nibabel >objects for the I/O of these functions.
>But I may miss something: couldyou please help me?

We can do that there was no point beside it being sort of standalone function but I guess we can do that for both.

ivadomed/config/config_vertebral_labeling.json
lrouhier: Done

ivadomed/scripts/prepare_dataset.py
lrouhier: I'll update this I am already using os.path.join

ivadomed/scripts/prepare_dataset.py
charleygros: Yeah I would just recommend to replace image by label :-)

lrouhier: Hum I think i explained it wrong I meant like the suffix of the input image (e.g., acq-sag_T2w)

ivadomed/scripts/prepare_dataset.py
lrouhier: Done in 66a3132

ivadomed/postprocessing.py
lrouhier: So It was resolved when I switched to nibabel object.

docs/source/technical_features.rst
```diff
@@ -106,3 +107,5 @@ Loss functions
    additional feature compared to the published reference, is that the
    background volume can be weighted by the inverse of its area, which
    could be of interest in high class imbalance scenarios.
+-  `Adaptive wing loss <https://arxiv.org/abs/1904.07399>`__. Loss function used to detect key points with Gaussian representation of the target.
+-  Loss Combination: Linear combination of any other implemented lost.
```
andreanne-lemay, line 111: lost -> loss

ivadomed/utils.py
andreanne-lemay, line 391: This line should work what was the issue?

lrouhier: Might be fixed in new PR but it used to fails because [None] was not considered as None.

docs/source/technical_features.rst
lrouhier: Corrected

ivadomed/loader/loader.py
```diff
@@ -42,6 +42,7 @@ def load_dataset(data_list, bids_path, transforms_params, model_params, target_s
         dataset_type (str): Choice between "training", "validation" or "testing".
         requires_undo (bool): If True, the transformations without undo_transform will be discarded.
         object_detection_params (dict): Object dection parameters.
+        soft_gt (bool): If True, ground truth will be non-binarized images encoded in float32.
```
jcohenadad, line 45: why forcing float32 encoding? doesn't it work for e.g. uint8?

ivadomed/losses.py
```diff
@@ -320,3 +323,115 @@ def forward(self, input, target):
...
+
+class AdapWingLoss(nn.Module):
+    """
+    Adaptive Wing loss as in
+    Used for heatmap ground truth.
+
+    ..seealso::
+        Wang, Xinyao, Liefeng Bo, and Li Fuxin. "Adaptive wing loss for robust face alignment via heatmap regression."
+        Proceedings of the IEEE International Conference on Computer Vision. 2019.
+
```
jcohenadad, line 348: unnecessary line

ivadomed/losses.py
```diff
@@ -320,3 +323,115 @@ def forward(self, input, target):
...
+
+    Args:
+        theta (float): Threshold between linear and non linear loss.
+        alpha (float): Used to adapt loss shape to input shape and make loss smooth at 0 (background).
+        It needs to be slightly above 2 to maintain ideal properties.
+        omega (float): Multiplicating factor for non linear part of the loss.
+        epsilon (float): factor to avoid gradient explosion. It must not be too small
+
```
jcohenadad, line 356:
```suggestion
```

ivadomed/losses.py
jcohenadad: @lrouhier did you see my comment?

ivadomed/models.py
```diff
@@ -961,6 +962,129 @@ def forward(self, inputs):
...
+        conv2_out = self.conv2(x)
+        conv3_out = self.conv3(x)
+        output = torch.cat([conv1_out, conv2_out, conv3_out], 1)
+        output = self.MP(output)
+        return output
+
+
+class Countception(nn.Module):
+    """Countception model
+
```
jcohenadad, line 1014: A brief description of the model would be welcome

ivadomed/models.py
```diff
@@ -961,6 +962,129 @@ def forward(self, inputs):
...
+    .. seealso::
+        Paul Cohen, Joseph, et al. "Count-ception: Counting by fully convolutional redundant counting."
+        Proceedings of the IEEE International Conference on Computer Vision Workshops. 2017.
+
+    Args:
+        in_channel (int): number of channel on input image
```
jcohenadad, line 1020:
```suggestion
        in_channel (int): number of channels on input image
```

ivadomed/models.py
```diff
@@ -961,6 +962,129 @@ def forward(self, inputs):
...
+        out_channel (int): number of channels on output image
+        use_logits (bool): boolean to change output
+        logits_per_output (int): number of outputs of final convolution which will multiplied by the number of channels
+    """
+
+    def __init__(self, in_channel=3, out_channel=1, use_logits=False, logits_per_output=12, name='CC'):
+        super(Countception, self).__init__()
```
jcohenadad, line 1027: docstrings?

ivadomed/postprocessing.py
```diff
@@ -161,3 +162,48 @@ def mask_predictions(predictions, mask_binary):
     assert predictions.shape == mask_binary.shape
     assert np.array_equal(mask_binary, mask_binary.astype(bool))
     return predictions * mask_binary
+
+
+def coordinate_from_heatmap(nifti_image, thresh=0.3):
+    """
+    Retrieve coordinates of local maxima in a soft segmentation.
+    Args:
+        nifti_image (nibabel object): nifti image of the soft output.
```
jcohenadad, line 171: the concept of input/output is relative to where we are (in this function, it is an input, not an output, so it could be confusing)
```suggestion
        nifti_image (nibabel object): nifti image of the soft segmentation.
```

ivadomed/postprocessing.py
```diff
@@ -161,3 +162,48 @@ def mask_predictions(predictions, mask_binary):
...
+    Creates a nifti object with single-voxel labels. Each label has a value of 1. The nifti object as the same
+    orientation as the input.
+    Args:
+        nifti_image (nibabel object): Path to the image which affine matrix will be used to generate a new image with
+        labels.
+        coord_list (list): list of coordinates. Each element is [x, y, z]. Orientation should be the same as the image
+
+    Returns:
+        nib: A nifti object containing the singe-voxel label of value 1. The matrix will be the same size as
+        fname_image's.
```
jcohenadad, line 196:
```suggestion
        `nifti_image`.
```

ivadomed/postprocessing.py
```diff
@@ -161,3 +162,48 @@ def mask_predictions(predictions, mask_binary):
...
+    """
+    Creates a nifti object with single-voxel labels. Each label has a value of 1. The nifti object as the same
+    orientation as the input.
+    Args:
+        nifti_image (nibabel object): Path to the image which affine matrix will be used to generate a new image with
+        labels.
+        coord_list (list): list of coordinates. Each element is [x, y, z]. Orientation should be the same as the image
+
+    Returns:
+        nib: A nifti object containing the singe-voxel label of value 1. The matrix will be the same size as
```
jcohenadad, line 195:
```suggestion
        nib_pred: A nifti object containing the singe-voxel label of value 1. The matrix will be the same size as
```

ivadomed/preprocessing.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_midslice_average(path_im, ind, slice_axis=0):
+    """
+    Extract an average 2D slice out of a 3D volume. This image is generated by
+    averaging the 7 slices in the middle of the volume
+    Args:
+        path_im (string): path to image
+        ind (int): index of the slice around which we will average
+        slice_axis (int): Slice axis according to RAS convention
```
jcohenadad, line 14: does it matter that it is according to the RAS convention? here, it only refers to a dimension, right?

ivadomed/preprocessing.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Returns:
+        nifti: a single slice nifti object containing the average image in the image space.
+
+    """
+    image = nib.load(path_im)
+    image_can = nib.as_closest_canonical(image)
```
jcohenadad, line 21: is this necessary, given that the slice dimension is given as input? why do you need to do that?

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python
+##############################################################
+#
+# This script is used to create a dataset with (i) mid-sagittal image and
+# (ii) heatmap of disc labels associated with the mid-sagittal image.
+#
+# Usage: python scripts/automate_training.py -p bids_path -s T2w -a -1
```
jcohenadad, line 7: obsolete?

ivadomed/models.py
```diff
@@ -961,6 +962,129 @@ def forward(self, inputs):
...
+        output = torch.cat([conv1_out, conv2_out, conv3_out], 1)
+        output = self.MP(output)
+        return output
+
+
+class Countception(nn.Module):
+    """Countception model
+
+    .. seealso::
+        Paul Cohen, Joseph, et al. "Count-ception: Counting by fully convolutional redundant counting."
```
jcohenadad, line 1016: "Cohen" is the name, "Joseph Paul" are first names. Just put the initials 
```suggestion
        Cohen JP et al. "Count-ception: Counting by fully convolutional redundant counting."
```

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
```diff
@@ -0,0 +1,156 @@
...
+            if arr[x, y, z] < 30 and arr[x, y, z] != 1:
+                list_label_image.append([x, y, z, arr[x, y, z]])
+        elif aim > 0:
+            if arr[x, y, z] == aim:
+                list_label_image.append([x, y, z, arr[x, y, z]])
+    list_label_image.sort(key=lambda x: x[3])
+    return list_label_image
+
+
+def extract_mid_slice_and_convert_coordinates_to_heatmaps(bids_path, suffix, aim=-1):
```
jcohenadad, line 88: how come `bids_path` does not appear on the [generated doc](https://ivadomed.org/en/lr-deep_vertebral_labeling/scripts.html#ivadomed.scripts.prepare_dataset_vertebral_labeling.extract_mid_slice_and_convert_coordinates_to_heatmaps)?

ivadomed/loader/loader.py
lrouhier: With float32 we will have a mask with continuous value which is what we are looking for with the soft ground truth.

jcohenadad: ah, yes obviously uint8 would not work. I was thinking more of other floats, e.g. float64. I am just wondering why this specific float32 is expected, and if it is, why not asserting it in your code?

ivadomed/models.py
lrouhier: The init method can be documented at either class level or with the [method.](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). The rest of the file has documentation at class level which is why the documentation is above the method.

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
lrouhier: Yes! Came from a misunderstanding about automatically generated documentation

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
lrouhier: I am not sure what you mean as I am seeing it 
image

ivadomed/losses.py
jcohenadad: docstrings needed

ivadomed/models.py
jcohenadad: ah, i didn't see the definition above. 
but "name" seems to be missing

ivadomed/preprocessing.py
lrouhier: This is linked to the slice_axis (so if one is in RAS orientation the other one should be as well). I did that have a known orientation so that I don't have to check which axis is RL. It can be done. I just find it easier to always aligned in RAS, as most of the code does. 
When processing batch of images, it avoid finding the 'RL' axis [here](https://github.com/ivadomed/ivadomed/blob/48e9d9b2b17f60fff3c87d270652b402db7b8e78/ivadomed/scripts/prepare_dataset_vertebral_labeling.py#L88) which is what I was thinking about when coding this.

ivadomed/loader/loader.py
lrouhier: We transform our tensor in float (to understand float32 following pytorch [naming convention](https://pytorch.org/docs/stable/tensors.html)) so greater precision does not seem relevant. I will add it to the documentation.

ivadomed/losses.py
lrouhier: Yes sorry I forgot to answer. I can add it if you want, however, there is an automated doc for forward in all losses [here](https://ivadomed.org/en/latest/api_ref.html#module-ivadomed.losses) (there is no docstring for the other forward methods in the doc)

ivadomed/preprocessing.py
jcohenadad: i see. makes sense

ivadomed/preprocessing.py
jcohenadad: addressed in https://github.com/ivadomed/ivadomed/pull/271#pullrequestreview-436913166

ivadomed/losses.py
jcohenadad: cool! how is this auto doc generated?

ivadomed/loader/loader.py
jcohenadad: so why not asserting if it is expected?

ivadomed/models.py
lrouhier: indeed

ivadomed/losses.py
lrouhier: Comes from sphinx According to @Drulex 
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

jcohenadad: yes, i know we are using sphinx, but my question was more: where is the native doc for "forward()" defined, and how is it populated for all the instances of this function&

ivadomed/loader/loader.py
lrouhier: Not sure if I understand. Uint8 is used to reduce the ram storage when we have binary mask.

lrouhier: Oh right right! I will modify the documentation

ivadomed/losses.py
lrouhier: It is just below (as there is no parameter, it is just a description)

ivadomed/losses.py
lrouhier: I don't know, it is not defined in our code. There is a similar doc for each model forward method. My best guess is sphinx doing some auto-completion. I am looking to it, and I asked @Drulex if he knows how it is created.

ivadomed/models.py
```diff
@@ -961,6 +962,133 @@ def forward(self, inputs):
...
+        output = torch.cat([conv1_out, conv2_out, conv3_out], 1)
+        output = self.MP(output)
+        return output
+
+
+class Countception(Module):
+    """Countception model.
+    Fully convolutional model using inception module and used for keypoints detection.
+    It extracts several patches within each image, every pixel is therefore processed by the network several times
+    allowing to average over the error and avoid false negatives.
```
jcohenadad, line 1016:
```suggestion
    The inception model extracts several patches within each image. Every pixel is therefore processed by the 
    network several times, allowing to average multiple predictions and minimize false negatives.
```

ivadomed/losses.py
Drulex: This one is a mystery.. Looking..

Drulex: It's from the `PyTorch` documentation. Multiple classes in `losses.py` inherit ` torch.nn.modules.module` which already has a `forward` method. The overridden `forward` method in `losses.py` doesn't have a docstring, but it looks like `Sphinx` uses the docstring from the base class. See https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.forward

lrouhier: Thanks for tracking this !

jcohenadad: awesome! thank you for revealing the mystery @Drulex 
good to know for the future

ivadomed/prepare_dataset.py
jcohenadad: @lrouhier many of my comments are unanswered-- so i don't know if you have implemented the feature or if you missed my comments-- the usual thing to do is to link to the commit where the requested changed is made, or to answer something if the requested change is rejected/requires further discussion

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
jcohenadad: here: 
https://github.com/ivadomed/ivadomed/blob/0f02ac410f7734822f0a60a1aee783ced8e45d30/ivadomed/scripts/prepare_dataset_vertebral_labeling.py#L88

but not here:
image

ivadomed/models.py
lrouhier: done in: 59e663a

ivadomed/prepare_dataset.py
lrouhier: Yes sorry. It was done in: 306221d

ivadomed/prepare_dataset.py
lrouhier: I think it can be done in subsequent PR with ivadomed/maths (as this is a convolution)

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
jcohenadad: @lrouhier nevermind, this seems to be a bug (not specific with your code)-- issue opened https://github.com/ivadomed/ivadomed/issues/318

ivadomed/prepare_dataset.py
lrouhier: Done in: 839deb4

ivadomed/prepare_dataset.py
lrouhier: Done in 839deb4

ivadomed/prepare_dataset.py
lrouhier: Removed after previous suggestion

ivadomed/prepare_dataset.py
lrouhier: Done in 319a44c

ivadomed/losses.py
lrouhier: done in: cf549d9 and corrected in ceb746d

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
lrouhier: Removed in cbdad4d

ivadomed/scripts/prepare_dataset_vertebral_labeling.py
lrouhier: Oh right I didn't see that. Now that you say it, all first arguments are missing for the script documentation.

ivadomed/prepare_dataset.py
jcohenadad: > I think it can be done in subsequent PR with ivadomed/maths (as this is a convolution)

agreed, can you pls open an issue and link it here?

jcohenadad: like for the previous function, it would make sense to move it to ivadomed/math. Could you link the issue you are about to open here?

lrouhier: issue open [here](https://github.com/ivadomed/ivadomed/issues/319) about the creation of ivadomed/maths

lrouhier: issue about the creation of ivadomed/maths.py [here](https://github.com/ivadomed/ivadomed/issues/319)

jcohenadad: @lrouhier it would be helpful if you could, in the issue, already list some of the functions already identified in this thread

lrouhier: I am doing it right now.

rezazad68:
Thanks for sharing the code. Following the required files for vertebral disc labeling, how can I generate the `straight` and `heatmap3` files?  
Link to the config file: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config_vertebral_labeling.json

jcohenadad:
> Thanks for sharing the code. Following the required files for vertebral disc labeling, how can I generate the `straight` and `heatmap3` files?
> Link to the config file: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config_vertebral_labeling.json

The straightening is part of the `sct_label_vertebrae` routine, which is [being refactored to include the DL-based disc detection](https://github.com/neuropoly/spinalcordtoolbox/pull/2679/files). That being said, I would encourage you to try _without_ straightening the cord, maybe you will end up with equally-performant results (which would be great, because straightening takes time).

rezazad68:
Thanks for the prompt reply. I have generated the heatmap3 files using the [prepare_dataset_vertebral_labeling](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/prepare_dataset_vertebral_labeling.py) code with aim=3.  Using the config_vertebral_labeling.json file [please check the parameters](https://github.com/rezazad68/temp/blob/main/config_vertebral_labeling.json) I have run the training section. However, the dice score seems near zero. Could you please update me with the possible solution? Thanks.

Link to training process:
https://prnt.sc/w9eggp

jcohenadad:
@lrouhier help needed here 🙏 

@rezazad68 while waiting for @lrouhier 's answer, I suggest you try to bypass the heatmap generation, ie: instead of applying a gaussian mask to the labels, try to _estimate_ those labels using another architecture/loss more adapted to single labels? check the pose marching model (there is an issue open on ivadomed), i don't think they use heatmap for that

PR #272, opened 2020-06-05
Adapt loader to slice-wise classification task

charleygros:
## Aim:
Adapt 2D loader to handle slice-wise classification task.

## Changes:
- Add a config file, `config_classification.json` dedicated to this task.
    - For now, the only change is the addition of this dict, which will contains the classifier params:
```json
    "NAME_CLASSIFIER_1": {
        "applied": true
    },
```
the `"applied": true` triggers the classification task given that this model name is listed [here](https://github.com/neuropoly/ivado-medical-imaging/blob/47db357d37e72a9174b0ee501c63f52fed26a580/ivadomed/loader/loader.py#L14), as part of the classifiers available on the repo'.
- the main consequence of triggering the classification task is that the ground truth is not a mask (array) but a label: does the slice contain the object of interest? Yes / No --> 1 / 0

## Comments:
- the implementation is robust to multi_label classification: eg `target_suffix=["_c2.nii.gz", "_c3.nii.gz"]`, then `stack_gt` is a 2 elements list, with values 0 or 1 ([here](https://github.com/neuropoly/ivado-medical-imaging/blob/47db357d37e72a9174b0ee501c63f52fed26a580/ivadomed/loader/loader.py#L281))
- in `config_classification.json`, I setup transformations for images only (ie not applied to GT) since we deal here with labels, not images. But, if classification task is detected, the loader forces no transformations on the labels anyway ([here](https://github.com/neuropoly/ivado-medical-imaging/blob/47db357d37e72a9174b0ee501c63f52fed26a580/ivadomed/loader/loader.py#L396)).

## TODO before merging
After PR approval but before merging:
- Fill wiki config doc'

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,7 +138,7 @@ future.
 
 Ensure that you are the original author of your changes, and if that is
 not the case, ensure that the borrowed/adapted code is compatible with
-the [project's license](https://ivado-medical-imaging.readthedocs.io/en/latest/license.html).
+the [project's license](https://ivadomed.org/en/latest/index.html#license).
 
 
 ## Committing
--- a/ivadomed/config/config_classification.json
+++ b/ivadomed/config/config_classification.json
@@ -1,0 +1,106 @@
+{
+    "command": "train",
+    "gpu": 6,
+    "log_directory": "sct_testing_lesion_ax",
+    "debugging": false,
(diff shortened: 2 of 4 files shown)


charleygros:
Wiki updated [here](https://github.com/neuropoly/ivado-medical-imaging/wiki/Configuration-file#architecture).

PR #274, opened 2020-06-05
Add config-file documentation to RTD

charleygros:
## Aim
Move config-file documentation from GH Wiki to RTD.

## Changes
- Move config-file documentation from GH Wiki to RTD
- Link this documentation from "Getting started"
- Add RTD link to `README.md`

## TODO once merged:
- Remove doc from [wiki](https://github.com/neuropoly/ivado-medical-imaging/wiki/configuration-file)


docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
+# Configuration file
```
jcohenadad, line 1: i would definitely reference this file in the `index.rst`, between getting_started.md and tutorials.md.

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+
+* [General parameters](#general-parameters)
```
jcohenadad, line 3: by referencing the file in the index.rst, the TOC will be automatically-generated, so it can be removed from here (less maintenance, less prone to errors)

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+* [Transformations](#transformations)
+* [Examples](#examples)
+
+## General parameters
+#### command
+Run the specified command. Choices: ``"train"``, ``"test"``, ``"eval"``, to train, test and evaluate a model respectively.
+#### gpu
+Integer. ID of the GPU to use.
+#### log_directory
+Folder name where output files (e.g. trained model, predictions, results) are saved.
```
jcohenadad, line 18:
```suggestion
Folder name that will contain the output files (e.g., trained model, predictions, results).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### debugging
+Bool. Allows extended verbosity and intermediate outputs.
```
jcohenadad, line 20:
```suggestion
Bool. Extended verbosity and intermediate outputs.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Loader parameters
+#### bids_path
+String. Relative path of the BIDS folder.
```
jcohenadad, line 24: doesn't it work for absolute paths?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### target_suffix
+Suffix list of the derivative file containing the ground-truth of interest (e.g. [`"_seg-manual"`, `"_lesion-manual"`]). If the list has a length greater than 1, then a multi-class model will be trained.
+#### contrasts
+- `train_validation`: list of image modalities loaded for the training and validation. If `multichannel`, then this list represent the different channels of the input tensors. Otherwise, the modalities are mixed and the model has only one input channel.
```
jcohenadad, line 28:
```suggestion
- `train_validation`: list of image modalities loaded for the training and validation. If `multichannel=True`, this list represents the different channels of the input tensors. Otherwise, the modalities are mixed and the model only has one input channel.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `test`: list of image modalities loaded in the testing dataset. Same comment than for `train_validation` regarding `multichannel`.
```
jcohenadad, line 29: is it possible to give an example?  people can get confused as to what is expected here. Modality can be interpreted as "image modality", e.g. MRI, ultrasound, CT, etc. I guess you meant "contrast" here? Like "T1w", "T2w", etc.. if so i would give examples

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `balance`: Dict. Used to under-represent some modalities in the dataset: e.g. `{"T1w": 0.1}` will include only 10% of the available `T1w` images into the training/validation/test set. Please set `multichannel` to `false` if you are using this parameter.
```
jcohenadad, line 30:
```suggestion
- `balance`: Dict. Enables to weight the importance of specific channels (or contrasts) in the dataset: e.g. `{"T1w": 0.1}` means that only 10% of the available `T1w` images will be included into the training/validation/test set. Please set `multichannel` to `false` if you are using this parameter.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### multichannel
+Bool. Indicated if more than a modality is used by the model. See details in both `train_validation` and `test` for the modalities that are input.
```
jcohenadad, line 32: see my comment above about the term "modality". Should we use "channel" everywhere, or "contrast" (maybe the latter is more explicit?)

```suggestion
Bool. Used to specify if more than one channel (or contrast) is used by the model. See details in both `train_validation` and `test` about the input channels.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### slice_axis
+Choice between `"sagittal"`, `"coronal"`, and `"axial"`.
+This parameter decides the slice orientation on which the model will train.
```
jcohenadad, line 35:
```suggestion
Sets the slice orientation for on which the model will be used.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### slice_filter
+Dict:
+- `filter_empty_input`: Bool. Filter empty images if `true`.
```
jcohenadad, line 38:
```suggestion
- `filter_empty_input`: Bool. Filter empty images.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `filter_empty_mask`: Bool. Filter empty GT mask slices if `true`.
```
jcohenadad, line 39: i'm a bit confused about two things here:
- what does "filter" mean here? Maybe you can describe
- here the word "slice" is used whereas in the previous line the word "images" is used. Are we talking about the same thing? if so i would use the same term (maybe slices).
```suggestion
- `filter_empty_mask`: Bool. Filter empty ground truth mask slices.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### roi
+Dict. of parameters about the region of interest
+- `suffix`: suffix of the derivative file containing the ROI used to crop (e.g. `"_seg-manual"`) with `ROICrop` as transform. Please use `null` if you do not want to use a ROI to crop.
```
jcohenadad, line 42: not sure i understand how this dict should be defined. An example of the value is given ("_seg-manual"), but what would be the key of the dict?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `slice_filter_roi`: int. It filters ROI mask slices with less than this threshold of non zero voxels. Active when using `"ROICrop"` and inputing ROI file.
```
jcohenadad, line 43: what is meant by "it filters"?  Did you mean the following?
"Voxels contained in the ROI mask and which value are strictly below the value given in this parameter will be replaced by zero.".

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Split dataset
+#### fname_split
+Filename of a joblib file containing the list of training / validation / test subjects, used to ensure reproducible experiments
```
jcohenadad, line 47:
```suggestion
Filename of a joblib file containing the list of training / validation / test subjects. This file can later be used to re-train a model using the same data splitting scheme.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### random_seed
+Seed used by the random number generator to split the dataset between training/validation/testing.
+#### center_test
+List. List of centers to only include in the testing dataset.
```
jcohenadad, line 51: i assume a list of string is expected? we should probably explain where and how the centers are defined (in the participants.tsv file? elsewhere?)

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### method
+Choice between `"per_patient"` (i.e. shuffle all subjects then splits) or `"per_center"` (split subject according to their acquisition centers).
```
jcohenadad, line 53: same comment as before-- where is the definition of a "patient" (is it the suffix of a BIDS-based dataset, e.g. will "sub-03" be considered a "patient") and "center"?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### train_fraction
+Number between `0` and `1` representing the fraction of the dataset used as training set.
```
jcohenadad, line 55: mention a float is expected (to follow the convention above, where the parameter type is specified at the beginning of each parameter's description)

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### test_fraction
+Number between `0` and `1` representing the fraction of the dataset used as test set. This parameter is only used if the `method` is `"per_patient"`.
```
jcohenadad, line 57: same comment about float

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Training parameters
+#### batch_size
+Integer.
```
jcohenadad, line 61:
```suggestion
Strictly positive integer.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### loss
+- `name`: Name of the loss function: Choice among the classes that are available [here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/losses.py).
```
jcohenadad, line 63: i would define the hyperlink to the documentation file `losses.md`. I'm not 100% sure how to do the cross-ref in md files, so you'll need to figure out.

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- Other parameters that could be needed in the Loss function definition: see attributes of the Loss function of interest (e.g. `"gamma": 0.5` for `FocalLoss`).
+#### training_time
+- `num_epochs`: int.
```
jcohenadad, line 66:
```suggestion
- `num_epochs`: Strictly positive integer.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `early_stopping_epsilon`: Threshold (percentage) for an improvement in the validation loss to be considered meaningful
```
jcohenadad, line 67:
```suggestion
- `early_stopping_epsilon`: percentage. Threshold on the validation loss below which training will stop.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `early_stopping_patience`: Number of epochs after which the training is stopped if the validation loss improvement not meaningful (i.e. less than `early_stopping_epsilon`)
```
jcohenadad, line 68:
```suggestion
- `early_stopping_patience`: Strictly positive integer. Number of epochs after which the training is stopped if the validation loss improvement not meaningful (i.e. less than `early_stopping_epsilon`)
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### scheduler
+- `initial_lr`: Float. Initial learning rate.
+- `scheduler_lr`:
+1. `name`: Choice between: `"CosineAnnealingLR"`, `"CosineAnnealingWarmRestarts"` and `"CyclicLR"`. Please find documentation [here](https://pytorch.org/docs/stable/optim.html).
+2. Other parameters that are needed for the scheduler of interest (e.g. `"base_lr": 1e-5, "max_lr": 1e-2` for `"CosineAnnealingLR"`).
+#### balance_samples
+Bool. If `true`, then positive and negative GT samples are balanced in both training and validation datasets.
```
jcohenadad, line 75:
```suggestion
Bool. Balance positive and negative ground truths in both the training and the validation datasets.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### mixup_alpha
+Float. Alpha parameter of the Beta distribution, see [original paper](https://arxiv.org/pdf/1710.09412.pdf).
```
jcohenadad, line 77:
```suggestion
Float. Alpha parameter of the Beta distribution, see [original paper](https://arxiv.org/abs/1710.09412).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+#### transfer_learning
+- `retrain_model`: Filename of the pretrained model (`path/to/pretrained-model`). If `null`, then no transfer learning is performed and the network is trained from scratch.
```
jcohenadad, line 79:
```suggestion
- `retrain_model`: Filename of the pretrained model (`path/to/pretrained-model`). If `null`, no transfer learning is performed and the network is trained from scratch.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+- `retrain_fraction`: Float between 0. and 1. Controls the fraction of the pre-trained model that will be fine-tuned. For instance, if set to 0.5, then the second half of the model will be fine-tuned, while the first layers will be frozen.
+
+## Architecture
+Architectures for both segmentation and classification are available and described [here](models.rst).
+If the selected architecture is listed [here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/loader/loader.py#L14), then a classification task is run, ie the ground-truth are labels extracted from `target`, instead of arrays for the segmentation task.
```
jcohenadad, line 84: same comment about redirecting to a file "loader" from the documentation

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+### default_model (Mandatory)
+Define the default model (`Unet`) and mandatory parameters that are common to all available architectures (listed [here](models.rst)). If a tailored model is defined (see next section), then the default parameters are merged with the parameters that are specific to the tailored model.
+- `name`: `Unet` (default)
+- `dropout_rate`: float (e.g. 0.4).
+- `batch_norm_momentum`: float (e.g. 0.1).
+- `depth`: int, number of down-sampling operations.
+Note:
+- `in_channel` is automatically defined with `multichannel` and `contrast/training_validation`.
+- `out_channel` is automatically defined with `target_suffix`.
+### Tailored model (optional)
```
jcohenadad, line 94: it would clarify to add a space before each new heading

docs/source/configuration_file.md
```diff
@@ -0,0 +1,139 @@
...
+Here are defined the tailored model and the parameters that are specific to it (ie not defined in the default model). See examples:
+- [FiLMedUnet](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/config/config.json#L64)
```
jcohenadad, line 96: even when not pointing to another file in the documentation (docs/), i would make a relative link. E.g.: if you create a branch where the json file changes, this absolute link to the master branch will be wrong

docs/source/configuration_file.md
charleygros: corrected, thanks!

docs/source/configuration_file.md
charleygros: Done :-)

docs/source/configuration_file.md
charleygros: DOne :-)

docs/source/configuration_file.md
charleygros: Agree ++. Added suggestion + use contrasts instead of modalities in the code.

docs/source/configuration_file.md
charleygros: Corrected with contrasts, thanks!

docs/source/configuration_file.md
charleygros: Filter: discard a slice from the dataset: ie not used for training etc
`filter_empty_input` discard a slice if all voxel intensities (image not GT) (eg T2w) are all zeros.
I tried to make it clearer. You are right, it was not clear at all.

docs/source/configuration_file.md
charleygros: I tried to clarify it. An example would be:
```json
"roi": {
    "suffix": "_seg-manual",
    "slice_filter_roi": 10
}
```

docs/source/configuration_file.md
charleygros: What about:
"It filters (i.e. discards from the dataset) slices where the ROI mask has with less than this number of non-zero voxels."

docs/source/configuration_file.md
charleygros: Excellent point. Will correct this.

docs/source/configuration_file.md
charleygros: Corrected too. Info extracted from `bids_dataset/participants.tsv`.

docs/source/configuration_file.md
charleygros: Yes agree about "L14". So the idea would be to give the link to a RTD page instead of the itself code: Have I understood correctly?

docs/source/configuration_file.md
charleygros: Agree! Corrected :-)

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
+# Configuration file
+
+## General parameters
+
+#### command
```
jcohenadad, line 5: i would follow the logical headings order (##, then ###, then ####), otherwise it will impact auto-TOC generation
```suggestion
### command
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+Sets the slice orientation for on which the model will be used.
+
+#### slice_filter
+Dict. Discard a slice from the dataset if it meets a condition, see below.
+- `filter_empty_input`: Bool. Discard slices where all voxel intensities are zeros.
+- `filter_empty_mask`: Bool. Discard slices where all voxel labels are zeros.
+
+#### roi
+Dict. of parameters about the region of interest
+- `suffix`: String. Suffix of the derivative file containing the ROI used to crop (e.g. `"_seg-manual"`) with `ROICrop` as transform. Please use `null` if you do not want to use a ROI to crop.
```
jcohenadad, line 44: what if a user wants to use an image called "awesomeimage_null.nii"? why not considering an empty string as null?
```suggestion
- `suffix`: String. Suffix of the derivative file containing the ROI used to crop (e.g. `"_seg-manual"`) with `ROICrop` as transform. Please use `null` if you do not want to use an ROI to crop.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `slice_filter_roi`: int. It filters (i.e. discards from the dataset) slices where the ROI mask has with less than this number of non zero voxels. Active when using `"ROICrop"`.
```
jcohenadad, line 45:
```suggestion
- `slice_filter_roi`: int. If the ROI mask contains less than `slice_filter_roi` non-zero voxels, the slice will be discarded from the dataset. This feature helps with noisy labels, e.g., if a slice contains only 2-3 labeled voxels, we do not want to use these labels to crop the image. This parameter is only considered when using `"ROICrop"`.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Split dataset
+
+#### fname_split
+Filename of a joblib file containing the list of training/validation/testing subjects. This file can later be used to re-train a model using the same data splitting scheme.
```
jcohenadad, line 50:
```suggestion
String. File name of the log (joblib) that contains the list of training/validation/testing subjects. This file can later be used to re-train a model using the same data splitting scheme.
```

jcohenadad, line 50: is "joblib" a known jargon? does that correspond to a log file?  Is that related to this? https://joblib.readthedocs.io/en/latest/

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### random_seed
+Seed used by the random number generator to split the dataset between training/validation/testing.
```
jcohenadad, line 53: what is expected? float? string? any range?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### center_test
+List of strings. List of centers to only include in the testing dataset. If used, please include a column `institution_id` in your `bids_dataset/participants.tsv`.
```
jcohenadad, line 56:
```suggestion
List of strings. Each string corresponds to an institution/center to only include in the testing dataset (not validation). If used, the file `bids_dataset/participants.tsv` needs to contain a column `institution_id`, which associates a subject with an institution/center.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### method
+Choice between `"per_patient"` (i.e. shuffle all subjects then splits, using the `participant_id` column from `my_bids_dataset/participants.tsv`) or `"per_center"` (split subjects according to their acquisition centers, using the `institution_id` column from `my_bids_dataset/participants.tsv`).
```
jcohenadad, line 59:
```suggestion
`{'per_patient', 'per_center'}`. `"per_patient"`: all subjects are shuffled, then split between train/validation/test, regardless their institution. `"per_center"`: all subjects are split so as not to mix institutions between the train, validation and test sets. The latter option enables to ensure the model is working across domains (institutions). Note: the institution information is contained within the `institution_id` column in the `participants.tsv` file.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### train_fraction
+Float. Between `0` and `1` representing the fraction of the dataset used as training set.
+
+#### test_fraction
+Float. Between `0` and `1` representing the fraction of the dataset used as test set. This parameter is only used if the `method` is `"per_patient"`.
```
jcohenadad, line 65: why couldn't we use that for "per_center"?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Training parameters
+
+#### batch_size
+Strictly positive integer.
+
+#### loss
+- `name`: Name of the loss function: Choice among the classes that are available [here](../../ivadomed/losses.py).
```
jcohenadad, line 73:
```suggestion
- `name`: Name of the [loss function Class](../../ivadomed/losses.py).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- Other parameters that could be needed in the Loss function definition: see attributes of the Loss function of interest (e.g. `"gamma": 0.5` for `FocalLoss`).
+
+#### training_time
+- `num_epochs`: Strictly positive integer.
+- `early_stopping_epsilon`: Float. If the validation loss difference during one epoch (i.e. `abs(validation_loss[n] - validation_loss[n-1]` where n is the current epoch) is inferior to this epsilon for `early_stopping_patience` consecutive epochs, then  the early stopping of the training is triggered.
```
jcohenadad, line 78:
```suggestion
- `early_stopping_epsilon`: Float. If the validation loss difference during one epoch (i.e. `abs(validation_loss[n] - validation_loss[n-1]` where n is the current epoch) is inferior to this epsilon for `early_stopping_patience` consecutive epochs, then training stops.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `early_stopping_patience`: Strictly positive integer. Number of epochs after which the training is stopped if the validation loss improvement not meaningful (i.e. less than `early_stopping_epsilon`)
```
jcohenadad, line 79:
```suggestion
- `early_stopping_patience`: Strictly positive integer. Number of epochs after which the training is stopped if the validation loss improvement is smaller than `early_stopping_epsilon`.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `initial_lr`: Float. Initial learning rate.
+- `scheduler_lr`:
+1. `name`: Choice between: `"CosineAnnealingLR"`, `"CosineAnnealingWarmRestarts"` and `"CyclicLR"`. Please find documentation [here](https://pytorch.org/docs/stable/optim.html).
+2. Other parameters that are needed for the scheduler of interest (e.g. `"base_lr": 1e-5, "max_lr": 1e-2` for `"CosineAnnealingLR"`).
+
+#### balance_samples
+Bool. Balance positive and negative labels in both the training and the validation datasets.
+
+#### mixup_alpha
+Float. Alpha parameter of the Beta distribution, see [original paper](https://arxiv.org/abs/1710.09412).
```
jcohenadad, line 91:
```suggestion
Float. Alpha parameter of the Beta distribution, see [original paper on the Mixup technique](https://arxiv.org/abs/1710.09412).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### transfer_learning
+- `retrain_model`: Filename of the pretrained model (`path/to/pretrained-model`). If `null`, no transfer learning is performed and the network is trained from scratch.
+- `retrain_fraction`: Float between 0. and 1. Controls the fraction of the pre-trained model that will be fine-tuned. For instance, if set to 0.5, then the second half of the model will be fine-tuned, while the first layers will be frozen.
```
jcohenadad, line 95:
```suggestion
- `retrain_fraction`: Float between 0. and 1. Controls the fraction of the pre-trained model that will be fine-tuned. For instance, if set to 0.5, the second half of the model will be fine-tuned while the first layers will be frozen.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Architecture
+Architectures for both segmentation and classification are available and described [here](models.rst).
+If the selected architecture is listed [here](../../ivadomed/loader/loader.py#L14), then a classification task is run, ie the ground-truth are labels extracted from `target`, instead of arrays for the segmentation task.
```
jcohenadad, line 99:
```suggestion
Architectures for both segmentation and classification are available and described in the [Models](models.rst) section.
If the selected architecture is listed in the [loader.py](../../ivadomed/loader/loader.py#L14) file, a classification (not segmentation) task is run. In the case of a classification task, the ground truth will correspond to a single label value extracted from `target`, instead being an array (the latter being used for the segmentation task).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+### default_model (Mandatory)
+Define the default model (`Unet`) and mandatory parameters that are common to all available architectures (listed [here](models.rst)). If a tailored model is defined (see next section), then the default parameters are merged with the parameters that are specific to the tailored model.
+- `name`: `Unet` (default)
```
jcohenadad, line 103:
```suggestion
Define the default model (`Unet`) and mandatory parameters that are common to all available architectures (listed in the [Models](models.rst) section). If a tailored model is defined (see next section), the default parameters are merged with the parameters that are specific to the tailored model.
- `name`: `Unet` (default)
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `dropout_rate`: float (e.g. 0.4).
+- `batch_norm_momentum`: float (e.g. 0.1).
+- `depth`: int, number of down-sampling operations.
```
jcohenadad, line 106:
```suggestion
- `depth`: Strictly positive integer. Number of down-sampling operations.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `retrain_fraction`: Float between 0. and 1. Controls the fraction of the pre-trained model that will be fine-tuned. For instance, if set to 0.5, then the second half of the model will be fine-tuned, while the first layers will be frozen.
+
+## Architecture
+Architectures for both segmentation and classification are available and described [here](models.rst).
+If the selected architecture is listed [here](../../ivadomed/loader/loader.py#L14), then a classification task is run, ie the ground-truth are labels extracted from `target`, instead of arrays for the segmentation task.
+
+### default_model (Mandatory)
+Define the default model (`Unet`) and mandatory parameters that are common to all available architectures (listed [here](models.rst)). If a tailored model is defined (see next section), then the default parameters are merged with the parameters that are specific to the tailored model.
+- `name`: `Unet` (default)
+- `dropout_rate`: float (e.g. 0.4).
```
jcohenadad, line 104:
```suggestion
- `dropout_rate`: Float (e.g. 0.4).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `batch_norm_momentum`: float (e.g. 0.1).
```
jcohenadad, line 105:
```suggestion
- `batch_norm_momentum`: Float (e.g. 0.1).
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `depth`: int, number of down-sampling operations.
+Note:
+- `in_channel` is automatically defined with `multichannel` and `contrast/training_validation`.
+- `out_channel` is automatically defined with `target_suffix`.
```
jcohenadad, line 109: how are they automatically defined? what value will/should they have? What do these parameters correspond to?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Tailored model (optional)
```
jcohenadad, line 111: section "default_model" is written with snake_case convention, why not having `tailored_model` here?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+Here are defined the tailored model and the parameters that are specific to it (ie not defined in the default model). See examples:
```
jcohenadad, line 112: This section includes fancier models and parameters that are not defined in the `default_model`:

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- [FiLMedUnet](../../ivadomed/config/config.json#L64)
```
jcohenadad, line 113: I find it problematic to include a link to a specific line. If the config.json file is updated in the future, i guaranty that someone will forget to update this line number in the documentation, so i would rather remove the link.

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+    - `metadata`: choice between `"without"`, `"mri_params"`, and `"contrast"`. If `"mri_params"`, then vectors of [FlipAngle, EchoTime, RepetitionTime, Manufacturer] are input to the FiLM generator. If `"contrast"`, then image contrasts (according to `config/contrast_dct.json`) are input to the FiLM generator.
```
jcohenadad, line 114: related note: we use by convention double quote "" for verbose/written text, and single quote '' for variables/dict values (like in this case).
```suggestion
    - `metadata`: `{'without', 'mri_params', 'contrast'}`. `mri_params`: Vectors of [FlipAngle, EchoTime, RepetitionTime, Manufacturer] are input to the FiLM generator. `contrast`: Image contrasts (according to `config/contrast_dct.json`) are input to the FiLM generator.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- [HeMISUnet](../../ivadomed/config/config_spineGeHemis.json#L64)
+    - `missing_contrast`: Bool.
```
jcohenadad, line 116:
```suggestion
- `HeMISUnet`
    - `missing_contrast`: Bool.
```

jcohenadad, line 116: any description?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- [UNet3D](../../ivadomed/config/config_tumorSeg.json#L65)
```
jcohenadad, line 117:
```suggestion
- `UNet3D`
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+    - `length_3D`: tuple indicating the size of the subvolumes or volume used for unet 3D model (depth, width, height).
```
jcohenadad, line 118:
```suggestion
    - `length_3D`: (Integer, Integer, Integer). Size of the subvolumes or volume used for unet 3D model: (depth, width, height).
```

jcohenadad, line 118: what is a "subvolume"?

jcohenadad, line 118: depth, width and height might be confusing: do they correspond to the first, second and third dimension of a numpy array, respectively? If so i would use that terminology instead (no confusion possible)

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+    - `padding_3D`: size of the overlapping per subvolume and dimensions (e.i `padding:0`). Note: In order to be used, each dimension of an input image needs to be a multiple of length plus 2 * padding and a multiple of 16. To change input image size use the following transformation `CenterCrop3D`.
```
jcohenadad, line 119: i read it 3 times and i still don't understand what this parameter does. It is mentioned "size of the overlapping", but the parameter is called "padding", so is that an overlap (i.e. number of voxels, in each dimension, that overlap between two sliding windows), or a padding (i.e., number of voxels added in each dimension). 

Also, it is still unclear to me what a subvolume is.

"(e.i `padding:0`)" → did you mean "(e.g `padding:0`)"?

to clarify, i would maybe give a concrete example of a volume of size XxYxZ, and what would happen to the subvolume and to the padding applied to the subvolume

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+Here are defined the tailored model and the parameters that are specific to it (ie not defined in the default model). See examples:
+- [FiLMedUnet](../../ivadomed/config/config.json#L64)
+    - `metadata`: choice between `"without"`, `"mri_params"`, and `"contrast"`. If `"mri_params"`, then vectors of [FlipAngle, EchoTime, RepetitionTime, Manufacturer] are input to the FiLM generator. If `"contrast"`, then image contrasts (according to `config/contrast_dct.json`) are input to the FiLM generator.
+- [HeMISUnet](../../ivadomed/config/config_spineGeHemis.json#L64)
+    - `missing_contrast`: Bool.
+- [UNet3D](../../ivadomed/config/config_tumorSeg.json#L65)
+    - `length_3D`: tuple indicating the size of the subvolumes or volume used for unet 3D model (depth, width, height).
+    - `padding_3D`: size of the overlapping per subvolume and dimensions (e.i `padding:0`). Note: In order to be used, each dimension of an input image needs to be a multiple of length plus 2 * padding and a multiple of 16. To change input image size use the following transformation `CenterCrop3D`. 
+    - `attention_unet`: indicates if attention gates are used in the Unet's decoder.
```
jcohenadad, line 120:
```suggestion
    - `attention_unet`: Bool. Use attention gates in the Unet's decoder.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Testing parameters
+- `binarize_prediction`: Indicates if predictions from the model are soft (prediction from 0 to 1) or binarized (either 1 or 0 according to the threshold 0.5).
```
jcohenadad, line 123:
```suggestion
- `binarize_prediction`: Bool. Binarize output predictions using a threshold of 0.5. If 'false', output predictions are float between 0 and 1. 
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### uncertainty
+Uncertainty computation is performed if `n_it>0` and at least `epistemic` or `aleatoric` is `True`.
```
jcohenadad, line 126: I've seen written 'true' and 'True' (same for false/False). I suggest to use a single convention.

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `epistemic`: Bool.
+- `aleatoric`: Bool.
+- `n_it`: int, number of Monte Carlo iterations.
```
jcohenadad, line 129:
```suggestion
- `n_it`: Integer. Number of Monte Carlo iterations. Set to 0 for no uncertainty computation.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+    - `padding_3D`: size of the overlapping per subvolume and dimensions (e.i `padding:0`). Note: In order to be used, each dimension of an input image needs to be a multiple of length plus 2 * padding and a multiple of 16. To change input image size use the following transformation `CenterCrop3D`. 
+    - `attention_unet`: indicates if attention gates are used in the Unet's decoder.
+
+## Testing parameters
+- `binarize_prediction`: Indicates if predictions from the model are soft (prediction from 0 to 1) or binarized (either 1 or 0 according to the threshold 0.5).
+
+#### uncertainty
+Uncertainty computation is performed if `n_it>0` and at least `epistemic` or `aleatoric` is `True`.
+- `epistemic`: Bool.
+- `aleatoric`: Bool.
```
jcohenadad, line 128: I would add a short description and a reference for each type. Also, I believe that both cannot be true (or can they)? what happens in this case?

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `n_it`: int, number of Monte Carlo iterations.
+
+## Transformations
+Indicate the transformation in the same order you would like them to be applied to your samples. For each transformation, please indicate the parameters as well as the following two optional entries:
```
jcohenadad, line 132:
```suggestion
Transformations applied during data augmentation. Transformations are sorted in the order they are applied to the image samples. For each transformation, the following parameters are customizable:
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `RandomRotation` (parameters: `degrees`)
+- `RandomAffine` (parameters: `translate`)
+- `RandomShiftIntensity` (parameters: `shift_range`)
+- `ElasticTransform` (parameters: `alpha_range`, `sigma_range`, `p`)
+- `Resample` (parameters: `wspace`, `hspace`, `dspace`)
+- `AdditionGaussianNoise` (parameters: `mean`, `std`)
+- `DilateGT` (parameters: `dilation_factor`) float, controls the number of iterations of ground-truth dilation depending on the size of each individual lesion, data augmentation of the training set. Use `0` to disable.
+- `HistogramClipping` (parameters: `min_percentile`, `max_percentile`)
+- `Clage` (parameters: `clip_limit`, `kernel_size`)
+- ` RandomReverse`
```
jcohenadad, line 150:
```suggestion
- `RandomReverse`
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `CenterCrop2D` (parameters: `size`)
+- `ROICrop2D` (parameters: `size`)
+- `NormalizeInstance`
+- `RandomRotation` (parameters: `degrees`)
+- `RandomAffine` (parameters: `translate`)
+- `RandomShiftIntensity` (parameters: `shift_range`)
+- `ElasticTransform` (parameters: `alpha_range`, `sigma_range`, `p`)
+- `Resample` (parameters: `wspace`, `hspace`, `dspace`)
+- `AdditionGaussianNoise` (parameters: `mean`, `std`)
+- `DilateGT` (parameters: `dilation_factor`) float, controls the number of iterations of ground-truth dilation depending on the size of each individual lesion, data augmentation of the training set. Use `0` to disable.
```
jcohenadad, line 147: if it sets the number of iterations, shouldn't it be integer instead?
```suggestion
- `DilateGT` (parameters: `dilation_factor`) Float. Controls the number of iterations of ground-truth dilation depending on the size of each individual lesion, data augmentation of the training set. Use `0` to disable.
```

docs/source/configuration_file.md
```diff
@@ -0,0 +1,159 @@
...
+- `HistogramClipping` (parameters: `min_percentile`, `max_percentile`)
+- `Clage` (parameters: `clip_limit`, `kernel_size`)
+- ` RandomReverse`
+
+## Examples
```
jcohenadad, line 152: excellent idea to include this!

charleygros:
TODO: replace `null` by `""` when string is asked.

docs/source/configuration_file.md
charleygros: Added short description + refs + clarification than they can both be true (ie both DropOut and Data aug are on).

charleygros:
Todo: add cross-ref losses.md in API REF

docs/source/configuration_file.md
charleygros: Very good point, corrected with `true` and `false` as it is the json convention.

docs/source/configuration_file.md
charleygros: What about: "If the validation loss difference during one epoch (i.e. `abs(validation_loss[n] - validation_loss[n-1]` where n is the current epoch) is inferior to this epsilon for `early_stopping_patience` consecutive epochs, the training will stop."

docs/source/configuration_file.md
charleygros: Not a log file per se: we use it to save numpy, scipy or sklearn objects as alternative to pickle, cf [here](https://scikit-learn.org/stable/modules/model_persistence.html#persistence-example).

docs/source/configuration_file.md
charleygros: Int is prefered, cf [here](https://docs.python.org/3/library/random.html#random.seed).
I am correcting this.

docs/source/configuration_file.md
charleygros: When "per_center": the testing dataset is defined by the "center_test", ie "center_test" controls the fraction of the dataset that is used for testing.
While "test_fraction" does this when using "per_patient".
The documentation is not clear at all on that aspect, thank you for raising it. I am suggesting a new wording.

docs/source/configuration_file.md
charleygros: These parameters do not need to be defined in the config file --> I removed them to avoid confusion. But I explained What and How in `target_suffix` and `contrasts/training_validation`.

docs/source/configuration_file.md
charleygros: `"default_model"` is the actual key of the dict that needs to be present in the config file, eg [here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/config/config.json#L58).
I agree this is not very clear, I am trying to clarify this.

docs/source/configuration_file.md
charleygros: I tried to re word this section.

docs/source/configuration_file.md
charleygros: What about:
```
- `length_3D`: (Integer, Integer, Integer). Size of the 3D patches used as model's input tensors, along the first, second and third dimension.
```

docs/source/configuration_file.md
charleygros: Example provided [here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/loader/loader.py#L423).
This parameter (and its behaviour) is likely to change is a very near future with PR #240 --> I am removing it for now and opening an issue to document it once refactored: issue #279.

docs/source/configuration_file.md
charleygros: 3D patch --> I have used this word instead

charleygros:
TODO @charleygros :
- [ ]  replace null by "" when string is asked.
- [ ] define the hyperlink to the documentation file losses.md

docs/source/configuration_file.md
jcohenadad: ah! so it is related to "https://joblib.readthedocs.io/en/latest/". I would add a hyperlink to "https://joblib.readthedocs.io/en/latest/" to avoid confusion

docs/source/configuration_file.md
```diff
@@ -0,0 +1,158 @@
...
+- `suffix`: String. Suffix of the derivative file containing the ROI used to crop (e.g. `"_seg-manual"`) with `ROICrop` as transform. Please use `null` if you do not want to use an ROI to crop.
+- `slice_filter_roi`: int. If the ROI mask contains less than `slice_filter_roi` non-zero voxels, the slice will be discarded from the dataset. This feature helps with noisy labels, e.g., if a slice contains only 2-3 labeled voxels, we do not want to use these labels to crop the image. This parameter is only considered when using `"ROICrop"`.
+
+## Split dataset
+
+#### fname_split
+String. File name of the log (joblib) that contains the list of training/validation/testing subjects. This file can later be used to re-train a model using the same data splitting scheme.
+
+#### random_seed
+Int. Seed used by the random number generator to split the dataset between training/validation/testing. The use of the same seed ensure the same split between the sub-datasets, which is useful to reproduce results.
```
jcohenadad, line 53:
```suggestion
Int. Seed used by the random number generator to split the dataset between training/validation/testing. The use of the same seed ensures the same split between the sub-datasets, which is useful to reproduce results.
```
the use of "Int." and "Integer" is inconsistent, but let's not spend too much additional time on this. At some point, it is possible that the json will be moved to a python class, with docstrings following strict convention for defining types, so no need to be "too" strict at this point

docs/source/configuration_file.md
jcohenadad: gotcha 😉

docs/source/configuration_file.md
jcohenadad: ahhh! i understand now. OK, what you did (i.e. remove "Tailored model") is now much much clearer.

docs/source/configuration_file.md
```diff
@@ -0,0 +1,158 @@
...
+- `retrain_fraction`: Float between 0. and 1. Controls the fraction of the pre-trained model that will be fine-tuned. For instance, if set to 0.5, the second half of the model will be fine-tuned while the first layers will be frozen.
+
+## Architecture
+Architectures for both segmentation and classification are available and described in the [Models](models.rst) section.
+If the selected architecture is listed in the [loader.py](../../ivadomed/loader/loader.py#L14) file, a classification (not segmentation) task is run. In the case of a classification task, the ground truth will correspond to a single label value extracted from `target`, instead being an array (the latter being used for the segmentation task).
+
+### default_model (Mandatory)
+Dict. Define the default model (`Unet`) and mandatory parameters that are common to all available architectures (listed in the [Models](models.rst) section). If a tailored model is defined (see next section), the default parameters are merged with the parameters that are specific to the tailored model.
```
jcohenadad, line 102: "(see next section)" does not work anymore

```suggestion
Dict. Define the default model (`Unet`) and mandatory parameters that are common to all available architectures (listed in the [Models](models.rst) section). For more specific models (see below), the default parameters are merged with the parameters that are specific to the tailored model.
```

docs/source/configuration_file.md
jcohenadad: i think you can even go with: "- `length_3D`: (Integer, Integer, Integer). Size of the 3D patches used as model's input tensors.".  It is fairly self-explanatory that the 1st, 2nd and 3rd arguments refer to the 1st, 2nd and 3rd dimension, although it was not clear before when the terms width, depth, etc. were used (because these terms imply a certain convention on the object itself)

docs/source/configuration_file.md
charleygros: I have opened an issue [here](https://github.com/neuropoly/ivado-medical-imaging/issues/280).

docs/source/configuration_file.md
charleygros: Issue opened [here](https://github.com/neuropoly/ivado-medical-imaging/issues/281).

dev/visualize_transforms.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
jcohenadad, line 1: that seems super useful! why not exposing it in the doc + package?

jcohenadad: also, should it be part of another PR? (this one is already getting fairly large)

charleygros: Oh yes my bad: I started to develop #282 on the wrong branch. I will take care of this when merging.

PR #276, opened 2020-06-06
Add preprocessing

andreanne-lemay:
Since the transforms refactor, we observed an increase in training time. After a time profiling of the code, the culprit seems to be the new resampling with scipy.ndimage.resize that's taking a significant amount of time during the interpolation.  In the current design, the resampling is made for every image at each iteration which is unecessary. Resampling and cropping should be done once per image at the beginning. This PR introduces preprocessing transforms.


ivadomed/loader/loader.py
```diff
@@ -358,27 +365,21 @@ def __getitem__(self, index):
         """
         seg_pair_slice, roi_pair_slice = self.indexes[index]
 
-        # Clean transforms params from previous transforms
-        # i.e. remove params from previous iterations so that the coming transforms are different
-        metadata_input = imed_loader_utils.clean_metadata(seg_pair_slice['input_metadata'])
-        metadata_roi = imed_loader_utils.clean_metadata(roi_pair_slice['gt_metadata'])
-        metadata_gt = imed_loader_utils.clean_metadata(seg_pair_slice['gt_metadata'])
```
charleygros, line 365: You removed it so that the preprocessing params (eg crop_params) are not wiped away. But it may involve that all the random transform will be the same at each `getitem` call since params already saved in metadata eg [here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/transforms.py#L562). What do you think?
A solution would be to remove `crop_params` from this list, while keeping these lines of `clean_metadata`. I am happy to help with this.

ivadomed/loader/loader.py
```diff
@@ -307,22 +312,24 @@ def __init__(self, filename_pairs, slice_axis=2, cache=True, transform=None, sli
         """
         self.indexes = []
         self.filename_pairs = filename_pairs
-        self.transform = transform
+        if isinstance(transform, list):
+            self.prepro_transforms, self.transform = transform
+        else:
+            self.transform = transform
+            self.prepro_transforms = None
```
charleygros, line 319: That's for cases where we are not going through `load_dataset` where we have:
```
tranform_lst = [prepro_transforms if len(preprocessing_transforms) else None, transforms]
```
Correct? :-)

ivadomed/preprocessing.py
```diff
@@ -0,0 +1,46 @@
+import copy
```
charleygros, line 1: I am wondering if we should not move these to `transforms.py` as there are only functions that split / apply transforms. Not too sure. Happy to discuss :-)

ivadomed/preprocessing.py
```diff
@@ -0,0 +1,46 @@
...
+
```
charleygros, line 2: Another annoying comment but I wanted to ask:
What about: to keep everything as before in config files and loader and only modify `transforms.py` as follows:
- for all `__call__` of our preprocessing transforms (ie Resample and Crop): just before the return: add a metadata in this sample saying that "all good, preprocessing performed on this sample, no need to run it again"
- add a decorator, eg `@preprocessing` which check the metadata of the input sample --> if already done --> skip

andreanne-lemay: We'd still have perform the transforms in the load_filenames, because if we don't the images are not preprocessed -> if we keep the code as before we do the transform a first time, we change the metadata to indicate it has been done -> on the next getitem, the segpair retrived from index has not been cropped and resampled. I don't see how we can do this if we don't apply the specified transforms before calling getitem (in the load_filenames fonction), but I might not fully grasp what you're suggesting, so correct me if I'm wrong :)

ivadomed/loader/loader.py
andreanne-lemay: ahh yes I see! I only saw ROICrop (if crop_params in metadata...) , that's the first thing I tried -> removing crop_params from the list, but I was getting an error with ROICrop because the param was not updated here: https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/loader/utils.py#L287 . If I remove crop_params but just from the update_metadata list I think everything should work

ivadomed/preprocessing.py
```diff
@@ -0,0 +1,46 @@
...
+
+def apply_transforms(transforms, seg_pair, roi_pair=None):
+    if transforms is None:
+        return (seg_pair, roi_pair)
+
+    if roi_pair is not None:
+        stack_roi, metadata_roi = transforms(sample=roi_pair["gt"],
+                                             metadata=roi_pair['gt_metadata'],
+                                             data_type="roi")
```
charleygros, line 23: We would also need to spread the params from metadata_roi (where we compute the crop params for ROICrop to metdata_input and metadata_gt --> via `update_medaata`

ivadomed/loader/loader.py
andreanne-lemay: yes exactly :)

PR #290, opened 2020-06-12
Postprocessing: Adding Conditional Random Field (CRF) interrogation mark

### LINKED ISSUE #135: Post-processing

charleygros:
### Aim
Implement Conditional Random Field (CRF) method in the code base and evaluate its benefit as postprocessing tool.

### CRF
CRF are increasingly used in combination to Deep Learning approach, eg [Ref1](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6700294/), [Ref2](https://arxiv.org/abs/1610.02177), [Ref3](https://arxiv.org/abs/1805.04777).
For a long time impracticable in segmentation tasks, their use has recently been allowed thanks to this [paper](https://arxiv.org/abs/1210.5644). This approach has been implemented in this package, [pydcrf](https://github.com/lucasb-eyer/pydensecrf).
Note to be discussed: this is a (Cython-based) Python wrapper.

### Done
- Implement CRF as a postprocessing method:
    - Unary potential: per-voxel class-probabilities (ie soft output of CNN). It doesn't account for neighborhoods at all.
    - Pairwise potential: smooth out predictions
        1. Feature --> Bilateral potential: the input image. **Prior** = voxels with a similar intensity are likely to belong to the same class.
        2. Spatial --> Gaussian potential. **Prior** =  voxels with a similar intensity are likely to belong to the same class --> enforces more spatially consistent segmentation.
     - Run inference, KL divergence et output the MAP.

```python
def apply_crf(predictions, image, n_iterations=5, eps=1e-6):
    """
    Apply Conditional Random Fields to the soft predictions. 2D inputs, with shape: n_label, height, width.

    Args:
        predictions (np.array): Input 2D soft segmentation.
        image (np.array): Input 2D image.
        n_iterations (int):
        eps (float): To avoid log(0), need to clip 0 probabilities to a positive value
    Returns:
        Array.
    """
```

### Todo
- Add uncertainty as an additional pairwise **Prior**.
- Evaluate the performances of the method.

Fixes #135.

### DIFF
--- a/ivadomed/postprocessing.py
+++ b/ivadomed/postprocessing.py
@@ -3,6 +3,7 @@
 import functools
 import numpy as np
 import nibabel as nib
+import pydensecrf.densecrf as dcrf
 from scipy.ndimage.measurements import label
 from scipy.ndimage.morphology import binary_fill_holes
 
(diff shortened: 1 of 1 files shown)


charleygros:
CLosing PR as that's not the current priority but keeping the branch alive.

PR #295, opened 2020-06-13
Fix of Random Affine 

andreanne-lemay:
This PR restores the RandomAffine transform which includes, rotation, translation and scale transform. To avoid code duplication, I removed RandomRotation and RandomTranslation. Transformation matrices are used to apply the transforms.


ivadomed/transforms.py
```diff
@@ -575,44 +577,108 @@ def __init__(self, degrees):
                 "degrees should be a list or tuple and it must be of length 2."
             self.degrees = degrees
 
+        # Scale
+        if scale is not None:
+            assert isinstance(scale, list)
```
charleygros, line 582: What about adding 
```
, \
                "sclae should be a list."
```
and things aboutlengths as done for transalte.

ivadomed/transforms.py
```diff
@@ -1,15 +1,15 @@
...
 import math
 import numbers
 import random
-import copy
-import numpy as np
 
+import numpy as np
 import torch
-from scipy.ndimage import rotate, shift
+from scipy.ndimage import shift
```
charleygros, line 9: Do we still need this one too?

PR #308, opened 2020-06-16
Fix automate training

andreanne-lemay:
After the main refactoring, automate training was not fully adapted to the new code. 
To make the file work I had to do some changes to the main:
- add return values for training, testing and eval
- add parameter containing the config file to run_main


ivadomed/scripts/automate_training.py
```diff
@@ -82,8 +82,8 @@ def test_worker(config):
     try:
         # Save best test score
 
-        config["command"] = "eval"
-        test_dict, eval_df = ivado.cmd_eval(config)
+        config["command"] = "test"
+        test_dict = ivado.run_main(config)
```
charleygros, line 86: Could you help me here for this change?

andreanne-lemay: the return value of test (test_dict) was used here and no values of the config file were displayed are used in automate_training, so the easiest fix was to simply call the test command since we are only using the dice score.

charleygros: I think the idea to call eval was to, in the future, add a parameter that controls the metric you want to optimise  --> I am thinking of implementing this in this PR and then move back to eval: would it be okay with you?

PR #311, opened 2020-06-18
Fill docstrings

charleygros:
@andreanne-lemay 
- [x] loader folder
- [x] object detection folder
- [x] models
- [x] losses
- [x] metrics

@charleygros 
- [x] main
- [x] training
- [x] testing
- [x] evaluation
- [x] transforms
- [x] utils

**Done:**
- [x] scripts folder
- [x] postprocessing

and vice-versa for review.

Fixes #250.

PR #313, opened 2020-06-23
Add PR311 docstrings to RTD

charleygros:
We have added docstrings quite everywhere in the code base in PR #311 . Let's add it to the showcase!


jcohenadad:
> Can we change the default display so that Attributes appear in the same format as parameters for instance? 

agreed about the problem, and yes, we can definitely change the output style. I believe [adding a custom css](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_css_files) would be the solution (it's fairly easy to do).

Issue #320, opened 2020-06-25
Multiclass segmentation: DiceLoss vs MultiClassLoss


charleygros:
Knot bag story!

If you would like to reproduce this experiment:
- master commit [60c6999](https://github.com/ivadomed/ivadomed/tree/60c6999c18a9bb72f4d5063821c1e866dbe13a21)
- config file: `duke/projects/iXX/config_csfsc.json`

### Chapter 1:
Multi-class task training: Loss=MultiClassDiceLoss + Display of DiceLoss
```
Epoch 13 training loss: -0.5068.        Dice training loss: -0.1737.            
Epoch 13 validation loss: -0.5325.      Dice validation loss: -0.2179. 
```
The difference between the two is.. not normal.
Hummm

### Chapter 2:
Mmm let's have a go at training each class separetely...
Reaching 85 of Dice easily.
Hummm Hummm

### Chapter 3:
Mmm let's go back to multi-class and print [each component of the MultiClassDiceLoss](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/losses.py#L35):
```
tensor(-0.8983, device='cuda:7', grad_fn=<DivBackward0>)
tensor(-0.0949, device='cuda:7', grad_fn=<DivBackward0>)
Epoch 16 training loss: -0.5130.        Dice training loss: -0.1910.
```
Interesting...
Let's look at the images: First row = Gt // Second row = Pred // First Column = SC // Second Column = CSF
![Figure_0](https://user-images.githubusercontent.com/14353425/85722494-b0779c80-b735-11ea-843c-b4a348cfab01.png)
Looks like the RandomAffine is not the same for Class 1 vs Class 2.

### Chapter 4:
Looking at the translation params for one tuple (im, gt_class0, gt_class1) I got:
```
(-1.0, -13.0, 0.0)
(-1.0, -13.0, 0.0)
(1.0, -12.0, 0.0)
```
--> The transformation applied to gt_class1 is indeed not the same as im and gt_class0.

The idea was then to play with [update_metadata](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/loader/utils.py#L374).
Played around for a bit and ended up with (from [L385](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/loader/utils.py#L385)):
```python
    if metadata_src_lst and metadata_dest_lst:
        for idx in range(len(metadata_dest_lst)):
[...]

PR #321, opened 2020-06-25
Propagation of random affine parameters

andreanne-lemay:
See issue: #320 

Summary: Random affine parameters were not the same for multi-labels or multi-channels.

### DIFF
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -387,5 +387,6 @@ def update_metadata(metadata_src_lst, metadata_dest_lst):
             metadata_dest_lst = metadata_dest_lst + [SampleMetadata({})] * \
                                 (len(metadata_src_lst) - len(metadata_dest_lst))
         for idx in range(len(metadata_src_lst)):
-            metadata_dest_lst[idx]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
+            for i_dest in range(len(metadata_dest_lst)):
+                metadata_dest_lst[i_dest]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
     return metadata_dest_lst
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -383,9 +383,5 @@ def update_metadata(metadata_src_lst, metadata_dest_lst):
         list: updated metadata list.
     """
     if metadata_src_lst and metadata_dest_lst:
-        if len(metadata_src_lst) > len(metadata_dest_lst):
-            metadata_dest_lst = metadata_dest_lst + [SampleMetadata({})] * \
-                                (len(metadata_src_lst) - len(metadata_dest_lst))
-        for idx in range(len(metadata_src_lst)):
-            metadata_dest_lst[idx]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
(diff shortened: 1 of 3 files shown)


ivadomed/loader/utils.py
```diff
@@ -387,5 +387,6 @@ def update_metadata(metadata_src_lst, metadata_dest_lst):
             metadata_dest_lst = metadata_dest_lst + [SampleMetadata({})] * \
                                 (len(metadata_src_lst) - len(metadata_dest_lst))
         for idx in range(len(metadata_src_lst)):
-            metadata_dest_lst[idx]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
+            for i_dest in range(len(metadata_dest_lst)):
+                metadata_dest_lst[i_dest]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
```
charleygros: Not sure of something:
- if we go that way, we can remove the "if" thing just above, right?
- but still: at each iteration of idx, we are "overwritting" `metadata_dest_lst[i_dest]`. Not sure if that s what we want, but we may miss something.

What about [solution in Chapter 4](https://github.com/ivadomed/ivadomed/issues/320#issue-645561146)? I can't think on a case where it would not work.

charleygros: Oh my bad, it is outdated

charleygros: Oh, it is outdated

ivadomed/transforms.py
```diff
@@ -31,6 +31,8 @@ def wrapper(self, sample, metadata):
         if isinstance(sample, list):
             list_data, list_metadata = [], []
             for s_cur, m_cur in zip(sample, metadata):
+                if len(list_metadata):
+                    imed_loader_utils.update_metadata([list_metadata[-1]], [m_cur])
```
charleygros: I see, that's a good idea.. to propagate params intra channels :+1:

ivadomed/loader/utils.py
```diff
@@ -387,6 +387,5 @@ def update_metadata(metadata_src_lst, metadata_dest_lst):
             metadata_dest_lst = metadata_dest_lst + [SampleMetadata({})] * \
                                 (len(metadata_src_lst) - len(metadata_dest_lst))
         for idx in range(len(metadata_src_lst)):
-            for i_dest in range(len(metadata_dest_lst)):
-                metadata_dest_lst[i_dest]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
+            metadata_dest_lst[idx]._update(metadata_src_lst[idx], TRANSFORM_PARAMS)
```
charleygros: With the change in the transform decorator --> we would need only to update metadata_src_lst[0] with metadata_dest_lst[0] :-)

ivadomed/transforms.py
```diff
@@ -31,7 +31,7 @@ def wrapper(self, sample, metadata):
         if isinstance(sample, list):
             list_data, list_metadata = [], []
             for s_cur, m_cur in zip(sample, metadata):
-                if len(list_metadata):
+                if len(list_metadata) > 0:
```
charleygros, line 34: Because of SampleMetadata({}) I imagine...

ivadomed/loader/utils.py
andreanne-lemay: Changed it :)

PR #327, opened 2020-06-26
Creating new file named maths for mathematical operation

lrouhier:
The Goal of this PR is to create a new file name maths.py that will be used to list useful mathematical functions for image manipulation.


charleygros:
I can't think of other function to put in this file at this point.
But this is a great initiative.. and I am sure that in the future we will found things to put in it, now that it exists.

ivadomed/maths.py
```diff
@@ -0,0 +1,40 @@
...
+        kernlen (int): size of kernel
+
+    Returns:
+        ndarray: a 2D array of size (kernlen,kernlen)
+    """
+
+    x = np.linspace(-1, 1, kernlen + 1)
+    kern1d = np.diff(scipy.stats.norm.cdf(x))
+    kern2d = np.outer(kern1d, kern1d)
+    return imed_transforms.rescale_values_array(kern2d / kern2d.sum())
```
charleygros, line 22: `rescale_values_array` could be a good candidate for this file!!!

ivadomed/maths.py
```diff
@@ -0,0 +1,40 @@
+import nibabel as nib
```
charleygros, line 1: Could you please add this file to the API / RST? thanks :)

lrouhier: Will do

ivadomed/maths.py
lrouhier: Good idea !!

lrouhier: done in 8aed599

ivadomed/maths.py
lrouhier: done in 064412a

charleygros: All good to me now :-)

docs/source/api_ref.rst
```diff
@@ -89,3 +89,9 @@ Utils API
 
 .. automodule:: ivadomed.utils
 
+
+Maths API
+++++++++++++++++++++
```
jcohenadad, line 94: i think the convention is to use the same number of characters as the text
```suggestion
++++++++++++++++++++
```

lrouhier: done in: 503aaa8

PR #333, opened 2020-07-06
Trained model directory

### LINKED ISSUE #322: Naming convention for models
Init debate: [this comment](https://github.com/neuropoly/spinalcordtoolbox/pull/2748#discussion_r445873840).

charleygros:
### Done:
- config files: add a new param, `model_name`, which is the name of the folder where the trained model and its configuration file are saved after training: `log_directory/model_name/model_name.onnx` and `log_directory/model_name/model_name.json`.
- RTD: document `model_name` in config file doc + make the model name convention explicit:
~~~
task_(animal)_region_(contrast)
~~~
with:
~~~
task = {seg, label, find}
animal = {human, dog, cat, rat, mouse, ...}
region = {sc, gm, csf, brainstem, ...}
contrast = {t1, t2, t2star, dwi, ...}
~~~
- main: save the best trained model (according to the validation loss), with ONNX format, in the appropriate folder and naming.

Fixes #322 .


ivadomed/utils.py
```diff
@@ -322,6 +322,7 @@ def mixup(data, targets, alpha, debugging=False, ofolder=None):
     lambda_ = max(lambda_, 1 - lambda_)  # ensure lambda_ >= 0.5
     lambda_tensor = torch.FloatTensor([lambda_])
 
+
```
lrouhier, line 325: Is the extra line needed?

ivadomed/main.py
```diff
@@ -172,6 +174,24 @@ def run_main(config=None):
...
+            os.makedirs(path_model)
+        else:
+            print('Model directory already exists: {}'.format(path_model))
+        # Copy model metadata
+        shutil.copyfile(os.path.join(log_directory, "config_file.json"),
+                        os.path.join(path_model, context["model_name"]+".json"))
+        # Copy model and convert it to ONNX format
+        shutil.copyfile(os.path.join(log_directory, "best_model.pt"),
+                        os.path.join(path_model, context["model_name"] + ".pt"))
+        convert_pytorch_to_onnx(fname_model=os.path.join(path_model, context["model_name"] + ".pt"),
```
andreanne-lemay, line 190: Here instead of saving, converting .pt file to .onnx, then deleting .pt file why not save it directly: `imed_utils.save_onnx_model(model, var_input, model_path.replace('pt', 'onnx'))`. This code could also be moved in function `train` to have access to the var_input variable.

docs/source/configuration_file.rst
```diff
@@ -21,6 +21,16 @@ log\_directory
 Folder name that will contain the output files (e.g., trained model,
 predictions, results).
 
+model\_name
+^^^^^^^^^^^^^^
+
+Folder name containing the trained model (ONNX format) and its configuration
```
andreanne-lemay, line 27: :+1:

ivadomed/main.py
charleygros: Yes, that's a bit of an overkill, I agree. The reason was: to do all this saving/packaging at the same place.
But I agree, I am going to move this to `train`.

charleygros: Ah and the other reason was: to avoid sending a new parameter to the training (`model_name`).. oh well

charleygros: Ah and last reason was: to avoid saving the best model twice (pt and onnx) at each epoch where we have a new best model.

charleygros: ahah at the end I am not too sure what is best

andreanne-lemay: The model doesn't have to be saved at every epoch, the saving + packaging can be put at the same place, at the end of `train` function after the `for` loop (so this covers, 2 out of 3 issues). As for the extra parameter, I agree it is not ideal. The model_name could be placed into `model_params` or `training_parameters` to avoid passing another param, but I don't know if it's better.

Another approach would be to leave the code where it is but create a dummy input (instead of using var_input).

I don't have a strong preference :) I let the choice up to you!

PR #338, opened 2020-07-08
Soft gt for 3d and adaptative loader

andreanne-lemay:


charleygros:
Oh, could you please also change the order in the undo_transform of resample? thx

andreanne-lemay:
> Oh, could you please also change the order in the undo_transform of resample? thx

Done :)! I tested it (verify if output is soft), everything looks good

PR #342, opened 2020-07-09
Apply preprocessing after filter ROI

charleygros:
We recently refactored the transforms so that some transforms are performed only once (so called "preprocessing", eg Crop) while the remaining are performed for each epoch (eg RandomAffine).
These preprocessing operations are done within `load_filename()`, before calling `filter_roi`. see [here](https://github.com/ivadomed/ivadomed/blob/270864fa3a1674e1ab4f2d2a57aa933a8ce8a49a/ivadomed/loader/loader.py#L100).
It indroduces a problem with the `ROICrop`: `ROICrop` is performed on potentially empty ROI slices --> error with the center of mass computation.

To avoid this, I am suggesting to split `load_filename` and preprocessing call in two different functions so that we could have the sucession: (1) load_filename, (2) filter_roi, (3) apply_prepro.

There is different ways to address this problem (eg call filter_roi within load_filename), I am happy to discuss with you the Pros and Cons before moving things around (for now, only done for the generic loader).

### DIFF
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -425,6 +425,11 @@ class MRI2DSegmentationDataset(Dataset):
                 # Note: we force here gt_type=segmentation since ROI slice is needed to Crop the image
                 slice_roi_pair = roi_pair.get_pair_slice(idx_pair_slice, gt_type="segmentation")
 
+                if self.slice_filter_roi:
+                    filter_roi = imed_loader_utils.filter_roi(slice_roi_pair['gt'], self.roi_thr)
+                    if filter_roi:
+                        continue
+
                 item = imed_transforms.apply_preprocessing_transforms(self.prepro_transforms,
                                                                       slice_seg_pair,
                                                                       slice_roi_pair)
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -727,6 +727,7 @@ class BidsDataset(MRI2DSegmentationDataset):
                  multichannel=False, object_detection_params=None, task="segmentation", soft_gt=False):
 
         self.bids_ds = bids.BIDS(root_dir)
+        self.roi_params = roi_params if roi_params is not None else {"suffix": None, "slice_filter_roi": None}
         self.soft_gt = soft_gt
         self.filename_pairs = []
         if metadata_choice == 'mri_params':
@@ -784,11 +785,11 @@ class BidsDataset(MRI2DSegmentationDataset):
                         if deriv.endswith(subject.record["modality"] + suffix + ".nii.gz"):
                             target_filename[idx] = deriv
 
-                    if not (roi_params["suffix"] is None) and \
-                            deriv.endswith(subject.record["modality"] + roi_params["suffix"] + ".nii.gz"):
+                    if not (self.roi_params["suffix"] is None) and \
+                            deriv.endswith(subject.record["modality"] + self.roi_params["suffix"] + ".nii.gz"):
                         roi_filename = [deriv]
 
-                if (not any(target_filename)) or (not (roi_params["suffix"] is None) and (roi_filename is None)):
+                if (not any(target_filename)) or (not (self.roi_params["suffix"] is None) and (roi_filename is None)):
                     continue
 
                 if not subject.has_metadata():
@@ -827,5 +828,5 @@ class BidsDataset(MRI2DSegmentationDataset):
                     self.filename_pairs.append((subject["absolute_paths"], subject["deriv_path"],
                                                 subject["roi_filename"], subject["metadata"]))
 
-        super().__init__(self.filename_pairs, slice_axis, cache, transform, slice_filter_fn, task, roi_params,
+        super().__init__(self.filename_pairs, slice_axis, cache, transform, slice_filter_fn, task, self.roi_params,
                          self.soft_gt)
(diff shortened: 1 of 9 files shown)


ivadomed/loader/loader.py
```diff
@@ -51,7 +51,7 @@ def load_dataset(data_list, bids_path, transforms_params, model_params, target_s
     slice_filter_params and object_detection_params see :doc:`configuration_file`.
     """
     # Compose transforms
-    tranform_lst, _ = imed_transforms.preprare_transforms(transforms_params, requires_undo)
+    tranform_lst, _ = imed_transforms.preprare_transforms(copy.deepcopy(transforms_params), requires_undo)
```
charleygros, line 54: @andreanne-lemay : I ve added this deepcopy otherwise, when we arrive [here](https://github.com/ivadomed/ivadomed/blob/270864fa3a1674e1ab4f2d2a57aa933a8ce8a49a/ivadomed/loader/loader.py#L103), the ROICrop (for instance) is no longer in this list because of the `del` that are performed within `prepare_transforms`.

andreanne-lemay:
I think the easiest way to resolve this issue would be to apply filter_roi within load_filenames. It is also more direct and requires less iterations over all data, because the way I see it if we successively do (1) load_filename, (2) filter_roi, (3) apply_prepro, we'll have to load data in a handler (first loop on data), go through all slices of every image to filter roi (2nd loop) then go another time across all data to do preprocessing, But there might be a more efficient way to do this that I didn't consider :). I don't see the cons to call filter_roi in load_filenames, have you thought of any?

charleygros:
I would agree. The only reason that prevented me to do this is... to add another param to the loader... I confess I get a bit phobic about this.
This param could be pass along roi_suffix in a dict "roi_param" to the loader for instance. I will modify this.

andreanne-lemay:
Yes, I understand your aversion for adding to adding to new parameters ;)! But, why do we need to add a new parameter in the config file, can't we just use the already existing param `slice_filter_roi`?

charleygros:
> Yes, I understand your aversion for adding to adding to new parameters ;)! But, why do we need to add a new parameter in the config file, can't we just use the already existing param `slice_filter_roi`?

I meant to add an argument to the loader (class) :-)

ivadomed/loader/loader.py
lrouhier: I may have a stupid question but ... why is the name `preprare_transforms` instead of `prepare_transforms` ?

charleygros: Not a stupid question --> Rather a "stupid charley" :-)

andreanne-lemay: Hahah wow :') this is actually my fault, I defined the function this way :'). I updated the code!

ivadomed/loader/loader.py
```diff
@@ -371,18 +375,28 @@ class MRI2DSegmentationDataset(Dataset):
...
         self.filename_pairs = filename_pairs
         self.prepro_transforms, self.transform = transform
         self.cache = cache
         self.slice_axis = slice_axis
         self.slice_filter_fn = slice_filter_fn
         self.n_contrasts = len(self.filename_pairs[0][0])
+        if roi_params is None:
+            roi_params = {"suffix": None, "slice_filter_roi": None}
+        self.roi_thr = roi_params["slice_filter_roi"]
+        if roi_params["suffix"] is not None and isinstance(self.roi_thr, int):
```
andreanne-lemay, line 396: This could be simplified to `self.slice_filter_roi = roi_params["suffix"] is not None and isinstance(self.roi_thr, int)`

ivadomed/loader/loader.py
```diff
@@ -413,6 +427,11 @@ def load_filenames(self):
                 # Note: we force here gt_type=segmentation since ROI slice is needed to Crop the image
                 slice_roi_pair = roi_pair.get_pair_slice(idx_pair_slice, gt_type="segmentation")
 
+                if self.slice_filter_roi:
```
andreanne-lemay, line 430: Suggestion:
```python
if self.slice_filter_roi and imed_loader_utils.filter_roi(slice_roi_pair['gt'], self.roi_thr):
     continue
```

ivadomed/loader/utils.py
```diff
@@ -165,35 +165,26 @@ def imed_collate(batch):
...
-            continue
-        if np.count_nonzero(roi_data) <= nb_nonzero_thr:
-            continue
-
-        filter_indexes.append((segpair, slice_roi_pair))
-
-    # Update dataset
-    ds.indexes = filter_indexes
-    return ds
+    return False
```
andreanne-lemay, line 187: Suggestion:
```python
return not np.any(roi_data) or np.count_nonzero(roi_data) <= nb_nonzero_thr
```
 instead of 
```python
    if not np.any(roi_data):
        return True
    if np.count_nonzero(roi_data) <= nb_nonzero_thr:
        return True

    return False
```

andreanne-lemay:
I kept having a failed RandomShiftIntensity test, so I also commented this test for now, until we sort the problem with this test or with the class :)

ivadomed/loader/loader.py
charleygros: zuper

ivadomed/loader/loader.py
charleygros: shinny

ivadomed/loader/utils.py
charleygros: brilliant

PR #357, opened 2020-07-13
Documentation: metric more formal defintion

charleygros:
### Done:
- Dice
- Precision
- Recall
- Accuracy
- Specificity
- Multi class Dice

### DIFF
--- a/ivadomed/metrics.py
+++ b/ivadomed/metrics.py
@@ -65,7 +65,7 @@ def numeric_score(prediction, groundtruth):
     Returns:
         float, float, float, float: FP, FN, TP, TN
     """
-    FP = np.float(np.sum(prediction * (1.0- groundtruth)))
+    FP = np.float(np.sum(prediction * (1.0 - groundtruth)))
     FN = np.float(np.sum((1.0 - prediction) * groundtruth))
     TP = np.float(np.sum(prediction * groundtruth))
     TN = np.float(np.sum((1.0 - prediction) * (1.0 - groundtruth)))
--- a/ivadomed/metrics.py
+++ b/ivadomed/metrics.py
@@ -46,10 +46,17 @@ class MetricManager(object):
 def numeric_score(prediction, groundtruth):
     """Computation of statistical numerical scores:
 
-    * FP = False Positives
-    * FN = False Negatives
-    * TP = True Positives
-    * TN = True Negatives
+    * FP = Soft False Positives
+    * FN = Soft False Negatives
+    * TP = Soft True Positives
+    * TN = Soft True Negatives
+
+    Robust to hard or soft input masks. For example::
+        prediction=np.asarray([0, 0.5, 1])
+        groundtruth=np.asarray([0, 1, 1])
+        Leads to FP = 1.5
+
+    Note: It assumes input values are between 0 and 1.
 
     Args:
         prediction (ndarray): Binary prediction.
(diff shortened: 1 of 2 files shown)


charleygros:
A kind of important related question is: we have a soft dice but a hard precision, recall etc. We could move to a soft version of it by replacing:
```python
    FP = np.float(np.sum((prediction == 1) & (groundtruth == 0)))
```
by:
```python
    FP = np.float(np.sum((prediction > 0) & (groundtruth == 0)))
```
But... is it what we want?

lrouhier:
> A kind of important related question is: we have a soft dice implementation but a hard IoU, hard precision, hard recall etc. 

I see what you mean but I know that some times soft output have not 0 as background but rather 1 x 10^-9, then you will have a huge number of false positives, yet, the image can look great visually so I don't know if these metrics make sens with soft outputs ... (you will likely have 0 false negatives as well).  Maybe a soft thresholding function will be needed for your suggestion to work? What I mean by 'soft_tresholding' is:

~~~
for all element in image: 
         if element value is lower than threshold
                 element takes the value 0 
         else
                 element keeps it float value
~~~

Maybe the use of MSE and other metrics used in regression will provide more insight?

ivadomed/metrics.py
```diff
@@ -68,7 +68,8 @@ def numeric_score(prediction, groundtruth):
 def dice_score(im1, im2, empty_score=np.nan):
     """Computes the Dice coefficient between im1 and im2.
 
-    Compute a soft Dice coefficient between im1 and im2.
+    Compute a soft Dice coefficient between im1 and im2, ie equals twice the sum of the two masks product, divided by
```
lrouhier, line 71:
```suggestion
    Compute a soft Dice coefficient between im1 and im2, it equals twice the sum of the two masks product, divided by
```

charleygros:
> I see what you mean but I know that some times soft output have not 0 as background but rather 1 x 10^-9, then you will have a huge number of false positives, yet, the image can look great visually so I don't know if these metrics make sens with soft outputs ... (you will likely have 0 false negatives as well). Maybe a soft thresholding function will be needed for your suggestion to work? What I mean by 'soft_tresholding' is:

I see what you mean, that's a good point. I think that the reason we implemented [this line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/utils.py#L84).

lrouhier:
> I see what you mean, that's a good point. I think that the reason we implemented [this line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/utils.py#L84): in case a soft output is required, we still discard the "noise".

Yes exactly. I just think this needs to be taken into consideration before computing metrics. But then do the metrics still make sens? like if you have a 0.1 instead of a 1 for a specific voxel, it will still be considered a true positive (if I am not mistaken). I am going to look for metrics used with soft output :)

andreanne-lemay:
> A kind of important related question is: we have a soft dice implementation but a hard IoU, hard precision, hard recall etc. We could move to a soft version of it by replacing:
> 
> ```python
>     FP = np.float(np.sum((prediction == 1) & (groundtruth == 0)))
> ```
> 
> by:
> 
> ```python
>     FP = np.float(np.sum((prediction > 0) & (groundtruth == 0)))
> ```
> 
> But... is it what we want?

Why not something like: np.float(np.sum(prediction > 0) * prediction * int(groundtruth == 0))) This way FP of [0, 0.5, 1] and [0, 1, 1]  gives 1.5 instead of 1 or 2 (which is the options described above). This way its a "soft" FP value.

charleygros:
Great idea @andreanne-lemay. Herebelow what I implemented:
```python
    FP = np.float(np.sum(prediction * (1.0- groundtruth)))
    FN = np.float(np.sum((1.0 - prediction) * groundtruth))
    TP = np.float(np.sum(prediction * groundtruth))
    TN = np.float(np.sum((1.0 - prediction) * (1.0 - groundtruth)))
```
Does it sound good?

PR #362, opened 2020-07-15
Automate training improvement

andreanne-lemay:
New features:
- flag `-l` or `--all-logs` that keeps all log directory for every iteration if used
- eval computation (before only test)
- generation of `average_eval.csv` which contains the average of eval dataframe (one per condition tested). The average is made on the n iterations.

### DIFF
--- a/ivadomed/scripts/automate_training.py
+++ b/ivadomed/scripts/automate_training.py
@@ -151,11 +151,12 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
                 "default_model": {"depth": [2, 3, 4]}
 
         fixed_split (bool): If True, all the experiments are run on the same training/validation/testing subdatasets.
-                            Flag: --fixed_split
+                            Flag: --fixed-split
         all_combin (bool): If True, all parameters combinations are run. Flag: --all-combin
         n_iterations (int): Controls the number of time that each experiment (ie set of parameter) are run.
-                            Flag: --n_iteration, -n
-        run_test (bool): If True, the trained model is also run on the testing subdataset. flag: --run_test
+                            Flag: --n-iteration, -n
+        run_test (bool): If True, the trained model is also run on the testing subdataset. flag: --run-test
+        all_logs (bool): If True, all the log directories are kept for every iteration. Flag: --all-logs, -l
     """
     # Load initial config
     with open(config, "r") as fhandle:
--- a/ivadomed/scripts/automate_training.py
+++ b/ivadomed/scripts/automate_training.py
@@ -233,9 +233,10 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
                 config["random_seed"] = seed
                 if all_logs:
                     if i:
-                        config["log_directory"] = config["log_directory"].replace("_n=" + str(i - 1), "_n=" + str(i))
+                        config["log_directory"] = config["log_directory"].replace("_n=" + str(i - 1).zfill(2),
+                                                                                  "_n=" + str(i).zfill(2))
                     else:
-                        config["log_directory"] += "_n=" + str(i)
+                        config["log_directory"] += "_n=" + str(i).zfill(2)
         validation_scores = pool.map(train_worker, config_list)
         val_df = pd.DataFrame(validation_scores, columns=[
             'log_directory', 'best_training_dice', 'best_training_loss', 'best_validation_dice',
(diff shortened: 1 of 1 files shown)


ivadomed/scripts/automate_training.py
```diff
@@ -88,6 +89,9 @@ def test_worker(config):
         test_dict = ivado.run_main(config)
         test_dice = test_dict['dice_score']
 
+        config["command"] = "eval"
+        df_results = ivado.run_main(config)
```
charleygros, line 93: EZ PZ

ivadomed/scripts/automate_training.py
```diff
@@ -220,13 +224,18 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
...
             # Set seed for iteration
             seed = random.randint(1, 10001)
             for config in config_list:
                 config["random_seed"] = seed
-
+                if all_logs:
+                    if i:
+                        config["log_directory"] = config["log_directory"].replace("_n=" + str(i - 1), "_n=" + str(i))
+                    else:
+                        config["log_directory"] += "_n=" + str(i)
```
charleygros, line 238: Does not really matter but I have not much to say.. so I am going to suggest it: `str(i)` --> `str(i).zfill(2)`

ivadomed/scripts/automate_training.py
```diff
@@ -242,17 +251,36 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
...
+            test_results = pool.map(test_worker, config_list)
+
+            df_lst = []
+            # Merge all eval df together to have a single excel file
+            for j, result in enumerate(test_results):
+                df = list(result).pop()
+                id = result[0].split("_n=")[0]
+                rows = df.index.values
+                for idx, row in enumerate(rows):
+                    df.rename({row: id + "_" + row}, axis='index', inplace=True)
```
charleygros, line 263: zuper

ivadomed/scripts/automate_training.py
```diff
@@ -242,17 +251,36 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
...
+            combined_df = val_df.set_index('log_directory').join(test_df.set_index('log_directory'))
             combined_df = combined_df.reset_index()
 
         else:
             combined_df = val_df
 
         results_df = pd.concat([results_df, combined_df])
         results_df.to_csv("temporary_results.csv")
+        eval_df /= n_iterations
+        eval_df.to_csv("average_eval.csv")
```
charleygros, line 283: Not sure I followed correctly: what are the columns / rows of this df at this end?

ivadomed/scripts/automate_training.py
```diff
@@ -278,7 +306,7 @@ def main():
     args = parser.parse_args()
     # Run automate training
     automate_training(args.config, args.params, bool(args.fixed_split), bool(args.all_combin), int(args.n_iterations),
-                      bool(args.run_test))
+                      bool(args.run_test), args.all_logs)
```
charleygros, line 309: To add in RTD? : 3

ivadomed/scripts/automate_training.py
andreanne-lemay: Here is the general look:
image

Each dataframe has the same format as the one for one training, so rows are each test subj and this for each condition we are testing. Rows are all the metrics :). Glad to make modifications to make it more convenient

ivadomed/scripts/automate_training.py
andreanne-lemay: Absolutely! will do :)

ivadomed/scripts/automate_training.py
charleygros: Yes, that's what I thought --> that's why I got confused by the "/= n_iterations" --> how is it used?

charleygros: And is there a reference to the log_directory_idXX in the df? could be handy too!

andreanne-lemay: n_iterations is to compute the average (before that it's the sum) :3 . For the reference, I put the name of the log_directory as prefix for each subject (row names), is this what you mean?

ivadomed/scripts/automate_training.py
andreanne-lemay: Updated :D!

ivadomed/scripts/automate_training.py
andreanne-lemay: Done!

ivadomed/scripts/automate_training.py
charleygros: Oooo got you: there is two df involved: the one where eval_df are appended (ie screenshot above) and the final df (average_eval.csv) which is the average of each metric. All guccci

PR #364, opened 2020-07-17
Optimize binarization

andreanne-lemay:
Following a time profiling, of soft vs hard labels, the [function `as_tensor`](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/loader/loader.py#L476) showed to be time-consuming. 
image

This PR optimize this operation, to reduce training time.


andreanne-lemay:
I am also questioning the choice of the 0.1 threshold, why 0.1 rather than 0.5?

charleygros:
> I am also questioning the choice of the 0.1 threshold, why 0.1 rather than 0.5?

Yeah or 0.35?... that's what makes me so ... uncomfortable with thresholds.. ..
0.5 is however more consensual I agree, let's change it to 0.5

PR #370, opened 2020-07-21
Split dataset with no test center specified

andreanne-lemay:
Currently, if no test center is specified, `split_dataset` is equivalent to `per_patient` split. This PR chooses at random test centers, if no center is specified.

### DIFF
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -48,6 +48,7 @@ def split_dataset(path_folder, center_test_lst, split_method, random_seed, train
         # make sure that subjects coming from some centers are unseen during training
         if len(center_test_lst) == 0:
             centers = list(df['institution_id'])
+            test_frac = test_frac if test_frac >= 1 / len(centers) else 1 / len(centers)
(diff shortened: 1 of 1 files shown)


ivadomed/loader/utils.py
```diff
@@ -46,6 +46,10 @@ def split_dataset(path_folder, center_test_lst, split_method, random_seed, train
     X_val = []
     if split_method == 'per_center':
         # make sure that subjects coming from some centers are unseen during training
+        if len(center_test_lst) == 0:
+            centers = list(df['institution_id'])
+            center_test_lst, _ = train_test_split(centers, train_size=test_frac, random_state=random_seed)
```
charleygros, line 51: Question: what happens if test_frac=0.2 but there is only 3 centers?

andreanne-lemay: There will we an error indicating that the set will be empty (test_frac should be changed), I could also add a condition so that the minimum test_frac has to be 1/n_centers, what do you think?

charleygros: Great!

PR #372, opened 2020-07-21
Make uncertainty compatible with 4d data

### LINKED ISSUE #369: Uncertainty quantification is not robust to 4D inputs
Uncertinty is computes on a single 3D, however the output is (H,W,D,C) now to accomodate for multichannel. This does not seem to work anymore and needs to be fixed. By looping over the channels it should be possible to make it work on multichannel/multilabel images as well.

lrouhier:
Following development, output of previous steps gives out 4D nifti images with height, width, depth and channels. The uncertainty function was designed for 3D images (height, width, depth). 
 It needs to be adapted for these new conditions. 

example: 
git checkout lr/increase_cov 
download data from https://github.com/ivadomed/data-testing
expand data in testing_data/ 
run pytest testing/test_scripts.py --trace 
to check every output (indeed visualize_transforms has an issue as well which will be fixed in another PR) 

fix #369


ivadomed/utils.py
```diff
@@ -235,14 +239,23 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+        data_im_l = []
+        for i in range(data_im.shape[3]):
+            data_im_tmp, _ = label(data_im[:, :, :, i], structure=bin_struct)
+            data_im_l.append(data_im_tmp)
         data_l_lst.append(data_im_l)
         del nib_im
+    data_l_lst = np.array(data_l_lst)
+    # channel went first due to the 'append' function
+    data_l_lst = np.reshape(data_l_lst, (data_l_lst.shape[0], data_l_lst.shape[2], data_l_lst.shape[3],
+                                         data_l_lst.shape[4], data_l_lst.shape[1]))
```
charleygros, line 251: I tend to use "transpose" for these kind of things, eg [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/testing.py#L225).
Is it doing the same thing?

ivadomed/utils.py
```diff
@@ -235,14 +239,23 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
 
     # loop across all structures of data_hard_l
     for i_l in range(1, n_l + 1):
         # select the current structure, remaining voxels are set to zero
-        data_i_l = (data_hard_l == i_l).astype(np.int)
+        data_i_l = (np.array(data_hard_l) == i_l).astype(np.int)
+        # channel went first with 'append' function for data_hard_l so we reshape this array
+        data_i_l = np.reshape(data_i_l, (data_i_l.shape[1], data_i_l.shape[2], data_i_l.shape[3], data_i_l.shape[0]))
```
charleygros, line 258: same comment

ivadomed/utils.py
```diff
@@ -218,7 +218,11 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
     # load hard segmentation and label it
     nib_hard = nib.load(fname_hard)
     data_hard = nib_hard.get_fdata()
-    bin_struct = generate_binary_structure(3, 2)  # 18-connectivity
+    data_hard_l = []
+    for i in range(data_hard.shape[3]):
+        bin_struct = generate_binary_structure(3, 2)  # 18-connectivity
+        data_hard_tmp, n_l = label(data_hard[:, :, :, i], structure=np.array(bin_struct))
```
jcohenadad, line 224: this will break with 2d images. how is ivadomed project compatible with 2d input so far? it would be good to keep this in mind for new code, anticipating microscopy data in the near future

ivadomed/utils.py
```diff
@@ -218,7 +218,11 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
     # load hard segmentation and label it
     nib_hard = nib.load(fname_hard)
     data_hard = nib_hard.get_fdata()
-    bin_struct = generate_binary_structure(3, 2)  # 18-connectivity
+    data_hard_l = []
+    for i in range(data_hard.shape[3]):
+        bin_struct = generate_binary_structure(3, 2)  # 18-connectivity
```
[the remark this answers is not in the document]
lrouhier: Yes right!

ivadomed/utils.py
lrouhier: I think so. I am just permuting two axis so yes.

ivadomed/utils.py
lrouhier: To do--> change code so that it always consider the last dimension (wether it is 3D or 4D)

ivadomed/utils.py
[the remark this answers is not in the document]
lrouhier: Done in bfa4d53

ivadomed/utils.py
```diff
@@ -218,7 +218,11 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+        data_hard_tmp, n_l = label(data_hard[:, :, :, i], structure=np.array(bin_struct))
+        data_hard_l.append(data_hard_tmp)
     data_hard_l, n_l = label(data_hard, structure=bin_struct)
```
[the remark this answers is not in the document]
lrouhier: done in bfa4d53

ivadomed/utils.py
lrouhier: done in 946dd6a

ivadomed/utils.py
lrouhier: done in 946dd6a

ivadomed/utils.py
lrouhier: done in bfa4d53. Changed something for swapping axis with condition (dim=3 or dim=4), a bit unsure about that (done in 946dd6a)

ivadomed/utils.py
```diff
@@ -236,14 +239,28 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+        data_im_l = []
+        for i in range(data_im.shape[-1]):
+            data_im_tmp, _ = label(data_im[..., i], structure=bin_struct)
+            data_im_l.append(data_im_tmp)
         data_l_lst.append(data_im_l)
         del nib_im
+    data_l_lst = np.array(data_l_lst)
+    # channel went first due to the 'append' function
+    if len(data_l_lst.shape) == 4:
+        data_l_lst = np.transpose(data_l_lst, (0,2,3,4,1))
```
jcohenadad, line 251: i'm a bit confused here. it would be useful to add a comment on the assumed dimensions for `data_l_lst` (dim1 is X, dim2 is Y, etc.)

lrouhier: Done in f4716ad , phrasing slightly changed in 51ae6d2

ivadomed/utils.py
```diff
@@ -236,14 +239,34 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
         nib_im = nib.load(fname)
         data_im = nib_im.get_fdata()
         data_lst.append(data_im)
-        data_im_l, _ = label(data_im, structure=bin_struct)
+        data_im_l = []
+        for i in range(data_im.shape[-1]):
+            data_im_tmp, _ = label(data_im[..., i], structure=bin_struct)
```
charleygros, line 244: Not sure: do we need a `np.array(bin_struct)` as in L225? which can also be put directly at L223 I think, if needed.

ivadomed/utils.py
```diff
@@ -236,14 +239,34 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+            data_im_l.append(data_im_tmp)
         data_l_lst.append(data_im_l)
         del nib_im
+    data_l_lst = np.array(data_l_lst)
+    # channel went first due to the 'append' function
+    # Befor transpose we are supposed to have MC_simulation X Channel X Height X Width X Depth
+    # After transpose we expect to have MC_simulation X Height X Width X Depth X Channel
+    # this is wanted because the network output has a shape of Height X Width X Depth X Channel
```
charleygros, line 252: The network has C x H x W x D. The rest is correct to me.

ivadomed/utils.py
```diff
@@ -236,14 +239,34 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    # So the rest of the code is designed to use something like this
+    if len(data_l_lst.shape) == 4:
+        data_l_lst = np.transpose(data_l_lst, (0, 2, 3, 4, 1))
+    elif len(data_l_lst.shape) == 3:
+        data_l_lst = np.transpose(data_l_lst, (0, 2, 3, 1))
```
charleygros, line 257: I would add a "else" as safety.

ivadomed/utils.py
```diff
@@ -236,14 +239,34 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
 
     # loop across all structures of data_hard_l
     for i_l in range(1, n_l + 1):
```
charleygros, line 260: What if the different channel have a different n_l? Here n_l is the one of the last processed channel (see L225)

ivadomed/utils.py
```diff
@@ -236,14 +239,34 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
         # select the current structure, remaining voxels are set to zero
-        data_i_l = (data_hard_l == i_l).astype(np.int)
+        data_i_l = (np.array(data_hard_l) == i_l).astype(np.int)
```
charleygros, line 262: I am not too sure now, see above comment.

ivadomed/utils.py
```diff
@@ -236,14 +239,34 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+        # channel went first with 'append' function for data_hard_l in line 225/226 (because we looped over channels)
+        # before transpose shape is supposed to be Channel X Height X Width X depth
+        # After transpose shape is expected to be  Height X Width X Depth X Channel
+        if len(data_i_l.shape) == 4:
+            data_i_l = np.transpose(data_i_l, (1, 2, 3, 0))
+        elif len(data_i_l.shape) == 3:
+            data_i_l = np.transpose(data_i_l, (1, 2, 0))
```
charleygros, line 269: Same comment about else

ivadomed/utils.py
lrouhier: I see what you mean, I will need to loop over the n_l I guess.

lrouhier: Another solution would be to check if the n_l are actually different before looping over them, don't know if it really slows the program down.

lrouhier: But then: I am a bit lost a line 287, because it won't have the same shape as data_l_lst[i_mc]

ivadomed/utils.py
```diff
@@ -236,14 +241,42 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
         data_im = nib_im.get_fdata()
         data_lst.append(data_im)
-        data_im_l, _ = label(data_im, structure=bin_struct)
+        data_im_l = []
+        for i in range(data_im.shape[-1]):
+            data_im_tmp, _ = label(data_im[..., i], structure=bin_struct)
+            data_im_l.append(data_im_tmp)
         data_l_lst.append(data_im_l)
         del nib_im
+    data_l_lst = np.array(data_l_lst)
```
jcohenadad, line 250: calling it _lst is confusing (now it becomes an array)

jcohenadad, line 250: why not using direct np array manipulation instead of going through a list?

ivadomed/utils.py
```diff
@@ -236,14 +241,42 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    # channel went first due to the 'append' function
+    # Befor transpose we are supposed to have MC_simulation X Channel X Height X Width X Depth
+    # After transpose we expect to have MC_simulation X Height X Width X Depth X Channel
+    # this is wanted because the network output has a shape of  Channel X Height X Width X Depth which is saved as
+    # nifti with shape Height X Width X Depth X Channel. We are using the second shape to save image afterward.
+
```
jcohenadad, line 256:
```suggestion
```

ivadomed/utils.py
```diff
@@ -236,14 +241,42 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
-        data_im_l, _ = label(data_im, structure=bin_struct)
+        data_im_l = []
+        for i in range(data_im.shape[-1]):
+            data_im_tmp, _ = label(data_im[..., i], structure=bin_struct)
+            data_im_l.append(data_im_tmp)
         data_l_lst.append(data_im_l)
         del nib_im
+    data_l_lst = np.array(data_l_lst)
+    # channel went first due to the 'append' function
+    # Befor transpose we are supposed to have MC_simulation X Channel X Height X Width X Depth
```
jcohenadad, line 252:
```suggestion
    # Before the transpose below, the variable data_l_lst has the shape: MC_simulation X Channel X Height X Width X Depth
```

ivadomed/utils.py
```diff
@@ -236,14 +241,42 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    # After transpose we expect to have MC_simulation X Height X Width X Depth X Channel
```
jcohenadad, line 253:
```suggestion
    # After transpose, the shape becomes: MC_simulation X Height X Width X Depth X Channel
```

ivadomed/utils.py
```diff
@@ -236,14 +241,42 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    # this is wanted because the network output has a shape of  Channel X Height X Width X Depth which is saved as
```
jcohenadad, line 254:
```suggestion
    # The rationale for doing this, is that the network output has the shape: Channel X Height X Width X Depth which is saved as
```

jcohenadad:
@lrouhier can you please give more context to the current PR body https://github.com/ivadomed/ivadomed/pull/372#issue-454040937
~~~
This PR aims at fixing uncertainty issue with 4D output that comes from multichannel and general prediction.
~~~

what issues are you referring to? can you give examples?

ivadomed/utils.py
jcohenadad: after reading this explanation, i still don't understand the context for this chunk of code: do you want data_l_lst shape to match the output shape of the network? if so, how is "MC_simulation" the same as "Channel"?
Also, why talking about the nifti output afterwards-- is it relevant here?

ivadomed/utils.py
```diff
@@ -236,14 +241,42 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+    # nifti with shape Height X Width X Depth X Channel. We are using the second shape to save image afterward.
+
+    if len(data_l_lst.shape) == 4:
+        data_l_lst = np.transpose(data_l_lst, (0, 2, 3, 4, 1))
+    elif len(data_l_lst.shape) == 3:
+        data_l_lst = np.transpose(data_l_lst, (0, 2, 3, 1))
+    else:
+        print("object does not appear to be 2D or 3D, please check the network output")
+        exit(1)
```
jcohenadad, line 263: use logging in API code-- is that already a written convention @Drulex ?
also, i think we should avoid these hard exit in APIs (instead, raise, and the parent function takes care of the rest)

Drulex: I don't see it in this project's wiki, but you can refer to sct's guidelines: https://github.com/neuropoly/spinalcordtoolbox/wiki more specifically the Logging and Exceptions sections.

jcohenadad: can someone pls update the wiki (maybe redirect to sct wiki for now until someone can dedicate the time to make proper wiki-- thx)

Drulex: Perhaps we should have an section on API code guidelines. Some generic things come to mind:
- No printing, use logging instead
- Don't use `exit`, raise and let parent handle
- Avoid system calls
- Deal with low-level objects instead of files
- Include unit tests
- Make interfaces generic and re-usable

jcohenadad: awesome! let's start with this-- thx!

ivadomed/utils.py
lrouhier: It is confusing indeed, the fact is I want them both to have a similar shape (like HxWxDxC) and we will loop over the MC simulation (so only the last 4 dimensions matters). We use HxWxDxC because that is the format of the input images (nifti_masks), we therefore have an output the same shape as the input (data ready to be save in nifti). 

I mentionned the network because of a mistake on the workflow of the function. I'll remove it.

ivadomed/utils.py
lrouhier: It was previsouly a list so I moved with that before realising I needed to transpose it, which is more convenient with an array. I'll change that.

ivadomed/utils.py
lrouhier: I am going to remove the bit about the network beacause it is confusing

ivadomed/utils.py
lrouhier: is what is written in 5efc0fe clearer ?

ivadomed/utils.py
```diff
@@ -260,8 +261,7 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
     elif len(data_l_lst.shape) == 3:
         data_l_lst = np.transpose(data_l_lst, (0, 2, 3, 1))
     else:
-        print("object does not appear to be 2D or 3D, please check the network output")
-        exit(1)
+        raise Exception("object does not appear to be 2D or 3D, please check input")
```
jcohenadad, line 264: there is probably more specific exception to use, like ValueError? @Drulex ?
also: the else is confusing as it checks for dim=4 or 3, and the message tells to input dim=2 or 3
maybe you can do the check before?  for better code clarity

ivadomed/utils.py
```diff
@@ -276,8 +276,7 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
         elif len(data_i_l.shape) == 3:
             data_i_l = np.transpose(data_i_l, (1, 2, 0))
         else:
-            print("object does not appear to be 2D or 3D, please check the network output")
```
jcohenadad: why do you need to check this again-- doesn't it originate from the same variable as above?

ivadomed/utils.py
lrouhier: Done in ed6cf059e810dbd62b1742659c084cc131105587

ivadomed/utils.py
lrouhier: done in ed6cf059e810dbd62b1742659c084cc131105587

ivadomed/utils.py
lrouhier: It was a list to match data_lst (that stores the image and not the masks) [L240](https://github.com/ivadomed/ivadomed/blob/65037dffb4361a0686e02ec753caba4d095e39e3/ivadomed/utils.py#L240)  , but I guess we can use np.array for the whole process

ivadomed/utils.py
Drulex: Yes `ValueError` is appropriate. @lrouhier see usage of built-in exceptions here https://docs.python.org/3/library/exceptions.html#ValueError

Also make the message more clear, for example: `raise ValueError("Input data of shape {} is not valid. Shape must be 3 or 4".format(data_l_lst.shape))`

ivadomed/utils.py
lrouhier: data_l_lst stems from "data" that comes from images in "fname_lst".  (see [here](https://github.com/ivadomed/ivadomed/blob/65037dffb4361a0686e02ec753caba4d095e39e3/ivadomed/utils.py#L241) , [here](https://github.com/ivadomed/ivadomed/blob/65037dffb4361a0686e02ec753caba4d095e39e3/ivadomed/utils.py#L249))
data_i_l from "data_hard" that comes from images in "fame_hard".  (see [here](https://github.com/ivadomed/ivadomed/blob/65037dffb4361a0686e02ec753caba4d095e39e3/ivadomed/utils.py#L270), and [here](https://github.com/ivadomed/ivadomed/blob/65037dffb4361a0686e02ec753caba4d095e39e3/ivadomed/utils.py#L221)).
It is linked but it is not the same files so I think it is better to check both.

ivadomed/utils.py
lrouhier: started in 65037df

ivadomed/utils.py
lrouhier: Done in ed6cf05

ivadomed/utils.py
lrouhier: done in 65037df. Thanks for the tips @Drulex ! (the link is not obvious in https://docs.python.org/3/tutorial/errors.html)

lrouhier:
> @lrouhier can you please give more context to the current PR body [#372 (comment)](https://github.com/ivadomed/ivadomed/pull/372#issue-454040937)
> 
> ```
> This PR aims at fixing uncertainty issue with 4D output that comes from multichannel and general prediction.
> ```
> 
> what issues are you referring to? can you give examples?

The issue occured when I use the tests. I modified the PR body. I will add the log of the error in the issue.

ivadomed/utils.py
jcohenadad: nope-- comment added in the commit

ivadomed/utils.py
jcohenadad: > awesome! let's start with this-- thx!

i meant the wiki-- can someone do it?

ivadomed/utils.py
jcohenadad: right-- but i would still check them at the beginning of the code-- also see my other comment about the check dim discrepancy

ivadomed/utils.py
Drulex: I can take care of it

lrouhier:
> so, you create data_l_lst of shape MC_simulation X Height X Width X Depth X Channel
> then you say that you need nifti masks of shape: Height X Width X Depth X Channel
> then you say "We want an output of the same shape."
> 
> but these are not the same shape...?

[here](https://github.com/ivadomed/ivadomed/blob/bd6261fffaf41ce241bca4588daa73b10f0e5934/ivadomed/utils.py#L287) 
we are combining what we created (data_i_l and data_l_lst) but we are doing so for each simultion (so we loop over the first dimension). 
We need to have the same shape as the input because [here] (https://github.com/ivadomed/ivadomed/blob/bd6261fffaf41ce241bca4588daa73b10f0e5934/ivadomed/utils.py#L292) we are using the image from [here](https://github.com/ivadomed/ivadomed/blob/bd6261fffaf41ce241bca4588daa73b10f0e5934/ivadomed/utils.py#L244) that are just loaded mask that are stacked in a list. Each mask has the shape HxWxDxC. We are still comparing simulation by simulation.

ivadomed/utils.py
lrouhier: > the else is confusing as it checks for dim=4 or 3, and the message tells to input dim=2 or 3
>maybe you can do the check before? for better code clarity

Basically I am doing this because the 4th dimension is the channel, but your input is not a commonly called "4D image", it is a 3D image with a channel dimension. The dimension will always be there but it's not what is usually considered a 4D image (e.g., it's not DTI or something like that). I thought that it would be less confusing for end user (that may not be used to considering channels as a dimension). I can change that.

ivadomed/utils.py
lrouhier: I can check them for the erorr right but would still need an if statement here (to make the right transpose)

ivadomed/utils.py
Drulex: WIP -> https://github.com/ivadomed/ivadomed/wiki/Coding-Best-Practices

jcohenadad: fantastic! thank you so much @Drulex -- we could use that template for other projects as well

jcohenadad: or maybe we should centralize (instead of duplicating across projects) such guidelines... 🤔  thoughts?

ivadomed/utils.py
lrouhier: Manipulating array is a bit more complicated, because the initialisation needs to have the right shape. Since we have not loaded the image when we initialize the data_l_lst objects, the shape is unknown (we can do it, it just increase the number of IO in the code). 
see https://numpy.org/doc/stable/reference/generated/numpy.empty.html. I think going through a list is more effcient. I can change the name though.

jcohenadad: @Drulex do you have some thoughts about best practice here?

ivadomed/utils.py
lrouhier: There are could be some instruction specific to ivadomed (where to add code, how to name import (e.g., import ivadomed.utils as imed_utils), So I think the template approach makes more sens.

ivadomed/utils.py
lrouhier: Changed for right numbers in ad094bc

ivadomed/utils.py
Drulex: Not exactly sure about this specific case since I am not familiar with the topic at hand, but it looks like the lists are used exclusively as containers for the actual arrays and labels which seems ok.

ivadomed/utils.py
```diff
@@ -231,33 +235,60 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
     data_iou, data_cv, data_avgUnc = np.zeros(data_hard.shape), np.zeros(data_hard.shape), np.zeros(data_hard.shape)
 
     # load all MC simulations and label them
-    data_lst, data_l_lst = [], []
+    data_lst = []
+    data_l_arr = np.array()
```
Drulex, line 239: I'm not sure you can do this. `Array()` has mandatory args (see https://numpy.org/doc/stable/reference/generated/numpy.array.html?highlight=array#numpy.array). Has this code run?

lrouhier: It doesn't run, I just pushed it to save it during my reflexion, because I was unsure of the way to go (hence the previous discussion)

ivadomed/utils.py
```diff
@@ -231,33 +235,60 @@ def structurewise_uncertainty(fname_lst, fname_hard, fname_unc_vox, fname_out):
...
+        elif len(data_i_l.shape) == 3:
+            data_i_l = np.transpose(data_i_l, (1, 2, 0))
+        else:
+            raise ValueError("Input data of shape {} is not valid. image must be 3D or 4D".format(len(data_l_arr.shape)))
 
         # select the current structure in each MC sample
         # and store it in data_mc_i_l_lst
         data_mc_i_l_lst = []
         # loop across MC samples
         for i_mc in range(len(data_lst)):
```
Drulex, line 283: The pythonic way to iterate over paired values from two containers is to use `zip()` (see https://docs.python.org/3/library/functions.html#zip)

Code would look cleaner, something like:
```
for img, label in zip(data_lst, data_l_arr):
    # in a nutshell, img = data_lst[i] and label = data_l_arr[i]
    data_i_inter = data_i_l * label
    i_mc_l = np.max(data_i_inter)

    if i_mc_l > 0:
        data_mc_i_l = np.copy(img)
        data_mc_i_l[label != i_mc_l] = 0.
    else:  # no structure in this sample
        data_mc_i_l = np.zeros(img.shape)
    data_mc_i_l_lst.append(data_mc_i_l)

It's purely a python thing, but makes the code more readable than having matrices with indexes all over the place.
```

lrouhier: I get what you mean. It might be a bit out of scope for this PR.

PR #383, opened 2020-07-24
Find optimal threshold with ROC analysis

### LINKED ISSUE #368: Optimal bin threshold based on ROC analysis
The choice of a bin threshold is always so critical... and can be so subjective when based on expert judgement or horoscope predictions.

A more robust way would be to run a ROC analysis using the trained model and the validation sub dataset.
We could use the param `binarise_predictions`, see [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config.json#L75), to trigger this analysis.
Note we do not want to make it mandatory as it is quite time consuming and unnecessary in some applications.
So an approach would be, when calling the testing command:
- Run the ROC analysis on the validation dataset. We would need to specify the increment used for the thresholds.
- Get the optimal threshold from this analysis and use it to binarise the predictions of the testing dataset.

Note: It could be a good opportunity to bring into production the threshold-based post-processing using uncertainty, which is based on the same analysis.

charleygros:
I hate thresholds... but I have to live with them.. oh well..
At least let's try to find one that is not completely subjective.. with a ROC analysis. That's what is done in this pull request.

### At training time:
- new flag in the CLI: `-r`  `--roc_increment` --> controls the increment of thresholds between 0 and 1 that are considered during the ROC analysis. For instance: `--roc_increment 0.1` --> 0, 0.1, 0.2, 0.3 ... 0.9, 1.0
- If this flag is used in the command, then a ROC analysis is performed after the training using the validation sub dataset: for each batch, we run the model and aggregate the sensititivy and specificity computed on the predictions that have been thresholds with different values (0, 0.1, 0.2 etc). For each threshold, we compute the mean sensititivy (TPR) and 1-Specificity (FPR) to find the optimal threshold by maximising the TPR-FPR difference and plot this curve (saved in `log_directory/roc.png`).
- We update `log_directory/config_file.json` with the optimal threshold (`binarize_predictions`) which could be further used at testing time.

### At testing time:
- Validation metrics are computed on data binarized with `binarize_predictions`. In case no ROC has been performed and `binarize_predictions=-1`, the metrics are computed on soft predictions (ie no binarization).
- Output niftis are binarized using `binarize_predictions` (which is the optimal threshold if the ROC analysis has ben run) if `binarize_nifitis=true`. This allows us to potentially compute metrics on hard predictions while keeping soft nifitis.

### When automate_training:
- At testing time, we use `log_directory/config_file.json` in case `binarize_predictions` has been updated after training.

## TODO:
- Update config file if the change `binarize_predictions` & `binarize_niftis` is approved.

Fixes #368.


charleygros:
```
ivadomed -c config_roc.json -r 0.05
```
Gives:
roc

THe green dot is related to the optimal threshold.

jcohenadad:
awesome! do i interpret correctly that the optimal thr from the above figure is 0.95?
what is the -r ?

docs/source/configuration_file.rst
```diff
@@ -302,9 +302,12 @@ UNet3D (Optional)
 Testing parameters
 ------------------
 
--  ``binarize_prediction``: Bool. Binarize output predictions using a
-   threshold of 0.5. If ``false``, output predictions are float between
-   0 and 1.
+- ``binarize_predictions``: Float. Threshold (between 0 and 1) used to binarize
+    the predictions before computing the validation metrics. To use soft predictions
+    (i.e. no binarisation, float between 0 and 1) for metric computation, indicate -1.
+-  ``binarize_niftis``: Bool. Binarize output predictions using ``binarize_predictions``
```
andreanne-lemay, line 308: Why do we need this new parameter `binarize_niftis`? If binarize_predicitons is set to `-1` the soft output is used (equivalent to `binarize_niftis = False` I believe). If the a roc_increment is provided than there will automatically be value and it implies we want to binarize. I feel like `binarize_niftis` is redundant with `binarize_niftis`. Let me know your thoughts, I might have forgotten a use case :)

charleygros:
> awesome! do i interpret correctly that the optimal thr from the above figure is 0.95?

You can't read the correct threshold from the figure. Each dot is a threshold (eg 0.1, 0.6) that has been tested: for each tested tresholds, we binarize the predictions and measure the mean TPR & FDR --> new dot on the plot. The dot which minimizes TPR-FDR turns green (ie optimal threshold).

The optimal threshold is indicated in the terminal output:
```
Running ROC analysis to find optimal threshold
	Optimal threshold: 0.35
	Saving plot: spineGeneric/roc.png
```
and used to update `log_directory/config_file.json` with the optimal threshold (`binarize_predictions` param) which could be further used at testing time.

charleygros:
> what is the -r ?

```
usage: ivadomed -c CONFIG [-g GIF] [-r ROC_INCREMENT] [-h]

MANDATORY ARGUMENTS:
  -c CONFIG, --config CONFIG
                        Path to configuration file.

OPTIONAL ARGUMENTS:
  -g GIF, --gif GIF     Generates a GIF of during training, one frame per
                        epoch for a given slice. The parameter indicates the
                        number of 2D slices used to generate GIFs, one GIF per
                        slice. A GIF shows predictions of a given slice from
                        the validation sub-dataset. They are saved within the
                        log directory.
  -r ROC_INCREMENT, --roc_increment ROC_INCREMENT
                        A ROC analysis is performed at the end of the training
                        using the trained model and the validation sub-dataset
                        to find the optimal binarization threshold. The
                        specified value indicates the increment between 0 and
                        1 used during the ROC analysis (e.g. 0.1). ROC plot is
                        saved under "log_directory/roc.png" and the optimal
                        threshold in "log_directory/config_file.json as
                        "binarize_predictions" parameter.
  -h, --help            Shows function documentation.
```

docs/source/configuration_file.rst
charleygros: Explained [above](https://github.com/ivadomed/ivadomed/pull/383#issue-456074565):

> Output` niftis are binarized using binarize_predictions (which is the optimal threshold if the ROC analysis has ben run) if binarize_nifitis=true. This allows us to potentially compute metrics on hard predictions while keeping soft nifitis.

Which is exactly what we want to do in our current experiments: compute Dice on hard but keep nifitis soft for visualisation.

andreanne-lemay: This will create discrepancy between the output labels and metrics (metrics won't be representative of the outputs in pred_masks), personally I would keep a single parameter since visualization is more for debugging (and classic argument our config file is pretty populated). But either way I'm happy :)

jcohenadad:
> > awesome! do i interpret correctly that the optimal thr from the above figure is 0.95?
> 
> You can't read the correct threshold from the figure. Each dot is a threshold (eg 0.1, 0.6) that has been tested: for each tested tresholds, we binarize the predictions and measure the mean TPR & FDR --> new dot on the plot. The dot which minimizes TPR-FDR turns green (ie optimal threshold).
> 
> The optimal threshold is indicated in the terminal output:
> 
> ```
> Running ROC analysis to find optimal threshold
> 	Optimal threshold: 0.35
> 	Saving plot: spineGeneric/roc.png
> ```
> 
> and used to update `log_directory/config_file.json` with the optimal threshold (`binarize_predictions` param) which could be further used at testing time.

ok I see. So this is somewhat different than our [other approach for finding a threshold](https://github.com/sct-pipeline/deepseg-threshold). I guess there is no single way to find "the best threshold" (whatever that means), but we should keep this in mind, in case at some point we want to change the method. In the present case, the "best threshold" is biased towards the ground truth (which has been generated mostly using older models, which were been generated using propseg, which is biased towards contrast). Just something to keep in mind.

docs/source/configuration_file.rst
charleygros: I hear your point :-) very valid.

But with allowing a soft output (while computing the metrics on hard masks, ie the right way) we can still binarise it further if we want (with a python script or fsleyes etc) whereas the other way does not allow this (it will be bin, I can't reconstruct the soft). These niftis, in ivadomed routine, are mainly for debugging / visualisation, not for "production" such as in SCT.

Not sure what's best here.

charleygros:
> ok I see. So this is somewhat different than our [other approach for finding a threshold](https://github.com/sct-pipeline/deepseg-threshold). I guess there is no single way to find "the best threshold" (whatever that means), but we should keep this in mind, in case at some point we want to change the method.

Yes absolutely! I made quite clear (I think) in the documentation how this "golden" threshold is computed so that the used method is clearly identified.
This other method would be super great for SC segmentation with multi contrast. But not appropriated for GM, lesions or tumor tasks. That's why I first implemented this other method (ROC).

> In the present case, the "best threshold" is biased towards the ground truth (which has been generated mostly using older models, which were been generated using propseg, which is biased towards contrast). Just something to keep in mind.

:+1:

docs/source/configuration_file.rst
jcohenadad: can we pls discuss it in a meeting

ivadomed/testing.py
```diff
@@ -221,8 +227,12 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
                                                         fname_out=fname_pred,
                                                         slice_axis=slice_axis,
                                                         kernel_dim='3d',
-                                                        bin_thr=0.5 if testing_params["binarize_prediction"] else -1)
-                    preds_npy_list.append(output_nii.get_fdata().transpose(3, 0, 1, 2))
+                                                        bin_thr=bin_thr)
+                    output_data = output_nii.get_fdata().transpose(3, 0, 1, 2)
+                    if testing_params["binarize_predictions"] >= 0:
+                        output_data = threshold_predictions(output_data, thr=testing_params["binarize_predictions"])
```
andreanne-lemay, line 233: The thresholding with `binarize_predicitons` should also be in the `evaluation.py`. The command `eval` uses directly the output in the pred_masks folder to compute the metrics, but with `binarize_niftis` we don't necessarily want to have the metrics from the generated masks. This would be another argument in favor of having only one param (only `binarize_predictions` in the config file :}

andreanne-lemay: updated evaluation.py here 7aa3e21

ivadomed/main.py
```diff
@@ -35,13 +35,20 @@ def get_parser():
                                     ' The parameter indicates the number of 2D slices used to generate GIFs, one GIF '
                                     'per slice. A GIF shows predictions of a given slice from the validation '
                                     'sub-dataset. They are saved within the log directory.')
+    optional_args.add_argument('-t', '--thr_increment', required=False, type=float,
```
jcohenadad, line 38: i would suggest to use `--thr-increment`

ivadomed/testing.py
charleygros: As we decided to have only one param... I can undo this, right? Since the predictions will always be generated according to testing_params. If binarise_prediction = -1 --> metrics_testing, niftis and metrics_eval are soft, otherwise they are hard, all with the same threshold. Are we on the same page? :-)

ivadomed/main.py
charleygros: Done in [b44b8be](https://github.com/ivadomed/ivadomed/pull/383/commits/b44b8bef5c948ab3f88f3d59b6af2a84ce1d1b3f)

ivadomed/testing.py
andreanne-lemay: Yes I totally agree :D!

PR #384, opened 2020-07-24
Normalized ReLU

andreanne-lemay:
Add option to use a Normalized ReLU as final activation. To use it, the parameter "relu" can be added this way in the config file:
For 2D:
``` 
"default_model": {
        "name": "Unet",
        "dropout_rate": 0.3,
        "bn_momentum": 0.9,
        "depth": 4
        "relu": true
    }
```

For 3D:
```
    "UNet3D": {
        "applied": true,
        "length_3D": [128, 128, 128],
        "stride_3D": [64, 64, 64],
        "relu": true
    }
```

This parameter is optional, removing it will automatically apply the sigmoid activation.

### DIFF
--- a/docs/source/configuration_file.rst
+++ b/docs/source/configuration_file.rst
@@ -262,7 +262,7 @@ the default parameters are merged with the parameters that are specific
 to the tailored model. - ``name``: ``Unet`` (default) -
 ``dropout_rate``: Float (e.g. 0.4). - ``batch_norm_momentum``: Float
 (e.g. 0.1). - ``depth``: Strictly positive integer. Number of
-down-sampling operations.
+down-sampling operations. - ``relu`` (optional): Bool, sets final activation to normalized ReLU (relu between 0 and 1).
 
 FiLMedUnet (Optional)
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


docs/source/configuration_file.rst
```diff
@@ -297,7 +297,9 @@ UNet3D (Optional)
    translation of 1 voxel in the 1st dimension, 2 voxels in the 2nd
    dimension and 3 voxels in the 3rd dimension at every iteration until
    the whole input matrix is covered.
--  ``attention_unet``: Bool. Use attention gates in the Unet's decoder.
+-  ``attention_unet`` (optional): Bool. Use attention gates in the Unet's decoder.
+- ``n_filters`` (optional): Int. Number of filters in the first convolution of the UNet. This number of filters will be doubled at each convolution.
+- ``relu`` (optional): Bool. Sets final activation to normalized ReLU (relu between 0 and 1).
```
charleygros, line 302: I would have thought `relu` is only needed in the default model, see update [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/main.py#L114).

andreanne-lemay: Bien vu! updated :)

PR #379, opened 2020-07-22
Create Ivadomed download function

lrouhier:
Creation of download_data. This function will be used to download binary files from multiple URL. It is based on spinalcordtoolbox's download function. 

Done: 
* Script creation
* Addition of entrypoint

fix #346

PR #385, opened 2020-07-27
Create coverage and improve testing

lrouhier:
This PR need to be merged after #390  , #379, and (data-testing PR [#2](https://github.com/ivadomed/data-testing/pull/2))

It aims at adding badge for coverage as well as adapting tests for new data and new functionnalities in the ivadomed packages. 

changes: 
* Adapt each files that used to call sub-test001 to now call sub-unf01, which is the new test subject
* Added test for function in ivadomed/object_detectionà
* Create test for metric when there is an error
* Create test for untested model (Countception)
* Create test for new loss functions (AdapWingLoss, L2loss, FocalDiceLoss and LossCombination)
* Separate test for scripts.


testing/test_HeMIS.py
```diff
@@ -23,7 +23,7 @@
 BN = 0.1
 N_EPOCHS = 10
 INIT_LR = 0.01
-PATH_BIDS = 'testing_data'
+PATH_BIDS = 'testing_data/'
```
jcohenadad, line 26: now windows compatible-- shouldn't we use os.pathsep instead?

lrouhier: You mean windows subsystem for linux? 
If not, I can track other similar issue.

jcohenadad: if in one day, ivadomed works with windows, this will break-- in general, it is good practice to use OS-agnostic separators in code

jcohenadad: i should also add, that you should not need to add a separator, if down the line paths are constructed in a pythonic way (e.g. using pathlib or os.path.join(), which should always be the case)

lrouhier: Oh sorry I think I understood the first command wrong. The separator will be removed during cleaning (because it is not needed, I just changed the data path and I didn't pay cloe enough attention when I changed it back).

lrouhier:
Previosu testing data (located in this repo) will need to be removed in order to reduce repo size and because we won't use it anymore.

lrouhier:
For the review: Please discard changes done on files from testing_data. this was done to allow some tests to pass in the begining to establish coverage. The new testing data is located in it own [repo](https://github.com/ivadomed/data-testing).

testing/test_sampler.py
```diff
@@ -76,13 +77,14 @@ def test_sampler(transforms_dict, train_lst, target_lst, roi_params):
                               collate_fn=imed_loader_utils.imed_collate,
                               num_workers=0)
     neg_percent, pos_percent = _cmpt_label(train_loader)
-    assert neg_percent > pos_percent
+    assert abs(neg_percent - pos_percent) > 20
```
lrouhier, line 80: the images changes hence the change in the balance

testing/test_sliceFilter.py
```diff
@@ -76,4 +76,4 @@ def test_slice_filter(transforms_dict, train_lst, target_lst, roi_params, slice_
         assert cmpt_neg == 0
         assert cmpt_pos != 0
     else:
-        assert cmpt_neg > cmpt_pos
+        assert cmpt_neg != 0 and cmpt_pos != 0
```
lrouhier, line 79: Images changes so the balance as well.

testing_data/bounding_box_dict.json
```diff
@@ -0,0 +1,21 @@
...
+  "testing_data/sub-test001/anat/sub-test001_T2w.nii.gz":
+    [
+      7,
+      386,
+      113,
+      231,
+      1,
+      16
+    ]
+}
```
lrouhier, line 21: Not relevant as it is in ivadomed/data-testing

testing_data/hyperparameter_opt.json
```diff
@@ -0,0 +1,5 @@
+{
```
lrouhier, line 1: Not relevant, new used file is in ivadomed/data-testing

testing_data/model_config.json
```diff
@@ -1,39 +1,36 @@
 {
```
lrouhier, line 1: Not relevant, new used file is in ivadomed/data-testing

testing_data/model_config_test.json
```diff
@@ -0,0 +1,122 @@
+{
```
lrouhier, line 1: Not relevant, new used file is in ivadomed/data-testing

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+    json.dump(initial_config, file_conf)
+
+
+def test_automate_training():
+    command = "ivadomed_automate_training -c testing_data/model_config_auto.json " \
+              "-p testing_data/hyperparameter_opt.json -n 1 --fixed-split"
+    subprocess.check_output(command, shell=True)
+
+    command = "ivadomed_automate_training -c testing_data/model_config_auto.json " \
+              "-p testing_data/hyperparameter_opt.json -n 1 --run-test --all-combin"
```
andreanne-lemay, line 174: We could add the flag `-t 0.1` for extra coverage :)

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+    subprocess.check_output(command, shell=True)
+
+
+def test_create_json_3d_unet_test():
```
andreanne-lemay, line 178: Idem

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+def test_create_model_unet3d():
+    model = imed_models.UNet3D(in_channel=1, out_channel=1, n_filters=8)
+    torch.save(model, "model_unet_3d_test.pt")
+    os.makedirs("3d_test")
+    command = "cp model_unet_3d_test.pt 3d_test/best_model.pt"
+    subprocess.check_output(command, shell=True)
+
+
+def test_testing_unet3d():
+    subprocess.check_output(["ivadomed -c testing_data/model_config_3d.json"], shell=True)
```
andreanne-lemay, line 241: Linked to my previous comment but feels redundant with `test_training_time.py`

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+            },
+        "CenterCrop": {"size": [48, 48, 16], "preprocessing": True},
+        "NumpyToTensor": {},
+        "NormalizeInstance": {"applied_to": ["im"]}
+    }
+    json.dump(initial_config, file_conf)
+
+
+def test_create_model_unet3d():
+    model = imed_models.UNet3D(in_channel=1, out_channel=1, n_filters=8)
```
andreanne-lemay, line 233: Not sure what this is for?

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+    os.makedirs("3d_test")
+    command = "cp model_unet_3d_test.pt 3d_test/best_model.pt"
+    subprocess.check_output(command, shell=True)
+
+
+def test_testing_unet3d():
+    subprocess.check_output(["ivadomed -c testing_data/model_config_3d.json"], shell=True)
+
+
+def append_list_as_row(file_name, list_of_elem):
```
andreanne-lemay, line 244: Idem

testing_data/bounding_box_dict.json
andreanne-lemay: Should we delete it then?

testing_data/hyperparameter_opt.json
andreanne-lemay: Delete :)?

testing_data/model_config.json
andreanne-lemay: I know I am repeating myself but why not completely deleting testing_data?

testing/test_script.py
lrouhier: This is to add row to the csv file (particpant.csv)

testing/test_script.py
lrouhier: Basically to increase coverage by testing the creation of 3D unet (and basic forward)

testing/test_script.py
lrouhier: Right !

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+    json.dump(initial_config, file_conf)
+
+
+def test_eval():
+    subprocess.check_output(["ivadomed -c testing_data/model_config_eval.json"], shell=True)
+
+
+def test_create_automate_training_json():
+    # modify train config
+    command = "cp testing_data/model_config_eval.json testing_data/model_config_auto.json"
```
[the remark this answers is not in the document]
lrouhier: Yeah but also, this came after the first release of data-testing. Basically, the idea is to store a low number of file and modify them each time. otherwise each time you need a new test, or modify the config file, you will need to modify like 10 files and do a release in data-testing (+ change link in ivadomed_download_data)... So I am trying to figure out a way of doing our maintenance at one single place :)

testing/test_losses.py
```diff
@@ -207,3 +208,133 @@ def test_focaltverskyloss(params):
...
+    test focaldiceloss
+
+    Args:
+        params (tuple): containing input tensor, target tensor, expected value, loss function
+
+    """
+
+    input, target, expected_value, loss_fct = params
+    loss = loss_fct.forward(input, target)
+    assert isclose(loss.detach().cpu().numpy(), expected_value, rel_tol=1e-2)
```
charleygros, line 340: Awesome here too!

testing/test_metrics.py
```diff
@@ -0,0 +1,67 @@
+import ivadomed.metrics as imed_metrics
+import numpy as np
+
+
+def test_multi_class_dice_score():
```
charleygros, line 5: That's great!
At some point we may want to move to "pytest.mark.parametrize", similar to what is done in `test_losses.py`... opening an issue.

testing/test_sampler.py
```diff
@@ -39,14 +40,14 @@ def _cmpt_label(ds_loader):
...
+            "size": [128, 128],
             "preprocessing": True
         },
     "NumpyToTensor": {}
 }])
-@pytest.mark.parametrize('train_lst', [['sub-test001']])
+@pytest.mark.parametrize('train_lst', [['sub-unf01']])
 @pytest.mark.parametrize('target_lst', [["_lesion-manual"]])
-@pytest.mark.parametrize('roi_params', [{"suffix": "_seg-manual", "slice_filter_roi": 10}])
+@pytest.mark.parametrize('roi_params', [{"suffix": "_seg-manual", "slice_filter_roi": 16}])
```
charleygros, line 50: Why did you have to make that change?

testing/test_sampler.py
charleygros: Not sure to understand why the previous test is not relevant anymore. COuld you please help me?

testing/test_sampler.py
```diff
@@ -76,13 +77,17 @@ def test_sampler(transforms_dict, train_lst, target_lst, roi_params):
...
                                        sampler=imed_loader_utils.BalancedSampler(ds_train),
                                        shuffle=False, pin_memory=True,
                                        collate_fn=imed_loader_utils.imed_collate,
                                        num_workers=0)
-    neg_percent, pos_percent = _cmpt_label(train_loader_balanced)
-    assert abs(neg_percent - pos_percent) <= 20.
+
+    neg_percent_bal, pos_percent_bal = _cmpt_label(train_loader_balanced)
+    # We check if the loader is more balanced. The actual distribution comes from a probabilistic model
+    # This is however not very efficient to get clos to 50 %
```
charleygros, line 91: Why is it not efficient? What is not efficient?

testing/test_sampler.py
```diff
@@ -76,13 +77,17 @@ def test_sampler(transforms_dict, train_lst, target_lst, roi_params):
...
+    # in the case where we have 16 slices, with 87,5 % of one class (positive sample).
+    assert abs(neg_percent_bal - pos_percent_bal) < abs(neg_percent - pos_percent)
```
charleygros, line 93: 16 slices in total.. 14 positive, 2 negative --> we are expecting to have the same amount of neg slices and pos slices in the resulting dataset. Not sure to understand this assert. Not sure if that's actually testing the right thing.

testing/test_sliceFilter.py
charleygros: Do you mean: the new testing image as more pos than neg? ok
I feel this "assert" is a bit useless.. what about: ` cmpt_neg < cmpt_pos` --> ie no filtering has been done.

testing/test_transfer_learning.py
```diff
@@ -10,9 +10,10 @@
 OUT_CHANNEL = 1
 INITIAL_LR = 0.001
 
-@pytest.mark.parametrize('fraction', [0.1, 0.3, 0.5])
+
+@pytest.mark.parametrize('fraction', [0.1, 0.2, 0.3])
```
charleygros, line 14: That's because we are using a very tiny network.. is it correct?

testing/test_script.py
andreanne-lemay: But it wouldn't be necessary if all the data was in ivadomed/data-testing

testing/test_sliceFilter.py
lrouhier: Since the test is designed to filter negative slice, I am just checking if there a still some negative examples.

testing_data/bounding_box_dict.json
lrouhier: This one yes because I created it.

testing_data/model_config.json
lrouhier: It will be done after this PR. See #391

testing/test_script.py
lrouhier: Yes but then you will have a15 mb data-testing that juste stores identical images.

testing/test_sampler.py
lrouhier: The 'sampler' is not really working. It comes from the fact that this is a probabilistic choice and each slice has a heavy weight. Maybe the phrasing is wrong.

testing/test_script.py
andreanne-lemay: I don't see why this is a problem, this repo exists for this exact purpose, no?

jcohenadad: i would still encourage to keep this repos at a minimum size (it is downloaded very often by CI and local developers)

andreanne-lemay: We are talking about 4 MB extra, but if we wanted to keep this dataset to a minimal size we could use smaller images or dummy images (`sub-unf01_T2star.nii.gz` and `sub-unf01_T2star.nii.gz` weight about 1 MB each which represents pretty much all the dataset's weight). I just feel like doing a lot of extra code to create more data defies the purpose to have a testing dataset that should include all the data we need to test, plus it is not very elegant.  Cropping the images around the spinal cord would limit the size of the dataset and would allow us to have more test subjects.

lrouhier: Data were already cropped (eventhough it is not just around the spinal cord)

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+
+
+def test_testing_with_uncertainty():
+    # Test config. Uses Uncertainty
+    subprocess.check_output(["ivadomed -c testing_data/model_config_test.json"], shell=True)
+
+
+def test_training():
+    # Train config 
+    subprocess.check_output(["ivadomed -c testing_data/model_config.json"], shell=True)
```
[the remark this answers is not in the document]
lrouhier: Yes you are right, it is just that  `test_training_time.py` does not actually call the `training.py` file, so it is not considered as covered.

testing/test_script.py
lrouhier: Perhaps we should test every possible model that way (more extensive coverage of `models.py`, and closer to unit testing).

testing/test_script.py
jcohenadad: ok i need a bit of context here-- based on the discussion thread, i understand that the discussion is about duplicating some of the testing data in order to test what happens if ivadomed receives 2 subjects instead of 1, is that correct?

if this is indeed correct, then this would justify doing the copy on the fly by the testing code (ie not duplicating the data)

lrouhier: This is the context: 
for training a model with ivadomed, right now, you need to have a training set, a validation set, and a test set. We load images per subject. So If you want three sets, you will need to have three patients. 
For now, during the testing phase, I duplicate what I have for one subject (derivatives, images, metadata) to create 2 "new" ones. This allows me to run the full ivadomed workflow (integration tests) with 1 subject as the training set, 1 as the validation and 1 as the testing set.

jcohenadad: ok got it-- i'm not sure what's the best strategy here. I am leaning towards adding 2 other subjects in the testing-data repository, but with extreme cropping (to be greedy on the repos size)

testing/test_sampler.py
lrouhier: change was sue to misunderstanding removed in f6fed96

testing/test_transfer_learning.py
lrouhier: Yes that is the reason.

testing/test_script.py
lrouhier: Done in 
e28665b

testing/test_script.py
lrouhier: I think the main issue is that right now it is kind of an hybrid approach. We can either add subject or go with your suggestion from #392 and create everything (since we don't need to much anatomical meaning in the image considering the training is just 1 epochs we don't test performance)

testing_data/hyperparameter_opt.json
lrouhier: Done

testing_data/model_config_test.json
lrouhier: removed in b47b7d8

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+
+# def test_training_curve():
+# using the results from previous training
+#   command = "ivadomed_training_curve -i testing_script/ -o training"
+#  subprocess.check_output(command, shell=True)
+
+
+def test_create_eval_json():
+    # modify train config
+    command = "cp testing_data/model_config.json testing_data/model_config_eval.json"
```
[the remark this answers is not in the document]
lrouhier: Discussion here https://github.com/ivadomed/ivadomed/pull/385#discussion_r468890808

testing/test_script.py
```diff
@@ -1,18 +1,250 @@
...
+                            "T1w -d 1", shell=True)
+
+
+def test_compare_model():
+    # testing compare_model
+    command = "ivadomed_compare_models -df testing_data/temporary_results.csv -n 2"
+    subprocess.check_output(command, shell=True)
+
+
+def test_creation_dataset():
```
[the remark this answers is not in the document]
lrouhier: discussion here https://github.com/ivadomed/ivadomed/pull/385#discussion_r468890808

testing/test_sampler.py
lrouhier: The balance changed now And I wanted it to be robust to new images (in case we change them again)

testing/test_script.py
lrouhier: `test_training_time.py` didn't actually called the training function so tht's why I went with that. Maybe `test_training_time.py` could be removed in the future

testing_data/bounding_box_dict.json
lrouhier: Done

.github/workflows/run_tests.yml
```diff
@@ -6,12 +6,14 @@ name: Run tests
 on: [push, pull_request]
```
jcohenadad, line 6: we probably don't need both. @Drulex @kousu which one is preferred?

kousu: You want it on `pull_request` on all branches and `push` just on `master`. The syntax for this in Travis is different because Travis's `pull_request` option is a setting in their GUI

https://github.com/neuropoly/spinalcordtoolbox/blob/49cb4a16ade7c91533e669c8d1d2a777f5bbfe66/.travis.yml#L19-L22

I don't know off hand the syntax for that in Github Actions.

lrouhier: I think something like this : 
~~~
on:
  # Trigger the workflow on push or pull request,
  # but only for the master branch
  push:
    branches:
      - master
  pull_request
~~~
Should work after reading https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions .
I am going to try this

lrouhier: Ok si after getting it wrong a bunch of times I think the correct solution is in a863278

PR #386, opened 2020-07-27
Script to mesure reproducibility

andreanne-lemay:
This PR introduces a script to apply data augmentation to images and recompute metrics. The output file contains the mean and standard deviation of all metrics (avd, dice, lfdr, ltpr, mse, n_gt, n_pred, precision, recall, rvd, specificity, vol_gt, vol_pred, csa_gt, csa_pred, absolute_csa_diff, relative_csa_diff). The script takes as input one or multiple config files (-c), the number of iterations (-n) and the path (or paths) to the output file containing the results.

This PR also fixes the RandomAffine undo method. When applying `affine_transform` the translation is performed followed by the rotation and scaling transforms. To undo these transforms, the scaling and rotation transforms need to be undone, then undo translation.

### DIFF
--- a/dev/reproducibility.py
+++ b/dev/reproducibility.py
@@ -43,7 +43,7 @@ def compute_csa(config, df_results, logdir, bids):
         # Get GT csa
         gt_path = os.path.join(bids, "derivatives", "labels", subject.split("_")[0],
                                "anat", subject + config["loader_parameters"]["target_suffix"][0] + ".nii.gz")
-        os.system(f"sct_process_segmentation  -i {gt_path} -append 1 -perslice 1 -angle-corr 0 -o csa.csv")
+        os.system(f"sct_process_segmentation  -i {gt_path} -perslice 1 -angle-corr 0 -o csa.csv")
         df = pd.read_csv("csa.csv")
         # Take only the medial slice
         csa_gt = float(df["MEAN(area)"][len(df["MEAN(area)"]) // 2])
@@ -57,7 +57,7 @@ def compute_csa(config, df_results, logdir, bids):
         single_label_pred_path = "pred_single_label.nii.gz"
         nib.save(single_label_pred, single_label_pred_path)
         os.system(
-            f"sct_process_segmentation -i {single_label_pred_path} -append 1 -perslice 1 -angle-corr 0 -o csa.csv")
+            f"sct_process_segmentation -i {single_label_pred_path} -perslice 1 -angle-corr 0 -o csa.csv")
         df = pd.read_csv("csa.csv")
         # Take only the medial slice
         csa_pred = float(df["MEAN(area)"][len(df["MEAN(area)"]) // 2])
--- a/dev/reproducibility.py
+++ b/dev/reproducibility.py
@@ -12,7 +12,7 @@ from ivadomed import main as ivado
 
 def get_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument("-c", "--config", required=True, nargs="+", help="Base config file path.")
+    parser.add_argument("-l", "--log-directory", required=True, nargs="+", help="Log directory of trained model.")
     parser.add_argument("-b", "--bids-path", required=True, type=str, help="Bids path where are located the GT.")
     parser.add_argument("-n", "--iterations", default=10, type=int, help="Number of Monte Carlo iterations.")
     parser.add_argument("-o", "--output-path", nargs="+", dest="output_path", required=True,
--- a/dev/reproducibility.py
+++ b/dev/reproducibility.py
@@ -10,8 +10,9 @@ from ivadomed import main as ivado
 def get_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument("-c", "--config", required=True, help="Base config file path.")
-    parser.add_argument("-n", "--iterations", required=True, help="Number of Monte Carlo iterations.")
-    parser.add_argument("-o", "--output-path", required=True, help="Output directory to save final csv file.")
+    parser.add_argument("-n", "--iterations", default=10, type=int, help="Number of Monte Carlo iterations.")
+    parser.add_argument("-o", "--output-path", default="output_reproducibility.csv", type=str,
+                        help="Output directory to save final csv file.")
     return parser
 
 
--- a/dev/reproducibility.py
+++ b/dev/reproducibility.py
@@ -1,6 +1,8 @@
 import argparse
 import json
+import shutil
 import pandas as pd
+import os
 
 from ivadomed import main as ivado
 
@@ -18,6 +20,7 @@ def reproducibility_pipeline(config):
         context = json.load(fhandle)
 
     context["command"] = "eval"
+    shutil.rmtree(os.path.join(context["log_directory"], "pred_masks"))
     # RandomAffine will be applied during testing
     del context["transformation"]["RandomAffine"]["dataset_type"]
     return ivado.run_command(context)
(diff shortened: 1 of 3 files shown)


ivadomed/transforms.py
```diff
@@ -758,9 +758,13 @@ def undo_transform(self, sample, metadata=None):
         translation = - np.array(metadata['translation'])
 
         # Undo rotation
-        dict_params = {"rotation": [angle, axes], "scale": scale, "translation": translation}
+        dict_params = {"rotation": [angle, axes], "scale": scale, "translation": [0, 0, 0]}
+
         data_out, _ = self.__call__(sample, dict_params)
```
charleygros, line 763: Does it guarantee the rotation and scaling to be done in the right order? Or should we break it down as done with the translation?

dev/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+
+def get_results(context):
+    context["command"] = "eval"
+    pred_mask_path = os.path.join(context["log_directory"], "pred_masks")
+    if os.path.exists(pred_mask_path):
+        shutil.rmtree(pred_mask_path)
+
+    # RandomAffine will be applied during testing
+    if "dataset_type" in context["transformation"]["RandomAffine"]:
+        del context["transformation"]["RandomAffine"]["dataset_type"]
```
charleygros, line 31: That's just a safety thing to ensure this is run at eval time, right? which is normally not the case so I understand why needed here.

dev/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+        df_list = []
+        metrics = []
+        for i in range(int(args.iterations)):
+            df = get_results(context)
+            df = compute_csa(context, df)
+            metrics = list(df.columns)
+            df_list.append(np.array(df))
+
+        # Get average and std for each subject (intra subject), then average on all subjects
+        average = np.average(np.average(np.array(df_list), axis=0), axis=0)
```
charleygros, line 88: I may miss something, please correct me: In our case we would like to get the meanCSA per subjects, then compute the abs&rel CSA difference with the GT_raw per subject, then take the average and std of these for the final table. Is this still possible?

dev/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+        std = np.average(np.std(np.array(df_list, dtype=np.float), axis=0), axis=0)
```
charleygros, line 89: Yep zuper!

dev/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+    return ivado.run_command(context)
+
+
+def compute_csa(config, df_results):
+    subject_list = list(df_results.index)
+    df_results = df_results.assign(csa_pred="", csa_gt="", absolute_csa_diff="", relative_csa_diff="")
+    for subject in subject_list:
+        # Get GT csa
+        gt_path = os.path.join(config["loader_parameters"]["bids_path"], "derivatives", "labels", subject.split("_")[0],
+                               "anat", subject + config["loader_parameters"]["target_suffix"][0] + ".nii.gz")
```
charleygros, line 41: Here.. that's a bit tricky because in our case we do not want to compute the diff with the GT_down.. but with the GT_raw, which is the best GT we have for the CSA.

dev/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+        os.system("rm csa.csv")
+
+        # Get prediction csa
+        pred_path = os.path.join(config["log_directory"], "pred_masks", subject + "_pred.nii.gz")
+        pred_nii = nib.load(pred_path)
+        # Keep only first label to compute csa
+        single_label_pred = nib.Nifti1Image(pred_nii.get_fdata()[..., 0], pred_nii.affine)
+        single_label_pred_path = "pred_single_label.nii.gz"
+        nib.save(single_label_pred, single_label_pred_path)
+        os.system(f"sct_process_segmentation  -i {single_label_pred_path} -append 1 -perslice 0 -o csa.csv")
```
charleygros, line 54: :+1:

jcohenadad: for our study we probably want to estimate CSA over very small number of slices (1-5). the more slices, the less benefits we will observe with the soft vs hard seg

jcohenadad: so you could do perslice 1 or perslice 0 -z 5:7 (for example)

charleygros: Should we go with: median slice (in IS) for each image?

charleygros: > so you could do perslice 1 or perslice 0 -z 5:7 (for example)

Some images (eg ucl) have a very low number of slices (eg 3).
What about measuring on all the slices, then pick the median slice in the csv? or the 3 median slices (median +/- 1)?

dev/reproducibility.py
andreanne-lemay: ahh yes absolutely! I'll add a param for bids path

dev/reproducibility.py
andreanne-lemay: Yes, if I understand correctly what you are saying, I believe what you are describing is what I implemented:
1) Compute meanCSA on multiple slices for prediction and GT of a single patient
2) Compute abs and rel CSA diff (per subject)
3) Redo step 1 and 2 n times
4) Average within a single subject  across the n iterations (which is the first average `np.average(np.array(df_list), axis=0)`) (and measure std at this step across a single subject)
5) Average across patients to get a mean average and mean std

dev/reproducibility.py
andreanne-lemay: Exact, to avoid a config file where RandomAffine would only be applied during training (which is usually the case)

dev/reproducibility.py
charleygros: My question / issue will be fixed with this: https://github.com/ivadomed/ivadomed/pull/386#discussion_r461963974 :-)

dev/reproducibility.py
jcohenadad: medial slice sounds good

ivadomed/transforms.py
andreanne-lemay: yes absolutely! Thank you for this catch: I updated the code. For the RandomAffine undo, the transform matrix is created in the opposite order compared when doing the transform for scaling and rotation

dev/reproducibility.py
andreanne-lemay: Updated here: 83bde28

dev/reproducibility.py
andreanne-lemay: Updated here: 8361dcd

andreanne-lemay:
> use python for that

Updated here:  1059ad1



> i would use `-angle-corr 0` to prevent adding variance related to centerline fitting (same thing below)

Updated here: 3b05bfb

dev/reproducibility.py
```diff
@@ -43,11 +43,11 @@ def compute_csa(config, df_results, logdir, bids):
         # Get GT csa
         gt_path = os.path.join(bids, "derivatives", "labels", subject.split("_")[0],
                                "anat", subject + config["loader_parameters"]["target_suffix"][0] + ".nii.gz")
-        os.system(f"sct_process_segmentation  -i {gt_path} -append 1 -perslice 1 -o csa.csv")
+        os.system(f"sct_process_segmentation  -i {gt_path} -append 1 -perslice 1 -angle-corr 0 -o csa.csv")
```
jcohenadad, line 46: why explicitly using "append 1" if you delete the file afterwards? (it confuses code readers)

andreanne-lemay: Absolutely, it is unnecessary here. Updated here:  c2d8388

Issue #380, opened 2020-07-23
Add test for training_curve script


charleygros:
Once testing_data moved to the new repo.

lrouhier:
I tried to implement it (called CLI). The issue is that training curve needs tensorflow, which is not in the requirements... We need to modify that I suppose.

jcohenadad:
> We need to modify that I suppose.

you mean: add the requirement or change something in training_curve to avoid that dependency?

TF is a heavy dependency, buggy, so i'd rather avoid adding it

lrouhier:
I meant that we need to change one or the other. We are looking into a soliution that would not require to use tensorflow as a dependency.

PR #396, opened 2020-08-13
Modify scripts/training_curve.py to avoid tensorflow dependency

lrouhier:
related to #380. 

Previous implementation used TensorFlow to iterate through event en retrieves scalar.  Tensorflow is a rather heavy dependency. The changes proposed in this PR allow us to do something similar with tensorboard.

Changes: 
* Created `tensorboard_retrieve_event` function, that create df (the same way `get_data` worked)
* Removed `get_data`
* Changed the main function to call `tensorboard_retrieve_event`.

### DIFF
--- a/ivadomed/scripts/training_curve.py
+++ b/ivadomed/scripts/training_curve.py
@@ -4,7 +4,7 @@ import os
 import argparse
 import numpy as np
 from collections import defaultdict
-from tensorflow.python.summary.summary_iterator import summary_iterator
+#from tensorflow.python.summary.summary_iterator import summary_iterator
 import pandas as pd
 import matplotlib.pyplot as plt
 from textwrap import wrap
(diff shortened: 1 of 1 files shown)


ivadomed/scripts/training_curve.py
```diff
@@ -211,6 +182,55 @@ def run_plot_training_curves(input_folder, output_folder, multiple_training=Fals
...
+    Function that retrieves data from tensorboard summary event
+    Args:
+        dpath (str): log directory where the event are located
+
+    Returns:
+        df: a panda dataframe where the columns are the metric or loss and the row are the epochs.
+
+    """
+    list_metrics = ['dice_score', 'multiclass dice_score', 'hausdorff_score', 'precision_score', 'recall_score',
+                    'specificity_score', 'intersection_over_union', 'accuracy_score']
```
charleygros, line 196: Could you please add a TODO: something to remind us to find a way for not hardcoding this list? somehow.. thx

charleygros: Also add a note that elements of this list are in the same order as in the training file...

lrouhier: done in e17adda

PR #398, opened 2020-08-14
Filter image slices with std == 0

charleygros:
Some background images have not zero values --> they were not filtered [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/utils.py#L838).
We are implemented here a solution to discard these slices from the dataset if `filter_im=True`.


ivadomed/utils.py
```diff
@@ -834,7 +834,7 @@ def __call__(self, sample):
                 return False
 
         if self.filter_empty_input:
-            if not np.all([np.any(img) for img in input_data]):
+            if not np.all([np.any(img) for img in input_data]) or not np.all([img.std() != 0 for img in input_data]):
```
andreanne-lemay, line 837: The double negation is a bit confusing, without negations I find it clearer.

```suggestion
            if np.any([np.all(img) for img in input_data]) or np.any([img.std() == 0 for img in input_data]):
```

charleygros: Actually... `np.any([img.std() == 0 for img in input_data])` covers `np.any([np.all(img) for img in input_data])` eheh we can make it even simpler I think.

andreanne-lemay: Yes you're absolutely right!

PR #401, opened 2020-08-18
Find optimal threshold on raw data

charleygros:
Two changes:
- Instead of doing the binarization threshold search on the transformed data (ie after Resampling / Cropping), we do it on the Raw data
- Instead of only considering the validation sub-dataset, we now use training+validation to have a better estimation of this threshold

### DIFF
--- a/ivadomed/main.py
+++ b/ivadomed/main.py
@@ -207,7 +207,7 @@ def run_command(context, n_gif=0, thr_increment=None):
             # Choice of optimisation metric
             metric = "recall_specificity" if model_params["name"] in imed_utils.CLASSIFIER_LIST else "dice"
             # Model path
-            model_path = os.path.join(log_directory, context["model_name"] + "best_model.pt")
+            model_path = os.path.join(log_directory, "best_model.pt")
             # Training dataset without data augmentation
             ds_train = imed_loader.load_dataset(**{**loader_params,
                                                    **{'data_list': train_lst,
--- a/ivadomed/main.py
+++ b/ivadomed/main.py
@@ -200,7 +200,7 @@ def run_command(context, n_gif=0, thr_increment=None):
             n_gif=n_gif,
             debugging=context["debugging"])
 
-    if thr_increment and (command == 'train' or testing_params["binarize_prediction"] < 0):
+    if thr_increment:
         # LOAD DATASET
         if command != 'train':  # If command == train, then ds_valid already load
             # Get Validation dataset
--- a/ivadomed/main.py
+++ b/ivadomed/main.py
@@ -190,6 +190,10 @@ def run_command(context, n_gif=0, thr_increment=None):
             debugging=context["debugging"])
 
         if thr_increment:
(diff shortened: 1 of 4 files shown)


ivadomed/testing.py
```diff
@@ -243,11 +247,99 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
+                        num_workers=0)
+
+    # Run inference
+    preds_npy, gt_npy = run_inference(loader, model, model_params,
+                                      testing_params,
+                                      ofolder=None,
+                                      cuda_available=cuda_available)
+
+    # Make sure the GT is binarized
+    gt_npy = [threshold_predictions(gt, thr=0.5) for gt in gt_npy]
```
andreanne-lemay, line 309: I don't think a loop is necessary here

```suggestion
    gt_npy = threshold_predictions(gt_npy, thr=0.5)
```

ivadomed/testing.py
```diff
@@ -243,11 +247,99 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
+    # Move threshold
+    for thr in thr_list:
+        preds_thr = [threshold_predictions(copy.deepcopy(pred), thr=thr) for pred in preds_npy]
```
andreanne-lemay, line 312: same comment

```suggestion
        preds_thr = threshold_predictions(copy.deepcopy(preds_npy), thr=thr)
```

andreanne-lemay (approved):
Looks good to me :100:

ivadomed/main.py
```diff
@@ -189,8 +202,31 @@ def run_command(context, n_gif=0, thr_increment=None):
             thr_increment=thr_increment,
```
andreanne-lemay, line 202: This variable is no longer used
```suggestion
```

ivadomed/main.py
```diff
@@ -189,8 +202,31 @@ def run_command(context, n_gif=0, thr_increment=None):
...
             debugging=context["debugging"])
 
-        # Update threshold in config file
         if thr_increment:
+            # Choice of optimisation metric
```
andreanne-lemay, line 206: What do you think about putting all this block directly in testing, this way we don't need to retrain to have an optimal threshold?

ivadomed/main.py
```diff
@@ -210,6 +210,18 @@ def run_command(context, n_gif=0, thr_increment=None):
...
     if thr_increment:
+        # LOAD DATASET
+        # Get Validation dataset
+        ds_valid = imed_loader.load_dataset(**{**loader_params,
+                                               **{'data_list': valid_lst, 'transforms_params': transform_valid_params,
+                                                  'dataset_type': 'validation'}}, device=device,
+                                            cuda_available=cuda_available)
+        # Get Training dataset
+        ds_train = imed_loader.load_dataset(**{**loader_params,
+                                               **{'data_list': train_lst, 'transforms_params': transform_train_params,
```
charleygros, line 220: @andreanne-lemay --> needs transform_valid here, otherwise DA active! Will change this

PR #402, opened 2020-08-18
Fix reproducibility issue with split per center

charleygros:
When splitting dataset per center with same seed, we did not get the same sub datasets.

Please let me introduce you to the culprit: [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/loader/utils.py).
This line fails to achieve its purpose: sort alphabetically the centers. Therefore, at each run, the list is sorted differently --> resulting in different split output by sklearn function.

Minor change does the job. Next.


charleygros:
I took this opportunity to check the different behaviour (per_center and per_patient) for both SpineGeneric and GM --> looks all gravy now!

PR #406, opened 2020-08-20
Add test for mixup code

lrouhier:
Adding test to try image creation with mixup. 
Two tests are made, one for mixup and one for the image saving function associated with mixup (which is tested through mixup creation with debugging option)

### DIFF
--- a/testing/test_mixup.py
+++ b/testing/test_mixup.py
@@ -1,0 +1,26 @@
+import ivadomed.utils as imed_utils
+import torch
+
+
+def test_mixup():
+    inp = [[[[0 for i in range(40)] for i in range(40)]]]
+    targ = [[[[0 for i in range(40)] for i in range(40)]]]
+    for i in range(10):
+        for j in range(10):
+            targ[0][0][i][j]=1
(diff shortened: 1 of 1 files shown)


testing/test_mixup.py
```diff
@@ -0,0 +1,26 @@
...
+    for i in range(10):
+        for j in range(10):
+            targ[0][0][i][j]=1
+    inp = torch.tensor(inp).float()
+    targ = torch.tensor(targ).float()
+    # just testing if mixup function run
+    out = imed_utils.mixup(inp, targ, alpha=0.5)
+
+
+def test_mixup_save():
```
charleygros, line 17: @lrouhier --> What about combining these 2 two in just one test function with a `@pytest.mark.parametrize`?

charleygros (changes requested):

testing/test_mixup.py
lrouhier: Right good idea!

charleygros (approved):
:1st_place_medal:

PR #409, opened 2020-08-21
js divergence

olix86:


jcohenadad:
@olix86 can you please add a justification when closing a PR without merging it? Is the code wrong? If not, can it be used later?

olix86:
This is specific to the rater project and doesn't really apply to ivadomed in general, it is better suited for the repo specific to this project

PR #416, opened 2020-08-25
Resume training

charleygros:
When your 3-days training is suddenly killed by a fd:bmaer;hbùmrq,gqgj,qg error... it would be really really nice to resume it! This PR makes this dream come true!

DONE:
- add a flag `--resume` --> to use it: `ivadomed -c <config_file> --resume` --> will open the more recent checkpoint and update model, optimizer etc to resume the training
- this checkpoint file is saved when the validation loss is best than the current best validation loss
- `best_model.pt` is extracted from the last checkpoint and saved

TODO:
- add a test, otherwise @lrouhier would not be happy
- add this option to `automate_training` to be able to... resume it as well!
- test it on a real case scenario... and for this... sadly... rosenberg looks like in a mood to help us... :-(

### DIFF
--- a/ivadomed/training.py
+++ b/ivadomed/training.py
@@ -320,18 +320,19 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
     if os.path.isfile(resume_path):
         state = torch.load(resume_path)
         model_path = os.path.join(log_directory, "best_model.pt")
-        torch.save(model.load_state_dict(state['state_dict']), model_path)
+        best_model = model.load_state_dict(state['state_dict'])
+        torch.save(best_model, model_path)
         # Save best model as ONNX in the model directory
         try:
             # Convert best model to ONNX and save it in model directory
             best_model_path = os.path.join(log_directory, model_params["folder_name"],
                                            model_params["folder_name"] + ".onnx")
-            imed_utils.save_onnx_model(torch.load(model_path), input_samples, best_model_path)
+            imed_utils.save_onnx_model(best_model, input_samples, best_model_path)
         except:
             # Save best model in model directory
             best_model_path = os.path.join(log_directory, model_params["folder_name"],
                                            model_params["folder_name"] + ".pt")
-            torch.save(torch.load(best_model_path), best_model_path)
+            torch.save(best_model, best_model_path)
             logger.warning("Failed to save the model as '.onnx', saved it as '.pt': {}".format(best_model_path))
 
     # Save GIFs
--- a/ivadomed/training.py
+++ b/ivadomed/training.py
@@ -320,7 +320,7 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
     if os.path.isfile(resume_path):
         state = torch.load(resume_path)
         model_path = os.path.join(log_directory, "best_model.pt")
-        torch.save(model.load_state_dict(checkpoint['state_dict']), model_path)
+        torch.save(model.load_state_dict(state['state_dict']), model_path)
         # Save best model as ONNX in the model directory
         try:
             # Convert best model to ONNX and save it in model directory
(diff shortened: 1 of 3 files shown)


ivadomed/training.py
```diff
@@ -263,10 +282,15 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
 
             # UPDATE BEST RESULTS
             if val_loss_total_avg < best_validation_loss:
+                # Save checkpoint
+                state = {'epoch': epoch + 1, 'state_dict': model.state_dict(),
+                         'optimizer': optimizer.state_dict(), 'tensorboard_logger': writer,
```
andreanne-lemay, line 287: I am having a mistake here: `TypeError: cannot serialize '_io.BufferedWriter' object` It seems to be due to saving the writer. There is probably a work around this, I'll look into it :)

andreanne-lemay:
I didn't find a workaround for the SummaryWriter and surprisingly I didn't find anything online regarding this issue. It is still minor, the consequence is that we can't see the first epochs on tensorboard. For instance, I stopped the training at epoch 14, so when I restart I get something like this on tensorboard:
image

I added the scheduler, validation loss total and patience count in the checkpoint and fixed the epoch number inconsistencies when resuming a model. It works well on my end (except for what I mentioned above). Also another minor thing, even when resuming a model we get this print: `Initialising model's weights from scratch.`. I find this confusing, but very minor. I am not sure how (and if) we want to deal with this.

ivadomed/training.py
```diff
@@ -283,23 +316,32 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
     if model_params["name"] == "FiLMedUnet" and debugging:
         save_film_params(gammas_dict, betas_dict, contrast_list, model_params["depth"], log_directory)
 
-    try:
```
andreanne-lemay, line 286: I am not sure why we need to change these lines

ivadomed/training.py
```diff
@@ -283,23 +316,32 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
...
-    except:
-        # Save best model in model directory
-        best_model_path = os.path.join(log_directory, model_params["folder_name"], model_params["folder_name"] + ".pt")
-        torch.save(model, best_model_path)
-        logger.warning("Failed to save the model as '.onnx', saved it as '.pt': {}".format(best_model_path))
+    # Save best model in log directory
+    if os.path.isfile(resume_path):
+        state = torch.load(resume_path)
+        model_path = os.path.join(log_directory, "best_model.pt")
+        torch.save(state["model"], model_path)
```
andreanne-lemay, line 323: state['model'] doesn't exist here (we don't have any key named model, we only have the state dict)

ivadomed/training.py
```diff
@@ -123,7 +145,8 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
     begin_time = time.time()
 
     # EPOCH LOOP
-    for epoch in tqdm(range(1, num_epochs + 1), desc="Training"):
+    for epoch in tqdm(range(num_epochs), desc="Training", initial=start_epoch):
```
charleygros, line 148: @andreanne-lemay : Don't we need a `+1` here?

ivadomed/training.py
charleygros: If we don't have a validation dataset, then no checkpoint file, then no best model.

ivadomed/training.py
charleygros:
```suggestion
        torch.save(model.load_state_dict(checkpoint['state_dict']), model_path)
```

ivadomed/training.py
andreanne-lemay: Nope :), we have the right number of epochs when using only one param in range

andreanne-lemay (approved):

PR #417, opened 2020-08-25
Make tutorials standalone

andreanne-lemay:
A trained model has been added so all tutorials are standalone. In the same vein of thought, I added the download step to every tutorial so the users don't have to look other tutorials.


charleygros (approved):
:1st_place_medal:

PR #419, opened 2020-08-25
Add ivadomed overview image

charleygros:
https://ivadomed.org/en/cg-overview_figure/


jcohenadad (commented):
beautiful 😍
i'm thinking of maybe cropping a bit the top and bottom part (to remove some white)?

jcohenadad (commented):
also, i'm wondering how it would look like if you kept the ivadomed logo, and moved the image to the top of RTD (and remove the "ivadomed" text at the top)

charleygros:
> i'm thinking of maybe cropping a bit the top and bottom part (to remove some white)?

I ve cropped the image (without title) --> `images/overview.png` and used that one for the `README.md`.

charleygros:
> also, i'm wondering how it would look like if you kept the ivadomed logo, and moved the image to the top of RTD (and remove the "ivadomed" text at the top)

I ve added the image with overview + title --> `overview_title.png` and used that one for the top of RTD + changes that you suggested.
See new build.

jcohenadad (changes requested):
i prefer much better this new RTD version with the logo on top, however i would add white space between logo and overview.

also, i would use the same image for GH (no need to add another image on the repos).

charleygros:
> i prefer much better this new RTD version with the logo on top, however i would add white space between logo and overview.

Yes, agree! Corrected

> also, i would use the same image for GH (no need to add another image on the repos).

The only reason I did this was not to duplicate the logo (once at the top of the README, once in the figure).. New suggestion pushed :)

README.md
```diff
@@ -1,11 +1,11 @@
-![ivadomed Logo](https://github.com/ivadomed/ivadomed/raw/master/images/ivadomed_logo.png)
-
 Integrated framework for medical image analysis with deep learning.
```
jcohenadad, line 1: i would move that below the image and change the sentence for: "`ivadomed` is an integrated framework for medical image analysis with deep learning."

jcohenadad (changes requested):
the list was lost in translation:
image

jcohenadad (commented):
i would add a space before this title appears:
https://github.com/ivadomed/ivadomed/pull/419/files#diff-8aa79b83ed86e48962cc94787fc5cb56R15

and/or a horizontal line

README.md
```diff
@@ -1,11 +1,11 @@
-![ivadomed Logo](https://github.com/ivadomed/ivadomed/raw/master/images/ivadomed_logo.png)
-
-Integrated framework for medical image analysis with deep learning.
-
 [![Coverage Status](https://coveralls.io/repos/github/ivadomed/ivadomed/badge.svg?branch=master)](https://coveralls.io/github/ivadomed/ivadomed?branch=master)
 ![](https://github.com/neuropoly/ivadomed/workflows/Python%20package/badge.svg)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](LICENSE.md)
 
+![ivadomed Overview](images/overview_title.png)
+
```
jcohenadad, line 6: i would move the badges here

charleygros:
> the list was lost in translation:
> image

These lines have not been lost: they still are in `index.rst`. But I am scared I am missing something: are you suggesting to add this list to `README.md` as well?

README.md
charleygros: Done

charleygros:
> i would add a space before this title appears:
> https://github.com/ivadomed/ivadomed/pull/419/files#diff-8aa79b83ed86e48962cc94787fc5cb56R15
> 
> and/or a horizontal line

Done in [2681641](https://github.com/ivadomed/ivadomed/pull/419/commits/26816417477cdd30f8ac8f4fd2a037cc8484fd41).

jcohenadad:
> > the list was lost in translation:
> > image
> 
> These lines have not been lost: they still are in `index.rst`. But I am scared I am missing something: are you suggesting to add this list to `README.md` as well?

i meant that the bullet points were lost (not the text)-- see: https://ivadomed.org/en/latest/ in comparison

charleygros:
> i meant that the bullet points were lost (not the text)-- see: https://ivadomed.org/en/latest/ in comparison

Oh right!!
I played with it for a while.. but can't manage to have bullet points anymore, even if the PyCharm preview shows them.
It seems to work with Numbered list tho (last pushed commit).

jcohenadad:
@charleygros i recommend you try building the sphinx doc locally, it is much easier for debugging the doc

charleygros:
> @charleygros i recommend you try building the sphinx doc locally, it is much easier for debugging the doc
> 
> instructions [here](https://github.com/ivadomed/ivadomed/wiki/documentation)

Yeah, that's easier for debugging. But I still can't make it work :/
I ve asked the other kids if they could help here.

docs/source/index.rst
```diff
@@ -1,15 +1,20 @@
 ivadomed
 ========
```
charleygros, line 2: Yes the issue is coming from the title, but see comment [here](https://github.com/ivadomed/ivadomed/pull/419#pullrequestreview-475094131). .. :-(

andreanne-lemay: We need a title or else in the breadcrumb, it is shown that Contributors is the title of the page
image

But we could change ivadomed for something more general like Home Page or Summary

jcohenadad: hum... we should be able to omit the "home page" or "ivadomed" at the top, see eg [here](https://niftytorch.github.io/doc/)

jcohenadad: ^ ah, nevermind, they are using MkDocs, so there might be some subtle differences...

jcohenadad: i'm definitely not a big fan of the "home page", so let's try to find a workaround...

jcohenadad: doing some tests locally-- a solution would be to move the contributors/license in separate files and be displayed in the TOC

jcohenadad: oops... when doing that, that's what we get 😞 

image

jcohenadad: https://github.com/ivadomed/ivadomed/pull/419/commits/483acd51f993361cf219a23cd0e7d773ad01ba52 seems like a good compromise:

image

but now we're back to the list not displaying 😭

jcohenadad: ah! i have an idea: let's move the purpose to another section

docs/source/index.rst
```diff
@@ -1,6 +1,3 @@
-Home Page
-=========
```
charleygros: Yeah the result of our investigations was that a title is (for some reason) needed to have the bullet point displayed eheh :-(

docs/source/index.rst
jcohenadad: ok how about https://github.com/ivadomed/ivadomed/pull/419/commits/ad93601c4126fb73426270188129443e23358c35 ?

i actually find it better than previously-- it will allow us to also add more items in the list of purpose (two items was a bit sad)

charleygros: I also prefer to move the main heading from "comparison to other projects" to "purpose" :-)

charleygros:
@jcohenadad : Should we go with that and merge?

jcohenadad (approved):

PR #425, opened 2020-08-31
Updating documentation

### LINKED ISSUE #418: missing an entry for script in RTD?
https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/training_curve.py

lrouhier:
This PR aims at completing the documentation. 
This add documentation for training curve and download_data
The example DF needed for automate_training and  compare_models were also added 
A few indentation changes were done for better visuals. 

fix #418 

To see built documentation visit: https://ivadomed.org/en/lr-forgotten_doc/


docs/source/scripts.rst
```diff
@@ -34,4 +34,15 @@ ivadomed_extract_small_dataset
 
 .. autofunction:: ivadomed.scripts.extract_small_dataset.extract_small_dataset
 
+ivadomed_training_curve
+""""""""""""""""""""""""""""""
```
jcohenadad, line 38: needs to be the same number of chars as the title

ivadomed/scripts/download_data.py
```diff
@@ -120,28 +119,49 @@ def unzip(compressed, dest_folder):
...
-    :return: None
+    Download a data bundle from an URL and install it in the destination folder.
+
+    Usage example ::
+
+        ivadomed_download_data -d data_testing -o ivado_testing_data
+
+    Existing data bundle:
+
+        - `data_example_spinegeneric <https://github.com/ivadomed/data_example_spinegeneric/archive/r20200825.zip>`_: 10 randomly picked subject from `Spine Generic <https://github.com/spine-generic/data-multi-subject>`_. Used for Tutorial and example in Ivadomed.
```
jcohenadad, line 130: we should replace with an automatic generation from the DICT_URL

ideally DICT_URL would contain the description (nested dic)

lrouhier:
@lrouhier update 'Data' page with data_example_spine_generic

ivadomed/scripts/download_data.py
lrouhier: Ok so it's done after a long trial and error period see 
482a676 Visual is not really great. The issue is the following: The part that I add does not use the docstring convention so I had to improvise a bit (hence it is not a bullet list). But at least it's dynamic.

docs/source/scripts.rst
lrouhier: done in 384de6c

ivadomed/scripts/download_data.py
```diff
@@ -118,30 +127,58 @@ def unzip(compressed, dest_folder):
...
-    :return: None
+    Download a data bundle from an URL and install it in the destination folder.
+
+    Usage example ::
+
+        ivadomed_download_data -d data_testing -o ivado_testing_data
+
+
+    Existing data bundle: |br|
+    %s
```
jcohenadad, line 153: the doc looks very cool! but i'm curious how it works?? (where is this "%s" defined???)

lrouhier: here https://github.com/ivadomed/ivadomed/blob/482a6769a364b64f2051a6d0995b680133f45d78/ivadomed/scripts/download_data.py#L250

lrouhier:
> @lrouhier update 'Data' page with data_example_spine_generic

 done in 30d1d7e

ivadomed/scripts/download_data.py
```diff
@@ -210,6 +247,8 @@ def install_data(url, dest_folder, keep=False):
     shutil.rmtree(extraction_folder)
 
 
+install_data.__doc__%=create_string(DICT_URL)
```
jcohenadad, line 250: i would definitely add comments above this line of code to explain what it does

lrouhier: done in 100747d

ivadomed/scripts/download_data.py
lrouhier: Well small thing to correct, links are not working (I wanted to point to the repo and not the zip so I tried something but it does not work). I will fix that.

lrouhier: fixed in cf89b3d

lrouhier:
Olivier picked up that '--' were seen as one long dash in the doc, which is often used for flag (e.g, --output). It has been changed in 45231e7 and should be ok.

lrouhier:
> Olivier picked up that '--' were seen as one long dash in the doc, which is often used for flag (e.g, --output). It has been changed in [45231e7](https://github.com/ivadomed/ivadomed/commit/45231e7b55d27e21cadb3991d1f0ef674fcb946e) and should be ok.
> 
> Edit: some conf option needed to be deactivated, done in [cdb0cec](https://github.com/ivadomed/ivadomed/commit/cdb0cec08e0fc5459c490e6821e88b28c57bedc6)

It is weirdly working locally but not on the built branch...

lrouhier:
So after help from @kousu, it seems that RTD and us do not use the same sphinx version. Which explains the difference. I'll look into it.

ivadomed/scripts/download_data.py
kousu: What about using `install_data.__doc__.format(DICT_URL)`? Then you can rewrite `__doc__` to use named variables instead of bare `%s`s.

lrouhier: I tried something similar and I might have made a mistake because it didn'T work. Would it be better practice?

ivadomed/scripts/download_data.py
```diff
@@ -13,16 +13,26 @@
...
-    "mice_uqueensland_gm": ["https://github.com/ivadomed/mice_uqueensland_gm/archive/r20200622.zip "],
-    "mice_uqueensland_sc": ["https://github.com/ivadomed/mice_uqueensland_sc/archive/r20200622.zip"],
-    "findcord_tumor": ["https://github.com/ivadomed/findcord_tumor/archive/r20200621.zip"]
-    }
+    "data_example_spinegeneric": {
+        "url": ["https://github.com/ivadomed/data_example_spinegeneric/archive/r20200825.zip"],
+        "description": "10 randomly picked subject from "
+                       "`Spine Generic <https://github.com/spine-generic/data-multi-subject>`_. "
+                       "Used for Tutorial and example in Ivadomed."},
+    "data_testing": {"url": ["https://github.com/ivadomed/data-testing/archive/r20200807.zip"],
```
jcohenadad, line 22: pep8

ivadomed/scripts/download_data.py
```diff
@@ -210,6 +249,13 @@ def install_data(url, dest_folder, keep=False):
     shutil.rmtree(extraction_folder)
 
 
+# This line allows to format the install_data docstring
+# It can not be done in the function directly which is why it is here.
+# create_string is a custom function that convert our dict into a string
+# which is easier to add in the documentation.
```
jcohenadad, line 255:
```suggestion
# This line allows to format the `install_data()` docstrings, because this formatting
# cannot be done in the function directly. 
# `create_string()` is a custom function that converts our dict into a string
# which is easier to add in the documentation.
```

ivadomed/scripts/training_curve.py
```diff
@@ -84,34 +84,38 @@ def run_plot_training_curves(input_folder, output_folder, multiple_training=Fals
...
         - the metrics computed on the validation sub-dataset.
 
-    It could consider one log directory at a time, for example::
+    It could consider one log directory at a time, for example:
+
     .. image:: ../../images/plot_loss_single.png
         :width: 600px
         :align: center
 
     ... or multiple (using ``multiple_training=True``). In that case, the hard line represent the mean value across the
```
jcohenadad, line 97:
```suggestion
    ... or multiple (using ``multiple_training=True``). In that case, the hard line represents the mean value across the
```

jcohenadad (approved):

ivadomed/scripts/download_data.py
kousu: I forgot that you need to unpack the dict:

```
install_data.__doc__=install_data.__doc__.format(**DICT_URL)
```

But I didn't look at the rest of the context when I gave that suggestion.

PR #429, opened 2020-09-03
Version pin sphinx.

kousu:


kousu:
I got rid of the `requirements: requirements.txt` line because it's redundant, and my impression is requirements.txt files are slowly on their way to deprecation; we already load it in setup.py anyway, and `method: pip` implies doing `pip install .` which means using setup.py.

kousu:
Please double-check that the instructions I wrote work for you!

kousu:
Sorry this isn't ready yet it's still buggy.

lrouhier (approved):
looks good.

PR #430, opened 2020-09-03
split from right instead of left

olix86:
Changes nothings for most use cases, but when using `automate_training` to vary `target_suffix`, fixes a bug were uncertainty predictions are saved in the wrong folder due to the fact that the `target_suffix` is present twice in the path and therefore the split was done at the first occurence instead of the last


lrouhier (approved):
seems good !

PR #435, opened 2020-09-15
evaluate -- Bread crumb cleaning

charleygros:
I promise that I was expecting to make much more important changes in this PR...
I was working on a bug that ended up not being a bug (unfortunately or not eheh).
Anyway.. only a microscopic change here to review.


andreanne-lemay (approved):
Wouhou one parameter removed!

Issue #434, opened 2020-09-14
Wrong image size while using center crop


lrouhier:
While working with sct testing large, I cannot use a bigger batch size than one because there is a mismatch in images dimension in the batch. I am using a center crop after a resampling which should be enough to ensure same size images.  As the dimension of the image are really different across the dataset, some needs to be padded and others need to be cropped. 
After a little digging, I think that it comes from the fact that the dimensions are different between what is padded and what is cropped. 
What is really weird is that the difference in dimension at the time of the error was not constant.

charleygros:
Thanks @lrouhier for reporting this.
Could you please share or add in `duke/tmp/` the config file you used to raise this issue? Cheers

lrouhier:
The config file is now in duke/temp/lrouhier/Centercrop_issue, tell me if you have access but I think you should be able to access it. 

I launched two training sequentially, here are the errors I get: 
test 1:
`RuntimeError: stack expects each tensor to be equal size, but got [1, 250, 497] at entry 0 and [1, 250, 502] at entry 1`
test 2: 
`RuntimeError: stack expects each tensor to be equal size, but got [1, 250, 500] at entry 0 and [1, 250, 499] at entry 1`

As you can see what is really weird is that the difference in the size of the images in the batch varies from one training to the other.

lrouhier:
I created a second config file (name config_file_smaller_test.json), with a centercrop with a much smaller size (50x50) and it works, So it appears that if all images are cropped (because they are all bigger than 50x50) there does not seem to be a problem. Maybe it is more of a padding issue. Let me see what happens if I pad all the images.

lrouhier:
Here are some stat about the image: 
min size_x (RL) : 70
max size_x (RL): 71
min size_y (IS): 160
max size_y (IS): 733


I tried to pad only on the first axis (x) with a center crop of size (75,50), it worked. Now If I pad with center crop of size (250,50), I get an error. Same as before the size is not constant. Which means the images have different behavior.

Test 1 failed with different size in batch 249 and 250, 
Test 2 failed with different size in batch 249 and 251 

This is weird.  I will dig around the padding function to see what happens.

lrouhier:
[...]

PR #446, opened 2020-09-17
Al/remove preprocessing param

andreanne-lemay:
Fix related to issue #434 . "preprocessing" parameter is no longer necessary. By default, Resample, CenterCrop, and ROICrop transforms will only be applied once at the beginning of the training.


testing/test_bounding_box.py
```diff
@@ -22,7 +22,6 @@
             "log_directory": LOG_DIR
         },
         "transforms_params": {
-            "Resample": {"wspace": 0.75, "hspace": 0.75},
```
charleygros, line 25: Not needed anymore?

charleygros (approved):
Royal, merci!

testing/test_bounding_box.py
andreanne-lemay: Before the changes, Resample was not applied as preprocessing for this test, but now that it is automatically being applied as preprocessing, the bounding box dimensions were not good anymore since the dimensions corresponded to the resampled image :). This transform was not necessary for this test.

PR #447, opened 2020-09-17
Retrain model without resetting weights

lrouhier:
This PR aims at adding an option to retrain a model without resetting any weights.
This could be useful if you think you need to add more epochs to an ongoing training for example. This is designed as an optional field.

### DIFF
--- a/ivadomed/training.py
+++ b/ivadomed/training.py
@@ -87,8 +87,13 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
             100 - training_params["transfer_learning"]['retrain_fraction'] * 100.))
         old_model_path = training_params["transfer_learning"]["retrain_model"]
         fraction = training_params["transfer_learning"]['retrain_fraction']
+        if 'reset' in training_params["transfer_learning"]:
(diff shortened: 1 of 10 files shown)


ivadomed/training.py
```diff
@@ -87,8 +87,13 @@ def train(model_params, dataset_train, dataset_val, training_params, log_directo
             100 - training_params["transfer_learning"]['retrain_fraction'] * 100.))
         old_model_path = training_params["transfer_learning"]["retrain_model"]
         fraction = training_params["transfer_learning"]['retrain_fraction']
+        if 'reset' in training_params["transfer_learning"]:
+            reset = training_params["transfer_learning"]['reset']
+        else :
+            reset = True
```
charleygros, line 93: I would recommend having `reset` as mandatory if using the `transfer_learning`: I am just scared this parameter will be forgotten otherwise and yield errors :/ Let's make it explicit I reckon

charleygros (commented):
Would be good to document `reset` in `configuration_file.rst`. Thanks :-)

andreanne-lemay (approved):
Looks good to me :)!

PR #453, opened 2020-09-23
fix run_test gpu assignation

olix86:


andreanne-lemay (approved):
:100: Thanks, hopefully it will be faster now!

PR #456, opened 2020-09-24
Display version

### LINKED ISSUE #455: Fetch ivadomed version that is being used
In order to make sure what version users are using, it would be great to fetch and display version and git commit SHA, [as done for SCT](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/utils.py#L461)

charleygros:
This PR fetches the `ivadomed` version, either release version or git commit, and display it in the terminal each time a script is launched.

Note: adding the file `ivadomed/version.txt` were is indicated the release version, instead of `ivadomed/__init__.py`.

Fixes #455.


charleygros:
Message at the top of each scripts:
```
--
IVADOMED (git-cg/version-b26271a0bd80695b2c3ba758b5fd70888fe5db7c)

Log directory already exists: spineGeneric
Using GPU number 0
```

charleygros:
TODO before merging: change the docstrings of new functions (thanks heaps to SCT team eheh)

ivadomed/__init__.py
```diff
@@ -1 +1,2 @@
-__version__ = "2.2.1"
```
jcohenadad, line 1: is this change still compatible with RTD build (which fetches the version, i think..?)

ivadomed/scripts/convert_to_onnx.py
```diff
@@ -50,4 +50,5 @@ def main():
 
 
 if __name__ == '__main__':
+    imed_utils.init_ivadomed()
```
lrouhier, line 53: I performed a test with this script and the version did not show.. would it be possible to put that directly in the main function ?

ivadomed/__init__.py
Drulex: Looks ok https://readthedocs.org/projects/ivado-medical-imaging/builds/11961354/

ivadomed/scripts/convert_to_onnx.py
charleygros: You are very correct! Thanks ++
Corrected, please let me know if the problem remains, cheers

charleygros:
> TODO before merging: change the docstrings of new functions (thanks heaps to SCT team eheh)

Done.

andreanne-lemay (approved):
Works well on my end :)! Thank you for this change, my only comment would be why the two "--", it is not super aesthetic, but really minor.
image

jcohenadad:
how does everyone feel about displaying “ivadomed” (without capital letters), as it is the official writing of the software package?

charleygros:
> my only comment would be why the two "--", it is not super aesthetic, but really minor.

Yep, removed.

> how does everyone feel about displaying “ivadomed” (without capital letters), as it is the official writing of the software package?

Yep, agreed, corrected.

PR #458, opened 2020-09-28
Custom final activation

andreanne-lemay:
Add the option to set the final activation layer to either 'sigmoid', 'relu' or 'softmax'.


charleygros (approved):
Looks good to me!

Issue #439, opened 2020-09-17
Uncertainty as post-processing step


andreanne-lemay:
Currently uncertainty computation is available on ivadomed, but the generated information is not used to improve the segmentation.

Related tasks:
- Pipeline to find the best uncertainty for a given task
- Post-processing function to modify predictions with uncertainty
- Implement new forms of uncertainty

PR #459, opened 2020-09-28
Uncertainty as post-processing step

andreanne-lemay:
Main changes:
- Save images after post-processing step so the predictions are representative of the metrics in evaluation_3Dmetrics.csv
- Use a threshold on uncertainty as post-processing step


ivadomed/evaluation.py
```diff
@@ -46,7 +47,12 @@ def evaluate(bids_path, log_directory, target_suffix, eval_params):
         fname_pred = os.path.join(path_preds, subj_acq + '_pred.nii.gz')
         fname_gt = [os.path.join(bids_path, 'derivatives', 'labels', subj, 'anat', subj_acq + suffix + '.nii.gz')
                     for suffix in target_suffix]
+        fname_uncertainty = os.path.join(path_preds, subj_acq + '_soft.nii.gz')
```
charleygros, line 50: I think we should be a bit clearer here as this file is not technically the uncertainty measure but rather the non-binarized average of the MC predictions.
An alernative would be to pass as parameter the suffix (eg here "soft") to control this --> eg I want to do some post processing using the CV metric --> I will have in my config file:
```
         "postprocessing": {"thr": 0.1, "suffix": "_unc-cv.nii.gz"}
```
It can also be in another PR... I am just putting this on the table.

charleygros: But .. if we go that route.. it would be a bit different from here as we will use the unc file to remove some voxels on the predictions (so we will have two files on board) etc.

ivadomed/utils.py
```diff
@@ -143,7 +143,8 @@ def run_uncertainty(ifolder):
         fname_soft = os.path.join(ifolder, subj_acq + '_soft.nii.gz')
         # find Monte Carlo simulations
         fname_pred_lst = [os.path.join(ifolder, f)
-                          for f in os.listdir(ifolder) if subj_acq + '_pred_' in f and '_painted' not in f]
+                          for f in os.listdir(ifolder) if subj_acq + '_pred_' in f and
+                          ('_painted' not in f) and ('_color' not in f)]
```
charleygros, line 147: ++ !

ivadomed/evaluation.py
charleygros: Throwing other thoughts: that could be quite time consuming.. but we could also imagine a post processing where the unc_thr is optimized on the train+valid (as done for the vox_thr) then used to perform some post-processing. (which would require to run the MC on the entire Train+Valid then perform the unc_thr optimization. Another question would be: do we optimise the vox_thr or the unc_thr first? or both simultaneously? ...etc if so how? eheh Let's brainstorm about it tomorrow!

andreanne-lemay: > I think we should be a bit clearer here as this file is not technically the uncertainty measure but rather the non-binarized average of the MC predictions.
> An alernative would be to pass as parameter the suffix (eg here "soft") to control this --> eg I want to do some post processing using the CV metric --> I will have in my config file:
> 
> ```
>          "postprocessing": {"thr": 0.1, "suffix": "_unc-cv.nii.gz"}
> ```
> 
> It can also be in another PR... I am just putting this on the table.

Yes, I like the idea, I'll do the change.



> Throwing other thoughts: that could be quite time consuming.. but we could also imagine a post processing where the unc_thr is optimized on the train+valid (as done for the vox_thr) then used to perform some post-processing. (which would require to run the MC on the entire Train+Valid then perform the unc_thr optimization. Another question would be: do we optimise the vox_thr or the unc_thr first? or both simultaneously? ...etc if so how? eheh Let's brainstorm about it tomorrow!

For this part, I'd do it in another PR since it requires a lot of changes (related to pipeline idea presented here #439) . Also currently (as you mentioned) running 10 MC iterations is extremely long even only on the test set. If we want this to be viable on train+validation set, we need to concentrate on time optimization which might require some refactoring.

andreanne-lemay:
image

I ran uncertainty thresholding (with thr=0.5) on csf segmentation model which means I removed all pixels with uncertainty > 0.5 . I tried with aleatoric, epistemic, both, voxel-wise and structure-wise uncertainty. Here is a glimpse of the results. Apologies for the uneven screenshot size, but it is always the same slice in the image (subj strasbourg04 slice 8).

ivadomed/main.py
```diff
@@ -282,10 +282,12 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
         # PREDICTION FOLDER
         path_preds = os.path.join(log_directory, 'pred_masks')
         # If the prediction folder does not exist, run Inference first
-        if not os.path.isdir(path_preds):
-            print('\nRun Inference\n')
-            context["command"] = "test"
-            run_command(context)
+        if os.path.isdir(path_preds):
+            context['testing_parameters']['uncertainty']['n_it'] = 0
```
andreanne-lemay, line 286: Re-run prediction since the _pred.nii.gz file is the prediction with previous post-processing step (not raw prediction), but MC iterations are not re-run.

jcohenadad:
Thank you for https://github.com/ivadomed/ivadomed/pull/459#issuecomment-701546786 @andreanne-lemay, this is super interesting! 

based on these preliminary results, for segmentation tasks on big objects (ie cord, csf, but _not_ MS lesions), i would be inclined to not rely on the uncertainty for post-processing, but just apply CRF on the prediction to remove small objects (ie: what is currently done in deepseg_sc). It seems like using the uncertainty might "erode" a bit of the prediction, which might be a non-desired effect.

andreanne-lemay:
> Thank you for [#459 (comment)](https://github.com/ivadomed/ivadomed/pull/459#issuecomment-701546786) @andreanne-lemay, this is super interesting!
> 
> based on these preliminary results, for segmentation tasks on big objects (ie cord, csf, but _not_ MS lesions), i would be inclined to not rely on the uncertainty for post-processing, but just apply CRF on the prediction to remove small objects (ie: what is currently done in deepseg_sc). It seems like using the uncertainty might "erode" a bit of the prediction, which might be a non-desired effect.

Yes, I agree. For large structures, structure-wise uncertainty seems to be more adequate to remove random blobs of predictions (e.i. bottom right predictions of the spinal cord).

charleygros:
Really cool ++

Yup, we really see a potential / benefit of using an uncertainty_threshold in addition to (or instead?) a pred_threshold (top row). Something like:
- structure-wise to remove blobs (but on that example it looks like voxel-wise both_unc can do this job too) or false positive MS lesions
- voxel-wise to refine big objects borders

ivadomed/postprocessing.py
```diff
@@ -211,7 +211,7 @@ def label_file_from_coordinates(nifti_image, coord_list):
 
 def threshold_uncertainty(data, uncertainty, thr):
```
charleygros, line 212: What about calling it "mask_predictions" or something like this? Or .. to keep this name but rename `threshold_predictions` with `binarize_predictions`? To make clear the difference between the two (below thr = 0 and keep other values VS below=0, above=1).
Also, I am thinking that this function could be used for other purposes than uncertainty, eg #463 --> `mask_predictions(data, data, thr=1e-2)`.

charleygros (approved):
Great +++!

PR #465, opened 2020-10-06
Remove "eval" cmd

andreanne-lemay:
Remove "eval" command and keep only "test" since "eval" requires "test".


charleygros (commented):
Pow.. makes the whole thing so much easier!
My only comment would be to modify the tutorials as well

docs/source/tutorials/one_class_segmentation_2d_unet.rst
```diff
@@ -175,11 +175,11 @@ Evaluate model
 --------------
 
 To test the trained model on the testing sub-dataset and compute evaluation metrics, open your config file and
-set ``command`` to ``eval``:
+set ``command`` to ``test``:
```
andreanne-lemay, line 178: @charleygros I did modify the tutorials, did I miss a spot where it needs to be updated :)?

charleygros: My bad @andreanne-lemay , for some reasons I thoughts there were more changes to do in the tutos but I ve just checked and it is all golden!

charleygros (approved):

PR #472, opened 2020-10-13
Adapt all metrics to multiclass predictions

andreanne-lemay:
Only Dice score was properly adapted to multiclass predictions. The other metrics computed the score for all the predictions, instead of by class.

2D predictions were also not adapted to multiclass.


ivadomed/evaluation.py
```diff
@@ -444,21 +444,24 @@ def run_eval(self):
             dict, ndarray: dictionary containing evaluation results, data with each object painted a different color
         """
         dct = {}
-
+        data_gt = self.data_gt.copy()
```
charleygros, line 447: Have you checked if still okay with one-class segmentation tasks? Otherwise happy to have a look at it!

ivadomed/testing.py
```diff
@@ -348,3 +343,14 @@ def threshold_analysis(model_path, ds_lst, model_params, testing_params, metric=
         imed_metrics.plot_roc_curve(tpr_list, fpr_list, optimal_idx, fname_out)
 
     return optimal_threshold
+
+
+def get_gt(filenames):
+    gt_lst = []
```
charleygros, line 349:
```suggestion
    """Get ground truth data as numpy array.
    
    Args:
        - filenames (list): List of ground truth filenames, one per class.
    Returns:
        ndarray: 4D numpy array.
    """
    gt_lst = []
```

charleygros (approved):

ivadomed/testing.py
andreanne-lemay: Yes thanks for this :D!

PR #481, opened 2020-10-19
Version format

### LINKED ISSUE #480: Error while releasing new version
Version number no longer appears in the __init__.py file. The version number is now parsed differently causing an error during deployment:

Now:
image

Before:
image

charleygros:
Fixes #480.

The issue was coming from the parsing of the version number.

I am suggesting that we have `dev` in `version.txt` on master --> and change it to the release version (e.g., `2.3.0`, without quotes or ..) when generating the release, following our older bro', SCT: [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/version.txt) and [here](https://github.com/neuropoly/spinalcordtoolbox/blob/4.3/version.txt).


andreanne-lemay (approved):
:100:

jcohenadad:
@charleygros or someone can you please update the wiki section release workflow so we don’t forget to update version.txt

charleygros:
> @charleygros or someone can you please update the wiki section release workflow so we don’t forget to update version.txt

Done here #483

jcohenadad:
> I am suggesting that we have dev in version.txt on master --> and change it to the release version (e.g., 2.3.0, without quotes or ..) when generating the release, following our older bro', SCT: here and here.

i'm coming a bit late... but a difference with SCT, is that SCT has a `release` branch, where we can update `version.txt` separately from `master` (which is kept "dev"), while `ivadomed` does not. So, I've drafted a procedure for creating a release, but I realize now that the drafted procedure will not work because each tag/release points to master. So what is the current strategy?  Change version.txt to e.g. vX.Y.Z, release, and then put back master's version.txt to "dev"? That sounds prone to error.

PR #483, opened 2020-10-19
Documenting release workflow

charleygros:
Related to: [here](https://github.com/ivadomed/ivadomed/pull/481#issuecomment-712485479).


jcohenadad:
i meant the procedure on the wiki, ie: what are the steps to follow to create a new release.

i'm working on it-- should be ready in 5min

jcohenadad:
from: https://github.com/ivadomed/ivadomed/pull/481

> I am suggesting that we have dev in version.txt on master --> and change it to the release version (e.g., 2.3.0, without quotes or ..) when generating the release, following our older bro', SCT: here and here.

a difference with SCT, is that SCT has a `release` branch, where we can update `version.txt` separately from `master` (which is kept "dev"), while `ivadomed` does not. So, I've [drafted a procedure](https://github.com/ivadomed/ivadomed/wiki/create-new-release) for creating a release, but I realize now that the drafted procedure will not work because each tag/release points to master. So what is the current strategy?  Change version.txt to e.g. vX.Y.Z, release, and then put back master's version.txt to "dev"? That sounds prone to error and high maintenance. 

For example, after releasing v2.3.1 the current [`version.txt`](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/version.txt) in the `master` branch has the value `2.3.1` instead of `dev`, as suggested in https://github.com/ivadomed/ivadomed/pull/481#issue-506321312. 

So to remove one step, we could create a branch `release`.

Altenatively, if we want more automation, we are currently testing a procedure in [AxonDeepSeg](https://github.com/neuropoly/axondeepseg/pull/364), wherein:
- a release is automatically created by adding a trigger in the PR commit message
- pypi is pushed
- changelog is automatically created
- the init.py version is automatically bumped (based on the release trigger: minor/major/patch)

we could test out this workflow for a bit, and if we find it useful we could replicate it for ivadomed.

charleygros:
> a difference with SCT, is that SCT has a `release` branch, where we can update `version.txt` separately from `master` (which is kept "dev"), while `ivadomed` does not.

Mmm I see. I forgot about this difference when I suggested the `dev` version.

> So what is the current strategy?

Before #481 : Update the `version.txt` at each new release.

> Change version.txt to e.g. vX.Y.Z, release, and then put back master's version.txt to "dev"? That sounds prone to error and high maintenance.

Since #481: correct. And Agree: prone to error :/

> So to remove one step, we could create a branch `release`.
> 
> Altenatively, if we want more automation, we are currently testing a procedure in [AxonDeepSeg](https://github.com/neuropoly/axondeepseg/pull/364), wherein.

Both solutions sound good to me. So.. in the meantime, would you recommend to go with post#481 strategy (ie updating version in txt only at release time)? or to go with SCT strategy (ie to create a `release` branch)?

jcohenadad:
easiest solution would be to keep version.txt as the current release number (now 2.3.1). It will evolve with master, but if users have cloned the repos (vs stable release) it will display the git commit so we can properly debug in case users have issue.

when next release comes, we bump version.txt, and the story continues

docs/source/contributing.rst
```diff
@@ -279,6 +279,8 @@ modules or data/binaries (which requires to re-run installer for people
 working on the dev version), MAJOR version when there are major
 incompatible API changes, Beta releases follow the following convention:
 
-MAJOR.MINOR.PATCH-beta.x (with x = 0, 1, 2, etc.) Stable version is
-indicated in the file version.txt. For development version (on master),
-the version is "dev".
+MAJOR.MINOR.PATCH-beta.x (with x = 0, 1, 2, etc.)
+
+Stable version is indicated in the file version.txt. For development 
+version (on master), the version is "dev". On the release branch, edit
```
charleygros, line 285:
```suggestion
Stable version is indicated in the file version.txt. On the release branch, edit 
```

charleygros:
> easiest solution would be to keep version.txt as the current release number (now 2.3.1). It will evolve with master, but if users have cloned the repos (vs stable release) it will display the git commit so we can properly debug in case users have issue.
> 
> when next release comes, we bump version.txt, and the story continues

Yep sounds great! I changed `dev` to `2.3.1`.

jcohenadad:
in https://github.com/ivadomed/ivadomed/pull/483/commits/eff7a0c985a0a0056ace279160a9d995c80098f0 I redirected the release procedure in the wiki as it addressed for internal contributors

jcohenadad (approved):

PR #484, opened 2020-10-22
Configuration File Manager

andreanne-lemay:
Add ConfigurationManager class ("ivadomed/config_manager.py") to help with retrocompatibility when the configuration file is modified. The default config file is defined in "config/config_default.json". From now on, it will only be necessary to update "config/config_default.json" and not all config files.

### DIFF
--- a/ivadomed/config_manager.py
+++ b/ivadomed/config_manager.py
@@ -69,7 +69,7 @@ class ConfigurationManager(object):
         if self.updated_config['debugging']:
             self._display_differing_keys()
 
-        return self.default_config
+        return self.updated_config
 
     def _display_differing_keys(self):
         print('Adding the following keys to the configuration file')
--- a/ivadomed/config_manager.py
+++ b/ivadomed/config_manager.py
@@ -43,8 +43,16 @@ class ConfigurationManager(object):
         Returns:
             dict: Updated configuration dict.
         """
+        if self.context['debugging']:
+            self._display_differing_keys()
         self.default_config.update(self.context)
+
         return self.default_config
+
+    def _display_differing_keys(self):
+        for key in self.default_config:
+            if key not in self.context:
+                print(f'Adding the following key in configuration file:\n {key}: {self.default_config[key]}')
 
     def _validate_path(self):
         """Ensure validity of configuration file path.
--- a/ivadomed/config_manager.py
+++ b/ivadomed/config_manager.py
@@ -1,7 +1,7 @@
 import json
 import os
 
-__ivadomed_dir__ = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+from . import __ivadomed_dir__
 
 
 def _load_json(config_path):
@@ -9,8 +9,7 @@ def _load_json(config_path):
 
     Args:
         config_path (str): Path to json file.
-    Returns:
-        dict: config dictionary.
+
     """
     with open(config_path, "r") as fhandle:
         default_config = json.load(fhandle)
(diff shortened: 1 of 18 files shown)


andreanne-lemay:
For now I only did the minimal code to make it work. I did not integrate the config documentation in the class, since I wasn't sure where to include it.

ivadomed/config_manager.py
```diff
@@ -0,0 +1,51 @@
...
+        self.context = _load_json(context_path)
+
+    def get_config(self):
+        """Get updated configuration file with all parameters from the default config file.
+
+        Returns:
+            dict: Updated configuration dict.
+        """
+        self.default_config.update(self.context)
+        return self.default_config
```
charleygros, line 44: Great!
What about adding something like: if debbugging is True, then display the keys that differ between the default and the context?

charleygros (approved):

ivadomed/config_manager.py
```diff
@@ -0,0 +1,52 @@
+import json
+import os
+
+__ivadomed_dir__ = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
Drulex, line 4: Why not `from ivadomed import __ivadomed_dir__`? Redefining `__ivadomed_dir__` could lead to potential problems if for example a file that defines it is moved somewhere else in the package at some point in time (e.g `config_manager` is refactored in `ivadomed/config/config_manager.py`). This variable is not supposed to change so there is not need to redefine it.

I think the most appropriate location to define `__ivadomed_dir__` in the `__init__.py` of the top-level package -> `ivadomed/ivadomed/__init__.py`. This file will never move and always represents the top-level package directory.

andreanne-lemay: That's what I did initially (I imported `__ivadomed_dir__` from ivadomed/utils.py), but I had problems with circular import (eb78051). I tried moving `__ivadomed_dir__` to ivadomed/ivadomed/__init__.py, but I currently have the same issue during testing (https://github.com/ivadomed/ivadomed/pull/484/checks?check_run_id=1309819496). @Drulex I am not sure how to fix this, do you have an idea how to fix this? Thank you for your help!

Drulex: Yes will look.

Drulex: Oh I hadn't noticed that `ivadomed/__init__.py` imports `utils.py` because of the git code and then `utils.py` also needs `__ivadomed_dir__`. It's not a big deal, we can just put `__ivadomed_dir__` in `utils.py` and expose it in `ivadomed/__init__.py` the same way as `__version__`.

The problem I see now is that `utils.py` is importing from `config_manager.py`. In fact it is importing from a lot of other files which is a bad idea considering that this is a utilities files (normally code should depend on this, not the other way around).

This is a bigger problem than moving `__ivadomed_dir__` around. I suggest we revert the recent commits and I'll create a separate issue for this.

andreanne-lemay: Good thank you for looking into this, I opened an issue here: #489

ivadomed/config_manager.py
andreanne-lemay: Updated here: 7e42580

ivadomed/config_manager.py
```diff
@@ -43,9 +43,17 @@ def get_config(self):
         Returns:
             dict: Updated configuration dict.
         """
+        if self.context['debugging']:
+            self._display_differing_keys()
         self.default_config.update(self.context)
+
         return self.default_config
 
+    def _display_differing_keys(self):
```
charleygros, line 93: great

charleygros (approved):

charleygros (approved):
All goodie

PR #486, opened 2020-10-23
Option to override postprocessing in segment volume

andreanne-lemay:
When using segment_volume with SCT (sct_deepseg), postprocessing steps can be chosen by command-line. This PR allows a postprocessing override if command-line postprocessing is chosen.

This PR is rooted on #484 . #484 should be merged before this one.

### DIFF
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -414,7 +414,7 @@ def save_mixup_sample(ofolder, input_data, labeled_data, lambda_tensor):
     plt.close()
 
 
-def segment_volume(folder_model, fname_image, fname_prior=None, gpu_number=0):
+def segment_volume(folder_model, fname_image, fname_prior=None, gpu_number=0, options=None):
     """Segment an image.
 
     Segment an image (`fname_image`) using a pre-trained model (`folder_model`). If provided, a region of interest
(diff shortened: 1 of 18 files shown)


ivadomed/utils.py
```diff
@@ -432,6 +432,7 @@ def segment_volume(folder_model, fname_image, fname_prior=None, gpu_number=0):
             e.g. spinal cord centerline, used to crop the image prior to segment it if provided.
             The segmentation is not performed on the slices that are empty in this image.
         gpu_number (int): Number representing gpu number if available.
+        options (dict):
```
charleygros, line 435: What do you think about adding `fname_prior` to `options`? It kind of makes sense to me as this file is quite option..al. But happy either ways.

ivadomed/utils.py
```diff
@@ -446,6 +447,21 @@ def segment_volume(folder_model, fname_image, fname_prior=None, gpu_number=0):
...
+        if 'largest' in options and options['largest']:
+            postpro['keep_largest'] = {}
+        if 'fill_holes' in options and options['fill_holes']:
+            postpro['fill_holes'] = {}
+        if 'remove_small' in options and ('mm' in options['remove_small'] or 'vox' in options['remove_small']):
+            unit = 'mm3' if 'mm3' in options['remove_small'] else 'vox'
+            thr = int(options['remove_small'].replace(unit, ""))
+            postpro['remove_small'] = {"unit": unit, "thr": thr}
+
+        context['postprocessing'] = postpro
```
charleygros, line 463: Zuper

charleygros (approved):

ivadomed/utils.py
andreanne-lemay: I agree! I integrated your suggestion here: 4fe58a0 and edf5d9b

Issue #462, opened 2020-10-01
Can't convert FiLM UNet to .onnx 


andreanne-lemay:
Warning while trying to save FiLM UNet as .onnx
![image](https://user-images.githubusercontent.com/49137243/94845282-77faf500-03ed-11eb-98e1-9b47dc069f2e.png)

PR #498, opened 2020-10-30
Training without test set

andreanne-lemay:
Some dataset, like the ones in medical decathlon (http://medicaldecathlon.com/) have a testing set without labels. The following changes allow training with only training and validation set, and allows to segment test images without the need of GT.

Changes include:
- Add option to train without test set.
- Add new command "segment" to simply perform segmentation on test set.


charleygros:
>     * Add option to train with test set.

What do you mean here? :-)

ivadomed/loader/utils.py
```diff
@@ -66,9 +66,16 @@ def split_dataset(df, center_test_lst, split_method, random_seed, train_frac=0.8
 
     elif split_method == 'per_patient':
         # Separate dataset in test, train and validation using sklearn function
-        X_train, X_remain = train_test_split(df['participant_id'].tolist(), train_size=train_frac,
-                                             random_state=random_seed)
-        X_test, X_val = train_test_split(X_remain, train_size=test_frac / (1 - train_frac), random_state=random_seed)
+        if test_frac == 1:
```
charleygros, line 69:
```suggestion
        # In case we want to use the entire dataset for testing purposes
        if test_frac == 1:
```

ivadomed/loader/utils.py
```diff
@@ -66,9 +66,16 @@ def split_dataset(df, center_test_lst, split_method, random_seed, train_frac=0.8
...
+            X_test = df['participant_id'].tolist()
+        else:
+            X_train, X_remain = train_test_split(df['participant_id'].tolist(), train_size=train_frac,
+                                                 random_state=random_seed)
+            if test_frac == 0:
```
charleygros, line 74:
```suggestion
            # In case the entire dataset is used to train / validate the model
            if test_frac == 0:
```

ivadomed/loader/utils.py
```diff
@@ -66,9 +66,16 @@ def split_dataset(df, center_test_lst, split_method, random_seed, train_frac=0.8
...
+                X_val = X_remain
+            else:
+                X_test, X_val = train_test_split(X_remain, train_size=test_frac / (1 - train_frac),
+                                                 random_state=random_seed)
```
charleygros, line 78: So.. if I understand well:
- you will have two distinct datasets: one for training / validation, another one for testing
- at training time, you will put `test_frac=0`
- at testing tome, `test_frac=1`

Another solution.. would be to generate a `joblib` file containing the customed split and use it with `fname_split`. So that no need to modify this param between the training / testing... dunno what's easier.

charleygros: Update: Mmmm I ve just seen the changes in `main.py`... Will think further about this split

ivadomed/main.py
```diff
@@ -286,6 +288,19 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
...
         return df_results, pred_metrics
 
+    if command == 'segment':
+        bids_ds = bids.BIDS(context["loader_parameters"]["bids_path"])
+        bids_subjects = [s for s in bids_ds.get_subjects() if s.record["subject_id"] in test_lst]
+        for subject in bids_subjects:
+            fname_img = subject.record["absolute_path"]
+            pred = imed_utils.segment_volume(os.path.join(context['log_directory'], context['model_name']),
+                                             fname_image=fname_img,
+                                             gpu_number=context['gpu'])
```
charleygros, line 298: We should add here the post-processing params like done in SCT?

charleygros (commented):
Great :-) love it! It s going to be super useful!
Lil' documentation of it once we reach a consensus? :3

ivadomed/main.py
andreanne-lemay: they can still be applied [through the config file ](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/utils.py#L577), but we might want to modify the config file in packaged model.

ivadomed/loader/utils.py
andreanne-lemay: Yes I agree the workflow could be re-thought, not super convenient

andreanne-lemay:
> > ```
> > * Add option to train with test set.
> > ```
> 
> What do you mean here? :-)

Oups *without* I updated the description, thank you!

ivadomed/loader/utils.py
andreanne-lemay: The case where we would want to use "segment" rather than "test" is probably only when we have unlabeled datasets (in a different BIDS folder than the labeled dataset). Maybe when "segment" is chosen, the whole dataset can be segmented, this way we avoid having to change test_frac at training and testing time.

ivadomed/main.py
charleygros: Oh! Yup! I was "we need to feed `options` for the postpro" but in this application we will use the postpro params defined in the config file! All good :-)

ivadomed/loader/utils.py
charleygros: > Maybe when "segment" is chosen, the whole dataset can be segmented, this way we avoid having to change test_frac at training and testing time.

Agree!

ivadomed/main.py
```diff
@@ -292,9 +292,18 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
...
         subj_lst = bids_ds.participants.content['participant_id'].tolist()
         bids_subjects = [s for s in bids_ds.get_subjects() if s.record["subject_id"] in subj_lst]
+
+        # Add postprocessing to packaged model
+        path_model = os.path.join(context['log_directory'], context['model_name'])
+        path_model_config = os.path.join(path_model, context['model_name'] + ".json")
+        model_config = imed_config_manager.load_json(path_model_config)
+        model_config['postprocessing'] = context['postprocessing']
+        with open(path_model_config, 'w') as fp:
+            json.dump(model_config, fp, indent=4)
```
charleygros, line 302: So if I understand correctly here: it allows the user to `segment` with postpro params that are potentially different from the ones defined in the initial config file (ie the one used to train the model). Is it correct?

docs/source/configuration_file.rst
```diff
@@ -24,8 +24,8 @@ General parameters
 command
 ^^^^^^^
 
-Run the specified command. Choices: ``"train"`` or ``"test"``,
-to train and evaluate a model respectively.
+Run the specified command. Choices: ``"train"``, ``"test"`` or ``segment``,
+to train, evaluate a model, or segment a dataset respectively.
```
charleygros, line 28: What about something like:
```
Run the specified command. Choices: ``"train"``: train a model on a training/validation sub-datasets, ``"test"``: evaluate a trained model on a testing sub-dataset, ``"segment"``: segment a entire dataset using a trained model.
```

ivadomed/main.py
andreanne-lemay: Yes, exact: the postprocessing applied is the one in the current config file. During training, postprocessing steps are not applied, so it is likely at inference the user might want to modify these steps.  Would we want to resave the config file in the packaged model to update de postprocessing steps?

docs/source/configuration_file.rst
andreanne-lemay: Updated here: f127378

charleygros: Great, thanks!

ivadomed/main.py
charleygros: > Would we want to resave the config file in the packaged model to update de postprocessing steps?

Do you mean: update the config file in the packaged model with the postprocessing params input by the user? If so: no I don't think so: the user may have different needs than the packaged model and it could create confusion. I would leave it as you implemented. All good to me!

ivadomed/utils.py
```diff
@@ -449,7 +449,7 @@ def segment_volume(folder_model, fname_image, gpu_number=0, options=None):
     # Load model training config
     context = imed_config_manager.ConfigurationManager(fname_model_metadata).get_config()
 
-    if any(pp in options for pp in ['thr', 'largest', ' fill_holes', 'remove_small']):
+    if options is not None and any(pp in options for pp in ['thr', 'largest', ' fill_holes', 'remove_small']):
```
charleygros, line 452: @andreanne-lemay : So.. `['thr', 'largest', ' fill_holes', 'remove_small']` are the param names from the SCT API and in the herebelow lines you are doing a matching between the ivadomed param names (ie in the config file) and these param names. Nothing crucial: but I would suggest to make this matching in SCT instead, so that these param names are consistent through `ivadomed`: what do you think? :-)

charleygros: The other user case of this function is when using `command=segment` but in that case we are using a config file so the param names are matching well (thanks to our new Config manager <3 ).

ivadomed/utils.py
```diff
@@ -536,7 +554,13 @@ def segment_volume(folder_model, fname_image, gpu_number=0, options=None):
     for i_batch, batch in enumerate(data_loader):
         with torch.no_grad():
             img = cuda(batch['input'], cuda_available=cuda_available)
-            preds = model(img) if fname_model.endswith('.pt') else onnx_inference(fname_model, img)
+
+            if ('FiLMedUnet' in context and context['FiLMedUnet']['applied']) or \
+                    ('HeMISUnet' in context and context['HeMISUnet']['applied']):
+                metadata = imed_training.get_metadata(batch["input_metadata"], model_params)
+                preds = model(img, metadata)
```
charleygros, line 561: No need to `onnx` because we are always going with `.pt` for FiLM models? [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/training.py#L340)

ivadomed/utils.py
andreanne-lemay: If I understand well you are suggesting to rename 'thr' and 'largest'  to 'binarize_prediction' and 'keep_largest'  in SCT so all names are consistent (in SCT and ivadomed)? If so, I agree with this change. I opened an issue to fix this inconsistency in another PR: https://github.com/ivadomed/ivadomed/issues/506

ivadomed/utils.py
andreanne-lemay: Yes currently there is an issue: FiLM models can't be saved in onnx format so it can only be saved in .pt  see https://github.com/ivadomed/ivadomed/issues/462

ivadomed/utils.py
jcohenadad: to be consistent with 'remove_small' maybe we should use the terminology 'keep_large' instead of 'keep_largest'?

also, i like 'thr' in that it is intuitive that a non-binary answer is expected (ie: the threshold), whereas when i see "binarize_prediction" I am tempted to assign a binary response (true/false). How about 'threshold_prediction'?

charleygros: Yes, I meant that it would be great if the names are consistent in SCT and ivadomed. So zuper for #506.
But.. it may not be always possible (for X reason, for instance with `thr`) and in that case I was suggesting to move at these lines do (ie match SCT_keyname and ivadomed_keyname) in the SCT function instead of in this ivadomed function (happy to re explain via phone :-) ).

Re. the specific case of `thr` vs `binarize_prediction` vs `threshold_prediction`: I am quite happy any ways: Just wanted to bring on the table the following info: this threshold in `ivadomed` currently refers to a threshold on the soft predictions. We are also developing (and soon into production?) a postprocessing which thresholds the soft predictions using the uncertainty map --> just to keep in mind when naming `thr` as we may have soon another threshold which uses the uncertainty map.

charleygros (approved):

PR #500, opened 2020-10-30
set h5py version in requirements.txt

### LINKED ISSUE #499: dependency issue: h5py version 3.0.0 cause test crash
a new release from h5py was done today ( v 3.0.0 , 30/10/2020). It causes the test to crash it needs to be investigated (perhaps during hdf5 refactoring) for now we could fix the version to 2.9.

lrouhier:
New release from h5py (v3.0.0) cause some crash in the adaptative loader (hemis). 
This Pr fixed the package at version 2.10.0 to avoid this issue. 

fix #499

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 bids-neuropoly>=0.2
-h5py
+h5py==2.10.0
 joblib
 matplotlib>=3.3.0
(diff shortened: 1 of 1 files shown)


andreanne-lemay (approved):
:100:

PR #503, opened 2020-11-04
Create sample to balance metadata

andreanne-lemay:
Currently, there is a data sampler to balance batches containing the class of interest. This PR adds a sampler to balance batches according to a given metadata.

```balance_samples: false ``` becomes 
```balance_samples: {"applied": true, "type": disease}``` (to balance according to the disease of the subject)

PRs #498 and #491 are merged in this branch

### CODE REVIEW

File: ivadomed/loader/utils.py [hunk 1]
```diff
@@ -365,12 +369,20 @@ def _get_label(dataset, idx):
         Returns:
             int: 0 or 1.
         """
-        # For now, only supported with single label
-        sample_gt = np.array(dataset[idx]['gt'][0])
-        if np.any(sample_gt):
-            return 1
+        if metadata != 'gt':
```
charleygros, line 372: What about adding above this: two lines that check if `metadata` is either `gt` either in `dataset[idx]['input_metadata'][0].keys()`, otherwise throw an error?

testing/test_script.py
```diff
@@ -364,6 +365,9 @@ def test_film_contrast():
     with open(film_config, "r") as fhandle:
         context = json.load(fhandle)
     # Modify params
+    context["training_parameters"]["balance_samples"] = {}
+    context["training_parameters"]["balance_samples"]["applied"] = True
+    context["training_parameters"]["balance_samples"]["type"] = balance_type
     context["loader_parameters"]["contrast_params"]["training_validation"] = ["T2w", "T1w", "T2star"]
     context["FiLMedUnet"]["applied"] = True
     context["FiLMedUnet"]["film_layers"] = 2 * [1] * context["default_model"]["depth"] + [0, 0]
```
charleygros, line 373: Great!

charleygros (approved):
Smooth and easy! Many thanks!
Would you simply mind adding a line about this new feature in the RTD/config_file.. and then ready to go :-)

ivadomed/loader/utils.py
andreanne-lemay: > Smooth and easy! Many thanks!
> Would you simply mind adding a line about this new feature in the RTD/config_file.. and then ready to go :-)

Updated here: c43e47b

andreanne-lemay: I added a metadata check in the loader/loader.py : 9096673. Doing the check directly in the BalancedSampler was useless since if the metadata is not 'gt' or a column from participants.tsv an error was already thrown in loader/loader.py, so I added a custom error :}

PR #509, opened 2020-11-06
Some fixes to logging

### LINKED ISSUE #508: Wrong use of logging module
Root cause of: https://github.com/neuropoly/spinalcordtoolbox/issues/2993

API modules should not configure the root logger, instead they should simply define a module-scoped logger and use it:
```python
logger = logging.getLogger(__name__)
logger.info("something...")
```

It is always the caller that should configure root logger based on its needs.

This gets the root logger instead of the module scope logger.
https://github.com/ivadomed/ivadomed/blob/3c9d32c41ef4222848f0dfb024e89c11f7264046/ivadomed/utils.py#L17
It should be `logger = logging.getLogger(__name__)`

This is used to configure the root logger, not the module scoped logger
https://github.com/ivadomed/ivadomed/blob/3c9d32c41ef4222848f0dfb024e89c11f7264046/ivadomed/utils.py#L18

`logging.basicConfig` configures the root logger. `logger.setLevel()` can be used to configure the module-scoped logger (assuming it is defined correctly)

Drulex:
- Don't configure root logger from API modules, let the caller handle that.
- Fix one instance of logger getting root logger instead of module logger

Closes #508 

Fix for https://github.com/neuropoly/spinalcordtoolbox/issues/2993

### DIFF
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -25,7 +25,6 @@ __numpy_type_map = {
 
 TRANSFORM_PARAMS = ['elastic', 'rotation', 'scale', 'offset', 'crop_params', 'reverse', 'translation', 'gaussian_noise']
 
-logging.basicConfig(level=logging.INFO, format='%(message)s')
 logger = logging.getLogger(__name__)
 
 
--- a/ivadomed/training.py
+++ b/ivadomed/training.py
@@ -22,7 +22,6 @@ from ivadomed import visualize as imed_visualize
 from ivadomed.loader import utils as imed_loader_utils
 
 cudnn.benchmark = True
-logging.basicConfig(level=logging.INFO, format='%(message)s')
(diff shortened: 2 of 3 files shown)


jcohenadad:
@Drulex how do you conveniently test this PR within SCT? Do you modify a local copy of the requirements to point ivadomed to a specific git commit and then reinstall SCT's dependencies, or is there a faster way to do it?

Drulex:
> @Drulex how do you conveniently test this PR within SCT? Do you modify a local copy of the requirements to point ivadomed to a specific git commit and then reinstall SCT's dependencies, or is there a faster way to do it?

From within the `venv` navigate to `ivadomed` and `pip install -e .` It installs `ivadomed` in the virtual environment from your filesystem`

jcohenadad (approved):
working like a charm 🎉

PR #517, opened 2020-11-10
paper.md: Fixed broken link

jcohenadad:
i had to use "latest" instead of "stable" for the RTD build because there is no release (yet) with the latest change (section: "use cases")


andreanne-lemay (commented):
:+1:

andreanne-lemay (approved):

PR #522, opened 2020-11-11
Support multiclass if first class missing

andreanne-lemay:
The current code throws an error if the first target class defined in the config file doesn't exist for a given subject. This PR aims to fix this.


ivadomed/evaluation.py
```diff
@@ -152,7 +152,7 @@ def __init__(self, data_pred, data_gt, dim_lst, params=None):
                                             ['gt'] * len(label_gt_size_lst) + ['pred'] * len(label_pred_size_lst)])
 
         else:
-            self.label_size_lst = [[[], []] * self.n_classes]
+            self.label_size_lst = [[[], []]] * self.n_classes
```
charleygros, line 155: eheh

ivadomed/loader/loader.py
```diff
@@ -267,8 +267,13 @@ def get_pair_metadata(self, slice_index=0, coord=None):
                     "data_type": 'gt',
                     "crop_params": {}
                 }))
-            elif len(gt_meta_dict):
-                gt_meta_dict.append(gt_meta_dict[0])
+            else:
+                gt_meta_dict.append(None)
```
charleygros, line 271: Could we add a comment here to explain why we are doing this trick? :-)

ivadomed/testing.py
```diff
@@ -170,14 +170,14 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
-                fname_ref = metadata_idx[0]['gt_filenames'][0]
+                fname_ref = list(filter(None, metadata_idx[0]['gt_filenames']))[0]
 
                 # NEW COMPLETE VOLUME
                 if pred_tmp_lst and (fname_ref != fname_tmp or last_sample_bool) and task != "classification":
                     # save the completely processed file as a nifti file
                     if ofolder:
                         fname_pred = os.path.join(ofolder, fname_tmp.split('/')[-1])
-                        fname_pred = fname_pred.rsplit(testing_params['target_suffix'][0],1)[0] + '_pred.nii.gz'
+                        fname_pred = fname_pred.rsplit("_", 1)[0] + '_pred.nii.gz'
```
charleygros, line 180: In which case `target_suffix` was incorrect?
I am not sure if `.rsplit("_", 1)[0]` will always be robust :/

andreanne-lemay: we take the first target_suffix (`target_suffix[0]), but it might be the second target suffix in fname_pred, if the first target doesn't exist. If the folder respects BIDS conventions, this line will work (remove target_suffix et replace by _pred.nii.gz). I think we can assume the files respect this convention.

charleygros: I was confused with `rsplit` behaviour, made some tests: you are right, all good :-)

ivadomed/loader/loader.py
andreanne-lemay: Added here :-) b840602

charleygros (approved):

PR #524, opened 2020-11-12
Multiclass and multichannel support for segment volume

andreanne-lemay:
This PR aims at supporting the segmentation with `segment_volume` of multichannel models (e.i., allow more than one image as input) and multiclass models (e.i., generate as much output files as the number of classes).


ivadomed/inference.py
```diff
@@ -278,7 +280,19 @@ def segment_volume(folder_model, fname_image, gpu_number=0, options=None):
                                        bin_thr=-1,
                                        postprocessing=context['postprocessing'])
 
-    return pred_nib
+                pred_list = split_classes(pred_nib)
+                target_list = context['loader_parameters']['target_suffix']
+
+    return pred_list, target_list
```
charleygros, line 286: Add `target_list` in the docstring? :-)

ivadomed/inference.py
```diff
@@ -278,7 +280,19 @@ def segment_volume(folder_model, fname_image, gpu_number=0, options=None):
...
+
+
+def split_classes(nib_prediction):
+    pred = nib_prediction.get_fdata()
```
charleygros, line 290:
```suggestion
    """Split a 4D nibabel multi-class segmentation file in multiple 3D nibabel binary segmentation files.
    
    Args:
        nib_prediction (nibabelObject): 4D nibabel object.
    Returns:
        list of nibabelObject.
     """
    pred = nib_prediction.get_fdata()
```

ivadomed/inference.py
```diff
@@ -278,7 +280,19 @@ def segment_volume(folder_model, fname_image, gpu_number=0, options=None):
...
+    pred_list = []
+    for c in range(pred.shape[-1]):
+        class_pred = nib.Nifti1Image(pred[..., c].astype('float32'), nib_prediction.affine)
```
charleygros, line 293: Is `float32` needed / desired here?
Also: I am not sure: are the nibabel files saved this way proper 3D objects? or 4D with empty last channel?
Happy to check, let me know :-)

ivadomed/main.py
```diff
@@ -318,20 +318,34 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
...
-            fname_img = subject.record["absolute_path"]
+            if context['loader_parameters']['multichannel']:
+                fname_img = []
+                contrasts = context['loader_parameters']['contrast_params']['testing']
+                # Keep contrast order
+                for c in contrasts:
+                    for s in bids_subjects:
+                        if subject.record['subject_id'] == s.record['subject_id'] and s.record['modality'] == c:
+                            fname_img.append(s.record['absolute_path'])
+                            bids_subjects.remove(s)
```
charleygros, line 329: Do you think we should check here if a given subject has all requested contrasts? To avoid adding a subject with missing contrast. It would be detected further down the line I would say but.. the sooner the better?

ivadomed/inference.py
andreanne-lemay: I added this to comply to this test: https://github.com/ivadomed/ivadomed/blob/master/testing/test_segment_volume.py#L139
Or else it is in float64 which I believe is overkill.

3D images can directly be saved with nibabel :), but a 4D with the last dimension to identify channels works as well.

ivadomed/main.py
andreanne-lemay: Yes good idea, I'll add it :)

ivadomed/inference.py
charleygros: Sick!

ivadomed/inference.py
andreanne-lemay: Added here :-) 4d9b7c8

ivadomed/main.py
andreanne-lemay: Added it here : 6ecb85d

ivadomed/inference.py
andreanne-lemay: Updated here :-) 4d9b7c8

charleygros (approved):

PR #533, opened 2020-11-20
automate_training: add new parameter to change multiple params

andreanne-lemay:
Two options are currently available to change hyperparameters with automate_training:

default: One parameter will be changed one at the time from parameters specified in the hyperparameters file (specified by `-p`)
`--all-combin`: Will train on all combinations of parameters specified in the hyperparameters file (specified by `-p`)

This PR adds a new option:
`-m` or `--multi-params`: If True, more than one parameter will be change at the time from the hyperparameters file. All the first elements from the hyperparameters list will be applied, then all the second, etc.


lrouhier (approved):
looks good !

PR #534, opened 2020-11-20
Select subjects for training based on metadata

andreanne-lemay:
New parameter added to `loader_parameters` to select subjects for training based on metadata.

"subject_selection": {"n": [10, 5], "metadata": ["task", "task"], "value": ["spleen", "kidney"]},

The following would isolate 10 subjects with the "task" "spleen" and 5 subjects with the "task" "kidney" for training, validation and testing. This avoids having to create manually subdatasets.


ivadomed/loader/utils.py
```diff
@@ -107,12 +108,24 @@ def get_new_subject_split(path_folder, center_test, split_method, random_seed,
...
 
     Returns:
         list, list list: Training, validation and testing subjects lists.
     """
     # read participants.tsv as pandas dataframe
     df = bids.BIDS(path_folder).participants.content
+    if subject_selection is not None:
+        # Verify subject_selection format
+        if not (len(subject_selection["metadata"]) == len(subject_selection["n"]) == len(subject_selection["value"])):
+            raise ValueError("All lists in subject_selection parameter should have the same length.")
```
charleygros, line 121: Should we add a test about the presence of the correct keys in the dict? ie metadata, n , value
To be done here or in the config_manager?

charleygros (commented):
Looks all good to me. Documentation in the RTD with example would be appreciated :-)

ivadomed/loader/utils.py
andreanne-lemay: The keys are added in the config_manager: `"subject_selection": {"n": [], "metadata": [], "value": []}` when keys are missing :3

andreanne-lemay: Documentation added here :) 8a8bd78

charleygros (approved):

Issue #512, opened 2020-11-09
Retrocompatibility - Models and is_2d attribute


charleygros:
Retro-compatibility issue when using an old `ivadomed` model, eg [exivvo_t2star](https://github.com/ivadomed/sc-gm_t2star_exvivo).

Following error when using `master`:
```
AttributeError: 'UpConv' object has no attribute 'is_2d'
```
occuring at [this line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/models.py#L285).

Reason: at the time of this model training: the attribute `is_2d` did not exist, which triggers the above error when calling `forward()`.

Wondering if there is a way to avoid this.
If not... Is there an alternative to "run the model with the `ivadomed` version it has been generated with".
If not.. we really need to enforce the `ivadomed` version within the model json file. Otherwise it would be very messy to dive into history and found out this information.

andreanne-lemay:
we could add something like: if not hasattribute(self, is_2d): self.is_2d = True, this way it becomes retrocompatible

PR #535, opened 2020-11-23
Make is_2d retrocompatibility

andreanne-lemay:
See issue #512 

This PR also corrects a bug in testing.py for 3D


lrouhier (approved):
Looks good !

PR #537, opened 2020-11-24
Split_dataset: consider center_list when per_patient is used

andreanne-lemay:
The splitting method `per_center` overwrites the use of `center_test`. This PR allows using the `per_patient` splitting method with the parameter `center_test`.


olix86 (approved):
👍

Issue #525, opened 2020-11-13
Providing support for EEG modality in IVADOMED


mpompolas:
@jcohenadad 
This is to start the conversation for opening up the toolbox to EEG.
As mentioned to Guillaume (GitHub handle?) on Slack, I imagine the ivadomed toolbox to be accessed at a later stage of the analysis pipeline, since in electrophysiology there are multiple pre-processing steps needed for making comparisons across conditions.

I attach a diagram to make this a bit more clear. Ivadomed comes into play after the last step:
![image](https://user-images.githubusercontent.com/23224563/99029630-f5636a80-2540-11eb-8a29-eb8426773e26.png)

Assuming that we want to perform classification among conditions, the easiest way would be to create an importer for different types of analysis features (for our purposes we can just think of them just as 2D or 3D matrices). 
For example, to compare conditions in:
1. The time domain (nTrials x Channels x TimeSamples)
2. The frequency domain (nTrials x Channels x FFTbins)
3. Time-frequency decompositions (nTrials x Channels x FFTbins x TimeSamples)

Any suggestions are welcome

jcohenadad:
My first reaction would be: can the described pre-processing pipeline in https://github.com/ivadomed/ivadomed/issues/525#issue-742138726 follow the BIDS convention? I.e. processing outputs would be placed under the derivatives folder, and be used as input in ivadomed. If we go with a BIDS folder as input, this will minimize the amount of refactoring needed on the `ivadomed` side.

So, if we can "bidsify" the 2D/3D matrices mentioned in https://github.com/ivadomed/ivadomed/issues/525#issue-742138726, that would be great.

mpompolas:
Yes, I agree. That would be the easiest first step.
I'll reach out to Guillaume to get an example dataset to create the BIDS folders based on the way he stores his metadata.

If we are satisfied with the way these example files handle metadata we can use them as a guide for whoever wants to use the software.
If not, we do one more iteration on this.

Regarding applications, classification is a standard approach (comparison of patients/controls for example).
What would be also very interesting (I'm not sure how we can do this yet with ivadomed - I only used statistics for that), would be to also find the exact time or/and frequency that differentiates conditions

mpompolas:
I made a BIDSified version for mne epochs:
![image](https://user-images.githubusercontent.com/23224563/100142537-0cb82700-2e62-11eb-93c9-1b0221d64ba5.png)


This is on my computer right now. Will create a repo

[...]

Issue #474, opened 2020-10-14
Migrate to HDF5 loader


andreanne-lemay:
ivadomed's adaptive loader currently uses HDF5 to load data and parse metadata. HDF5 allows to load data in RAM or "on the fly". Do we want to migrate ivadomed's loader to HDF5?

charleygros:
After a brief look at `hdf5` documentation and Anthime's code.. I think my preference would be to: (1) generate this pandas dataframe with filenames and images infos by checking data / derivatives availability / suitability (2) split (3) load data as numpy (4) perform preprocessing (5) save preprocessed data etc as HDF5.

charleygros:
A Con of HDF5, I ve just thought about could be the inference on a single image (ie `segment_volume`) --> do we want to systematically move to HDF5 even for this case (ie when used by SCT / ADS) --> it sounds like an "overkill" to me (but maybe it's not?). --> Should we keep the option not to convert data to HDF5 for this case?

jcohenadad:
> After a brief look at hdf5 documentation and Anthime's code.. I think my preference would be to: (1) generate this pandas dataframe with filenames and images infos by checking data / derivatives availability / suitability (2) split (3) load data as numpy (4) perform preprocessing (5) save preprocessed data etc as HDF5.
> A Con of HDF5, I ve just thought about could be the inference on a single image (ie segment_volume) --> do we want to systematically move to HDF5 even for this case (ie when used by SCT / ADS) --> it sounds like an "overkill" to me (but maybe it's not?). --> Should we keep the option not to convert data to HDF5 for this case?

wouldn't it be easy to _not_ have the option to save (ie bypass step 5) and simply use the variables already in RAM to perform prediction?

charleygros:
> wouldn't it be easy to _not_ have the option to save (ie bypass step 5) and simply use the variables already in RAM to perform prediction?

Good idea!!

andreanne-lemay:
After digging into it, I agree with Charley's suggestion.

jcohenadad:
I'm not sure how relevant this is, but I've just came across [this blog against HDF5](https://cyrille.rossant.net/moving-away-hdf5/). Food for thoughts...

andreanne-lemay:
Very interesting post! My thoughts on this:
[...]

Issue #304, opened 2020-06-15
Adapt the loader for histology data


mariehbourget:
- Use BIDS (following link)?
- Alternatives:
-> Loader that accept multiple data structures
-> Task specific loader
- Deal with dataset splits and patches and dataset splits (ADS currently creates new files on disk when splitting and extracting patches)
- Deal with large histology images (see OpenSlide) ex:
-> TEM - 3762 × 2286 - 6.6 Mb
-> SEM - 1901 × 1801 - 2.4 Mb

mariehbourget:
I ran some preliminary tests (Unet2D, 1 contrast, 1 label, Center Crop) with the SEM example dataset in PNG format.
This is a breakdown of the 4 main issues I found so far. Specific smaller issues and PR will be open to address them separately.

1. Sample filename and split

- BIDS microscopy uses filename `sub-label_sample-label_modality.png` with multiple samples per subject.
- BIDS microscopy has an additional file `samples.tsv` to list unique samples ID with participants ID.
- ivadomed splits per_patient (microscopy would need a split per sample).
- ivadomed uses `bids_neuropoly` to fetch subjects and filenames.
-- The subject lists are based on unique participants ID.
-- The parsing of the filenames returns subject_id.

2. File format

- ivadomed and bids_neuropoly expects Nifti files (extension nii and nii.gz).
- ivadomed uses nibabel library and functions to load files and metadata from header.
- Microscopy files will most likely be PNG or OME-TIFF format (not confirmed).

3. 3D vs. 2D

- Some for loops failed with a single 2D slice

4. Unknown issue with metadata when applying preprocessing transforms (still to investigate)

charleygros:
>     * BIDS microscopy uses filename `sub-label_sample-label_modality.png` with multiple samples per subject.

Would it be possible to drop here a screenshot (ie organisation & hierarchy) of one of your BIDS dataset? so that we could be more familiar with it.

>     * ivadomed uses `bids_neuropoly` to fetch subjects and filenames.

I would be happy to absorb the code of `bids_neuropoly` within `ivadomed`. And as you suggested, we should pull a pair (filename_image, filename_gt) instead of `Participant-ID`. Or.. alternatively.. to adjust the code for your needs.. could we pull `ParticipantYY_sampleXX` instead of `ParticipantYY`.

>     1. File format

Yeah .. I see! thanks for explaining this. So it will impact the `nib.load` in the loader and the `pred_to_nii` current functions.

[...]

Issue #532, opened 2020-11-20
Empty .json file within /duke/sct_testing/large


mpompolas:
duke/sct_testing/large/sub-spineGeneric011/anat/sub-spineGeneric011_T2star.json

Ivadomed training on the /large dataset failed.

Whoever received the original data should make a check that this .json was not altered at a later copying phase.

If the original .json is indeed empty we should either try to populate it (harder solution - probably just check the rest of the json files from the same center and generalize) or just remove from the dataset the *T2star.json and the corresponding NIFTI (easy and lazy solution) so the training can run.

andreanne-lemay:
The problem originates from the bids library. To avoid having to change your dataset, in the class `Subject`, method `metadata`, the code can be modified to this (bids.py line 87):
```
    def metadata(self):
        if not self.has_metadata():
            raise RuntimeError("This subject does not have a metadata file.")
        metadata_path = self.__get_metadata_path()
        with open(metadata_path, "r") as fp:
            try:
                json_content = json.load(fp)
            except:
                print("{} not found".format(metadata_path))
                json_content = {}
        return json_content
```
rather than this:

```
    def metadata(self):
        if not self.has_metadata():
            raise RuntimeError("This subject does not have a metadata file.")
        metadata_path = self.__get_metadata_path()
        with open(metadata_path, "r") as fp:
            json_content = json.load(fp)
        return json_content
``` 

We plan to stop depending on this library in the future (during the loader change).

mpompolas:
so the .json are not used anywhere in the software?
No need to track down the empty ones?

charleygros:
> so the .json are not used anywhere in the software?
> No need to track down the empty ones?

The `.json` are used during the loading, to access metadata.
Empty `json` eg for derivatives are fine. But currently requested for an image to be included in the process.
Note: the change suggested by Andréanne [here](https://github.com/ivadomed/ivadomed/issues/532#issuecomment-731280877) has not been implemented yet.
The missing `json` files should be tracked and added.

PR #538, opened 2020-11-24
New loader: generate dataframe

### LINKED ISSUE #519: Generate dataframe for new loader
## Context

Following this [comment](https://github.com/ivadomed/ivadomed/issues/474#issuecomment-708688892), the first step is to generate a dataframe with filenames and images infos by checking data/derivatives availability and associated metadata.

It also aims to remove the dependency to bids_neuropoly.

## Steps to generate dataframe

1. Get a list of all files in BIDS folders starting with `sub*` and ending with specified extensions, including derivatives (ignore json or other files). Questions:
    * Should the extensions be specified as: all extensions compliant with BIDS for supported data_type (anat and microscopy) or extensions specified by user in the config file?
    * Should derivatives include: all files or only files with specified `seg-<label>` (target_suffix) in config file?

2. Parse all entities in filename (`sub`, `ses`, `sample`, `run`, `ce`, `down`, `acq`, `rec`) for supported modalities + `suffix` for modality + `seg-<label>` for derivatives

3. Create a dataframe with 1 line for each file with: `filename`, `all entities’ labels`, `modality`, `data_type`, `filetype`, `seg-<label>`, `absolute_path`, `dataset_root`

4. Read TSV files if present (`participants`, `samples`, `sessions`, `scans`), add columns to dataframe.

5. Read JSON sidecar files, add columns to dataframe.

6. Read appropriate metadata directly from file, depending on the filetype, add columns to dataframe.

## Other considerations

Somewhere, we should check derivatives availability. Currently, the check is done after the split when loading the dataset, both in traditional and adaptative loader. What is best?

We might want to take a look at or use [pybids](https://github.com/bids-standard/pybids) which is a package maintained by BIDS to interact with BIDS datsets.

## Next step

Once the dataframe is done, the next step would be the split which could be done by filtering one or several columns of the dataframe.

mariehbourget:
Fixes #519 
It may also include changes to support EEG modality (#525).

It aims to generate a dataframe by indexing filenames and metadata in a bids path.

Changes have been tested with `anat` and `microscopy` datasets.

The changes are:
1 new function `create_bids_dataframe` in `ivadomed/loader/utils.py`.
1 new requirement: `pybids>=0.12.4`
3 new files:
`ivadomed/config/config_bids.json` : custom bids_config file needed for dealing with additional microscopy entities in `pybids`.
`ivadomed/config/config_new_loader.json` : ivadomed config file with 2 new `loader parameters`:  `bids_config` and `extensions`.
`dev/df_new_loader.py` : Script used to test the new `create_bids_dataframe` function.

I left a lot of comments in the code to explain the different steps and choices made.

I was able to keep the validation (by the BIDS-validator) by forcing the indexing of files in `sub-*` folders at the root of the bids path and of files specific to microscopy (`samples.tsv`, `samples.json`).

To discuss:
- Do we want to index derivatives? (depends on where and when we want to check for availability and do the dataset split)
- Currently, the files are indexed and filtered by `extensions` and `target_suffix`. We could also filter by `data_type` (anat, microscopy, etc) and/or modality suffix (ex: T1w, SEM, etc) in case multiple data type are available in a dataset.

TODO:
- Include parsing of `sessions.tsv` and `scans.tsv` files (not done at the moment, only `participants.tsv` and `samples.tsv`)
- Read appropriate metadata from image file and add to dataframe. Is it pertinent at this stage? Could also be done after the split when loading images to numpy arrays.
- Evaluate changes to make to include EEG.

### DIFF
--- a/dev/df_new_loader.py
+++ b/dev/df_new_loader.py
@@ -25,7 +25,6 @@ loader_params = context["loader_parameters"]
 loader_params["contrast_params"]["contrast_lst"] = loader_params["contrast_params"]["training_validation"]
 
 # CHOOSE TO INDEX DERIVATIVES OR NOT
-# To discuss: depending on how derivatives availibility is checked and split is done afterwards.
 # As per pybids, the indexing of derivatives works only if a "dataset_description.json" file
 # is present in "derivatives" or "labels" folder with minimal content:
 # {"Name": "Example dataset", "BIDSVersion": "1.0.2", "PipelineDescription": {"Name": "Example pipeline"}}
@@ -35,6 +34,6 @@ derivatives = True
 df = imed_loader_utils.create_bids_dataframe(loader_params, derivatives)
 
 # SAVE DATAFRAME TO CSV FILE
-path_csv = "test.csv"
+path_csv = "test_df_new_loader.csv"
 df.to_csv(path_csv)
 print(df)
--- a/dev/df_new_loader.py
+++ b/dev/df_new_loader.py
@@ -22,11 +22,13 @@ from ivadomed.loader import utils as imed_loader_utils
 path_config_file = "ivadomed/config/config_new_loader.json"
 context = imed_config_manager.ConfigurationManager(path_config_file).get_config()
 loader_params = context["loader_parameters"]
+loader_params["contrast_params"]["contrast_lst"] = loader_params["contrast_params"]["training_validation"]
 
 # CHOOSE TO INDEX DERIVATIVES OR NOT
 # To discuss: depending on how derivatives availibility is checked and split is done afterwards.
 # As per pybids, the indexing of derivatives works only if a "dataset_description.json" file
-# is present in "derivatives" or "labels" folder.
+# is present in "derivatives" or "labels" folder with minimal content:
+# {"Name": "Example dataset", "BIDSVersion": "1.0.2", "PipelineDescription": {"Name": "Example pipeline"}}
 derivatives = True
 
 # CREATE DATAFRAME
--- a/dev/df_new_loader.py
+++ b/dev/df_new_loader.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+##############################################################
+#
+# This script is used to test the dataframe of the new loader
+#
+# Usage: python dev/df_new_loader.py
+#
+##############################################################
+
+# IMPORTS
+import os
+import pandas as pd
+from ivadomed import config_manager as imed_config_manager
+from ivadomed.loader import utils as imed_loader_utils
+import bids as pybids
+
+# GET LOADER PARAMETERS FROM IVADOMED CONFIG FILE
+# The loader parameters have 2 new fields: "bids_config" and "extensions".
+# "bids_config" is mandatory for microscopy until BEP is merged and pybids is updated, the file is
+# in ivadomed/config/config_bids.json.
+# "bids_config" is optional for anat, can be the same as microscopy or set to null.
+# "extensions" is used to filter which files are to be indexed, in case multiple file types are present.
+path_config_file = "/home/mhbourget/ivadomed/ivadomed/config/config_new_loader.json"
+context = imed_config_manager.ConfigurationManager(path_config_file).get_config()
+loader_params = context["loader_parameters"]
+
+# CHOOSE TO INDEX DERIVATIVES OR NOT
+# To discuss: depending on how derivatives availibility is checked and split is done afterwards.
+# As per pybids, the indexing of derivatives works only if a "dataset_description.json" file
+# is present in "derivatives" or "labels" folder.
+derivatives = True
+
+# CREATE DATAFRAME
+df = imed_loader_utils.create_bids_dataframe(loader_params, derivatives)
+
+# SAVE DATAFRAME TO CSV FILE
+path_csv = "./test.csv"
+df.to_csv(path_csv)
--- a/dev/df_new_loader.py
+++ b/dev/df_new_loader.py
@@ -12,15 +12,14 @@ import os
 import pandas as pd
 from ivadomed import config_manager as imed_config_manager
 from ivadomed.loader import utils as imed_loader_utils
-import bids as pybids
 
 # GET LOADER PARAMETERS FROM IVADOMED CONFIG FILE
 # The loader parameters have 2 new fields: "bids_config" and "extensions".
 # "bids_config" is mandatory for microscopy until BEP is merged and pybids is updated, the file is
 # in ivadomed/config/config_bids.json.
-# "bids_config" is optional for anat, can be the same as microscopy or set to null.
+# "bids_config" is optional for anat
 # "extensions" is used to filter which files are to be indexed, in case multiple file types are present.
-path_config_file = "/home/mhbourget/ivadomed/ivadomed/config/config_new_loader.json"
+path_config_file = "ivadomed/config/config_new_loader.json"
 context = imed_config_manager.ConfigurationManager(path_config_file).get_config()
 loader_params = context["loader_parameters"]
 
@@ -34,5 +33,6 @@ derivatives = True
 df = imed_loader_utils.create_bids_dataframe(loader_params, derivatives)
 
 # SAVE DATAFRAME TO CSV FILE
-path_csv = "./test.csv"
+path_csv = "test.csv"
 df.to_csv(path_csv)
+print(df)
--- a/ivadomed/config/config_new_loader.json
+++ b/ivadomed/config/config_new_loader.json
@@ -19,7 +19,7 @@
             "slice_filter_roi": null
         },
         "contrast_params": {
-            "training_validation": [],
+            "training_validation": ["T1w", "SEM"],
             "testing": [],
             "balance": {}
         },
--- a/ivadomed/config/config_new_loader.json
+++ b/ivadomed/config/config_new_loader.json
@@ -1,7 +1,7 @@
 {
     "command": "train",
     "gpu": 0,
-    "log_directory": "/home/mhbourget/ivadomed/log_directory/new_loader",
+    "log_directory": "log_directory/new_loader",
     "model_name": "model_name",
     "debugging": false,
     "object_detection_params": {
@@ -9,8 +9,8 @@
         "safety_factor": [1.0, 1.0, 1.0]
     },
     "loader_parameters": {
-        "bids_path": "/home/mhbourget/ivadomed/data_example_microscopy_sem",
-        "bids_config": "/home/mhbourget/ivadomed/ivadomed/config/config_bids.json",
+        "bids_path": "data_example_microscopy_sem",
+        "bids_config": "ivadomed/config/config_bids.json",
         "target_suffix": ["_seg-manual", "_seg-axon-manual"],
         "extensions": [".png", ".nii.gz"],
         "roi_params": {
--- a/ivadomed/config/config_new_loader.json
+++ b/ivadomed/config/config_new_loader.json
@@ -12,7 +12,8 @@
         "bids_path": "data_example_microscopy_sem",
         "bids_config": "ivadomed/config/config_bids.json",
         "target_suffix": ["_seg-manual", "_seg-axon-manual"],
-        "extensions": [".png", ".nii.gz"],
+        "extensions": [".nii.gz", ".png"],
+        "data_types": ["anat", "microscopy"],
         "roi_params": {
             "suffix": null,
             "slice_filter_roi": null
--- a/ivadomed/config/config_new_loader.json
+++ b/ivadomed/config/config_new_loader.json
@@ -3,7 +3,7 @@
     "gpu": 0,
     "log_directory": "log_directory/new_loader",
     "model_name": "model_name",
-    "debugging": false,
+    "debugging": true,
     "object_detection_params": {
         "object_detection_path": null,
         "safety_factor": [1.0, 1.0, 1.0]
--- a/ivadomed/config/config_new_loader.json
+++ b/ivadomed/config/config_new_loader.json
@@ -1,0 +1,88 @@
+{
+    "command": "train",
+    "gpu": 0,
+    "log_directory": "/home/mhbourget/ivadomed/log_directory/new_loader",
+    "model_name": "model_name",
+    "debugging": false,
+    "object_detection_params": {
+        "object_detection_path": null,
+        "safety_factor": [1.0, 1.0, 1.0]
+    },
+    "loader_parameters": {
+        "bids_path": "/home/mhbourget/ivadomed/data_example_microscopy_sem",
+        "bids_config": "/home/mhbourget/ivadomed/ivadomed/config/config_bids.json",
+        "target_suffix": ["_seg-manual", "_seg-axon-manual"],
+        "extensions": [".png", ".nii.gz"],
+        "roi_params": {
+            "suffix": null,
+            "slice_filter_roi": null
+        },
+        "contrast_params": {
+            "training_validation": [],
+            "testing": [],
+            "balance": {}
+        },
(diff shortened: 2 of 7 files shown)


ivadomed/config/config_new_loader.json
```diff
@@ -0,0 +1,88 @@
...
+    "gpu": 0,
+    "log_directory": "/home/mhbourget/ivadomed/log_directory/new_loader",
+    "model_name": "model_name",
+    "debugging": false,
+    "object_detection_params": {
+        "object_detection_path": null,
+        "safety_factor": [1.0, 1.0, 1.0]
+    },
+    "loader_parameters": {
+        "bids_path": "/home/mhbourget/ivadomed/data_example_microscopy_sem",
```
olix86, line 12: same

ivadomed/config/config_new_loader.json
```diff
@@ -0,0 +1,88 @@
+{
+    "command": "train",
+    "gpu": 0,
+    "log_directory": "/home/mhbourget/ivadomed/log_directory/new_loader",
```
olix86, line 4: I'd use a relative path such as
``` "log_directory": "log_directory/new_loader",```

that way it's easier for others to use it

File: ivadomed/loader/utils.py [hunk 3]
```diff
@@ -501,3 +503,81 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+    print(df)
+
+    # Add metadata from participants.tsv
+    fname_participants = bids_path + "/participants.tsv"
+    df_participants = pd.read_csv(fname_participants, sep='\t')
+    df['participant_id'] = "sub-" + df['subject']
+    df = pd.merge(df, df_participants, on='participant_id', suffixes=("_x", None), how='left')
+
+    # Add metadata from samples.tsv, if present
+    fname_samples = bids_path + "/samples.tsv"
```
olix86, line 565: Maybe consider [os.join](https://www.geeksforgeeks.org/python-os-path-join-method/)

ivadomed/loader/utils.py
```diff
@@ -501,3 +503,81 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
     trans_orient = nib.orientations.ornt_transform(ras_orientation, ref_orientation)
     # apply transformation
     return nib.orientations.apply_orientation(arr_ras, trans_orient)
+
+def create_bids_dataframe(loader_params, derivatives):
+    """Create a dataframe containing all BIDS image files in a bids_path and their metadata.
+
+    Args:
+        loader_params (dict): Loader parameters, see :doc:`configuration_file` for more details.
+        derivatives (bool): If True, derivatives are indexed.
```
andreanne-lemay, line 512: This is great :100: ! It makes it flexible for the dataset where we don't have derivatives. Concerning your question, I think we should index the derivatives if this option is set to True, it will be easier for the loading if everything is in the df. This function should also verify availability (in my opinion), so if we want to index derivatives, only the subjects with the right target(s) and contrast(s) should be in the final df.

ivadomed/loader/utils.py
```diff
@@ -501,3 +503,81 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+
+    # Suppress a Future Warning from pybids about leading dot included in 'extension' from version 0.14.0
+    # The config_bids.json file used matches the future behavior
+    # TODO: when reaching version 0.14.0, remove the following line
+    pybids.config.set_option('extension_initial_dot', True)
+
+    # Initialize BIDSLayoutIndexer and BIDSLayout
+    # validate=True by default for both indexer and layout, BIDS-validator is not skipped
+    # Force index for subfolders starting with string "sub", and samples tsv and json files (for microscopy)
+    force_index = [d for d in os.listdir(bids_path) if (os.path.isdir(os.path.join(bids_path, d))) and (d.startswith("sub"))]
```
andreanne-lemay, line 532: I understand why the indexing of 'samples.tsv' and 'samples.json' are forced, but why do we have to force the indexing of folders starting by "sub" won't they already be indexed by pybids?

ivadomed/loader/utils.py
```diff
@@ -501,3 +503,81 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+    force_index.extend(['samples.tsv', 'samples.json'])
+    indexer = pybids.BIDSLayoutIndexer(force_index=force_index)
+    layout = pybids.BIDSLayout(bids_path, config=bids_config, indexer=indexer, derivatives=derivatives)
```
andreanne-lemay, line 535: I see we have a fork of pybids, do we plan on overwriting some of their code? If so, we might want to handle the error when .json files exist, but are empty (https://github.com/ivadomed/ivadomed/issues/532). Currently, we still have an error in this case.

ivadomed/loader/utils.py
```diff
@@ -501,3 +503,81 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+    # Could be used for samples.tsv with modifications to pybids after BEP microscopy is merged in BIDS.
+
+    # TODO: check if other files are needed for EEG
+
+    # Drop columns with all null values
+    df.dropna(axis=1, inplace=True, how='all')
+
+    # TODO: read appropriate metadata from image files and add to dataframe
+
+    return df
```
andreanne-lemay, line 583: I tested it on MRI data and it worked very well! I love the final dataframe format!

ivadomed/loader/utils.py
```diff
@@ -501,3 +503,81 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+        loader_params (dict): Loader parameters, see :doc:`configuration_file` for more details.
+        derivatives (bool): If True, derivatives are indexed.
+
+    Returns:
+        df (pd.DataFrame): Dataframe containing all BIDS image files indexed and their metadata.
+    """
+
+    # Get bids_path, bids_config, target_suffix and extensions from loader parameters
+    bids_path = loader_params['bids_path']
+    bids_config = loader_params['bids_config']
```
andreanne-lemay, line 520: We should make this parameter optional in the config file (we could simply put None by default) and if it is None, we take the default file present in ivadomed/ivadomed/config, because in most cases we don't need a custom bids file.

andreanne-lemay (commented):
:1st_place_medal:

ivadomed/loader/utils.py
mariehbourget: Ok thanks, I also think it's better to check availability before the split, I'll look into that.

ivadomed/config/config_new_loader.json
mariehbourget: Fixed in e5eb063

ivadomed/config/config_new_loader.json
mariehbourget: Fixed in e5eb063

ivadomed/loader/utils.py
mariehbourget: Fixed in 7c10596

ivadomed/loader/utils.py
mariehbourget: Fixed in 4e5eaf7
Let me know if this is what you had in mind.

ivadomed/loader/utils.py
mariehbourget: This is because I kept the BIDS-validator "on", so files that are not BIDS-compliant are ignored. For microscopy, that means that sub folders at the root of the dataset are ignored because png and ometiff files are not compliant yet. So forcing "sub" is overkill for anat, but necessary for microscopy. When microscopy will be included in pybids and BIDS-validator, we will be able to remove the force_index.
This is not full proof because if non-compliant files are in the sub folders, they will be indexed to. I'll check if I can find a better condition.

ivadomed/loader/utils.py
mariehbourget: I was planning on making PR from the fork before I found the custom config file. For now, it seems easier to wait until BEP microscopy is merged before submitting changes to pybids. But the json file error has nothing to do with microscopy so yes, we could propose changes.

ivadomed/loader/utils.py
mariehbourget: Thanks!

ivadomed/loader/utils.py
```diff
@@ -505,3 +507,78 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
     trans_orient = nib.orientations.ornt_transform(ras_orientation, ref_orientation)
     # apply transformation
     return nib.orientations.apply_orientation(arr_ras, trans_orient)
+
```
jcohenadad, line 510: PEP8 asks for 2 lines between function defs

mariehbourget: Fixed in 85537da

ivadomed/loader/utils.py
mariehbourget: Filtering based on derivatives availability is done in b3c9d34. We now have to determine what is the behavior if no matching derivatives are found (currently raise an error and exit).

mariehbourget:
Update in implementation

DONE:
-	Added filtering based on modality suffix (contrast)
-	Added filtering based on available derivatives
-	Added parsing or `sessions.tsv` and` scans.tsv` files

TODO:
-	Find a better condition for force index of microscopy files
-	Determine behavior if no matching derivatives are found (currently raise error and exit)
-	Load metadata from files (I'm not sure if it is the best place, should we do it in the loader specific to each file format instead?)
-	Create a test dataset including : `microscopy` (png + ometiff), `anat` (2D + 3D nifty), `dwi` (4D nifty), `eeg` (files TBD)
-	Adapt behavior of `scans.tsv` parsing for `eeg` and `dwi`.
-	Write a test for the function

ivadomed/loader/utils.py
```diff
@@ -520,3 +522,116 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+                df_scans = pd.concat([df_scans, df_temp], ignore_index=True)
+    if not df_scans.empty:
+        df_scans['filename'] = df_scans['filename'].apply(os.path.basename)
+        df = pd.merge(df, df_scans, on=['filename'], suffixes=("_x", None), how='left')
+
+    # TODO: check if other files are needed for EEG
+
+    # If indexing of derivatives is true
+    # Get list of subject files, list of subject prefix filenames without extensions, and list of available derivatives
+    if derivatives:
```
jcohenadad, line 614: as discussed in meeting, you could inform with logger.debug if one derivative for a file is missing

ivadomed/loader/utils.py
mariehbourget: I updated the condition in 12a5b50. The indexing is now forced for sub-folders only if they include microscopy files based on extensions.

ivadomed/loader/utils.py
mariehbourget: Done in 7c3e8a2: warnings for missing one derivative for a subject, and missing all derivatives for a subject.

mariehbourget:
This PR is ready for review.

As discussed:
- force index is now only applied to microscopy files
- if debugging is true: there are warnings for missing one derivative or all derivatives for a subject.

ivadomed/config/config_new_loader.json
```diff
@@ -0,0 +1,88 @@
+{
```
charleygros, line 1: I am guessing we will use this file during the prototyping / development of the new loader before removing it when fully integrated. Is it the idea?

ivadomed/config/config_new_loader.json
```diff
@@ -0,0 +1,88 @@
...
+    "log_directory": "log_directory/new_loader",
+    "model_name": "model_name",
+    "debugging": true,
+    "object_detection_params": {
+        "object_detection_path": null,
+        "safety_factor": [1.0, 1.0, 1.0]
+    },
+    "loader_parameters": {
+        "bids_path": "data_example_microscopy_sem",
+        "bids_config": "ivadomed/config/config_bids.json",
```
charleygros, line 13: Does it mean this config_bids could potentially change across projects?

dev/df_new_loader.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
charleygros, line 1: Very well documented and useful. I am wondering if we should not move these lines to a proper testing function!

charleygros (approved):
Really well done @mariehbourget ! That was not a easy task to integrate this new hierarchy!
A small comment would be: to add some documentation about this new Galaxy in ReadTheDocs (e.g., re the new params in config file).

charleygros (commented):
I appreciated how well your code was commented / documented!
Can I help with something? eg run some tests? any idea?

ivadomed/config/config_new_loader.json
mariehbourget: Not really. This custom config is a modification to the default config of `pybids` to include microscopy data, since microscopy won't be merged with pybids for a while. It includes everything that is already in the default pybids config file + relevant entities for microscopy. It could be updated if microscopy BEP changes, or if we want to include other "non-bids" modality.
For MRI and any other "official" bids modality, the custom pybids config file is not necessary and should not be used.

dev/df_new_loader.py
mariehbourget: Yes absolutely! The purpose of this script was to test during development. I'll open another PR to include a proper testing function.

mariehbourget:
> Really well done @mariehbourget ! That was not a easy task to integrate this new hierarchy!
> A small comment would be: to add some documentation about this new Galaxy in ReadTheDocs (e.g., re the new params in config file).

Thank you @charleygros for your feedback. I added the new params in the config file. I also added the key "extensions" to the default config and modify how the filtering is done with the extensions in the function. That way, "extensions" is not mandatory for MRI (as it was before), and is necessary only if a specific dataset contains more than one possible file extensions (e.g. microscopy with png and ometiff). Could you review the changes? Thanks!

I was wondering if it is OK to merge the PR after that without the test, because the coverage will decrease, what do you think?
(I'm starting to work on the test dataset and test now, it will be in another PR)

ivadomed/loader/utils.py
```diff
@@ -570,12 +570,14 @@ def create_bids_dataframe(loader_params, derivatives):
...
-    df = (df[(~df['path'].str.contains('derivatives') & df['suffix'].str.contains('|'.join(contrast_lst))) |
-          df['filename'].str.contains('|'.join(target_suffix))])
+    # Drop rows with json, tsv and LICENSE files in case no extensions are provided in config file for filtering
+    df = df[~df['filename'].str.endswith(tuple(['.json', '.tsv', 'LICENSE']))]
+
+    # Update dataframe with subject files of chosen contrasts and extensions,
+    # and with derivative files of chosen target_suffix from loader parameters
+    df = (df[(~df['path'].str.contains('derivatives') & df['suffix'].str.contains('|'.join(contrast_lst)) &
+          df['extension'].str.contains('|'.join(extensions))) |
+          (df['filename'].str.contains('|'.join(target_suffix)))])
```
charleygros, line 580: :100:

charleygros:
> Thank you @charleygros for your feedback. I added the new params in the config file. I also added the key "extensions" to the default config and modify how the filtering is done with the extensions in the function. That way, "extensions" is not mandatory for MRI (as it was before), and is necessary only if a specific dataset contains more than one possible file extensions (e.g. microscopy with png and ometiff). Could you review the changes? Thanks!

I had a look at it: looks great! Many thanks for documenting this as well!

> I was wondering if it is OK to merge the PR after that without the test, because the coverage will decrease, what do you think?
> (I'm starting to work on the test dataset and test now, it will be in another PR)

Oh yeah, sure, makes complete sense! Let's first merge this PR, then introduce the testing dataset to properly test it. Cheers!

PR #542, opened 2020-11-30
fix lfdr

andreanne-lemay:
Even if the GT has n_obj (number of objects to zero), there can still be a value of LFDR. Example: GT has no object, but the prediction has one -> LFDR = 1 (and not np.nan)

### DIFF
--- a/ivadomed/evaluation.py
+++ b/ivadomed/evaluation.py
@@ -381,7 +381,7 @@ class Evaluation3DMetrics(object):
         lfp = self._get_lfp(label_size, class_idx)
 
         denom = ltp + lfp
-        if denom == 0 or n_obj == 0:
+        if denom == 0:
             return np.nan
 
         return lfp / denom


charleygros:
That's a discussion we had when we first implemented LFDR: I don't think that's a fix but rather a choice of implementation :-)
I believe that both have pros and cons!
We can have this discussion again during next meeting?

andreanne-lemay:
I don't understand why we don't want to track all false positives? In which case is it better to set to Nan the value of LFDR when the class is absent, but the model predicted this class (which seems to me like a clear false positive)?

charleygros:
> I don't understand why we don't want to track all false positives? In which case is it better to set to Nan the value of LFDR when the class is absent, but the model predicted this class (which seems to me like a clear false positive)?

Our train of mind was something like:
- the object-based metrics are computed on images where the GT has at least a lesion. I can provide REF of papers to back this up.
- the false positive detection on images without lesion (eg healthy subjects or MS patients with no SC lesion) are assessed with voxel based metrics, eg precision, dice

andreanne-lemay:
This metric can be used on other tasks than lesion segmentation (e.g., tumor, edema, or cavity segmentation), so if the model predicts a class that is not there, it is a clear false positive. Since on the lesion detection task, you suppose each GT has at least a lesion, the `if n_obj == 0` will never happen. But even in the case where GT would be 100% healthy, if a lesion is predicted, it is still a false positive. However, it doesn't forbid the user to go towards other voxel-based metrics.

charleygros:
I am not against changing its name (eg `ofdr` for object FDR) and possibly its behaviour to accommodate other tasks :-) I was only explaining why it has been implemented this way, in the context of MS lesion.

> Since on the lesion detection task, you suppose each GT has at least a lesion, the `if n_obj == 0` will never happen.

There is an important number of MS patients that do not have intramedullary lesions --> `n_obj == 0` will happen. That's why: we wanted to make this difference: object metric for images where GT says there is an object, voxel metric for small blobs of FPs.

> But even in the case where GT would be 100% healthy, if a lesion is predicted, it is still a false positive. However, it doesn't forbid the user to go towards other voxel-based metrics.

Correct, that's why we ran `sct_deepseg_lesion` on healthy subjects and MS patients without SC lesions and reported voxel-based metrics for those.

So to generalise, we could compute `ofdr` on all images (what you implemented here) --> the user can then chose to do their own stats by selecting the CSV rows they want (eg only images with `vol_gt > 0`).

charleygros (approved):

PR #547, opened 2020-12-02
Fix missing attribute softmax

andreanne-lemay:
In the `forward` method self.softmax can be called, but the attribute is not defined.


charleygros (approved):
eheh pretty useful

PR #549, opened 2020-12-04
Apply bounding box safety factor in segment volume

andreanne-lemay:
The bounding box safety factor is not applied when segmenting images with segment volume.


ivadomed/inference.py
```diff
@@ -140,6 +140,7 @@ def segment_volume(folder_model, fname_images, gpu_number=0, options=None):
             thr = int(options['remove_small'].replace(unit, ""))
             postpro['remove_small'] = {"unit": unit, "thr": thr}
 
+        postpro.update(context['postprocessing'])
```
andreanne-lemay, line 143: Combine postprocessing from the config file (if any) and the postprocessing set in SCT

charleygros: Makes sense!

charleygros: I am just wondering if we should not do it the other way around, ie update the configfile with the one provided by the SCT user!?

andreanne-lemay: I asked myself the same thing, but by default, `options` is already populated by the default values of sct_deepseg (keep_largest to false for example). Since my architecture is cascaded I might want different parameters for tumors and SC localization. So my train of thought was: most of the time `postprocessing` param in the config file will be empty (i.e {}), and we only put postprocessing if it is important and should not be changed (keep_largest for the SC localization since there is only one) or else we leave it empty for the user to choose. This way we can have postprocessing steps customized to the first step of a cascaded architecture as well.

ivadomed/inference.py
```diff
@@ -127,20 +127,30 @@ def segment_volume(folder_model, fname_images, gpu_number=0, options=None):
     # Load model training config
     context = imed_config_manager.ConfigurationManager(fname_model_metadata).get_config()
 
-    if options is not None and any(pp in options for pp in ['thr', 'largest', ' fill_holes', 'remove_small']):
+    postpro_list = ['binarize_prediction', 'keep_largest', ' fill_holes', 'remove_small']
```
andreanne-lemay, line 130: Names from SCT now fits the ivadomed names (see: https://github.com/neuropoly/spinalcordtoolbox/pull/3090)

ivadomed/inference.py
andreanne-lemay: Values in SCT are set by default to None, this way if no flags are called, the parameters in the config file will be applied.

charleygros (approved):
TipTop!!

PR #551, opened 2020-12-04
Implementation of max pooling

andreanne-lemay:
Binarization adapted to multiclass models. Takes the maximum prediction across classes to binarize. This was the sum of all classes (including the background) is always 1.

Apply it this way in the config file:
"postprocessing": {"binarize_maxpooling": {}}


charleygros (approved):
Yep love it!
2 sentenes about it in RTD and good to go!!

andreanne-lemay:
> Yep love it!
> 2 sentenes about it in RTD and good to go!!

Oups my bad! Updated here :) 5fd6091

PR #560, opened 2020-12-07
Updating comparison table after review

### LINKED ISSUE #557: Comparison with other projects - Enhance table
It would be nice to enhance the [RTD table](https://ivadomed.org/en/latest/purpose.html#comparison-with-other-projects) where we compare `ivadomed` with other projects.

Suggestions:
- break it down in multiple tables to avoid the horizontal scroll
- add hyperlink in repo's name

lrouhier:
fix #557 

This PR aims to address some comments that have been received during the Joss review. 
The previous comparison table that was made to compare ivadomed with existing software has been divided for easier reading. 
This creates new files that each correspond to a sub-section in purpose.rst


charleygros (approved):
:100:

Issue #556, opened 2020-12-07
Installation Python 3.9 fails


charleygros:
Installation with Python 3.9 does not work
```
pip install ivadomed
Collecting ivadomed
  Downloading ivadomed-2.6.0-py3-none-any.whl (139 kB)
     |████████████████████████████████| 139 kB 8.9 MB/s
  Downloading ivadomed-2.5.0-py3-none-any.whl (137 kB)
     |████████████████████████████████| 137 kB 17.9 MB/s
  Downloading ivadomed-2.4.0-py3-none-any.whl (132 kB)
     |████████████████████████████████| 132 kB 23.7 MB/s
  Downloading ivadomed-2.3.1-py3-none-any.whl (129 kB)
     |████████████████████████████████| 129 kB 24.1 MB/s
  Downloading ivadomed-2.2.1-py3-none-any.whl (127 kB)
     |████████████████████████████████| 127 kB 28.2 MB/s
  Downloading ivadomed-2.2.0-py3-none-any.whl (127 kB)
     |████████████████████████████████| 127 kB 23.9 MB/s
  Downloading ivadomed-2.1.0-py3-none-any.whl (110 kB)
     |████████████████████████████████| 110 kB 30.8 MB/s
  Downloading ivadomed-2.0.2-py3-none-any.whl (108 kB)
     |████████████████████████████████| 108 kB 29.9 MB/s
  Downloading ivadomed-2.0.1-py3-none-any.whl (108 kB)
     |████████████████████████████████| 108 kB 30.9 MB/s
  Downloading ivadomed-1.2.1-py3-none-any.whl (68 kB)
     |████████████████████████████████| 68 kB 1.8 MB/s
  Downloading ivadomed-1.2-py3-none-any.whl (68 kB)
     |████████████████████████████████| 68 kB 5.0 MB/s
  Downloading ivadomed-1.1-py3-none-any.whl (68 kB)
     |████████████████████████████████| 68 kB 4.6 MB/s
  Downloading ivadomed-1.0-py3-none-any.whl (68 kB)
     |████████████████████████████████| 68 kB 2.4 MB/s
  Downloading ivadomed-0.14-py3-none-any.whl (68 kB)
     |████████████████████████████████| 68 kB 4.8 MB/s
  Downloading ivadomed-0.13-py3-none-any.whl (68 kB)
     |████████████████████████████████| 68 kB 457 kB/s
  Downloading ivadomed-0.12-py3-none-any.whl (66 kB)
[...]

PR #561, opened 2020-12-07
Add installation instructions in README

charleygros:
This PR adds the installation instructions in `README.md` (already present in RTD).

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,35 @@
 
 The technical documentation is available [here](https://ivadomed.org).
 
+## Installation
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,35 @@
 
 The technical documentation is available [here](https://ivadomed.org).
 
+## Installation
+
+``ivadomed`` requires Python >= 3.6 and PyTorch >= 1.5.0. We recommend working under a virtual environment, which could be set as follows:
```
jcohenadad, line 15: do we want to keep the py>3.6 if we know it fails with py3.9?

charleygros: I was thinking we could fix this once we investigated #556 --> Future PR.

jcohenadad (approved):

PR #569, opened 2020-12-09
Save config file before training

andreanne-lemay:
If the training is interrupted before the end, the config file is not saved in the log directory. This PR save the config file before training to prevent this.


ivadomed/main.py
```diff
@@ -274,15 +285,9 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
 
         # Update threshold in config file
         context["postprocessing"]["binarize_prediction"] = {"thr": thr}
+        save_config_file(context, log_directory)
 
     if command == 'train':
-        # Save config file within log_directory and log_directory/model_name
```
charleygros, line 279: I was just wondering whether there was a reason why we put this "save config" in the `command = train` condition, ie not saving it at testing time.... But can't think of anything so... all good!

charleygros (approved):

PR #570, opened 2020-12-09
Remove small for multiclass

andreanne-lemay:
The current version of the code allows to choose only one threshold value, but this is not adapted to multiclass models.

Change:
A list of thresholds can be given to have different thresholds depending on the class.

Sister PR in SCT: https://github.com/neuropoly/spinalcordtoolbox/pull/3090


docs/source/configuration_file.rst
```diff
@@ -484,7 +484,8 @@ Dict. Remove small objects from the prediction. An object is defined as a group
 neighbors are connected to the center, diagonally-connected elements are not considered neighbors.
 
 - ``unit``: String. Either "vox" for voxels or "mm3". Indicates the unit used to define the minimal object size.
-- ``thr``: Int. Minimal object size.
+- ``thr``: Int or list. Minimal object size. If a list of threshold is chosen, the length should match the number of
+  predicted classes.
```
charleygros, line 488:
```suggestion
- ``thr``: Int or list. Minimal object size. If a list of thresholds is chosen, the length should match the number of
  predicted classes.
```

charleygros (approved):
Great ++

PR #572, opened 2020-12-11
visualization of multiple evaluation models

mpompolas:
Added the script that was described on the last meeting for comparing multiple evaluation models.
Users can run this from their command line as indicated below. Any number of inputs/folders is supported:

`python3 compare_testing_models.py --listfolders /home/nas/Desktop/logs/logs_NO_FILM_sctUsers /home/nas/Desktop/logs/logs_onlyT1w /home/nas/Desktop/logs/logs_onlyT2w`

or
`python3 compare_testing_models.py --listfolders /home/nas/Desktop/logs/logs_*`

The output will be a collection of ViolinPlots, mean values (red values on top) and p-values (ns: Not significant, *: significant, ***: very significant) as in the figure below:
image

I spoke too fast yesterday when I mentioned that it is an easy fix to make these plots based on the `pathology`. I thought that this information was passed down to `~/results_eval/evaluation_3Dmetrics.csv`

A way of making these plots based on pathology or any other metadata would be to just append more metadata columns on the evaluation_3Dmetrics.csv. As far as I saw, there are no metadata anywhere in the `logs` folders.
Alternatively, if you think that appending metadata on `~/results_eval/evaluation_3Dmetrics.csv` is a pain, we can just make a copy of the `participants.tsv` inside the `log` folder during training. 

Any suggestions?


charleygros:
The output is absolutely amazing @mpompolas !!! Many many thanks!!

Comments:
- Would it be possible to move the scripts in the script folder --> `ivadomed/scripts` ? as this script could be used by a larger community than the core `ivadomed` developers (`dev` folder)
- it could be renamed `ivadomed_compare_testing_models` with an entry [here](https://github.com/ivadomed/ivadomed/blob/master/setup.py#L53)
- and documented [here](https://github.com/ivadomed/ivadomed/blob/master/docs/source/scripts.rst).
- could be nice to add a parameter to select the metric of interest (with `dice_class0` by default)
- could be awesome to call this script within `ivadomed_automate_training` when `--run-test`, ie [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L282) --> we run X experiments --> it would plot the violin plots of the X trained models on the testing dataset!

> Alternatively, if you think that appending metadata on ~/results_eval/evaluation_3Dmetrics.csv is a pain, we can just make a copy of the participants.tsv inside the log folder during training

Absolutely, I agree, I think that the only way to pull this information and do the subject_name matching with evaluation_3Dmetrics.csv

mpompolas:
I followed @charleygros suggestion to add an optional selection of --metric to be displayed on the violinplots (default - dice_class0).
I took it a step further to also have an optional selection of metadata to be diplayed: for example, if I want to take a look at the performance of different models only on `MS` patients, I can define that with --metadata pathology MS.

The title of the plot will be the selected criteria for easy identification as is here:

image


This should be merged after https://github.com/ivadomed/ivadomed/pull/588 since it requires the presence of a copy of participants.tsv within the log folder.

@charleygros I will take a look at your suggestion now: 

> could be awesome to call this script within ivadomed_automate_training when --run-test, ie here --> we run X experiments --> it would plot the violin plots of the X trained models on the testing dataset!

charleygros:
Great additions @mpompolas ! Love it!!

My only comment: Could you please organise the script as the other `ivadomed/scripts`? ie:
- a function `main` which init_ivadomed, parser then call function, see [example](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L367).
- the function that is called by `main`: look example [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L151) --> with in particular a docstring conatining what you wrote from L14 and description of the parameters --> it is this function that you will call in `docs/source/scripts.rst` and in `setup.py`

The reason for this:
- it will automatically document your function in the RTD, see example [here](https://ivadomed.org/en/latest/scripts.html#ivadomed-automate-training)
- it will show the usage of your function when typing `ivadomed_visualize_and_compare_testing_models -h` in your terminal

I think you could also add an image of your violin plots in the docstring, example [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/visualize_transforms.py#L70), as it would be very nice to show it in the RTD.

Let me know if you have any questions / thoughts!

setup.py
```diff
@@ -51,6 +51,7 @@
             'ivadomed_extract_small_dataset=ivadomed.scripts.extract_small_dataset:main',
             'ivadomed_download_data=ivadomed.scripts.download_data:main',
             'ivadomed_training_curve=ivadomed.scripts.training_curve:main'
+            'ivadomed_visualize_and_compare_testing_models=ivadomed.scripts.ivadomed_visualize_and_compare_testing_models:main'
```
charleygros, line 54:
```suggestion
            'ivadomed_visualize_and_compare_testing_models=ivadomed.scripts.visualize_and_compare_testing_models:main'
```

docs/source/scripts.rst
```diff
@@ -48,5 +48,8 @@ ivadomed_download_data
 
 .. autofunction:: ivadomed.scripts.download_data.install_data
 
+ivadomed_visualize_and_compare_testing_models
+"""""""""""""""""""""""""""""""""""""""""""""
 
+.. autofunction:: ivadomed.scripts.ivadomed_visualize_and_compare_testing_models
```
charleygros, line 54:
```suggestion
.. autofunction:: ivadomed.scripts.visualize_and_compare_testing_models
```

You will also append the name of the major function of this script, see examples above.

mpompolas:
> My only comment: Could you please organise the script as the other ivadomed/scripts? ie:
> 
> a function main which init_ivadomed, parser then call function, see example.
> the function that is called by main: look example here --> with in particular a docstring conatining what you wrote from L14 and description of the parameters --> it is this function that you will call in docs/source/scripts.rst.
>
> I think you could also add an image of your violin plots in the docstring, example here, as it would be very nice to show it in the RTD.
> 

@charleygros Followed all your great suggestions regarding the changes at the code and also added an image for reference.

> could be awesome to call this script within ivadomed_automate_training when --run-test, ie here --> we run X experiments --> it would plot the violin plots of the X trained models on the testing dataset!

Regarding this, how many models do you usually compare?
The plot looks nice for up to 4 models, after that it gets a bit busy with the multiple comparisons between each combination of models (I refer to the horizontal lines that show significance).
If you compare more than 4 models I will just deactivate the significance lines only then.

The insertion will be here: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L308
right after all the models have their results.

charleygros:
> Regarding this, how many models do you usually compare?

Difficult to answer, it varies a lot depending user case.

> The plot looks nice for up to 4 models, after that it gets a bit busy with the multiple comparisons between each combination of models (I refer to the horizontal lines that show significance).
> If you compare more than 4 models I will just deactivate the significance lines only then.

Sounds good to me! Good idea!!

> The insertion will be here: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L308
> right after all the models have their results.

Yep! Agree

mpompolas:
Added interactive selection to display the subject and sequence for each datapoint selected on the violinplots (see red box in the figure to the right). If more than one datapoints are very close together when users click, all subjects/sequences in the selection vicinity will be displayed.

image

charleygros:
> Added interactive selection to display the subject and sequence for each datapoint selected on the violinplots (see red box in the figure to the right). If more than one datapoints are very close together when users click, all subjects/sequences in the selection vicinity will be displayed.

That is really really Gucci!

mpompolas:
The problem with the RTD not displaying the function was identified by @ahill187 as a warning on the RTD build regarding matplotlib.
The issue is solved after doing a rotation on multiple GUI frameworks until one is loaded properly. The list might need to be extended

charleygros (approved):

Issue #554, opened 2020-12-05
Add functional tests for CLIs


jcohenadad:
Currently, ivadomed only includes API unit tests. It would be good to also add functional tests for CLIs, as [done in SCT](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests#functional-tests-pytest).

PR #573, opened 2020-12-11
Merge 2 BIDS datasets by duplicating them into a new BIDS dataset

mpompolas:
## Description

Merging 2 BIDS datasets together from a CLI command.
Subjects' files are copied to a new directory.
The `participants.tsv` and `participants.json` from the initial folders are merged into one of each. If some columns/metadata don't exist for the other dataset they will be filled with `NaNs`.

Example call:
`python3 merge_participants_files.py --ifolders ~/first_Dataset/ ~/second_Dataset/ --ofolder ~/mergedDataset/`

If needed we can generalize this function to accommodate more than 2 Datasets.

The assumption here is that the sole purpose of the newly created dataset is to be used in training - not to substitute the original dataset folders. For this reason only the absolutely necessary files are copied and merged: 
`1. derivatives folder
2. subjects folders
3. participants.tsv
4. participants.json`

Any other suggestion for files to be merged from the BIDS folders?


jcohenadad (changes requested):
could you please:
- add a functional test for this CLI
  - we currently don't have CLI tests, but this is an opportunity to get it started #554 
- add an entry point in the setup.py so the script can be called like other CLI scripts
- use ivadomed prefix on the CLI (see the setup)
- [document this new CLI function in the RTD](https://ivadomed.org/en/latest/scripts.html)

mpompolas:
will do.
I designed https://github.com/ivadomed/ivadomed/pull/572 as a CLI as well so I assume you want similar treatment on that as well?

jcohenadad:
> I designed #572 as a CLI as well so I assume you want similar treatment on that as well?

if it makes sense, yes-- sorry i don't have the time to look at all possible implications, reelvance, etc. so i hope others from the team can cheap in--

charleygros:
Before looking at the code, two design questions:
- What is the advantage of creating a new dataset (which is the merge of 2 datasets)? There is the cons of duplication.

- And what about no creation of new BIDS dataset, but instead: merge the list of image filenames? suggestions:
(1) have a list of BIDS paths --> loop across the `root_dir` ([here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/loader/loader.py#L753))
OR (2) have two instances of [BidsDataset ](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/loader/loader.py#L749) and implement a way to merge the `filename_pairs`.

If we go that route, we could have a list of BIDS paths instead of a single BIDS path in the config file ([here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config.json#L12)) --> no need to run a python script and data duplication prior to this.
For instance:
```json
        "bids_path": [path/to/large, path/to/spinegeneric],
```

mpompolas:
> Before looking at the code, two design questions:
>
>What is the advantage of creating a new dataset (which is the merge of 2 datasets)? There is the cons of data duplication.
>
>And what about no creation of new BIDS dataset, but instead: merge the list of image filenames? suggestions: have a list of >BIDS paths and (1) loop across the root_dir (here) OR (2) have two instances of BidsDataset and implement a way to merge >the filename_pairs.
>
>If we go that route, we could have a list of BIDS paths instead of a single BIDS path in the config file (here) --> no need to run a >python script and data duplication prior to this.
>For instance:
>
>        "bids_path": ["path/to/large", "path/to/spinegeneric"],


I totally agree with you. As I mentioned two lab meetings ago,  `"bids_path": ["path/to/large", "path/to/spinegeneric"]` is on my todo list .

The initial version of this code had a flag if you want to duplicate the files or not since it's not ideal to copy files around.

Most of this script will be duplicated for achieving what you mentioned: 

> have a list of BIDS paths instead of a single BIDS path in the config file


So regarding the training pipeline, I will integrate what I described just above.

Regarding this script as an extra feature, I personally want to have an easy way to merge two BIDS datasets. I just need some feedback on what other kinds of files need _merging_ to make it more robust.


If this feature is something you guys think shouldn't belong in the toolbox we can remove it.

charleygros:
> I just need some feedback on what other kinds of files need merging to make it more robust.

Ok! :-)

These are good:
> For this reason only the absolutely necessary files are copied and merged: 1. derivatives folder 2. subjects folders 3. participants.tsv 4. participants.json

And obviously a `dataset_description.json`. So pretty much all files :-)

You could add (if relevant for your application, dunno) a parameter that filters some contrasts: eg `_flair.nii.gz` only --> so that the `*_t1w.nii.gz`, `*_pd.nii.gz` etc are not copied if you do not need them? Dunno just suggestion.

Other suggestion could include: number of subjects in the merged dataset? randomly picked in the 2 datasets? but again it depends on your application. Just throwing ideas.

charleygros:
> If this second (extra) feature is something you guys think shouldn't belong in the toolbox we can remove it.

Sounds good to me! Thanks for that! It could definetly be handy!

My previous message was just in case I was not clear during the lab meeting: I thought you were going to directly implement it in the `ivadomed` code base --> So I was surprised to see it in `dev`.
Now I understand you made this script for your own application and are offering to share it with the community via a `dev` script, which is great and much appreciated! All gravy :-)

charleygros (approved):
:100:

mariehbourget:
> Any other suggestion for files to be merged from the BIDS folders?

Sorry to be late about this. For microscopy, the files `samples.tsv` and `samples.json` should also be merged. They are at the same level as `participants.tsv`.
Also, you are using `bids_neuropoly` to read the content of `participants.tsv`. We are currently trying to remove this dependency with the new loader project and move to `pybids` (example [here](https://github.com/ivadomed/ivadomed/blob/b42d62002d0d4cc997994198deaf6ea2c151a77a/ivadomed/loader/utils.py#L590)). In this case, `pybids` may be overkill if you just want to read the content a `tsv` file.

ivadomed/scripts/merge_bids_datasets.py
```diff
@@ -0,0 +1,170 @@
...
+    parser.add_argument("-o", "--ofolder", required=False, default="./merged_bids_datasets",
+                        help="Output folder for merged datasets.",
+                        metavar=imed_utils.Metavar.folder)
+    parser.add_argument("-m", "--metadata", required=False, default=[], nargs=2,
+                        help="Select files within a dataset that match specific metadata.",
+                        metavar=imed_utils.Metavar.list)
+    parser.add_argument("--nsubjects", required=False, default=[],
+                        help="Select the total number of subjects to have on the merged dataset",
+                        metavar=imed_utils.Metavar.int)
+    parser.add_argument("--copy_files", required=False, default=True,
```
jcohenadad, line 37: lab convention
```suggestion
    parser.add_argument("--copy-files", required=False, default=True,
```

jcohenadad (commented):
Is there still a relevance for this PR now that it is possible to train a model by aggregating various BIDS datasets?

mpompolas:
What I can think of that we can borrow from this PR to include in the loader of ivadomed is:
1. Do the training on subjects that meet specific metadata. A new field in the config file would be needed for that.
2. Assign a maximum number of subjects from the input datasets to run ivadomed on. This would be particularly useful when dealing with large datasets while debugging.

mariehbourget:
> What I can think of that we can borrow from this PR to include in the loader of ivadomed is:
> 
> 1. Do the training on subjects that meet specific metadata. A new field in the config file would be needed for that.
> 2. Assign a maximum number of subjects from the input datasets to run ivadomed on. This would be particularly useful when dealing with large datasets while debugging.

There is something similar in the config file with "subject selection": https://ivadomed.org/en/latest/configuration_file.html#subject-selection

mpompolas:
Thanks @mariehbourget 
The existing loader is doing pretty much everything at this point then.
I will close this pull request

PR #579, opened 2020-12-13
Remove python version in virtualenv command

### LINKED ISSUE #574: Documentation error in virtualenv command
### Current behavior

In the RTD [Installation ](https://ivadomed.org/en/stable/installation.html#installation)section, we have the following command:
```
virtualenv venv-ivadomed --python=python3.6
```

### Expected behavior
```
virtualenv venv-ivadomed
```
i.e., be Python version agnostic

charleygros:
## Description
This PR removes the python version in the `virtualenv` command because it causes an error if a different Python is installed. It is clear in both `README.md` and `installation.rst` that Python >= 3.6 and < 3.9 is required: the Python version in the virtual env command can then be omitted.

## Linked issues
Fixes #574


andreanne-lemay (approved):
:+1:

PR #580, opened 2020-12-14
Add functional tests for CLI

### LINKED ISSUE #554: Add functional tests for CLIs

ahill187:
## Description

The purpose of this pull request is to add functional tests for the command line integration (CLI).
These tests will be placed in the folder:
```
testing/
|--- functional_tests/
     |--- test_{file_to_test_1}.py
     |--- test_{file_to_test_2}.py
```

I have created a testing data repo for functional tests called `data_functional_testing`: https://github.com/ivadomed/data_functional_testing

> **Update**: In a future pull request, I would like to merge `data_functional_testing` with `data-testing`, as they share
most of the same data. The reason I separated them was because initially I wasn't totally familiar with the testing procedure, and I thought it best to make a separate repo so as to avoid interfering with the other unit tests.

## Linked issues
Resolves: https://github.com/ivadomed/ivadomed/issues/554


jcohenadad (changes requested):
i know this PR is in draft mode, but i just wanted to point out that the big (18MB) file `testing/functional_tests/fixtures/spinegeneric_model.pt` will need to be removed from this branch, and use squash before merging so the fat does not accumulate in the master branch (or you can also rebase now). thanks 😊

ahill187:
@jcohenadad Oh thanks for the flag, I'll remove the `spinegeneric` file!

ahill187:
@jcohenadad Moved the large `*.pt` file out into a separate `data_functional_testing` repo to be downloaded instead:
https://github.com/ivadomed/ivadomed/pull/580/commits/bcdf72634bc88a6591f43622bfd04568bfc4dcf3

ivadomed/scripts/automate_training.py
```diff
@@ -260,13 +280,17 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
...
 
+    if output_dir and not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+    if not output_dir:
+        output_dir = ""
+
     # CUDA problem when forking process
     # https://github.com/pytorch/pytorch/issues/2517
-    mp.set_start_method('spawn')
+    ctx = mp.get_context("spawn")
```
gisellemartel, line 290: Nitpick: I am not sure if here single-quotes would be more appropriate? (Basing myself off of this: https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Single-or-double-quotes)?

ivadomed/scripts/automate_training.py
```diff
@@ -364,18 +396,17 @@ def automate_training(config, param, fixed_split, all_combin, n_iterations=1, ru
...
-    args = parser.parse_args()
+    args = imed_utils.get_arguments(parser, args)
 
     # Get thr increment if available
     thr_increment = args.thr_increment if args.thr_increment else None
 
-    # Run automate training
-    automate_training(args.config, args.params, bool(args.fixed_split), bool(args.all_combin), int(args.n_iterations),
-                      bool(args.run_test), args.all_logs, thr_increment, bool(args.multi_params))
+    automate_training(args.config, args.params, bool(args.fixed_split), bool(args.all_combin),
```
gisellemartel, line 407: Perhaps it would be cleaner if each argument was placed in its own line? see: https://github.com/ivadomed/ivadomed/wiki/code-style

gisellemartel (approved):
Excellent work! Tested locally and all the tests pass. Left a few questions about code style but besides that code look's great

ivadomed/scripts/automate_training.py
ahill187: Interesting! I agree with the logic behind using `" "` for strings with text, as that could contain apostrophes etc. I think maybe switching between the two adds some overhead, like it might be easier to just use `" "` for everything?

ivadomed/scripts/automate_training.py
ahill187: Good call!

ahill187: Fixed here: https://github.com/ivadomed/ivadomed/pull/580/commits/353dcaaf82ef7004812c968028b19c5d1cead2ae

jcohenadad (approved):
Fantastic contribution!! Thank you @ahill187 !!

ivadomed/scripts/automate_training.py
ak-gmartel: @ahill187 you make a good point, I tend to lean toward the preference of having either all single quote or all double quotes, so just having `" "` could be a good choice. Seems like this is not the case with the SCT project, but given this is a separate project perhaps we can define our own standard for this?

gisellemartel: Perhaps it would make more sense to stick with just `" "` for the ivadomed project in general. I did notice SCT has a hybrid policy on this, but I tend to have a preference more toward just have one standard, especially with potential refactoring overhead.

ahill187: Yah I agree!

PR #582, opened 2020-12-16
Load 2D GT slice as uint if not soft training

charleygros:
## Description
When we are running a soft training, we want to load data as float.
When we are running a hard training, we want to load data as uint.
This is the case for the 3D loader, but not the 2D loader --> that's what is fixed in this small PR.


andreanne-lemay (approved):
:1st_place_medal:

PR #546, opened 2020-12-02
Random sampling of expert annotations

charleygros:
This PR implements a way to randomly sample experts' annotation to be used as GT during the training. Has application with datasets where annotations from several experts are available.

## Rationale

This new feature supports different on-going projects playing with datasets with annotations from multiple experts (e.g., look at inter-rater variability effect on DL model training, alternatives to label fusion techniques).

Our hypothesis is that this sampling method could lead to a better model generalisation to expert variability, than label fusion techniques (e.g., STAPLE) which are commonly used.

## Behaviour

In simple words, let's say we have 3 experts that segmented the same tissue, i.e., for each slice, we have 3 different annotations of the same tissue. At each epoch, the model will be trained on one of the 3 available annotations of the same slice, randomly picked.

## Specific needs

- Compatible with single- or multi-class tasks.

## Implementation

- in the config file: if a list is provided for `target_suffix` instead of a string, then this sampling method is triggered among the provided annotations:
```json
        "target_suffix": [["_gmseg-manual-ucl", "_gmseg-manual-unf", "_gmseg-manual-vanderbilt", "_gmseg-manual-zurich"],
                            ["_wmseg-manual-ucl", "_wmseg-manual-unf", "_wmseg-manual-vanderbilt", "_wmseg-manual-zurich"]],
```
instead of:
```json
        "target_suffix": ["_gmseg-manual-ucl", "_wmseg-manual-ucl"],
```

- XX

PR #584, opened 2020-12-16
Add tests for create_bids_dataframe function

### LINKED ISSUE #563: Write a test for create_bids_dataframe function
We need to write a test function for the create_bids_dataframe function for the new loader.
First, we need to update the testing dataset [here](https://github.com/ivadomed/data-testing/issues/5), and make sure the current tests are still working with this new updated dataset.

mariehbourget:
Fixes #563 
This PR add tests for the `create_bids_dataframe` function for the new loader.

It also contains corrections to the function for:
- target_suffix as a nested list (as of PR #546)
- filtering when no contrast_params are provided in the config file

The new tests won't passed until PR [#6](https://github.com/ivadomed/data-testing/pull/6) on `data-testing repo` is merged and released, and the link to download updated [here](https://github.com/ivadomed/ivadomed/blob/5564e6345a0bffdee52631cae4f9d6f0fe50c122/ivadomed/scripts/download_data.py#L25).


mariehbourget:
This PR is ready for review.
Note: The PR [#6](https://github.com/ivadomed/data-testing/pull/6) in data-testing must be reviewed at the same time and merged first to update the download link here.
@charleygros or @andreanne-lemay, would you like to review?

testing/unit_tests/test_loader.py
```diff
@@ -0,0 +1,55 @@
...
+    loader_params = loader_parameters
+    loader_params["contrast_params"]["contrast_lst"] = loader_params["contrast_params"]["training_validation"]
+    derivatives = True
+    df = imed_loader_utils.create_bids_dataframe(loader_params, derivatives)
+    df = df.drop(columns=['path', 'parent_path'])
+    df = df.sort_values(by=['filename']).reset_index(drop=True)
+    df.to_csv("testing_data/microscopy_png/df_test.csv", index=False)
+    command = "diff testing_data/microscopy_png/df_test.csv testing_data/microscopy_png/df_ref.csv"
+    subprocess.check_output(command, shell=True)
+
```
andreanne-lemay, line 33: 2 new lines

testing/unit_tests/test_loader.py
```diff
@@ -0,0 +1,55 @@
...
+    # Test for target_suffix as a nested list
+    # Test for when no contrast_params are provided
+    loader_params = loader_parameters
+    loader_params["contrast_params"]["contrast_lst"] = loader_params["contrast_params"]["training_validation"]
+    derivatives = True
+    df = imed_loader_utils.create_bids_dataframe(loader_params, derivatives)
+    df = df.drop(columns=['path', 'parent_path'])
+    df = df.sort_values(by=['filename']).reset_index(drop=True)
+    df.to_csv("testing_data/microscopy_png/df_test.csv", index=False)
+    command = "diff testing_data/microscopy_png/df_test.csv testing_data/microscopy_png/df_ref.csv"
```
andreanne-lemay, line 31: Here, we run the diff of both files, but it doesn't test if they are the same, which I believe is the goal of this test?

testing/unit_tests/test_loader.py
```diff
@@ -0,0 +1,55 @@
...
+    # Test for multiple target_suffix
+    loader_params = loader_parameters
+    loader_params["contrast_params"]["contrast_lst"] = loader_params["contrast_params"]["training_validation"]
+    derivatives = True
+    df = imed_loader_utils.create_bids_dataframe(loader_params, derivatives)
+    df = df.drop(columns=['path', 'parent_path'])
+    df = df.sort_values(by=['filename']).reset_index(drop=True)
+    df.to_csv("testing_data/df_test.csv", index=False)
+    command = "diff testing_data/df_test.csv testing_data/df_ref.csv"
+    subprocess.check_output(command, shell=True)
```
andreanne-lemay, line 55: Same comment

ivadomed/loader/utils.py
```diff
@@ -545,6 +546,8 @@ def create_bids_dataframe(loader_params, derivatives):
     bids_path = loader_params['bids_path']
     bids_config = None if 'bids_config' not in loader_params else loader_params['bids_config']
     target_suffix = loader_params['target_suffix']
+    if any(isinstance(t, list) for t in target_suffix):
```
andreanne-lemay, line 549:
```suggestion
    # If `target_suffix` is a list of lists convert to list
    if any(isinstance(t, list) for t in target_suffix):
```

testing/unit_tests/test_loader.py
```diff
@@ -0,0 +1,55 @@
+import os
+import subprocess
+
+import pytest
+import pandas as pd
+
+from ivadomed.loader import utils as imed_loader_utils
+
```
andreanne-lemay, line 8: PEP8: 2 lines between functions/methods

andreanne-lemay (approved):
LGTM :100: , see minor comments

testing/unit_tests/test_loader.py
mariehbourget: The diff command exit code are 0 for identical and 1 for different, so:
If they are the same, the test passes (exit code 0).
If they are not, test test fails (exit code 1).

andreanne-lemay: Since a diff of 2 different files doesn't through an error, I'm pretty sure it is not supposed to make the test fail. I tried it locally and when running diff of two different files in pytest, my test passed. 
Also, `diff` is Linux specific, if we start testing in MacOS or Windows the test will fail. Using an assert on the two df or on a numpy conversion of the df would be simpler (no need to start subprocesses) and will trigger if there is a difference.

mariehbourget: Oh ok, good to know (diff worked as I was expecting on my linux station). I'll check how I can use assert, thanks!

jcohenadad: or there is that (also it's extra dependency): https://pypi.org/project/csv-diff/

andreanne-lemay: Oh actually you're right it throws an error, my bad! But I still think it would be better to use assert when comparing both files (because it doesn't require a subprocess, it's not os dependant) I found this: https://stackoverflow.com/questions/45195388/best-way-to-compare-pandas-dataframe-with-csv-file to force datatypes, then you can use `assert df1.equals(df2)`

charleygros (approved):

testing/unit_tests/test_loader.py
mariehbourget: I tested both `pandas.DataFrame.equals` and `pandas.testing.assert_frame_equal`. Even when forcing datatypes when reading the csv, some comparison still fail (seems related to a precision in float data). I tried all sorts of conversion and "almost equal" type of thing but was unable to make it work on all fields of the dataframes. I finally decided to use csv-diff as Julien suggested and added it to the requirements., it is much simpler and is working as expected.

PR #586, opened 2020-12-17
Automate_training: fix bug for multiple parameters

andreanne-lemay:
There was an issue with lists of parameters nested under the same dict key.
```
Example of hyperparameter:
{
  "training_parameters": {
    "batch_size": [2, 4, 8],
    "loss": [
      {"name": "DiceLoss"},
      {"name": "FocalLoss", "params": {"gamma": 0.5, "alpha" : 0.2}},
      {"name": "FocalLoss", "params": {"gamma": 0.2, "alpha" : 0.5}}
    ]
  },
  "default_model": {"depth": [2, 3, 4]}
}
```
Since ` batch_size` and `loss` are under `training_parameters` the flag -m doesn't work properly for this case. To fix this, a special case of `make_category` was added.

### DIFF
--- a/ivadomed/scripts/automate_training.py
+++ b/ivadomed/scripts/automate_training.py
@@ -106,7 +106,7 @@ def test_worker(config):
     return config["log_directory"], test_dice, df_results
 
 
-def make_category(base_item, keys, values, is_all_combin=False):
+def make_category(base_item, keys, values, is_all_combin=False, multiple_params=False):
     items = []
     names = []
 
@@ -120,7 +120,22 @@ def make_category(base_item, keys, values, is_all_combin=False):
 
             items.append(new_item)
             names.append(name_str)
+    elif multiple_params:
+        value_len = set()
+        for value in values:
+            value_len.add(len(value))
+        if len(value_len) != 1:
+            raise ValueError("To use flag --multi-params or -m, all hyperparameter lists need to be the same size.")
 
+        for v_idx in range(len(values[0])):
+            name_str = ""
+            new_item = copy.deepcopy(base_item)
+            for k_idx, key in enumerate(keys):
+                new_item[key] = values[k_idx][v_idx]
+                name_str += "-" + str(key) + "=" + str(values[k_idx][v_idx]).replace("/", "_")
+
+            items.append(new_item)
+            names.append(name_str)
     else:
         for value_list, key in zip(values, keys):
             for value in value_list:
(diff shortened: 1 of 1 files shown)


charleygros (approved):
Looks all good

My only hesitation is: WOuld it work if the first key is not a dict but a list parameters, eg:
```
    "gpu": [0, 1, 2],
```
PS: I know this example is silly eheh

Have you tried it yet? otherwise I am happy to give it a shot

andreanne-lemay:
> Looks all good
> 
> My only hesitation is: WOuld it work if the first key is not a dict but a list parameters, eg:
> 
> ```
>     "gpu": [0, 1, 2],
> ```
> 
> PS: I know this example is silly eheh
> 
> Have you tried it yet? otherwise I am happy to give it a shot

No it doesn't work, but it never did. [This line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L186) only works on dictionaries (because of the `.keys()`) . Since it doesn't concern the code introduced in this PR, I think we should fix this in a future PR. The only parameters that are not dicts are: gpu, command, log_directory, model_name, and debugging. I don't think this is a pressing matter since they are pretty much all irrelevant when it comes to do hyperparametrization, but it is good to keep this in mind! I'll open an issue.

charleygros:
> No it doesn't work, but it never did. [This line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L186) only works on dictionaries (because of the `.keys()`) . Since it doesn't concern the code introduced in this PR, I think we should fix this in a future PR. The only parameters that are not dicts are: gpu, command, log_directory, model_name, and debugging. I don't think this is a pressing matter since they are pretty much all irrelevant when it comes to do hyperparametrization, but it is good to keep this in mind! I'll open an issue.

I agree with all of these. Just wanted to kindly raise it :-)

andreanne-lemay:
> > No it doesn't work, but it never did. [This line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L186) only works on dictionaries (because of the `.keys()`) . Since it doesn't concern the code introduced in this PR, I think we should fix this in a future PR. The only parameters that are not dicts are: gpu, command, log_directory, model_name, and debugging. I don't think this is a pressing matter since they are pretty much all irrelevant when it comes to do hyperparametrization, but it is good to keep this in mind! I'll open an issue.
> 
> I agree with all of these. Just wanted to kindly raise it :-)

Opened here #587

PR #591, opened 2020-12-22
New loader: Adapt splitting methods

### LINKED ISSUE #564: Adapt splitting methods for new loader
## Context

The first step of the refactoring was to generate a bids dataframe: done in #519 and #538 for anat and microscopy data.
The dataframe includes one row per image file from a bids_path (with parsed entities and tsv/json metadata) for each subject and derivatives files, when derivatives are available according to the chosen target suffixes.

The second step is to adapt the splitting methods (train/valid/test) to split using the dataframe.

## Functions to adapt

1. Currently the function [get_new_subject_split](https://github.com/ivadomed/ivadomed/blob/7d3eef70cb24d5a3411927c372e614ced00a140e/ivadomed/loader/utils.py#L96) uses `bids_neuropoly` and `participants.tsv` content to list subjects and metadata. We aim to remove the dependency to bids_neuropoly, and use the metadata already loaded in the dataframe at the previous step.

2. Currently the function [split_dataset](https://github.com/ivadomed/ivadomed/blob/7d3eef70cb24d5a3411927c372e614ced00a140e/ivadomed/loader/utils.py#L34) contains 2 different methods for splitting: `per_patient `and `per_center`.
As discussed previously, for microscopy, we want to implement a `per_sample` method as well, because a same subject can have multiple samples. Ex:
    * sub-01_sample-01_SEM.png
    * sub-01_sample-02_SEM.png

## Next step
Load data as numpy array with different loaders for different file formats.

mariehbourget:
Fixes #564 

It aims to adapt the splitting methods to include different methods for microscopy (e.g. per sample), and to use the new bids dataframe feature (#538).

---
Here are the changes made in this PR so far:

**config/config_new_loader.json:**
- I replaced the  `center_test` param by `"data_testing": {"data_type": "subject", "data_value":[]}`  which can be generalized to isolate any data for testing.
- I removed the `method` param, as it is included in the new `data_testing` param.

**"_new" functions in loader/utils.py:**
- `get_subdatasets_subjects_list_new`
- `get_new_subject_split_new`
- `split_dataset_new`

These functions are copies of the originals with "_new" added at the end of the name.

Unlike the previous step with the dataframe, I couldn't replace the original functions because they are not independent from the current codebase.

In addition, the current functions return the split as lists of "subjects". For microscopy, a "subject" is not always unique because a same subject can have multiple samples. I chose to return lists of "filenames", as filenames are unique for sure in any cases. This is not compatible with the current loaders but could be adapted as well.

**dev/df_new_loader.py:**
- Updated to test both the dataframe and the split
---
I think we should discuss a more detailed strategy moving forward regarding the next step from here and how to manage both the current and the new loader in the codebase.
In the meantime, any feedback on the splitting methods I implemented here is welcome!


mariehbourget:
This PR is ready for review.

`split_dataset_new` now returns lists of the `data_type` chosen in the config file (e.g. “subject”, “participant_id”, “sample”, institution_id”, etc).
The data_type will be used in load_dataset to lookup in the dataframe and to pair filenames (image+gt). It will be compatible with multichannel training.

The PR also includes corrections to the dataframe relevant for the split:
-	include only subjects files with available `roi_suffix`, if roi_suffix is not None
-	add `participant_id` and `sample_id` columns : The original df only had “subject” and “sample”. For example:  `subject = “rat1”`, whereas `participant_id = ”sub-rat1”`. This will help for retrocompatibility and transition because current config file split per participant_id.

The PR does not include a test. It would require a `.joblib` file to be added to the data-testing dataset (with a new release). Since the implementation of the split may change in the short-term, I think it would be best to wait until we are farther in the development of the next step of the new loader project to add a test.

ivadomed/loader/utils.py
```diff
@@ -637,15 +780,23 @@ def create_bids_dataframe(loader_params, derivatives):
...
+        if roi_suffix is not None:
+            for p in prefix_fnames:
+                available = [d for d in deriv if p in d]
+                if roi_suffix in ('|'.join(available)):
+                    has_deriv.append(p)
+                else:
+                    logger.warning("Missing ROI derivatives for subject {}. Skipping subject.".format(p))
+        else:
+            for p in prefix_fnames:
+                available = [d for d in deriv if p in d]
```
andreanne-lemay, line 792: I did a test on my local station on MRI data (data_example_spinegeneric), and even though derivatives is set to True, the df don't store the path to the derivatives, so at this line `deriv` is empty. I get the following error: `RuntimeError: Derivatives not found.` Is it normal and will be implemented in future PR or maybe it is a bug from pybids?

mariehbourget: It is supposed to work. Do you have another warning about a missing dataset_description.json file? Because of pybids, the file is mandatory in the "derivatives" folder. See note [here](https://github.com/ivadomed/ivadomed/blob/c65ab5af2ef33619c09f686355606ca941dacdaf/dev/df_new_loader.py#L28).

ivadomed/loader/utils.py
```diff
@@ -164,6 +209,71 @@ def get_new_subject_split(path_folder, center_test, split_method, random_seed,
...
+    Randomly split dataset between training / validation / testing\
+        and save it in log_directory + "/split_datasets.joblib".
+
+    Args:
+        df (pd.DataFrame): Dataframe containing all BIDS image files indexed and their metadata.
+        data_testing (dict): Used to specify the data_type and data_value to include in the testing set.
+        random_seed (int): Random seed.
+        train_frac (float): Training dataset proportion, between 0 and 1.
+        test_frac (float): Testing dataset proportionm between 0 and 1.
+        log_directory (string): Output folder.
```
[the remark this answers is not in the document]
mariehbourget: The original function should be replaced by this one in the near future. I don't think it is worth changing at this point.

ivadomed/loader/utils.py
andreanne-lemay: Ah you're right this was it! There are no dataset_description.json file in derivatives (or labels) in our testing data. In a future PR, we should probably add a step: if there is no dataset_description.json in derivatives, but there is one in the main bids path, then simply copy it there. Most of our bids dataset don't have dataset_description at this location, so this would avoid the error.

ivadomed/loader/utils.py
```diff
@@ -95,6 +96,50 @@ def split_dataset(df, center_test_lst, split_method, random_seed, train_frac=0.8
...
+        data_value, _ = train_test_split(data, train_size=test_frac, random_state=random_seed)
+
+    X_test = df[df[data_type].isin(data_value)][data_type].unique().tolist()
+    X_remain = df[~df[data_type].isin(data_value)][data_type].unique().tolist()
+
+    # split using sklearn function
+    X_train, X_tmp = train_test_split(X_remain, train_size=train_frac, random_state=random_seed)
+    if X_test:  # X_test contains data from data_value unseen during the training, eg centers in SpineGeneric
+        X_val = X_tmp
+    else:  # X_test contains data from data_value seen during the training, eg centers in gm_challenge
```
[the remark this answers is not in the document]
mariehbourget: Good catch, thanks! This is from the generalization of the "per_center" method where the "test_fraction" was not used and valid/test were splitted 50/50. I'll fix it and also the case where the test_fraction is 0 (not working as expected at the moment).

ivadomed/loader/utils.py
mariehbourget: Good idea, I'll add it when dealing with the refactoring of the BIDSDataframe class (#595).

dev/df_new_loader.py
```diff
@@ -2,12 +2,14 @@
 ##############################################################
 #
 # This script is used to test the dataframe of the new loader
```
charleygros, line 4: We may want to move this script to a proper unit_test in the future :-)

ivadomed/config/config_new_loader.json
```diff
@@ -33,8 +33,7 @@
     "split_dataset": {
         "fname_split": null,
         "random_seed": 6,
-        "center_test": [],
-        "method": "per_patient",
+        "data_testing": {"data_type": "subject", "data_value":[]},
```
charleygros, line 36: So (my memory is failing again): the strategy is to dev the new loader with this config file. And once the new loader is all standing up on its feet --> we will do the conversion of the other config file to the new format, correct?

charleygros (approved):

ivadomed/config/config_new_loader.json
mariehbourget: Correct, this would happen after issue #596 is handled.

dev/df_new_loader.py
mariehbourget: This script will be obsolete soon when the new features are integrated in the pipeline following #596.

ivadomed/loader/utils.py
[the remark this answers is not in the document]
mariehbourget: Fixed in 35954b7.

mariehbourget:
@andreanne-lemay, @charleygros, I ran into an issue with the generalization of the split.

In the current `per_center` method. The documentation says that “all subjects are split so as not to mix institutions between the train/validation/test sets”. However, I think that a same center can be in train and valid sets, but a same center can not be in train and test or valid and test sets. Is it the case?

Because we decided to output a list of `data_type` (eg. `institution_id`) for the split, the function I wrote would not have allowed a same data_type into train and valid sets. For example, with 3 centers, all participants from center 1 would be in train set, all participants of center 2 in valid set and all participants from center 3 in test set. With 2 centers, the function would fail.

To fix this, I added a parameter called `split_output` in the config file. It controls what will be used to split and will be in the joblib file, either `partitipant_id` or `sample_id`. Again, we need sample here because of microscopy with multiple samples for a same subject.

The data_testing parameter remains the same and is used to isolate certain participants or samples in the test set. The `data_type` is needed only if different from `split_output`.

I'm not sure about this, seems a bit complicated to explain and not necessary user-friendly at first glance. Any thoughts or suggestions to improve this? (examples below) Thanks!

#### Example 1:
```
        "split_output": "participant_id",
        "data_testing": {"data_type": null, "data_value":[]},
```
Split is done per participant.

#### Example 2:
        "split_output": "participant_id",
        "data_testing": {"data_type": null, "data_value":["sub-unf01"]},
Split is done per participant, sub-unf01 is in the test set.

#### Example 3:
        "split_output": "participant_id",
        "data_testing": {"data_type": "institution_id", "data_value":["oxfordFmrib"]},
Split is done per partiticpant, all participants from oxfordFmrib are in the test set.
#### Example 4:
        "split_output": "sample_id",
        "data_testing": {"data_type": null, "data_value":["sample-data11"]},
Split is done per sample, sample-data11 is in the test set.

mariehbourget:
The PR is ready for re-review.

Here are the changes made following last meeting:

The `split_output` param was renamed `split_method`.
* This param specifies on which metadata (ex: participant_id, sample_id, etc) the train fraction is applied.

The `data_testing` param remains the same.
* It is optional and specifies `data_type` and `data_value` to include in the test dataset (ex: institution_id, disease, session, etc).
* If specified, the test fraction is applied to `data_type`.
* If not specified, `data_type` is the same as `split_method`.

New column in dataframe called `ivadomed_id`, where `ivadomed_id` is the filename minus the modality suffix and extension.
* Examples:
     * filename: `sub-cardiff03_T1w.nii.gz` --> ivadomed_id : `sub-cardiff03_T1w`
     * filename: `sub-rat3_ses-02_sample-data11_run-1_SEM.png` --> ivadomed_id: `sub-rat3_ses-02_sample-data11_run-1`
* In all cases, no matter the `split_method` or `data_type`, this is the information that is outputted in train, valid and test lists to the next step of the loader. 
* This was needed to deal with cases where there are multiple `data_type` for a same `split_method` (ex: multiple samples for a same participant in microscopy, multiple sessions for a same participant in longitudinal studies, etc), and to deal later with multichannel training.
* All modalities from a same image will be in the same dataset (train or valid ortest) whether the training is multichannel or not (as it is in the traditional loader).

#### Example 1:
```
        "split_method": "participant_id",
        "data_testing": {"data_type": null, "data_value":[]},
```
Split is done per `participant_id` according to train and test fractions.
Equivalent to `per_patient` method in the traditional loader.

#### Example 2:
```
        "split_method": "participant_id",
        "data_testing": {"data_type": institution_id, "data_value":[]},
```
The test set is split according to test fraction on` institution_id` list.
The remainder is split into train/valid sets according to train fraction on remaining `participant_id` list.
Participants from a same center can be mixed between train and validation sets.
Equivalent to `per_center` method when `center_test` is empty in the traditional loader.

#### Example 3:
```
        "split_method": "participant_id",
        "data_testing": {"data_type": institution_id, "data_value":[“oxfordFmrib”]},
```
The test set include all files from `oxfordFmrib`.
The remainder is split into train/valid sets according to train fraction on remaining `participant_id` list.
Equivalent to `per_center` method with `oxfordFmrib` in `center_test`.

#### Example 4:
```
        "split_method": "institution_id",
        "data_testing": {"data_type": null, "data_value":[]},
```
Split is done per institution_id` according to train and test fractions.
Participants from a same center cannot be mixed between train and validation sets.
This method does not exist in the traditional loader.

ivadomed/loader/utils.py
```diff
@@ -164,6 +230,73 @@ def get_new_subject_split(path_folder, center_test, split_method, random_seed,
...
+        for m, n, v in zip(subject_selection["metadata"], subject_selection["n"], subject_selection["value"]):
+            sampled_dfs.append(df[df[m] == v].sample(n=n, random_state=random_seed))
+
+        if len(sampled_dfs) != 0:
+            df = pd.concat(sampled_dfs)
+
+    # If balance, then split the dataframe for each categorical value of the "balance" column
+    if balance:
+        if balance in df.keys():
+            df_list = [df[df[balance] == k] for k in df[balance].unique().tolist()]
```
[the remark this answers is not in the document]
andreanne-lemay: Needed since derivatives are in the df and all metadata is set to nan.

ivadomed/loader/utils.py
```diff
@@ -194,6 +327,36 @@ def get_subdatasets_subjects_list(split_params, bids_path, log_directory, subjec
...
+    Returns:
+        list, list list: Training, validation and testing subjects lists.
+    """
+    if split_params["fname_split"]:
+        # Load subjects lists
+        old_split = joblib.load(split_params["fname_split"])
+        train_lst, valid_lst, test_lst = old_split['train'], old_split['valid'], old_split['test']
+    else:
+        train_lst, valid_lst, test_lst = get_new_subject_split_new(df=df,
+                                                               split_method=split_params['split_method'],
```
andreanne-lemay, line 348: idem indentation

ivadomed/loader/utils.py
```diff
@@ -164,6 +230,73 @@ def get_new_subject_split(path_folder, center_test, split_method, random_seed,
...
+    Returns:
+        list, list list: Training, validation and testing subjects lists.
+    """
+    if subject_selection is not None:
+        # Verify subject_selection format
+        if not (len(subject_selection["metadata"]) == len(subject_selection["n"]) == len(subject_selection["value"])):
+            raise ValueError("All lists in subject_selection parameter should have the same length.")
+
+        sampled_dfs = []
+        for m, n, v in zip(subject_selection["metadata"], subject_selection["n"], subject_selection["value"]):
```
andreanne-lemay, line 261: This function needs to change since in the dataframe, one line is not necessarily one participant. If I one 5 subject with `sex` `M` I'll get 3 lines that might be from the same subject but different modalities. This worked on my end:
```
random.seed(random_seed)
        for m, n, v in zip(subject_selection["metadata"], subject_selection["n"], subject_selection["value"]):
            participants = random.sample(df[df[m] == v]['participant_id'].unique().tolist(), n)
            for participant in participants:
                sampled_dfs.append(df[df['participant_id'] == participant])
```

dev/df_new_loader.py
```diff
@@ -34,11 +36,32 @@
...
+log_directory = context["log_directory"]
+if not os.path.isdir(log_directory):
+    print('Creating log directory: {}'.format(log_directory))
+    os.makedirs(log_directory)
+else:
+    print('Log directory already exists: {}'.format(log_directory))
+
+# SPLIT TRAIN/VALID/TEST (with "new" functions)
+train_lst, valid_lst, test_lst = imed_loader_utils.get_subdatasets_subjects_list_new(context["split_dataset"],
+                                                                                        df2,
```
andreanne-lemay, line 61: indentation (should be aligned with the param on the previous line). If you use PyCharm, ctrl+alt+L will fix all indentation and minor PEP8 errors automatically.

ivadomed/loader/utils.py
```diff
@@ -94,6 +95,71 @@ def split_dataset(df, center_test_lst, split_method, random_seed, train_frac=0.8
...
+    data_value = data_testing['data_value']
+
+    if not split_method in df:
+        raise KeyError("No split_method '{}' was not found in metadata".format(split_method))
+    if not data_type in df:
+        logger.warning("No data_type named '{}' was found in metadata. Not taken into account "
+                        "to split the dataset.".format(data_type))
+        data_type = split_method
+
+
```
andreanne-lemay, line 127:
```suggestion
```

andreanne-lemay (approved):
2 minor changes necessary to accomodate `balance` and `subject_selection`, but the rest LGTM. thanks a lot for this PR :pray: !

ivadomed/loader/utils.py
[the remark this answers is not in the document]
mariehbourget: Fixed as suggested in a3d03dd

ivadomed/loader/utils.py
mariehbourget: Thank you Andreanne for all the testing! I had missed the subject_selection part.
I fixed it as you suggested,.

Issue #590, opened 2020-12-22
HDF5 file reading/writing in `loader/adaptative.py`


ahill187:
## Issue description

In the file `loader/adaptative.py`, several of the classes/functions read or write to `*.hdf5` files. According to the documentation (https://docs.h5py.org/en/stable/high/file.html):

> HDF5 files work generally like standard Python file objects. They support standard modes like r/w/a, and should be closed
> when they are no longer in use.

Throughout the `adaptative.py` file, `*.hdf5` files are opened in read or write mode, but none of the files are being closed. This is leading to errors in this pull request, which adds files for testing: https://github.com/ivadomed/ivadomed/pull/580. Leaving files open can lead to issues, such as not being able to open a file because it hasn't been closed properly.

## Solution

The code just needs to be refactored to either:

```
f = h5py.File(hdf5_file_name, "w")
# do something here
f.close()
```

or 

```
with h5py.File(hdf5_file_name, "w") as f:
    # do something here
```

PR #592, opened 2020-12-22
Fix loader/adaptative.py code with reading/writing HDF5 files

ahill187:
## Description

In the file `loader/adaptative.py`, several of the classes/functions read or write to `*.hdf5` files. According to the documentation (https://docs.h5py.org/en/stable/high/file.html):

> HDF5 files work generally like standard Python file objects. They support standard modes like r/w/a, and should be closed
> when they are no longer in use.

Throughout the `adaptative.py` file, `*.hdf5` files are opened in read or write mode, but none of the files are being closed. This is leading to errors in this pull request, which adds files for testing: https://github.com/ivadomed/ivadomed/pull/580. Leaving files open can lead to issues, such as not being able to open a file because it hasn't been closed properly.

## This Pull Request Achieves...

1. In this pull request, I have fixed the `*.hdf5*` read/write issue by using:

```
with h5py.File(hdf5_file_name, "w") as f:
    # do something here
```

2. I renamed `Bids_to_hdf5` -> `BIDSToHDF5` and `HDF5_to_Bids` -> `HDF5ToBIDS` to be `PEP8` compliant (PascalCase)

## Linked issues
Resolves: https://github.com/ivadomed/ivadomed/issues/590

### DIFF
--- a/ivadomed/loader/adaptative.py
+++ b/ivadomed/loader/adaptative.py
@@ -17,11 +17,12 @@ from ivadomed.object_detection import utils as imed_obj_detect
 class Dataframe:
     """
     This class aims to create a dataset using an HDF5 file, which can be used by an adapative loader
-    to perform curriculum learning, Active Learning or any other strategy that needs to load samples in a specific way.
+    to perform curriculum learning, Active Learning or any other strategy that needs to load samples
+    in a specific way.
     It works on RAM or on the fly and can be saved for later.
 
     Args:
-        hdf5 (hdf5): hdf5 file containing dataset information
+        hdf5_file (hdf5): hdf5 file containing dataset information
         contrasts (list of str): List of the contrasts of interest.
         path (str): Dataframe path.
         target_suffix (list of str): List of suffix of targetted structures.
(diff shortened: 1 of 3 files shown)


andreanne-lemay (approved):
LGTM! Thank you for taking care of this :100:

ivadomed/loader/adaptative.py
```diff
@@ -17,11 +17,12 @@
...
     """
     This class aims to create a dataset using an HDF5 file, which can be used by an adapative loader
-    to perform curriculum learning, Active Learning or any other strategy that needs to load samples in a specific way.
+    to perform curriculum learning, Active Learning or any other strategy that needs to load samples
+    in a specific way.
     It works on RAM or on the fly and can be saved for later.
 
     Args:
-        hdf5 (hdf5): hdf5 file containing dataset information
+        hdf5_file (hdf5): hdf5 file containing dataset information
```
jcohenadad, line 25: it’s minor but we’re trying to stick to a convention for variables: https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Variables#variables

so in this case: file_hdf5, and same for path_hdf5 below.

which makes me realize: we should add this internal doc for ivadomed coding rules in our wiki (could you open an issue pls?). thanks!

ahill187: Oh cool good to know! And yes I can open an issue.
In this case, `hdf5_file` is actually an open file, so it isn't a `string`, it's a file object. Should I still call it `file_hdf5`? Or should we call it something else?

jcohenadad: ah! in that case it is _better_ to not call it `file_hdf5` (otherwise that would be super confusing). So yes, I think you can keep your original naming 😊

PR #541, opened 2020-11-25
Fix HeMIS related bugs

andreanne-lemay:
`task` attribute is missing from the adaptive loader

The adaptive loader was not adapted to multiclass training.

Changes include:
- change None to 'None' since it is not a supported data type for h5py
- store all GT 
- `target_lst` became `target_contrast`. `target_contrast` indicates which GT to use (from which contrast)
- `roi_lst` became `roi_contrast`. `roi_contrast` indicates which ROI to use (from which contrast)

Old behavior: when `target_lst`had a length > 1, all GTs were used (the same way we store multiclass GT). For example, `target_lst`: ["T1w", "T2w"] would lead to a GT of size 2xHxW, but the prediction of the model would be 1xHxW for single class models. On multichannel, we want to pick a single GT for each class (from one of the contrast) to have the same behavior we have on the regular loader.

The adaptive loader saves a single dataset file (.hdf5) so the same dataset is used for validation, training and testing. Changes introduced in this PR fixes this behavior.

Remaining issues with HeMIS for future PRs:
- The adaptive loader works only for HeMIS (2D, multichannel)
- We can only test on datasets with all input channels

PR #594, opened 2020-12-23
Add new filter to SliceFilter class

leandervaneekelen:
#### GitHub

#### PR contents

## Description
For most typical multi-class models, all ground truth must be available when calculating the loss after the forward pass during training (e.g. when training a two class fully convolutional U-Net or HeMIS architecture). Currently, the `SliceFilter` class in `loader/utils.py`, responsible for filtering slices from the dataset, doesn't have an option to filter away slices that are missing _one or more_ classes (it just throws away slices without _any_ ground truth). This leaves the possibility of slices where, for example, spinal cord is annotated but not CSF - which throws off loss calculation.

This PR introduces `filter_absent_mask` as an option to catch this edge case. Documentation in the docstring/RTD pages is updated to match.


leandervaneekelen:
Shape errors were caused by an (in hindsight) erroneous commit (44d2bf0) in #541 - without it, things work smoothly.

charleygros (approved):
Sounds good to me!
But it is good to keep this parameter `False` as default value: let's imagine a task to segment both SC and lesion at the same time: we will have a lot of slices where the lesion mask will be empty.

My only suggestion would be to rename it to `filter_absent_class` --> make the difference between this filter and `filter_empty_mask` even more clear. And use "class" instead of "mask" in the documentation. What do you reckon @leandervaneekelen ?

leandervaneekelen:
> But it is good to keep this parameter False as default value: let's imagine a task to segment both SC and lesion at the same time: we will have a lot of slices where the lesion mask will be empty.

Agreed, default setting is false.

> rename it to `filter_absent_class`

Good suggestion, I originally didn't know what to name this! I'll change it.

charleygros (approved):

PR #602, opened 2021-01-05
Make sure test_HeMIS runs tests in order

ahill187:
Pytest doesn't always run tests in the order in which they are presented in the script, so you need to specify the order to run them if the order matters. In `test_HeMIS.py`, one of the files created in one test affects the other test, so the order matters. This pull request specifies the order using `pytest.mark`


gisellemartel (approved):
Tested locally, all tests pass and the tests execute in the intended order.

PR #612, opened 2021-01-10
Clarified doc on config file and fixed few typos/formatting issues

jcohenadad:
Fixes a few issues.

### Additional questions

Feedback needed @charleygros @andreanne-lemay @olix86 @lrouhier @mariehbourget 

https://github.com/ivadomed/ivadomed/blob/fbfbfcc4d91f91d0d6e974001a0c87a48f33f059/docs/source/configuration_file.rst#L327

The default config.json file uses "bn_momentum" instead:

https://github.com/ivadomed/ivadomed/blob/0718d6c6fd8824cef794b01866b303c869901c13/ivadomed/config/config.json#L73

---

https://github.com/ivadomed/ivadomed/blob/fbfbfcc4d91f91d0d6e974001a0c87a48f33f059/docs/source/configuration_file.rst#L328

About the "relu" field: why not instead having a field called "final_activation", to allow this field to be explicit (otherwise, how would someone know that the default is sigmoid?). Also, this would be more consistent with the Unet model, which takes the final_activation value as input (not "relu"). 

---

https://github.com/ivadomed/ivadomed/blob/fbfbfcc4d91f91d0d6e974001a0c87a48f33f059/docs/source/configuration_file.rst#L329

I don't see any field `is_2D`. Maybe it is `is_dim` and it was later updated without updating this comment?

---

https://github.com/ivadomed/ivadomed/blob/fbfbfcc4d91f91d0d6e974001a0c87a48f33f059/docs/source/configuration_file.rst#L330

I find `is_dim` confusing, because "is_something" refers to the question: "is something true or false", so the logic seems different here: the answer is not a boolean, but "2D" or "3D". I suggest changing the field for "dim".

---


andreanne-lemay:
I agree with you. Your suggestions are already implemented, the documentation is simply not up to date: 
- There is no `is_dim` it is `is_2d` (as you suggested) https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config_default.json#L76. 
- The `relu` field doesn't exist anymore it is rather `final_activation` (as you suggested) https://github.com/ivadomed/ivadomed/blob/master/ivadomed/models.py#L476.

I'll fix this discrepancy between code and documentation tomorrow.

docs/source/configuration_file.rst
```diff
@@ -324,11 +324,12 @@ to the tailored architecture.
 
 - ``name``: ``Unet`` (default)
 - ``dropout_rate``: Float (e.g. 0.4).
-- ``batch_norm_momentum``: Float (e.g. 0.1).
+- ``bn_momentum``: Float (e.g. 0.1).
```
jcohenadad, line 327: could you also add a description about this field? (and maybe a ref)

docs/source/configuration_file.rst
```diff
@@ -324,11 +324,12 @@ to the tailored architecture.
...
 - ``dropout_rate``: Float (e.g. 0.4).
-- ``batch_norm_momentum``: Float (e.g. 0.1).
-- ``depth``: Strictly positive integer. Number of down-sampling operations. - ``relu`` (optional): Bool.
-  Sets final activation to normalized ReLU (relu between 0 and 1), instead of sigmoid. Only available when `is_2D=True`.
-- ``is_dim``: Indicates dimensionality of model (2D or 3D). If ``is_dim`` is ``False``, then parameters ``length_3D`` and
-  ``stride_3D`` for 3D loader need to be specified (see :ref:`Modified3DUNet <Modified3DUNet>`).
+- ``bn_momentum``: Float (e.g. 0.1).
+- ``depth``: Strictly positive integer. Number of down-sampling operations. 
+- ``final_activation`` (optional): Bool. Defines the final activation layer. Options: ``sigmoid`` (default), ``relu``
+  (normalized ReLU), or ``softmax``.
```
jcohenadad, line 330: removed "Bool"
```suggestion
- ``final_activation`` (optional): Final activation layer. Options: ``sigmoid`` (default), ``relu`` 
(normalized ReLU), or ``softmax``.
```

docs/source/configuration_file.rst
```diff
@@ -324,11 +324,12 @@ to the tailored architecture.
...
-- ``batch_norm_momentum``: Float (e.g. 0.1).
-- ``depth``: Strictly positive integer. Number of down-sampling operations. - ``relu`` (optional): Bool.
-  Sets final activation to normalized ReLU (relu between 0 and 1), instead of sigmoid. Only available when `is_2D=True`.
-- ``is_dim``: Indicates dimensionality of model (2D or 3D). If ``is_dim`` is ``False``, then parameters ``length_3D`` and
-  ``stride_3D`` for 3D loader need to be specified (see :ref:`Modified3DUNet <Modified3DUNet>`).
+- ``bn_momentum``: Float (e.g. 0.1).
+- ``depth``: Strictly positive integer. Number of down-sampling operations. 
+- ``final_activation`` (optional): Final activation layer. Options: ``sigmoid`` (default), ``relu`` 
+(normalized ReLU), or ``softmax``.
+- ``is_2d``: Indicates if the model is 2d, if not the model is 3d. If ``is_2d`` is ``False``, then parameters
```
jcohenadad, line 331: how do you feel about changing that for `dim`?  It could then generalize to 1d (optical signal, EEG traces) and 4d (cardiac flow data, DWI data, etc.). If you think it's relevant let's open an issue.

docs/source/configuration_file.rst
andreanne-lemay: Added here: 77150bb . I also changed the default value in our config files to match the one defined by PyTorch (9fc873f)

docs/source/configuration_file.rst
andreanne-lemay: Currently, there is no Conv4D PyTorch class, so even if the data is 1d or 2d, we'll treat it as 3d data. For the 1d signals, my understanding is that we would also treat it as 2d (or 3d) data: electrodes and signal / electrode (2d). For now, I'd go simply with 2d or 3d.

docs/source/configuration_file.rst
```diff
@@ -932,27 +932,33 @@ being used for the segmentation task).
            "dropout_rate": {
                "type": "float"
            },
-           "batch_norm_momentum": {
-               "type": "float"
+           "bn_momentum": {
+               "type": "float",
+               "$$description": [
+                    "Defines the importance of the running average: (1 - `bn_momentum`). A large running\n",
```
andreanne-lemay, line 938: The list of strings is not very convenient. @ahill187 would there be a way to automatically set the width of the column so we could simply enter a string in the description field instead of needing to manually assess where we want our new lines?

ahill187: @andreanne-lemay ah yes this is quite annoying haha. The current implementation of the package `sphinx-jsonschema` doesn't seem to allow for that, however I could see about writing a patch that fixes it?

andreanne-lemay: that would be great! But this is not an urgent matter, I can open an issue about this.

ahill187: Sounds good!

ahill187 (approved):

jcohenadad:
@ahill187 feel free to merge whenever you want

ahill187:
> @ahill187 feel free to merge whenever you want

Great, I'll merge once these tests pass :)

PR #622, opened 2021-01-11
Add the Twitter badge

nekohayo:
"There can be only one!"… said our competition.


jcohenadad (approved):
label was missing-- otherwise looks good 👍

PR #624, opened 2021-01-11
soft_gt param: only active after Data Augmentation

charleygros:
## Description
Currently `soft_gt` impacts the GT loading in two locations:
1. when loading the nifti data: if `True` then load in `float32`, else in `uint8`
2. after the Data augmentation operations and before feeding the network: if `True` then no thresholding, else 0.5-thresholding.

Proposition:
1. Removed --> load all data to float32 format --> allows to have a soft data as input of Data AUgmentation operations.
2. Kept.
--> `soft_gt` only impacts the GT that is fed to the network (which is actually what is described in the RTD.. ..)


andreanne-lemay (approved):
Looks good to me! In `__getitem__` I would cast the gt to uint8 when [soft_gt=True](URL) to keep the old behavior. Currently, it is only binarized an cast to np.int by the `threshold_predictions` function.

PR #626, opened 2021-01-12
Add a README for the Sphinx docs

ahill187:
## Description
The purpose of this pull request is to add a `README` file on getting started with `Sphinx` documentation. It describes the commands needed to run/build the docs, and also how to view them. The audience of this `README` is developers, who may not be totally familiar with `Sphinx` and how to build it.


jcohenadad:
This is a great idea, but it duplicates information with [our wiki](https://github.com/ivadomed/ivadomed/wiki/documentation).

We should probably only maintain this doc at one place, but the question is then: where?  tagging @joshuacwnewton because he is on top of our internal-doc across projects  😊

joshuacwnewton:
> This is a great idea, but it duplicates information with [our wiki](https://github.com/ivadomed/ivadomed/wiki/documentation).
> 
> We should probably only maintain this doc at one place, but the question is then: where? tagging @joshuacwnewton because he is on top of our internal-doc across projects blush

I agree with what @jcohenadad is saying. For NeuroPoly projects, I think the repo's Wiki should be the place a developer looks for internal documentation about the project. (In fact, SCT has a [similar Sphinx/RTD page](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) on its own internal Wiki.)

In general, my thoughts are similar to what [this blog post](https://www.danclarke.com/internal-developer-documentation) says about internal documentation. :slightly_smiling_face:

ahill187:
From a development standpoint, I find having basic documentation in `README`s a lot easier, because it is open in my editor and I can just reference the file quite quickly. When it is on a wiki, I have to navigate between a browser and my code. 
Secondly, having `README` files allows them to be tracked/version controlled via `GitHub`, which you don't have with the wiki.
I think having both is valid and useful, we don't necessarily have to choose?

joshuacwnewton:
> From a development standpoint, I find having basic documentation in `README`s a lot easier, because it is open in my editor and I can just reference the file quite quickly. When it is on a wiki, I have to navigate between a browser and my code.
> Also, having `README` files allows them to be tracked/version controlled via `GitHub`, which you don't have with the wiki.
> I think having both is valid and useful, we don't necessarily have to choose? :)
> @joshuacwnewton I agree with the blog post, documentation is super key!

Those are very fair points! One small point, though: the GitHub Wiki is indeed version controlled. The UI doesn't show it very well, but you can clone `https://github.com/neuropoly/spinalcordtoolbox.wiki.git` (ditto for ivadomed or any repo) and access each page via Markdown files. Each edit is its own commit, too. 

What I do is keep both attached together in my PyCharm window, so I don't have to do the browser-code-back-and-forth that you mention. 

Screenshot from 2021-01-12 13-39-48

gisellemartel:
I think I agree with @ahill187 on this one. While this change would require both the wiki and README to be maintained, I would assume changes to the sphinx build pipeline would occur rarely.

jcohenadad (approved):

ahill187:
> > From a development standpoint, I find having basic documentation in `README`s a lot easier, because it is open in my editor and I can just reference the file quite quickly. When it is on a wiki, I have to navigate between a browser and my code.
> > Also, having `README` files allows them to be tracked/version controlled via `GitHub`, which you don't have with the wiki.
> > I think having both is valid and useful, we don't necessarily have to choose? :)
> > @joshuacwnewton I agree with the blog post, documentation is super key!
> 
> Those are very fair points! One small point, though: the GitHub Wiki is indeed version controlled. The UI doesn't show it very well, but you can clone `https://github.com/neuropoly/spinalcordtoolbox.wiki.git` (ditto for ivadomed or any repo) and access each page via Markdown files. Each edit is its own commit, too.
> 
> What I do for each project is keep both attached together in a PyCharm window, so I don't have to do the browser-code-back-and-forth that you mention.
> 
> Screenshot from 2021-01-12 13-39-48

@joshuacwnewton Good to know, I didn't realize that!

PR #631, opened 2021-01-12
Remove the testing_data folder from ivadomed

ahill187:
## Description

It seems that the `testing_data` folder in the `ivadomed` package is redundant because we download all of our data from the repo `data-testing`? Is there any reason that we need it? If not, I think it makes sense to remove it.


jcohenadad:
let's see if tests pass 😬

andreanne-lemay (approved):
Looks like the [tests passed](https://github.com/ivadomed/ivadomed/runs/1689811395) :tada:

PR #636, opened 2021-01-14
Unit Test cleanup

ahill187:
## Description

Currently, the unit tests use/create a lot of files that aren't cleaned up afterwards or in between tests. This can result in unit tests running fine alone and failing in conjunction with other tests due to a file being modified. This pull request addresses the issue by creating a base file, `t_utils`, which handles file cleaning.

### `t_utils` and `tmp` folder
The code will now do the following:

1. Download `data-testing` once at the beginning of the test suite and save it as `data_testing`.
2. Before each test, the `setup_function` makes a new folder called `tmp`, and copies `data_testing` into that folder.
```
data_testing/
tmp/
|-- data_testing/
```
3. Any files created during the test will be saved in `tmp`.
4. After each test, the `teardown_function` deletes the `tmp` folder.

### `ivado_test` and `model_unet_test.pt`
The `model_unet_test.pt` file is recreated each time, so this pull request addresses moving it to the `data-testing` repo: https://github.com/ivadomed/data-testing/pull/8

### `test_HeMIS` and `mytestfile.hdf5`
The tests in `test_HeMIS` rely on each other currently via the creation of the file `mytestfile.hdf5`. This means that if one test fails, the other one will too. The tests should be independent of each other, so I have added the file to the `data-testing` repo: https://github.com/ivadomed/data-testing/pull/8


jcohenadad (approved):
LGTM!

PR #637, opened 2021-01-14
Add template for tests

ahill187:
## Description
I've added two test templates, one for each of `unit_tests` and `functional_tests`. The purpose of these files is to provide a template that people can use for creating new tests.


ahill187:
Whoops this is the wrong branch.

PR #638, opened 2021-01-14
Add test template files

ahill187:
## Description
I've added two test templates, one for each of unit_tests and functional_tests. The purpose of these files is to provide a template that people can use for creating new tests.

### DIFF
--- a/testing/functional_tests/t_template.py
+++ b/testing/functional_tests/t_template.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""Template for Fucntional Tests
+"""Template for Functional Tests
 
 This file is a template for writing functional tests. It will not be run during testing, but you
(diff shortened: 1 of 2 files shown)


testing/functional_tests/t_template.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+"""Template for Fucntional Tests
```
andreanne-lemay, line 2:
```suggestion
"""Template for Functional Tests
```

andreanne-lemay (approved):
Great addition! Very useful, thanks!

testing/functional_tests/t_template.py
ahill187: Oh whoops! Nice catch :)

PR #641, opened 2021-01-18
Add metavar to parser

### LINKED ISSUE #627: Display expected metavar in the parser for CLI scripts

It is sometimes unclear what input is expected for a given CLI flag, see for example #613.

For better clarity, we could make explicit the type of input expected by each flag. To do so, we could duplicate what is done in SCT: 
- create a [parser class which contains possible metavar types](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/utils/shell.py#L186)
- Use it in CLI parser calls, example of use [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_deepseg.py#L42)

<details><summary>See how usage is rendered</summary>

```
sct_deepseg -h
--

usage: sct_deepseg [-i <file> [<file> ...]] [-c <file> [<file> ...]] [-o <str>] [-task <str> [<str> ...]] [-list-tasks]
                   [-install-task {seg_sc_t2star,seg_mice_sc,seg_mice_gm,seg_tumor_t2,seg_tumor-edema-cavity_t1-t2}] [-thr <float>] [-r {0,1}]
                   [-largest KEEP_LARGEST] [-fill-holes {0,1}] [-remove-small REMOVE_SMALL [REMOVE_SMALL ...]] [-v <int>] [-h]

Segment an anatomical structure or pathologies according to the specified deep learning model.

INPUT/OUTPUT:
  -i <file> [<file> ...]
                        Image to segment. Can be multiple images (separated with space).
  -c <file> [<file> ...]
                        Type of image contrast. Indicates the order in which the images have been presented with -i. Optional if only one image
                        is specified with -i. The contrasts should be separated by spaces (e.g., -c t1 t2).
  -o <str>              Output file name. In case of multi-class segmentation, class-specific suffixes will be added. By default,suffix '_seg'
                        will be added and output extension will be .nii.gz.

TASKS:
  -task <str> [<str> ...]
                        Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task. To list
                        available tasks, run: sct_deepseg -list-tasks. To use a custom task, indicate the path to the ivadomed packaged model.
                        More than one path can be indicated for cascaded application of the models. See tutorials at https://ivadomed.org/ for
                        more details.
  -list-tasks           Display a list of tasks that can be achieved. (default: False)
  -install-task {seg_sc_t2star,seg_mice_sc,seg_mice_gm,seg_tumor_t2,seg_tumor-edema-cavity_t1-t2}
                        Install models that are required for specified task.

PARAMETERS:
  -thr <float>          Binarize segmentation with specified threshold. Set to 0 for no thresholding (i.e., soft segmentation). Default value is
                        model-specific and was set during optimization (more info at https://github.com/sct-pipeline/deepseg-threshold).
  -r {0,1}              Remove temporary files. (default: 1)
  -largest KEEP_LARGEST
                        Keep the largest connected-objects from the output segmentation. Specify the number of objects to keep.To keep all
                        objects, set to 0
  -fill-holes {0,1}     Fill small holes in the segmentation.
  -remove-small REMOVE_SMALL [REMOVE_SMALL ...]
                        Minimal object size to keep with unit (mm3 or vox). A single value can be provided or one value per prediction class.
                        Single value example: 1mm3, 5vox. Multiple values example: 10 20 10vox (remove objects smaller than 10 voxels for class
                        1 and 3, and smaller than 20 voxels for class 2).

MISC:
  -v <int>              Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode (default: 1)
  -h, --help            Show this help message and exit
```

</details>

ahill187:
## Description
This pull request adds a class called `Metavar` to `ivadomed.utils`. This class defines different input types for command line arguments. All of the command line tools have been updated to implement this class. The purpose is to give users a better idea of what the input type should be.

For example, in `ivadomed_automate_training`:
```
usage: ivadomed_automate_training [-h] -c <file> -p <file> [-n <int>] [--all-combin] [-m] [--run-test]
                                  [--fixed-split] [-l] [-t <float>]

optional arguments:
  -c <file>, --config <file>
                        Base config file path.
  -p <file>, --params <file>
                        JSON file where hyperparameters to experiment are listed.
  -n <int>, --n-iterations <int>
                        Number of times to run each config.
  --all-combin          To run all combinations of config
  -m, --multi-params    To change multiple parameters at once.
  --run-test            Evaluate the trained model on the testing sub-set.
  --fixed-split         Keep a constant dataset split for all configs and iterations
  -l, --all-logs        Keep all log directories for each iteration.
  -t <float>, --thr-increment <float>
                        A threshold analysis is performed at the end of the training using the trained model
                        and the validation sub-dataset to find the optimal binarization threshold. The
                        specified value indicates the increment between 0 and 1 used during the analysis (e.g.
                        0.1).
```

## Linked issues
https://github.com/ivadomed/ivadomed/issues/627

### DIFF
--- a/ivadomed/scripts/automate_training.py
+++ b/ivadomed/scripts/automate_training.py
@@ -24,7 +24,7 @@ import torch.multiprocessing as mp
 
 from ivadomed import main as ivado
 from ivadomed import config_manager as imed_config_manager
-from ivadomed.utils import init_ivadomed
+from ivadomed.utils import init_ivadomed, Metavar
 from ivadomed.loader import utils as imed_loader_utils
 from ivadomed.scripts.compare_models import compute_statistics
 
@@ -34,11 +34,15 @@ logging.basicConfig(filename=LOG_FILENAME, level=logging.DEBUG)
 
 def get_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument("-c", "--config", required=True, help="Base config file path.")
-    parser.add_argument("-p", "--params", required=True, help="JSON file where hyperparameters to experiment are "
-                                                              "listed.")
+    parser.add_argument("-c", "--config", required=True, help="Base config file path.",
+                        metavar=Metavar.file)
+    parser.add_argument("-p", "--params", required=True,
+                        help="""JSON file where hyperparameters to experiment are
+                                listed.""",
+                        metavar=Metavar.file)
     parser.add_argument("-n", "--n-iterations", dest="n_iterations", default=1,
-                        type=int, help="Number of times to run each config.")
+                        type=int, help="Number of times to run each config.",
+                        metavar=Metavar.int)
     parser.add_argument("--all-combin", dest='all_combin', action='store_true',
                         help="To run all combinations of config"),
     parser.add_argument("-m", "--multi-params", dest="multi_params", action='store_true',
@@ -50,9 +54,11 @@ def get_parser():
     parser.add_argument("-l", "--all-logs", dest="all_logs", action='store_true',
                         help="Keep all log directories for each iteration.")
     parser.add_argument('-t', '--thr-increment', dest="thr_increment", required=False, type=float,
-                        help="A threshold analysis is performed at the end of the training using the trained model and "
-                             "the validation sub-dataset to find the optimal binarization threshold. The specified "
-                             "value indicates the increment between 0 and 1 used during the analysis (e.g. 0.1).")
+                        help="""A threshold analysis is performed at the end of the training using
+                                the trained model and the validation sub-dataset to find the optimal
+                                binarization threshold. The specified value indicates the increment
+                                between 0 and 1 used during the analysis (e.g. 0.1).""",
+                        metavar=Metavar.float)
 
     return parser
 
--- a/ivadomed/scripts/convert_to_onnx.py
+++ b/ivadomed/scripts/convert_to_onnx.py
@@ -7,7 +7,7 @@ from ivadomed.utils import init_ivadomed, Metavar, save_onnx_model
 def get_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument("-m", "--model", dest="model", required=True, type=str,
-                        help="Path to .pt model.", metavar=Metavar.str)
+                        help="Path to .pt model.", metavar=Metavar.file)
     parser.add_argument("-d", "--dimension", dest="dimension", required=True,
                         type=int, help="Input dimension (2 for 2D inputs, 3 for 3D inputs).",
                         metavar=Metavar.int)
@@ -15,7 +15,7 @@ def get_parser():
                         help="Number of input channels of the model.",
                         metavar=Metavar.int)
     parser.add_argument("-g", "--gpu", dest="gpu", default=0, type=str,
-                        help="GPU number if available.", metavar=Metavar.str)
+                        help="GPU number if available.", metavar=Metavar.int)
     return parser
 
 
--- a/ivadomed/scripts/convert_to_onnx.py
+++ b/ivadomed/scripts/convert_to_onnx.py
@@ -1,7 +1,7 @@
 import argparse
 import torch
 
-from ivadomed.utils import init_ivadomed, Metavar
+from ivadomed.utils import init_ivadomed, Metavar, save_onnx_model
 
 
 def get_parser():
@@ -42,7 +42,7 @@ def convert_pytorch_to_onnx(model, dimension, n_channels, gpu=0):
     model_net = torch.load(model, map_location=device)
     dummy_input = torch.randn(1, n_channels, 96, 96, device=device) if dimension == 2 \
                   else torch.randn(1, n_channels, 96, 96, 96, device=device)
-    imed_utils.save_onnx_model(model_net, dummy_input, model.replace("pt", "onnx"))
+    save_onnx_model(model_net, dummy_input, model.replace("pt", "onnx"))
 
 
 def main():
--- a/ivadomed/scripts/convert_to_onnx.py
+++ b/ivadomed/scripts/convert_to_onnx.py
@@ -1,17 +1,21 @@
 import argparse
 import torch
 
-from ivadomed import utils as imed_utils
+from ivadomed.utils import init_ivadomed, Metavar
 
 
 def get_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument("-m", "--model", dest="model", required=True, type=str, help="Path to .pt model.")
+    parser.add_argument("-m", "--model", dest="model", required=True, type=str,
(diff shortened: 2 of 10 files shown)


ivadomed/scripts/automate_training.py
```diff
@@ -34,11 +34,15 @@
 
 def get_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument("-c", "--config", required=True, help="Base config file path.")
-    parser.add_argument("-p", "--params", required=True, help="JSON file where hyperparameters to experiment are "
-                                                              "listed.")
+    parser.add_argument("-c", "--config", required=True, help="Base config file path.",
+                        metavar=Metavar.file)
+    parser.add_argument("-p", "--params", required=True,
+                        help="""JSON file where hyperparameters to experiment are
```
andreanne-lemay, line 40: Just out of curiosity: why use the triple vs single quotes (""" vs "). I see some help you modified have single quotes  ( line 44) and some triple quotes (line 40).

ivadomed/scripts/convert_to_onnx.py
```diff
@@ -1,17 +1,21 @@
...
 
-from ivadomed import utils as imed_utils
+from ivadomed.utils import init_ivadomed, Metavar, save_onnx_model
 
 
 def get_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument("-m", "--model", dest="model", required=True, type=str, help="Path to .pt model.")
+    parser.add_argument("-m", "--model", dest="model", required=True, type=str,
+                        help="Path to .pt model.", metavar=Metavar.str)
```
andreanne-lemay, line 10:
```suggestion
                        help="Path to .pt model.", metavar=Metavar.file)
```
The path should point to a ".pt" file.

ivadomed/scripts/convert_to_onnx.py
```diff
@@ -1,17 +1,21 @@
...
-                        type=int, help="Input dimension (2 for 2D inputs, 3 for 3D inputs).")
+                        type=int, help="Input dimension (2 for 2D inputs, 3 for 3D inputs).",
+                        metavar=Metavar.int)
     parser.add_argument("-n", "--n_channels", dest="n_channels", default=1, type=int,
-                        help="Number of input channels of the model.")
-    parser.add_argument("-g", "--gpu", dest="gpu", default=0, type=str, help="GPU number if available.")
+                        help="Number of input channels of the model.",
+                        metavar=Metavar.int)
+    parser.add_argument("-g", "--gpu", dest="gpu", default=0, type=str,
+                        help="GPU number if available.", metavar=Metavar.str)
```
andreanne-lemay, line 18:
```suggestion
                        help="GPU number if available.", metavar=Metavar.int)
```
Should be a gpu number

ivadomed/scripts/training_curve.py
```diff
@@ -9,26 +9,30 @@
...
                              "deviation envelope.")
     parser.add_argument("-y", "--ylim_loss", required=False, type=str,
-                        help="Indicates the limits on the y-axis for the loss plots, otherwise these limits are "
-                             "automatically defined. Please separate the lower and the upper limit by a comma, eg -1,0."
-                             "Note: for the validation metrics: the y-limits are always 0.0 and 1.0.")
+                        help="""Indicates the limits on the y-axis for the loss plots, otherwise
+                                these limits are automatically defined. Please separate the lower
+                                and the upper limit by a comma, eg -1,0. Note: for the validation
+                                metrics: the y-limits are always 0.0 and 1.0.""",
+                        metavar=Metavar.str)
```
andreanne-lemay, line 33:
```suggestion
                        metavar=Metavar.float)
```

andreanne-lemay (approved):
Very useful! Thank you!

ivadomed/scripts/automate_training.py
ahill187: Oh it's just cause the triple quotes allow for multi-line quotes, so you don't have to use new quotes with each line :)

ivadomed/scripts/convert_to_onnx.py
ahill187: Oh good call! I'll change that

ivadomed/scripts/convert_to_onnx.py
ahill187: I'll change that one :)

ivadomed/scripts/training_curve.py
ahill187: Right! thanks will fix that :)

ahill187: Fixed here: https://github.com/ivadomed/ivadomed/pull/641/commits/bc1691bdb29c57f758096efd0731c57e21580d82

ivadomed/scripts/convert_to_onnx.py
ahill187: Fixed here: https://github.com/ivadomed/ivadomed/pull/641/commits/bc1691bdb29c57f758096efd0731c57e21580d82

ivadomed/scripts/convert_to_onnx.py
ahill187: Fixed here: https://github.com/ivadomed/ivadomed/pull/641/commits/bc1691bdb29c57f758096efd0731c57e21580d82

ahill187:
For booleans in the command line, I think it makes sense to use `store_true`, so I changed the derivatives flag in `extract_small_dataset` to:
```
parser.add_argument("-d", "--include-derivatives",
                    action="store_true",
                    dest="include_derivatives",
                    help="""If true, include derivatives/labels content.""")
```
(https://github.com/ivadomed/ivadomed/pull/641/commits/70462ae9657d1f4fa789a1a2fcf023685f724ba4)

However, this now switches the default to excluding the derivatives, whereas before the default was to include them. I'm not sure if this would make the command line more cumbersome? What percent of the time do we want to include/exclude derivatives? 
I also considered changing the variable to `exclude_derivatives`, as then the default would be including them. But that leads to a lot of double negatives in the code, which I thought made it less readable.
Any thoughts?

jcohenadad:
> However, this now switches the default to excluding the derivatives, whereas before the default was to include them. I'm not sure if this would make the command line more cumbersome? What percent of the time do we want to include/exclude derivatives?
I also considered changing the variable to exclude_derivatives, as then the default would be including them. But that leads to a lot of double negatives in the code, which I thought made it less readable.
Any thoughts?

This is an eternal debate we have in SCT. In those cases, I am for not using store_true and simply ask for a 1/0 input. It makes the usage and code very explicit, and we can decide to change default behaviour in the future

ahill187:
Okay, that's a good point! I can switch it back to what it originally was :)

PR #460, opened 2020-09-28
Use custom data for film

andreanne-lemay:
Add option to use custom data to feed into film.

There was an error in `training.py`, the metadata of the first batch of the image was used  for all images in batch.

PR #642, opened 2021-01-18
Fix store film params

andreanne-lemay:
## Description
During training, when the `debugging` parameter is set to `True`, FiLM parameters should be saved. The function `store_film_params` in the `training.py` file was not updated after FiLM was generalized to custom data (#460) making the code deprecated. This PR fix this compatibility issue.


ivadomed/training.py
```diff
@@ -458,23 +461,24 @@ def get_metadata(metadata, model_params):
...
 
-def store_film_params(gammas, betas, contrasts, metadata, model, film_layers, depth):
+def store_film_params(gammas, betas, metadata_values, metadata, model, film_layers, depth, film_metadata):
     """Store FiLM params.
 
     Args:
         gammas (dict):
         betas (dict):
-        contrasts (list): list of the batch sample's contrasts (eg T2w, T1w)
+        metadata_values (list): list of the batch sample's metadata_values (eg T2w, T1w)
```
charleygros, line 470: Not sure to understand the difference between `metadata_values` and `film_metadata`: could you please help me? :-)

charleygros: Oh, is `film_metadata` a dict key? so that we can select a specific metadata. If so, all good! Maybe make it clearer in docstring? Thanksss

charleygros (approved):

ivadomed/training.py
andreanne-lemay: film metadata is the metadata used for film e.g. contrast, tumor_type, task while metadata values is the value e.g. T1w, astrocytoma or spleen segmentation

andreanne-lemay: clarified here: 96638cf

PR #644, opened 2021-01-18
Standardize the gpu ID argument

### LINKED ISSUE #589: Clarify multi-GPU usage in documentation
Currently, training a model with the standard `ivadomed --config config.json` syntax doesn't support multi-GPU usage; this is only meant for `ivadomed_automate_training`. 

I found [this page](https://github.com/ivadomed/ivadomed/wiki/pipeline-training-and-model-comparison)  on the dev wiki explaining the usage of the `ivadomed_automate_training` and the syntax for multi-GPU usage; perhaps we can use this and link it in the [RTD page](https://ivadomed.org/en/latest/configuration_file.html#gpu) that covers the configuration file?

ahill187:
## Description
In our config files, generally there is a GPU argument which specifies the device ID to use:
```
"gpu": 0
```

However, in the config file for `automate_training`, the GPU argument is a list:

```
"gpu": [0,1,2]
```

For consistency, I think it makes sense to change the config files to all use the list method, as this will encompass both cases.

This pull request does the following:

1. Refactor terms such as `gpu_number` and `gpu` to either `gpu_id` or `gpu_ids`. 
  `gpu_id` (int): the GPU ID for one device
  `gpu_ids` (list)(int): a list of GPU IDs
2. Refactor config files and the `ivadomed.main`, which accepts config files as an argument, to use a list of GPU ids.

## Linked issues
https://github.com/ivadomed/ivadomed/issues/589

### DIFF
--- a/docs/source/configuration_file.rst
+++ b/docs/source/configuration_file.rst
@@ -29,10 +29,10 @@ sub-datasets, ``"test"``: evaluate a trained model on a testing sub-dataset, ``"
 segment a entire dataset using a trained model.
 
 
-gpu
-^^^
+gpu_ids
+^^^^^^^
 
-Integer. ID of the GPU to use.
+``List`` ``Integer``. List of IDs of one or more GPUs to use.
 
 log\_directory
 ^^^^^^^^^^^^^^
@@ -152,9 +152,9 @@ below.
 -  ``filter_absent_class``: Bool. Discard slices where all voxel
    labels are zero for one or more classes (this is most relevant for
    multi-class models that need GT for all classes at train time).
--  ``filter_classification``: Bool. Discard slices where all images fail 
+-  ``filter_classification``: Bool. Discard slices where all images fail
    a custom classifier filter. If used, ``classifier_path`` must also be
-   specified, pointing to a saved PyTorch classifier. 
+   specified, pointing to a saved PyTorch classifier.
 
 roi
 ^^^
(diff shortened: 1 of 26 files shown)


charleygros (approved):
Great idea!! Refactoring looks all gravy! Many thanks @ahill187 !

leandervaneekelen (approved):
This looks good! Thanks @ahill187 - one thing of note, #589 was intended more to point out that the stock train code doesn't support multi-GPU usage; could we explicitly point that out somewhere in the documentation, until we implement it? Maybe [here](https://ivadomed.org/en/latest/configuration_file.html) on the configuration file documentation, as that's what people are most likely to read while setting up models.

ahill187:
> This looks good! Thanks @ahill187 - one thing of note, #589 was intended more to point out that the stock train code doesn't support multi-GPU usage; could we explicitly point that out somewhere in the documentation, until we implement it? Maybe [here](https://ivadomed.org/en/latest/configuration_file.html) on the configuration file documentation, as that's what people are most likely to read while setting up models.

Oh good call! I'll add some docs for that :)

ahill187:
@leandervaneekelen Was this what you had in mind? https://github.com/ivadomed/ivadomed/pull/644/commits/5c261d4b74230392a7a367199755320a2935258b

leandervaneekelen:
Yep, nothing more than that is needed :)

PR #645, opened 2021-01-19
Change condition to save FiLM parameters 

andreanne-lemay:
## Description
FiLM parameters are only saved at the last epoch of the training. Two conditions can trigger the training to stop: the training reached the number of epochs defined by the user or the early stopping is triggered due to minimal improvement. The latter is not considered in the condition to save the FiLM parameters. If the training is stopped due to early stopping, the FiLM parameters are not saved. This PR fixes this issue.


olix86 (approved):
👍

PR #646, opened 2021-01-20
Fix bugs in plot film params

andreanne-lemay:
### Description
The t-sne and PCA were made for each batch instead of doing one with all data together. Before the results were highly impacted by the batch size. This PR fixes this issue.


dev/plot_film_parameters.py
```diff
@@ -22,40 +25,46 @@ def plot_histogram(data, layer_no, fname_out):
     :param layer_no: number of the layer to consider
     :param fname_out: directory to save the figure
     """
-    bins = np.linspace(-10, 10, 100)
+    bins = np.linspace(0, 1, 100)
```
charleygros, line 28: I thought we may expect negative values. Could you help me here?

dev/plot_film_parameters.py
```diff
@@ -22,40 +25,46 @@ def plot_histogram(data, layer_no, fname_out):
...
     :param data: input data, which are gammas or betas for one layer
-    :param contrast_images: numpy array with the contrast of all the images
+    :param metadata_values: numpy array with the metadata values of all the images
     :param layer_no: number of the layer to consider
     :param fname_out: directory to save the figure
     """
     pca_df = pd.DataFrame()
 
-    pca = PCA(n_components=3)
+    pca = PCA(n_components=2)
```
charleygros, line 53: 2 sounds good to me! No idea why 3 before

dev/plot_film_parameters.py
```diff
@@ -22,40 +25,46 @@ def plot_histogram(data, layer_no, fname_out):
...
+    # Dim 0 will be the number of examples
+    data = np.concatenate(list(data), axis=0)
+    metadata_values = np.array(functools.reduce(operator.iconcat, metadata_values, [])).ravel()
+
+    pca_result = pca.fit_transform(data)
+    pca_df2 = pd.DataFrame()
+    pca_df2['pca1'] = pca_result[:, 0]
+    pca_df2['pca2'] = pca_result[:, 1]
+    pca_df2['contrast'] = metadata_values
+    pca_df = pd.concat([pca_df, pca_df2])
```
charleygros, line 64: Great

charleygros (approved):
Beautiful

dev/plot_film_parameters.py
andreanne-lemay: Finally I kept the original FiLM implementation that restricts values from 0 to 1 with the sigmoid, it also gave me better results on my test dataset.

PR #647, opened 2021-01-20
Add functional test for automate_training run_test flag

ahill187:
## Description

The purpose of this pull request is to add functional testing for the `run_test` flag in the `automate_training.py` file.

### Issues with Testing Modules that Use Multiprocessing

I describe this here: https://github.com/ivadomed/ivadomed/wiki/tests. Not sure if anyone had any experience with this, or anything they wanted to add?

One thing to note is that when testing any of the command line scripts that use `multiprocessing`, you should use `script_runner` to call the module instead of `script_to_test.main(args)`. This is because if you do the latter, the processes aren't closed between the tests in your script.

For example, let's say you have a script, `mp_script.py`:
```
import multiprocessing as mp
import time

def sleep_for_a_bit(seconds):
    print(f"Sleeping for {seconds} second(s)")
    time.sleep(seconds)
    print("Done sleeping!")
    print(mp.current_process())

def main():
    pool = mp.Pool(processes=2)
    pool.map(func=sleep_for_a_bit, iterable=[1 for i in range(0, 2)])
    pool.close()    

if __name__ == "__main__":
    main()
```

Let's say you want to test this script using `script_to_test.main(args)`:

```
import mp_script

def test_mp_script_a():
    mp_script.main()


def test_mp_script_b():
    mp_script.main()
```

If you run this test, you will get something like:

```
Sleeping for 1 second(s)
Sleeping for 1 second(s)
Done sleeping!
Done sleeping!
<SpawnProcess name='SpawnPoolWorker-1' parent=91203 started daemon>
<SpawnProcess name='SpawnPoolWorker-2' parent=91203 started daemon>
Sleeping for 1 second(s)
Sleeping for 1 second(s)
Done sleeping!
Done sleeping!
<SpawnProcess name='SpawnPoolWorker-4' parent=91203 started daemon>
<SpawnProcess name='SpawnPoolWorker-3' parent=91203 started daemon>
```

Note here how between the two separate tests, `test_mp_script_a` and `test_mp_script_b`, the `SpawnPoolWorker`s aren't resetting? In this example, the tests would pass fine, because the script isn't doing anything other than sleeping. However, if your script relies on the processes to be distinct, there could be a problem, as in `automate_training.py`.

To correctly test, you will need to use `script_runner` from `pytest-console-scripts`:
```
import pytest

@pytest.mark.script_launch_mode('subprocess')
def test_mp_script_a(script_runner):
    ret = script_runner.run('mp_script')
    print(f"{ret.stdout}")
    print(f"{ret.stderr}")
    assert ret.success

@pytest.mark.script_launch_mode('subprocess')
def test_mp_script_b(script_runner):
    ret = script_runner.run('mp_script')
    print(f"{ret.stdout}")
    print(f"{ret.stderr}")
    assert ret.success
```

If you run this test, you should see something like:
```
Sleeping for 1 second(s)
Done sleeping!
<SpawnProcess name='SpawnPoolWorker-1' parent=91378 started daemon>
Sleeping for 1 second(s)
Done sleeping!
<SpawnProcess name='SpawnPoolWorker-2' parent=91378 started daemon>

Sleeping for 1 second(s)
Done sleeping!
<SpawnProcess name='SpawnPoolWorker-1' parent=91378 started daemon>
Sleeping for 1 second(s)
Done sleeping!
<SpawnProcess name='SpawnPoolWorker-2' parent=91378 started daemon>
```

As you can see, the `SpawnPoolWorker` is not being carried over between tests anymore.


gisellemartel (approved):
Testing locally and it works! Thanks for the very detailed explanation of multiprocessing and testing, very helpful!

ahill187:
Great! No problem, it took me a few days to figure that one out aha...

Issue #596, opened 2020-12-23
Refactor BidsDataset classes for new loader


mariehbourget:
## Context

This issue is part of the loader refactoring project to include microscopy data (#304) and to migrate to HDF5 loader (#474).

The step 1 of the refactoring was to generate a bids dataframe: done in #538 for anat and microscopy data.

The step 2 was to adapt the splitting methods (train/valid/test) to split using the dataframe and generalized the data used for splitting (subject, participant_id, sample, institution_id, disease, etc) : done in #591.

The step 3 is:
-	Refactor create_bids_dataframe function in a BidsDataframe class (in #595)
-	Adapt BidsDataset classes

## To do in this issue

1. The modifications will be done in different loader files temporarily.

2. Modify [BidsDataset](https://github.com/ivadomed/ivadomed/blob/96399de60bd7dda71de5bd8569d8ffac209bfb14/ivadomed/loader/loader.py#L750) and [BIDStoHDF5](https://github.com/ivadomed/ivadomed/blob/ec20c84713ddec47a34c4c16517910c61ebba563/ivadomed/loader/adaptative.py#L175) classes:
    * to use the new dataframe and the new split to create `filenames_pairs`
    * up to when the `load_filenames` method are called

At this point, the resulting `filename_pairs` should be the same as the `filename_pairs` in the current loader.

## Next steps

- Integrate the changes made for the new loader (steps 1-3) to the current codebase:
    * minor refactoring in `main`
    * update `data-testing dataset`
    * update `unit_tests`
    * update config and default config files
    * update documentation to reflect changes in config file for split
    * refactor other functions that use “path data” as an input, examples: `evaluate` and `run_segment_command`
- Add file format specific loaders in class and methods called by `load_filenames` (SegmentationPair, get_pair_shape, get_pair_metadata and get_pair_data) to load data as numpy array.

PR #648, opened 2021-01-20
New loader: BidsDataframe class

### LINKED ISSUE #595: BidsDataframe class for new loader
In #538 , `create_bids_dataframe` was implemented as a function.

After the merge of #591 , we should refactor the function as a method for a `BidsDataframe` class, with the bids_dataframe as an attribute.

The class would also include other relevant methods similar to `bids_neuropoly` (has_derivatives, get_derivatives, get_subjects, etc) as well as a method to save the dataframe as a csv file in the log directory.

And add a step: if derivatives is true and there is no dataset_description.json in derivatives, ~~copy the one from the main bids path.~~
EDIT: The dataset_description.json file AND its "minimal content" are mandatory in derivatives folder for `pybids` to index the derivatives. We cannot just copy the main dataset_description.json of the bids_path.
As a reference, a dataset_description.json file must be present in "derivatives" or "labels" folder with minimal content:
`{"Name": "Example dataset", "BIDSVersion": "1.0.2", "PipelineDescription": {"Name": "Example pipeline"}}`

mariehbourget:
Fixes #595 

This PR refactors the `create_bids_dataframe` function from `loader/utils.py` as a method of a new `BidsDataframe` class.

The class includes a `save` method, and the resulting dataframe is now saved in the `log_directory`.

It also includes the following methods, which will be re-used in the next step of the project (#596 ).
`get_subject_fnames`
`get_deriv_fnames`
`get_derivatives`


mariehbourget:
The PR is ready for review.

@gisellemartel, I have a question regarding a UserWarning from a dependency.
[Here](https://github.com/ivadomed/ivadomed/blob/698e186ec506e4b82f4938bd77b159c686ea6c1a/ivadomed/loader/utils.py#L797), when derivatives is True, the `pybids` dependency returns a UserWarning if a “dataset_description.json” file is not present in the folder `bids_path/derivatives`. Is there a way to catch this warning if it occurs, and then either stop the process or maybe write the missing file in the `bids_path/derivatives` folder?
(The warning question could be handled in a separate PR).

<details>
  <summary>Warning from pybids</summary>
 
``` 
/venv-ivadomed-38c/lib/python3.8/site-packages/bids/layout/validation.py:119: UserWarning: Derivative indexing was requested, but no valid datasets were found in the specified locations (['ivadomed/data_example_spinegeneric/derivatives']).
Note that all BIDS-Derivatives datasets must meet all the requirements for BIDS-Raw datasets (a common problem is to fail to include a 'dataset_description.json' file in derivatives datasets). 
Example contents of 'dataset_description.json': 
{"Name": "Example dataset", "BIDSVersion": "1.0.2", "PipelineDescription": {"Name": "Example pipeline"}}
```
</details>

ivadomed/loader/utils.py
```diff
@@ -722,149 +722,210 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+    def get_deriv_fnames(self):
+        """Return the list of derivative filenames in dataframe."""
+        return self.df[self.df['path'].str.contains('derivatives')]['filename'].tolist()
+
+    def get_derivatives(self, subject_fname, deriv_fnames):
+        """Return list of available derivative filenames for a subject filename.
+        Args:
+            subject_fname (str): Subject filename.
+            deriv_fnames (list of str): List of derivative filenames.
+        """
```
andreanne-lemay, line 918:
```suggestion

        Returns:
            list: derivative filenames
        """
```

ivadomed/loader/utils.py
```diff
@@ -722,149 +722,210 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
 
-    # Drop columns with all null values
-    df.dropna(axis=1, inplace=True, how='all')
+    def get_subject_fnames(self):
+        """Return the list of subject filenames in dataframe."""
+        return self.df[~self.df['path'].str.contains('derivatives')]['filename'].to_list()
 
-    return df
+    def get_deriv_fnames(self):
+        """Return the list of derivative filenames in dataframe."""
```
andreanne-lemay, line 910:
```suggestion
        """Get the list of derivative filenames in dataframe.
        
        Returns:
            list: derivative filenames.
        """
```

ivadomed/loader/utils.py
```diff
@@ -722,149 +722,210 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
 
-    # Reset index
-    df.reset_index(drop=True, inplace=True)
+        # Drop columns with all null values
+        self.df.dropna(axis=1, inplace=True, how='all')
 
-    # Drop columns with all null values
-    df.dropna(axis=1, inplace=True, how='all')
+    def get_subject_fnames(self):
+        """Return the list of subject filenames in dataframe."""
```
andreanne-lemay, line 906:
```suggestion
        """Get the list of subject filenames in dataframe.
        
        Returns:
            list: subject filenames.
        """
```

andreanne-lemay (approved):
Very clean! I love it!

gisellemartel:
> The PR is ready for review.
> 
> @gisellemartel, I have a question regarding a UserWarning from a dependency.
> [Here](https://github.com/ivadomed/ivadomed/blob/698e186ec506e4b82f4938bd77b159c686ea6c1a/ivadomed/loader/utils.py#L797), when derivatives is True, the `pybids` dependency returns a UserWarning if a “dataset_description.json” file is not present in the folder `bids_path/derivatives`. Is there a way to catch this warning if it occurs, and then either stop the process or maybe write the missing file in the `bids_path/derivatives` folder?
> (The warning question could be handled in a separate PR).
> 
> Warning from pybids

You could first check if the files exists and if so throw an Exception if the intention is to stop the process if the file is missing. If you plan on writing the file dynamically, I would assume a set of default values would need to be specified somewhere in order to populate this file. Might be worth opening another issue to discuss the specifics of this

ivadomed/loader/utils.py
```diff
@@ -722,149 +722,221 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+        # derivatives
+        self.derivatives = derivatives
+
+        # Create dataframe
+        self.create_bids_dataframe()
+
+        # Save dataframe as csv file
+        self.save(os.path.join(log_directory, "bids_dataframe.csv"))
+
+    def create_bids_dataframe(self):
```
gisellemartel, line 778: Overall, really great work! I do have a suggestion for this function. Since it is over 100 lines long, it may be worth the effort to extract some of the nested conditions into helper functions. This long-term would make maintenance of the class less costly and also reduce some of complexity and readability.

ivadomed/loader/utils.py
```diff
@@ -722,149 +722,221 @@ def reorient_image(arr, slice_axis, nib_ref, nib_ref_canonical):
...
+                    df_scans = pd.concat([df_scans, df_temp], ignore_index=True)
+        if not df_scans.empty:
+            df_scans['filename'] = df_scans['filename'].apply(os.path.basename)
+            self.df = pd.merge(self.df, df_scans, on=['filename'], suffixes=("_x", None), how='left')
+
+        # TODO: check if other files are needed for EEG and DWI
+
+        # If indexing of derivatives is true
+        # Get list of subject files with available derivatives
+        if self.derivatives:
```
gisellemartel, line 866: Suggestion, code nested within this condition could be placed in another function (perhaps something like `get_subject_file_list()`

mariehbourget:
> > The PR is ready for review.
> > @gisellemartel, I have a question regarding a UserWarning from a dependency.
> > [Here](https://github.com/ivadomed/ivadomed/blob/698e186ec506e4b82f4938bd77b159c686ea6c1a/ivadomed/loader/utils.py#L797), when derivatives is True, the `pybids` dependency returns a UserWarning if a “dataset_description.json” file is not present in the folder `bids_path/derivatives`. Is there a way to catch this warning if it occurs, and then either stop the process or maybe write the missing file in the `bids_path/derivatives` folder?
> > (The warning question could be handled in a separate PR).
> > Warning from pybids
> 
> You could first check if the files exists and if so throw an Exception if the intention is to stop the process if the file is missing. If you plan on writing the file dynamically, I would assume a set of default values would need to be specified somewhere in order to populate this file. Might be worth opening another issue to discuss the specifics of this

Thanks for your insights, I'll open a separate issue.

gisellemartel (approved):
Look's great!

ivadomed/loader/utils.py
mariehbourget: Thanks for your suggestions, I implemented it as `get_subjects_with_derivatives` and I create another method specific to adding tsv files metadata `add_tsv_metadata`. Let me know if that makes sense. Thanks!

PR #650, opened 2021-01-22
Reformat configuration_file.rst for docs

ahill187:
## Description

The purpose of this pull request is to refactor the `configuration_file.rst` into a more readable format. I have implemented the Python package [sphinx-jsonschema](https://sphinx-jsonschema.readthedocs.io/en/latest/index.html), which creates structured documentation of json files.

### Example

filmedunet

PR #652, opened 2021-01-25
Externalize command, log_directory, and bids_path fields from JSON config files to CLI

### LINKED ISSUE #615: config JSON file: suggestions for externalizing some fields
As much as I love the config JSON file, i feel that some fields could be externalized to ease the user experience. Namely:
- `command`: given that we often want to first train, and then test, it feels superfluous to edit a config file to change the "action" behaviour. I would instead pass that as a new positional argument in ivadomed CLI:
  ~~~
  ivadomed train -c config.json
  ivadomed test -c config.json
  ~~~
- `log_directory`, `bids_path`: In general, I find it cleaner to make config files path-agnostic, so that they can be readily used (eg. from tutorials or by collaborators), and have paths passed through CLI arguments:
  ```
  ivadomed train -c config.json --path-data PATH_TO_DATA --path-output PATH_OUTPUT
  ```

As per the example above, I would also suggest to change `log_directory` for `path_output`, because I personally find it more intuitive.

gisellemartel:
## Description

This PR addresses requested enhancements described in issue #615. Instead of specifying the command (train, test, or segment) via the config file, you can now do this via command line. The same follows for the setting the path of the log_directory and the bids_path. Here is an example of the new usage:

`ivadomed --train -c path/to/config/config.json --path-data path/to/data --path-output path/to/logs`

The following has been modified to adapt to these new changes:
- main.py 
- utils.py (new helper functions to parse arguments)
- automate_training.py (some minor revisions needed to account for the change of some arguments now being parsed from CLI instead of the config file)
- rst docs (namely the "usage" and "tutorials" sections)
- config files in `ivadomed/config`: the "command", "log_directory", and "bids_path" entries in the JSON files have been removed since they are now parsed via CLI

## Important Notes
- The wiki documentation will need to be updated once this PR is approved and merged. Changes to the wiki unfortunately could not be included in this PR since the wiki docs reside on a separate repository. 

## Linked issues
Resolves #615


ivadomed/main.py
```diff
@@ -463,6 +474,10 @@ def run_main():
     # Get context from configuration file
     path_config_file = args.config
     context = imed_config_manager.ConfigurationManager(path_config_file).get_config()
+    
+    context["command"] = imed_utils.get_command(args)
+    context["log_directory"] = imed_utils.get_log_directory(args)
```
ahill187, line 479: If we're refactoring to call it the `path_output`, maybe we should change to:
```
context["path_output"]
```
What do you think?

ivadomed/main.py
```diff
@@ -463,6 +474,10 @@ def run_main():
...
+    context['loader_parameters']["bids_path"] = imed_utils.get_bids_path(args)
```
ahill187, line 480: Same here, maybe `bids_path` should be `path_data`?

ivadomed/main.py
ahill187, line 479: I think you don't actually need the `get_log_directory` function or `get_bids_path` function, the parser takes care of checking whether or not the argument is present via the `required` flag; this could just be:
```
context["log_directory"] = args.log_directory
context["bids_path"] = args.get_bids_path
```

ivadomed/utils.py
```diff
@@ -313,6 +313,41 @@ def _version_string():
...
+            return "segment"
+        else:
+            return "train" # default to training
+    except AttributeError:
+        logger.warning("Issue parsing argument for command: ( --train | --test | --segment ). Will default to train command.")
+        return "train"
+
+
+def get_log_directory(args):
+    try:
```
ahill187, line 332: See above, I think you might not actually need `get_log_directory` and `get_bids_path`

docs/source/tutorials/one_class_segmentation_2d_unet.rst
```diff
@@ -114,7 +95,23 @@ Once the configuration file is ready, run the training:
 
 .. code-block:: bash
 
-   ivadomed -c config.json
+   ivadomed --train -c config.json --path-data path/to/bids/data --path-output path/to/output/directory
```
ahill187, line 98: Great explanation! This makes it a lot easier to use 💯

docs/source/tutorials/uncertainty.rst
```diff
@@ -74,7 +59,28 @@ Once the configuration file has been modified, run the inference with the follow
 
 .. code-block:: bash
 
-   ivadomed -c config.json
+   ivadomed --test -c config.json --path-data <PATH_TO_DATASET>/data_example_spinegeneric/trained_model --path-output path/to/output/directory
```
ahill187, line 62: Looks good!!

ivadomed/main.py
```diff
@@ -32,8 +32,20 @@ def get_parser():
 
     # MANDATORY ARGUMENTS
     mandatory_args = parser.add_argument_group('MANDATORY ARGUMENTS')
+    command_group = mandatory_args.add_mutually_exclusive_group(required=True)
```
ahill187, line 35: Cool command, I didn't know that was a feature!

ahill187 (commented):
Looks great! I like the idea of having `command`, `log_directory`, and `path_bids` as command line arguments, thanks for the refactor!

ivadomed/main.py
```diff
@@ -32,8 +32,20 @@ def get_parser():
...
+    
+    command_group.add_argument("--train", dest='train', action='store_true', 
+                                help="Perform training on data.")
+    command_group.add_argument("--test", dest='test', action='store_true', 
+                                help="Perform testing on trained model.")
+    command_group.add_argument("--segment", dest='segment', action='store_true', 
+                                help="Perform segmentation on data.")
     mandatory_args.add_argument("-c", "--config", required=True, type=str,
                                 help="Path to configuration file.")
+    mandatory_args.add_argument("-pd", "--path-data", required=True, type=str,
```
andreanne-lemay, line 45: Can we keep these arguments (path-data, path-output) optional and continue to support the version with all the parameters in the config file? Personally, I find it more convenient to have everything in a single config file, but I get why it can also be more convenient to externalize these flags. The best solution, in my opinion, would be to continue supporting all the options in the config file (for retrocompatibility and for dev convenience), but also add the option to overwrite some parameters in the CLI.

ivadomed/main.py
gisellemartel: I think that is a great idea actually! Would probably be easier to understand the code base overall by deciding on a uniform. I have noticed different variable names being used to refer to the same thing, so I think it would be worth the effort to refactor

ivadomed/main.py
gisellemartel: It's actually quite useful!

ivadomed/main.py
ahill187: Great! Yah makes it easier when they're all the same name :)

ivadomed/main.py
gisellemartel: Good points! I removed the entries from the config file because from my understanding the linked issue #615 implied to transition from config files to CLI flag. I can definitely change it to support both JSON files and and CLI flags. Wouldn't mind some other input from others as well before proceeding. Tagging @jcohenadad since he created the issue and also @ahill187 for visibility.

ahill187: Yah they're both good points, I see the advantage of supporting both methods, although having just one method is probably cleaner/simpler? Particularly the `train/test/segment` flags seem to make more sense as just command-line parameters? 
I suppose we could make the `path_bids` and `path_output` default to whatever directory you are currently in? I'm not sure if that helps at all though haha.

gisellemartel: One of my concerns would also be in terms of maintenance in the long-term. It may be confusing for future users if they can specify commands both via CLI and config files. Perhaps a better solution would be to support both in the short-term and consider deprecating the config option at a later date?

ahill187: Yah that's a good point about maintaining long-term, it's not ideal to have coupled code to maintain in tandem.

ahill187:
Also @gisellemartel, I'm going to merge this pull request: https://github.com/ivadomed/ivadomed/pull/650 once it updates with the master branch. I made quite a lot of changes to `configuration_file.rst`, let me know if you have any questions with merging because I think that might affect this branch a bit :)

ivadomed/main.py
andreanne-lemay: If we have to choose only one of the methods, I think it would be better to keep everything in the configuration file since I find it confusing and inconvenient to do the back and forth between both. But, of course, this is really personal. I still think both options can coexist, just like what is done with `sct_run_batch` on SCT (https://spinalcordtoolbox.com/en/stable/user_section/command-line.html#sct-run-batch).

ahill187: Perhaps we could get some more input from other users in the lab? We could discuss on `Slack` or do a poll or something, or discuss at the next lab meeting?

gisellemartel:
> Also @gisellemartel, I'm going to merge this pull request: #650 once it updates with the master branch. I made quite a lot of changes to `configuration_file.rst`, let me know if you have any questions with merging because I think that might affect this branch a bit :)

Thanks for the head's up!! :)

ivadomed/main.py
andreanne-lemay: Yes, absolutely we can discuss this at the next lab meeting on Wednesday.

andreanne-lemay:
Regarding whether we should keep the externalized flags in the config file(s) or not, I feel like these keys should stay (in at least one configuration file) for the following reasons:

1 - For convenience. People who are not using the flags will have to open the documentation and copy all the missing required keys from the doc each time they pull the code from ivadomed and use one of the available config files.
2 - The point of having flags is to not have to open the config file. If you use the flags, chances are you won't open the config file. If you do open it, then to me it makes more sense to change everything in the configuration file.
3 - As long as everything is clearly stated in the documentation, I don't see the issue with leaving the keys in the default config.
4 - 3 keys won't make the difference in making the configuration file lighter.
5 - We can create a minimal requirement configuration file (which would look like the current config_default.json) and another configuration with all the parameters.

gisellemartel:
Update: A few tests are failing since the testing data repos still use "log_directory" in their config files instead of "path_output". Will need to update data repositories first to be able to merge this PR.

docs/source/usage.rst
```diff
@@ -9,9 +9,28 @@ terminal:
 
 ::
 
-    ivadomed -c path/to/config_file.json
+    ivadomed [command] -c path/to/config_file.json --path-data path/to/bids/data --path-output path/to/output/directory
 
-where ``config_file.json`` is a configuration file, which parameters are
-described in the :ref:`configuration_file:Configuration File`.
+``[command]`` represents the following choice of flags: 
+
```
ahill187, line 15: Great explanation!

ivadomed/main.py
```diff
@@ -32,38 +32,50 @@ def get_parser():
...
+
+    command_group = optional_args.add_mutually_exclusive_group(required=False)
+    
+    command_group.add_argument("--train", dest='train', action='store_true', 
+                                help="Perform training on data.")
+    command_group.add_argument("--test", dest='test', action='store_true', 
+                                help="Perform testing on trained model.")
+    command_group.add_argument("--segment", dest='segment', action='store_true', 
+                                help="Perform segmentation on data.")
+    mandatory_args.add_argument("-c", "--config", required=True, type=str,
```
ahill187, line 47: If the config file is still part of the mandatory args, we should probably put it at the top with the other mandatory args?

ahill187 (commented):
Looks great!

ivadomed/main.py
gisellemartel: I placed them in this order to respect the convention of having the command argument before the config argument in CLI.

gisellemartel: i.e. `ivadomed --train -c path/to/config/config.json` vs. ivadomed -c path/to/config/config.json  --train

ahill187: Ah I see. Maybe if that's the case then, the headers "Optional Arguments" and "Mandatory Arguments" should be removed, since they don't correspond to the arguments anymore?

gisellemartel: Technically the mandatory and optional arguments are still placed within their respective groups, but the ordering in the code is not as straightforward. I could maybe instead remove the command group from the optional group and just make it its own group. Not sure if that would be more straightforward?

gisellemartel: Removing the mandatory args actually might be the simplest way around this. The only advantage of keeping of it would be in case we decide to support other mandatory args later on

ahill187: Right yah, I think the command group is okay because it's optional? It's just that because the config argument is now in the code block under the header ``# OPTIONAL ARGUMENTS``, the header for that section of code doesn't totally make sense now.

gisellemartel: I reorganized the parser arguments to try and make clearer

ahill187: Okay great!

ahill187 (approved):

PR #661, opened 2021-02-01
Allow for non-dictionary hyperparameters in automate_training

### LINKED ISSUE #587: Impossible to modify parameters that are not dict with automate_training
When using the CLI `ivadomed_automate_training`, 4 parameters cannot be changed in the hyperparameter file: `gpu`, `command`, `log_directory`, `model_name`, and `debugging`. The script assumes that the parameters changed will be dictionaries due to [this line](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/scripts/automate_training.py#L186), and the parameters mentioned above are not dictionaries.

This issue should be fixed. All parameters should have the option to be modified with `ivadomed_automate_training`.

see [comment](https://github.com/ivadomed/ivadomed/pull/586#pullrequestreview-554209935)

ahill187:
## Description
This pull request does 4 things:
1. Enable users to use hyperparameters which are not dictionaries in `automate_training`. For example, the parameter `command` currently won't work, because it is not a dictionary. We would like to allow for a hyperparameter config file which allows for a format such as:
```
    {
        'training_parameters': {
            'batch_size': [2, 64],
            'loss': [
                {'name': 'DiceLoss'},
                {'name': 'FocalLoss', 'gamma': 0.2, 'alpha': 0.5}
             ]
        },
        'default_model': {
            'depth': [2, 3, 4]
        },
        'command': ['train', 'test'] # this is not currently allowed
    }
```
2. Refactor to simplify the logic in creating the config lists. One issue was that two dictionaries were being zipped together, (`param_dict` and `names_dict`), and the order of this mattered. However, Python dictionaries do not conserve order, and so zipped dictionaries are not guaranteed to be in the right order.
3. Add unit tests for `automate_training` under `unit_tests/test_automate_training.py`
4. Add an example under the `Tutorials` section in the docs on how to use `ivadomed_automate_training`

## Linked issues
Resolves: https://github.com/ivadomed/ivadomed/issues/587


docs/source/tutorials/automate_training.rst
```diff
@@ -0,0 +1,170 @@
+Optimize Hyperparameters
```
gisellemartel, line 1: Really clear documentation, great workt!

testing/unit_tests/test_automate_training.py
```diff
@@ -0,0 +1,697 @@
+"""Unit Test for Automate Training
```
gisellemartel, line 1: This is only a suggestion, but I noticed the mocking of the config makes this file quite lengthy. Would it maybe make more sense to place the mocked data in another file and import it into this one to separate the test logic from the mock data?

ivadomed/scripts/automate_training.py
```diff
@@ -123,74 +129,496 @@ def test_worker(config):
...
+        keys = set([hyper_option.base_key for hyper_option in param_list])
+        for combination in list(itertools.combinations(param_list, len(keys))):
+            if keys_are_unique(combination):
+                new_config = copy.deepcopy(initial_config)
+                path_output = new_config["path_output"]
+                for hyper_option in combination:
+                    new_config = update_dict(new_config, hyper_option.option, hyper_option.base_key)
+                    path_output = path_output + hyper_option.name
+                new_config["path_output"] = path_output
+                config_list.append(new_config)
```
gisellemartel, line 248: This block of code is very similar to lines 242 - 248. I would suggest maybe refactoring some of this logic into a helper function to reduce code duplication

ivadomed/scripts/automate_training.py
```diff
@@ -123,74 +129,496 @@ def test_worker(config):
...
+        max_length = np.min([len(base_key_dict[base_key]) for base_key in base_key_dict.keys()])
+        for i in range(0, max_length):
+            new_config = copy.deepcopy(initial_config)
+            path_output = new_config["path_output"]
+            for key in base_key_dict.keys():
+                hyper_option = base_key_dict[key][i]
+                new_config = update_dict(new_config, hyper_option.option, hyper_option.base_key)
+                path_output = path_output + hyper_option.name
+            new_config["path_output"] = path_output
+            config_list.append(new_config)
```
gisellemartel, line 263: could maybe be refactored? (see previous comment)

gisellemartel (approved):
Really great work! Some minor suggestions for code quality.

testing/unit_tests/test_automate_training.py
ahill187: Yah I was actually wondering that myself, that's a good point! I should probably put it all in a json file or something and save it in the `data-testing` repo.

docs/source/tutorials/automate_training.rst
ahill187: Thanks!

ivadomed/scripts/automate_training.py
```diff
@@ -407,19 +768,18 @@ def main(args=None):
     parser = get_parser()
     args = imed_utils.get_arguments(parser, args)
```
andreanne-lemay, line 769: When not entering the right flag (e.g., I used -p instead of -ch), the message error is not informative: 
```    raise ArgParseException('Error parsing args')
ivadomed.utils.ArgParseException: Error parsing args
```
vs. when using `parser.parse_args()`:
```
usage: automate_training.py [-h] -c <file> -ch <file> [-pd <int>] [-n <int>]
                            [--all-combin] [-m] [--run-test] [--fixed-split]
                            [-l] [-t <float>] [-o OUTPUT_DIR]
automate_training.py: error: the following arguments are required: -ch/--config-hyper
```

why not directly using parser.parse_args? We should probably fix this new function `imed_utils.get_argments`, but what is its purpose?

ivadomed/scripts/automate_training.py
```diff
@@ -34,7 +40,7 @@ def get_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument("-c", "--config", required=True, help="Base config file path.",
                         metavar=imed_utils.Metavar.file)
-    parser.add_argument("-p", "--params", required=True,
+    parser.add_argument("-ch", "--config-hyper", dest="config_hyper", required=True,
```
andreanne-lemay, line 43: Could we keep the original flag name?

andreanne-lemay (approved):
I did a few tests on my local station and it works great thank you!

testing/unit_tests/test_automate_training.py
ahill187: https://github.com/ivadomed/data-testing/pull/10

ivadomed/scripts/automate_training.py
ahill187: Oh interesting! So the `imed_utils.get_arguments` function was implemented to allow for CLI testing, or to call the main function with a set of arguments. Basically it handles the case where you pass in the arguments instead of getting them from the command line. I agree though, this isn't a very good error message. I can open a separate issue to fix this?

ivadomed/scripts/automate_training.py
ahill187: I refactored the name `params` to `config_hyper` because I think it's more descriptive; `params` is a little general. I don't love `--ch` though haha, not sure if there's a better shorthand. `--h` would be better but that is for help...

ivadomed/scripts/automate_training.py
ahill187: https://github.com/ivadomed/ivadomed/issues/669

andreanne-lemay: Sounds great, thank you!

ivadomed/scripts/automate_training.py
andreanne-lemay: Yeah  I agree `-p` is not great. Okay, let's go with `-ch`. I was just asking for retrocompatibility since I have some files calling this scripts and it is annoying to have to change all of them (mostly in .sh files on ComputeCanada), but I am realizing I don't have that much files to change.

ahill187: Ah I see, maybe in future we should add some retrocompatibility/deprecation warning when changing flags?

docs/source/tutorials/automate_training.rst
charleygros: Really like it, thank you @ahill187 !

charleygros (approved):

charleygros:
>     1. Enable users to use hyperparameters which are not dictionaries in `automate_training`.

Very much needed, indeed, thank you!

>     1. Refactor to simplify the logic in creating the config lists. One issue was that two dictionaries were being zipped together, (`param_dict` and `names_dict`), and the order of this mattered. However, Python dictionaries do not conserve order, and so zipped dictionaries are not guaranteed to be in the right order.

We faced so many issues with that in the past in this script.. thanks for fixing it!!

ahill187:
> > ```
> > 1. Enable users to use hyperparameters which are not dictionaries in `automate_training`.
> > ```
> 
> Very much needed, indeed, thank you!
> 
> > ```
> > 1. Refactor to simplify the logic in creating the config lists. One issue was that two dictionaries were being zipped together, (`param_dict` and `names_dict`), and the order of this mattered. However, Python dictionaries do not conserve order, and so zipped dictionaries are not guaranteed to be in the right order.
> > ```
> 
> We faced so many issues with that in the past in this script.. thanks for fixing it!!

No problem!!

testing/unit_tests/test_automate_training.py
ahill187: So it looks like it isn't too trivial to parametrize tests from data that is loaded, especially because we download that data in the `setup_function`. Basically, if I load that data outside the `setup_function`, the data won't have been downloaded yet. And if we load it in the `setup_function`, it isn't global and the tests don't have access to it.

ivadomed/scripts/automate_training.py
ahill187: I see what you mean, I think though they're different enough that we would only be able to factor out a few lines (since the iteration is different). Like, we could refactor out:

```
new_config = update_dict(new_config, hyper_option.option, hyper_option.base_key)
path_output = path_output + hyper_option.name
```

but I don't think that would be worth a new function?

PR #663, opened 2021-02-03
Store FiLM parameters during testing instead of training

andreanne-lemay:
## Description

FiLM parameters are saved during the last epoch of the training. Moving the parameters storing during testing allows saving the parameters for models already trained that doesn't have their FiLM parameters saved (e.g. debugging set to false). Also, because of the early stopping, the parameters can be saved on the few last epochs (condition of last epoch is flawed for this case), even after the FiLM parameters have slightly changed. During testing, the FiLM parameters are final, hence gives a realistic portrait of the gammas and betas.

The two film-related functions `store_film_params` and `get_film_metadata_models` have been moved to the `film.py` file.


ivadomed/loader/film.py
```diff
@@ -217,3 +217,68 @@ def get_film_metadata_models(ds_train, metadata_type, debugging=False):
...
+        betas (dict):
+        metadata_values (list): list of the batch sample's metadata values (eg T2w, T1w, if metadata type used is
+        contrast)
+        depth (int):
+        ofolder (str):
+
+    """
+    # Convert list of gammas/betas into numpy arrays
+    gammas_dict = {i: np.array(gammas[i]) for i in range(1, 2 * depth + 3)}
+    betas_dict = {i: np.array(betas[i]) for i in range(1, 2 * depth + 3)}
```
charleygros, line 272: Could you please add a comment about where comes from the "+3"?

charleygros (approved):
Went through, all look good to me! Thanks for that

ivadomed/loader/film.py
andreanne-lemay: Thanks for noticing this! It was there before, but I believe it should be 2 * depth + 2 (aka the number of FiLM layers). I'll add a comment about this :) + fix it

PR #668, opened 2021-02-05
Add new keys config manager

andreanne-lemay:
In PR #652, two keys in the configuration file were changed. These changes are not retrocompatible with old configuration files. The new keys were added to the configuration manager to ensure retrocompatiblity.

### DIFF
--- a/ivadomed/config_manager.py
+++ b/ivadomed/config_manager.py
@@ -2,6 +2,7 @@ import json
 import os
 import collections.abc
 from ivadomed import utils as imed_utils
+import copy
 
 
 def update(d, u):
@@ -65,7 +66,7 @@ def load_json(config_path):
 
 
 # To ensure retrocompatibility for parameter changes in configuration file
-KEY_CHANGE_DICT = {'UNet3D': 'Modified3DUNet'}
+KEY_CHANGE_DICT = {'UNet3D': 'Modified3DUNet', 'bids_path': 'path_data', 'log_directory': 'path_output'}
 
 
 class ConfigurationManager(object):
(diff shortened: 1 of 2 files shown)


ivadomed/config_manager.py
```diff
@@ -65,7 +66,7 @@ def load_json(config_path):
 
 
 # To ensure retrocompatibility for parameter changes in configuration file
-KEY_CHANGE_DICT = {'UNet3D': 'Modified3DUNet'}
+KEY_CHANGE_DICT = {'UNet3D': 'Modified3DUNet', 'bids_path': 'path_data', 'log_directory': 'path_output'}
```
ahill187, line 69: Cool I didn't know we had this retrocompatibility feature!

ahill187 (approved):
Looks good, tested it locally and it works for me :) Good catch, thanks!

The only caveat I might add is an unlikely edge case, which would be the following example:

Say this was part of our config file:
```
    "training_parameters": {
        "loss": {
            "name": "DiceLoss"
        },
        "scheduler": {
            "initial_lr": 0.001,
            "lr_scheduler": {
                "name": "CosineAnnealingLR",
                "base_lr": 1e-5,
                "max_lr": 1e-2
            }
        }
```

Suppose we wanted to update the `name` key. The current implementation of changing the keys wouldn't know whether or not we were renaming the `name` key in `loss` or in `lr_scheduler`. I don't think it's a pressing issue, more of a niche case, we could open a separate issue to fix that in a different pull request :)

gisellemartel (approved):
Look's good!

ivadomed/config_manager.py
andreanne-lemay: Yes you're right! We should definitely keep this in mind!

ahill187: https://github.com/ivadomed/ivadomed/issues/670

PR #679, opened 2021-02-10
Shared weights for the two first FiLM generator layers

andreanne-lemay:
When using FiLM, only the second layer of the FiLM generator is shared. The two first layers should be shared since the input is always the same for all FiLM layers, hence, the same features should be extracted from the input. This PR shares the weights of the two first layers of the FiLM generator instead of one.


ivadomed/models.py
```diff
@@ -573,18 +573,18 @@ def __init__(self, n_features, n_channels, n_hid=64):
...
-        x = self.linear1(x)
-        x = self.sig(x)
-
         if shared_weights is not None:  # weight sharing
-            self.linear2.weight = shared_weights
+            self.linear1.weight = shared_weights[0]
+            self.linear2.weight = shared_weights[1]
 
+        x = self.linear1(x)
+        x = self.sig(x)
```
charleygros, line 581: I can't remember why, two years ago, I put these lines before the weight sharing. What are your thoughts about this?

andreanne-lemay: `self.linear1` used to not have shared weights, so putting before or after the weight sharing would give the same results. With the new code, we have to extract the shared weights, then use the layer. If we put `x = self.linear1(x)` before the weight sharing, the behavior would be the same as before (no shared weights for the first layer).

charleygros: Makes full sense!

charleygros (approved):

Issue #583, opened 2020-12-16
Limit file size on commit


alexfoias:
Problem: we would like to forbid commits that include files bigger than XX MB.
Possible implementations: 

- https://stackoverflow.com/questions/39576257/how-to-limit-file-size-on-commit/39578014#39578014
- https://gist.github.com/kiwidamien/a6a909ee196be8795b30431079074d64

Crossref: https://github.com/neuropoly/spinalcordtoolbox/issues/3100 https://github.com/shimming-toolbox/shimming-toolbox/issues/199 https://github.com/neuropoly/axondeepseg/issues/423

kousu:
https://github.com/neuropoly/spinalcordtoolbox/issues/2705

gisellemartel:
@alexfoias do we have a specific max file size? I am investigating solutions with GitHub actions following the previously linked discussion

alexfoias:
@jcohenadad Could you please let us know what file size limit should we use ?

jcohenadad:
how about 500kB?  Enough for us to catch non-text uploads

gisellemartel:
Closed in #682

ssinad:
I should mention that once a file has been committed, it stays in Git history forever unless the history is rewritten.

jcohenadad:
> I should mention that once a file has been committed, it stays in Git history forever unless the history is rewritten.

indeed, hence this PR

ssinad:
I was thinking of writing an integration test which would fail when the commits contain large files, but it is difficult to track once a commit has been made. Prevention is better than cure, in this case.

gisellemartel:
So the merged changes here will prevent any large files from being committed in the future, but I did not address any large files that may already be in the repository. If the intention was also to remove existing large files I can make another issue for that.

PR #682, opened 2021-02-14
Add pre-commit hooks to limit file size to 500KB 

gisellemartel:
## Description
Local commit hooks have been added using `pre-commit` to prevent files > 500KB from being committed. Pre-commit hooks are also utilized via Github actions. These changes are similar to what was implemented [here](https://github.com/shimming-toolbox/shimming-toolbox/pull/215)

## Linked issues
#583


ahill187 (approved):
Looks good to me!!

PR #683, opened 2021-02-14
Add DOI JOSS

charleygros:
- `README.md`: Add JOSS badge
- `index.rst`: citation: from arxiv to JOSS

Question: do we want to keep both zenodo and JOSS badges?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
   
 ![ivadomed Overview](https://raw.githubusercontent.com/ivadomed/doc-figures/main/index/overview_title.png)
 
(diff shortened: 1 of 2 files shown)


jcohenadad:
> Question: do we want to keep both zenodo and JOSS badges?

i would remove the zenodo one (the joss article is sexier and more informative)

charleygros:
> i would remove the zenodo one (the joss article is sexier and more informative)

Agree! Done :-)

andreanne-lemay (approved):
:tada: thanks for taking care of this!

PR #687, opened 2021-02-16
New loader: input from multiple BIDS datasets

mariehbourget:
This PR enable the new loader pipeline to run with inputs from multiple BIDS datasets.

It supports `path_data` as string or as a list.

The BidsDataframe is a merge of all files and metadata of the specified paths. If there are duplicates, it only keeps the first encounter (same as the implementation in current loader).


andreanne-lemay (approved):
LGTM :rocket:

testing/unit_tests/test_loader.py
```diff
@@ -38,22 +38,22 @@ def test_bids_df_microscopy_png(loader_parameters, log_directory):
...
     df_test = bids_df.df
     df_test = df_test.drop(columns=['path', 'parent_path'])
     # TODO: modify df_ref.csv file in data-testing dataset to include "participant_id"
     # "sample_id" and "ivadomed_id" columns, then delete next line
     df_test = df_test.drop(columns=['participant_id', 'sample_id', 'ivadomed_id'])
     df_test = df_test.sort_values(by=['filename']).reset_index(drop=True)
-    csv_ref = os.path.join(path_data, "df_ref.csv")
-    csv_test = os.path.join(path_data, "df_test.csv")
+    csv_ref = os.path.join(path_data[0], "df_ref.csv")
+    csv_test = os.path.join(path_data[0], "df_test.csv")
```
gisellemartel, line 49: Suggestion: could maybe add a test that can ensure the loader can read in multiple entries in `path_data`. I notice that the current tests always assume only one entry.

gisellemartel (approved):

testing/unit_tests/test_loader.py
mariehbourget: Good point, thanks for the suggestion! For now, I only added a comment in the test_loader.py to add the test later. There are already a few modifications to do to the tests but I want to do them all at once after I finished with issue 596, because of upcoming changes in my implementation that will require updates and a new release in the data-testing repo. Thanks again!

PR #688, opened 2021-02-19
Multiclass ignored during inference if n_input and n_output are different

charleygros:
In `inference.py`: only n_input classes are saved. This is a problem if n_input < n_output: eg input=dwi, output=gm, wm --> only gm segmented.
This mini-mini PR fixes this bug.

Note: the behavior is however correct in `testing.py`, which is a relief, see [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/testing.py#L184).


andreanne-lemay (approved):
Nice catch!

PR #691, opened 2021-02-23
New loader: Refactor BidsDataset classes

### LINKED ISSUE #596: Refactor BidsDataset classes for new loader

mariehbourget:
### Context

Fixes #596 

It aims to refactor `BidsDataset`, `Bids3DDataset` and `BIDStoHDF5` classes to use the new loader pipeline (i.e. BidsDataframe + new splitting methods) to create `filename_pairs`.
The resulting `filename_pairs` are the same as the `filename_pairs` in the current loader.

### Changes done in this PR:

As discussed previously, the changes are made in "new" files until they are integrated in the pipeline.
- Refactoring of `BidsDasaset` and `Bids3DDataset` in `loader/loader_new.py` file
- Refactoring of `load_bounding_box` and `generate_bounding_box_file` in `object_detection/utils_new.py` file.
- Refactoring of `BIDStoHDF5` and `HDF5Dataset` in `loader/adaptative_new.py`.
- Remove `ivadomed_id` column in `BidsDataframe`
- Split functions output lists of `filenames`  instead of list of `ivadomed_id`
- `BidsDataset` and `BIDStoHDF5` input lists of `filenames` instead of `participants_id` (I made it retrocompatible with partitcipant_id, old joblib files should work)
- All the other functions/classes in "_new" files are identical to the originals, except for some imports.

### To test the new pipeline:
- Since, this is not integrated with the current pipeline, you need to run the `dev/df_new_loader.py` script to load a validation dataset (works only for nifti files)
- The script uses `config_new_loader.json` config file. To use older config files, make sure to add the following parameters:
   - `"extensions": [".nii.gz"]` in loader_parameters
   -   `"split_method" : "participant_id"`, and `"data_testing": {"data_type": null, "data_value":[]}`, in split_dataset parameters
- I tested the `U-net 2D` with and without `roi_suffix`, `multichannel` and `bounding_box`, `FilmedUnet` with `mri_param metadata_choice` and `custom metadata_choice`, as well as `Unet-3D` and `Hemis`. It seems to work fine but it would be great to test with real experiment files in case I missed something.

### Next step:
- Integrate the changes made for the new loader in PR #538, #591, #648 and #687 to the current pipeline.

### DIFF
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -877,9 +877,8 @@ class BidsDataframe:
             # As per pybids, derivatives don't include parsed entities, only the "path" column
             df_next = layout.to_df(metadata=True)
 
-            # Add filename and parent_path columns
+            # Add filename column
             df_next['filename'] = df_next['path'].apply(os.path.basename)
-            df_next['parent_path'] = df_next['path'].apply(os.path.dirname)
 
             # Drop rows with json, tsv and LICENSE files in case no extensions are provided in config file for filtering
             df_next = df_next[~df_next['filename'].str.endswith(tuple(['.json', '.tsv', 'LICENSE']))]
@@ -900,11 +899,10 @@ class BidsDataframe:
             # Merge dataframes
             self.df = pd.concat([self.df, df_next], join='outer', ignore_index=True)
 
-        # Drop duplicated rows based on all columns except 'path and 'parent_path'
+        # Drop duplicated rows based on all columns except 'path'
         # Keep first occurence
         columns = self.df.columns.to_list()
         columns.remove('path')
-        columns.remove('parent_path')
         self.df = self.df[~(self.df.astype(str).duplicated(subset=columns, keep='first'))]
 
         # If indexing of derivatives is true
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -14,6 +14,7 @@ import nibabel as nib
 import bids as pybids  # "bids" is already taken by bids_neuropoly
 import itertools
 import random
+import copy
 
 __numpy_type_map = {
     'float64': torch.DoubleTensor,
@@ -819,7 +820,7 @@ class BidsDataframe:
         self.bids_config = None if 'bids_config' not in loader_params else loader_params['bids_config']
 
         # target_suffix and roi_suffix from loader parameters
-        self.target_suffix = loader_params['target_suffix']
+        self.target_suffix = copy.deepcopy(loader_params['target_suffix'])
         # If `target_suffix` is a list of lists convert to list
         if any(isinstance(t, list) for t in self.target_suffix):
             self.target_suffix = list(itertools.chain.from_iterable(self.target_suffix))
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -110,7 +110,7 @@ def split_dataset_new(df, split_method, data_testing, random_seed, train_frac=0.
         train_frac (float): Between 0 and 1. Represents the train set proportion.
         test_frac (float): Between 0 and 1. Represents the test set proportion.
     Returns:
-        list, list, list: Train, validation and test data_type list.
+        list, list, list: Train, validation and test filenames lists.
     """
 
     # Get data_type and data_value from split parameters
@@ -134,7 +134,7 @@ def split_dataset_new(df, split_method, data_testing, random_seed, train_frac=0.
         data_value = sorted(df[data_type].unique().tolist())
         test_frac = test_frac if test_frac >= 1 / len(data_value) else 1 / len(data_value)
         data_value, _ = train_test_split(data_value, train_size=test_frac, random_state=random_seed)
-    X_test = df[df[data_type].isin(data_value)]['ivadomed_id'].unique().tolist()
+    X_test = df[df[data_type].isin(data_value)]['filename'].unique().tolist()
     X_remain = df[~df[data_type].isin(data_value)][split_method].unique().tolist()
 
     # List dataset unique values according to split_method
@@ -148,9 +148,9 @@ def split_dataset_new(df, split_method, data_testing, random_seed, train_frac=0.
     # Split remainder in TRAIN and VALID sets according to train_frac_update using sklearn function
     X_train, X_val = train_test_split(X_remain, train_size=train_frac_update, random_state=random_seed)
 
-    # Convert train and valid sets from list of "split_method" to list of "ivadomed_id"
-    X_train = df[df[split_method].isin(X_train)]['ivadomed_id'].unique().tolist()
-    X_val = df[df[split_method].isin(X_val)]['ivadomed_id'].unique().tolist()
+    # Convert train and valid sets from list of "split_method" to list of "filename"
+    X_train = df[df[split_method].isin(X_train)]['filename'].unique().tolist()
+    X_val = df[df[split_method].isin(X_val)]['filename'].unique().tolist()
 
     # Make sure that test dataset is unseen during training
     # (in cases where there are multiple "data_type" for a same "split_method")
@@ -254,7 +254,7 @@ def get_new_subject_split_new(df, split_method, data_testing, random_seed,
         subject_selection (dict): Used to specify a custom subject selection from a dataset.
 
     Returns:
-        list, list list: Training, validation and testing subjects lists.
+        list, list list: Training, validation and testing filenames lists.
     """
     if subject_selection is not None:
         # Verify subject_selection format
@@ -344,7 +344,7 @@ def get_subdatasets_subjects_list_new(split_params, df, path_output, subject_sel
         subject_selection (dict): Used to specify a custom subject selection from a dataset.
 
     Returns:
-        list, list list: Training, validation and testing subjects lists.
+        list, list list: Training, validation and testing filenames lists.
     """
     if split_params["fname_split"]:
         # Load subjects lists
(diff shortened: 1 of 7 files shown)


mariehbourget:
This PR is ready for review. Details of the changes and procedure to test are listed in the comment above.

@andreanne-lemay, @charleygros I have a question regarding multichannel. In the current loader, the `subject_lst` is for example: `[sub-01/anat/T1w, sub-01/anat/T2star, sub-01/anat/T2w, sub-02/anat/T1w, sub-02/anat/T2star, sub-02/anat/T2w,]` in that order. In the resulting `filename_pairs`, the `target_filename` and `roi_filename` are always the last one to go through the for loop (here T2w). Am I understanding this correctly? Is there a way to specify which gt should be in the multichannel dictonary? For now, I reproduced this behavior by sorting the list of filenames at the beginning of the init in BidsDataset. Let me know if I'm missing something. Thanks!

charleygros:
> In the resulting `filename_pairs`, the `target_filename` and `roi_filename` are always the last one to go through the for loop (here T2w). Am I understanding this correctly? Is there a way to specify which gt should be in the multichannel dictonary? For now, I reproduced this behavior by sorting the list of filenames at the beginning of the init in BidsDataset. Let me know if I'm missing something. Thanks!

I think this is well understood: the reason is: we assume that all the images of a same subject share the same GT, so GT_T2w = GT_T1w etc within the region of interest.

ivadomed/loader/utils.py
```diff
@@ -790,15 +791,15 @@ def merge_bids_datasets(path_data):
...
+    This class aims to create a dataframe containing all BIDS image files in a list of path_data and their metadata.
 
     Args:
         loader_params (dict): Loader parameters, see :doc:`configuration_file` for more details.
         derivatives (bool): If True, derivatives are indexed.
         path_output (str): Output folder.
 
     Attributes:
-        path_data (str): Path to the BIDS dataset.
+        path_data (list): Paths to the BIDS datasets.
```
charleygros, line 802: Good catch

ivadomed/loader/utils.py
```diff
@@ -810,14 +811,15 @@ class BidsDataframe:
...
+        # paths_data from loader parameters
+        # TODO: when integrating in pipeline, remove format_path_data here (done before in main)
         self.paths_data = imed_utils.format_path_data(loader_params['path_data'])
 
         # bids_config from loader parameters
         self.bids_config = None if 'bids_config' not in loader_params else loader_params['bids_config']
 
         # target_suffix and roi_suffix from loader parameters
-        self.target_suffix = loader_params['target_suffix']
+        self.target_suffix = copy.deepcopy(loader_params['target_suffix'])
```
charleygros, line 822: If I don't get it wrong, we had issues with this in the past, and then @ahill187 suggested a better way to do this, didn't you?

charleygros (commented):
First quick review on the "old" files. All gravy!
I am not able to test the `*_new` files atm, but will do on Monday, Tassie time.

charleygros:
@mariehbourget: I tried to test the new loader with a personal config file and had the following error. I am sending you the config file in DM. I may have done something wrong in the config file, could you please double check? Cheers

```bash
Output path already exists: exvivo_gm_wm_multi
  File "/home/GRAMES.POLYMTL.CA/chgroc/venv-ivadomed/lib/python3.6/site-packages/pandas/core/indexes/base.py", line 2895, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 70, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 101, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1675, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 1683, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'subject'

The above exception was the direct cause of the following exception:

  File "dev/df_new_loader.py", line 68, in <module>
    bids_df = imed_loader_utils.BidsDataframe(loader_params, derivatives, path_output)
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/loader/utils.py", line 842, in __init__
    self.create_bids_dataframe()
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/loader/utils.py", line 895, in create_bids_dataframe
    df_next = self.add_tsv_metadata(df_next, path_data, layout)
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/loader/utils.py", line 937, in add_tsv_metadata
    df['participant_id'] = "sub-" + df['subject']
  File "/home/GRAMES.POLYMTL.CA/chgroc/venv-ivadomed/lib/python3.6/site-packages/pandas/core/frame.py", line 2902, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/GRAMES.POLYMTL.CA/chgroc/venv-ivadomed/lib/python3.6/site-packages/pandas/core/indexes/base.py", line 2897, in get_loc
    raise KeyError(key) from err
KeyError: 'subject'
```

mariehbourget:
> I may have done something wrong in the config file, could you please double check?

There are 2 kinds of issue here.
1. The files naming in the dataset is not completely BIDS compliant
2. The field "extensions" is present 2x in the config file, once with `[.nii.gz]` and once with `[]`.

About BIDS compliance, we chose earlier in the process to keep the `BIDS-validator` "on" in `pybids` for the indexing of the files. Here, because they are not compliant, they are not indexed, resulting in no subject files in the dataframe.

To make the dataset BIDS compliant:
1. `sub-label`: In BIDS, the dash ("-") is used to separate the key (entity) from the value (label). Label cannot contains dash, so `sub-XXXX-top` would have to be `sub-XXXXtop` in both folder names and filenames, in both subjects and derivatives.
2. `modality-suffix`: In BIDS, underscore ("_") is used to serapate key-value pairs and to isolate the modality suffix at the end of the filename. `dwi_T0000` would have to be `dwi`, and be changed in the contrast params of the config file.
3. Any `modality-suffix` cannot be in any `data_type` folder. Here the `dwi suffix` cannot be in `anat data type` folder, so the `anat` folder should be renamed `dwi`.

Workaround with BIDS-validator off (not the best solution in my opinion):
1. In [pybids indexer](https://github.com/ivadomed/ivadomed/blob/dbb4fc0c9fb98f1aa6650c81abfefb0a1643f1f9/ivadomed/loader/utils.py#L868), add `validate=False`.
2. `sub-XXXX-top` would still have to be renamed `sub-XXXXtop` in both folder names and filenames, there is not way around this because of the parsing of the entity key-value pairs in pybids.
3. The contrast params in config file should be `T0000` instead of `dwi_T0000`, i.e. string that is after the last underscore.

To make the error more explicit, I added a warning [here](https://github.com/ivadomed/ivadomed/blob/d6656d87291247d7ff6008894f36dd209805709d/ivadomed/loader/utils.py#L894), if no subject files are found in a given path_data, as well as an error if no subject files are found in any path data of the list [here](https://github.com/ivadomed/ivadomed/blob/d6656d87291247d7ff6008894f36dd209805709d/ivadomed/loader/utils.py#L907).

charleygros:
Many many thanks for the very comprehensive answer @mariehbourget !
I'm gonna try this out on another dataset! Brb

charleygros:
@mariehbourget : I tried two of my config files and it worked very well :-)
A simple Unet and a FiLMedUnet.
Awesome work!!

charleygros (approved):

ivadomed/loader/utils.py
mariehbourget: @ahill187 here I use "deepcopy" because I want to append roi_suffix to target_suffix inside the function but not outside. Is it OK? Is there a better way to do this? Thanks!

ahill187: Hmmm I can't remember! `deepcopy` looks okay to me!

Issue #670, opened 2021-02-05
KEY_CHANGE_DICT for retrocompatibility will fail for non-unique key names


ahill187:
This issue was mentioned here:  https://github.com/ivadomed/ivadomed/pull/668#pullrequestreview-584541738

### Current behavior

In `config_manager.py`, there is a `KEY_CHANGE_DICT`, which allows for retrocompatibility if/when we change the key names in the config file. For example, if I use a config file that uses the old key `UNet3D`, the `KEY_CHANGE_DICT` maps this to the new key and updates the config file accordingly:

```KEY_CHANGE_DICT = {'UNet3D': 'Modified3DUNet'}```

However, because the `KEY_CHANGE_DICT` does not specify the level of the JSON dict, and because nested dictionary keys do not need to be unique, the edge case where we needed to change a non-unique key would fail.

For example, say this was in our original config file:
```
    "training_parameters": {
        "loss": {
            "name": "DiceLoss"
        },
        "scheduler": {
            "initial_lr": 0.001,
            "lr_scheduler": {
                "name": "CosineAnnealingLR",
                "base_lr": 1e-5,
                "max_lr": 1e-2
            }
        }
```

Suppose we want to rename the key for `name` in the `loss` dictionary to `loss_name`. We could update the `KEY_CHANGE_DICT` as follows:

```KEY_CHANGE_DICT = {'UNet3D': 'Modified3DUNet', 'name': 'loss_name'}```

But then we run into the issue: which `name` key are we referring to? Do we mean the `name` key associated with the `loss` dictionary, or do we mean the `name` key associated with the `lr_scheduler` dictionary? In this example, we may end up updating the incorrect key.

## Issue Solution

I think we somehow need to add some information into the `KEY_CHANGE_DICT` to specify which part of the JSON tree we intend to update.

mariehbourget:
As discussed in #699, I added another [list](https://github.com/ivadomed/ivadomed/blob/a56e9d97ef935af0cdfab17d66e3a3e7b8d5debe/ivadomed/config_manager.py#L70) `KEY_SPLIT_DATASET_CHANGE_LST` and [function](https://github.com/ivadomed/ivadomed/blob/a56e9d97ef935af0cdfab17d66e3a3e7b8d5debe/ivadomed/config_manager.py#L132) `change_keys_values` to handle key/value changes for old split dataset params.

There is probably a better to do this. As a reference, the key/value change is:

[...]

PR #699, opened 2021-03-03
New loader: First integration into pipeline

### LINKED ISSUE #698: First integration of new loader in pipeline
## Context

The step 1 of the refactoring was to generate a bids dataframe: done in #538, #687 and #690 for anat and microscopy data.

The step 2 was to adapt the splitting methods (train/valid/test) to split using the dataframe and generalized the data used for splitting (subject, participant_id, sample, institution_id, disease, etc) : done in #591.

The step 3 was to refactor create_bids_dataframe function in a BidsDataframe class : done in #648, and to adapt BidsDataset classes : done in #691.

## To do in this issue

- Integrate the changes made for the new loader (steps 1-3) to the current codebase:
    * refactoring in `main`
    * update `data-testing dataset`
    * update `unit_tests`
    * update config and default config files
    * update documentation to reflect changes in config file for split
    * refactor other functions that use “path data” as an input, examples: `evaluate` and `run_segment_command`

## Next steps

- Add file format specific loaders in class and methods called by `load_filenames` (SegmentationPair, get_pair_shape, get_pair_metadata and get_pair_data) to load data as numpy array.

### LINKED ISSUE #701: Doc: wrong key in HeMis example
While working on the integration of the new loader, I found a wrong key in the config file example in documentation for [HeMis](https://ivadomed.org/en/latest/configuration_file.html#hemisunet) architecture.

`hdf5_path` should be `path_hdf5`.

I'll fix it with other changes to documentation in #699.

### LINKED ISSUE #709: GPU missing key error in Cascaded architecture tutorial
While running the tutorial `Cascaded architecture`, I have this error:
```
  File "/home/mhbourget/ivadomed/ivadomed/main.py", line 519, in run_main
    run_command(context=context,
  File "/home/mhbourget/ivadomed/ivadomed/main.py", line 347, in run_command
    ds_valid = get_dataset(loader_params, valid_lst, transform_valid_params, cuda_available, device, 'validation')
  File "/home/mhbourget/ivadomed/ivadomed/main.py", line 113, in get_dataset
    ds = imed_loader.load_dataset(**{**loader_params,
  File "/home/mhbourget/ivadomed/ivadomed/loader/loader.py", line 93, in load_dataset
    dataset = BidsDataset(path_data,
  File "/home/mhbourget/ivadomed/ivadomed/loader/loader.py", line 827, in __init__
    bounding_box_dict = imed_obj_detect.load_bounding_boxes(object_detection_params,
  File "/home/mhbourget/ivadomed/ivadomed/object_detection/utils.py", line 254, in load_bounding_boxes
    object_detection_params['gpu'],
KeyError: 'gpu'
```
The key `gpu` is missing in `object_detection_params` in the `config.json` used in the tutorial.
The key should be add to `config_default.json`, or a default value set in [load_bounding_box](https://github.com/ivadomed/ivadomed/blob/321a91c7e3c82e6296764895e39695b04a80c8af/ivadomed/object_detection/utils.py#L254).

mariehbourget:
### Context

Fixes #698, Fixes #701, Fixes #709
It aims to integrate the first 3 steps of the project into the current pipeline.

### Changes done in this PR:

- The content of `loader/loader.py`, `loader/adaptative.py` and `object_detection/utils.py` is replaced by the content of `loader/loader_new.py`, `loader/adaptative_new.py` and `object_detection/utils_new.py` from PR #596, `_new` files are deleted.
- In `loader/utils.py`, the content of `split_dataset`, `get_new_subject_split` and `get_subdatasets_subjects_list` is replaced by the content of the same `_new` functions from PR #591, `_new` functions are deleted.
- Refactoring of `main.py`, including the `run_segment_command` function.
- Refactoring of `evaluate` in `evaluation.py`.
- Remove `bids_neuropoly` dependency when able (`loader/loader.py`, `loader/adaptative.py`, `main.py`)
- Update `unit_tests` for new pipeline
- Update config and default config files
- Update documentation to reflect changes in config file for split
- Print the real fractions after the split
- Add `change_key_value` function in `config_manager.py` for retrocompatibility in config between `method/center_test` and `split_method/data_testing` keys and values

### Changes TODO in this PR:
- Address BIDS compatibility issues and `contrast_params` in `config_classification.json`, `config_sctTesting.json` and `config_vertebral_labeling.json`.
- Is `config/contrast_dct.json` still used?

### Changes that could be done in another PR after the merge:
- Update tutorial 1 examples to match new split (random seed is now applied to filenames and not participant_id so the split is different)
- Check if `merge_bids_dataset` is still used somewhere, should be replace by the BidsDataframe: this is now the only funtion that is dependent on `bids_neuropoly` in `loader/utils.py`.
- `test_loader.py` is passing but needs to be updated to include `participant_id` and `sample_id` in the dataframe, and to test the multiple folders feature.
- Refactor the for loop in `loader.py` to extract some of the logic into other functions for better readability. Might concern `adaptative.py` as well.


mariehbourget:
This PR is ready for review. The changes are listed in the first comment above. As the new loader is integrated in the pipeline, it can be tested with the current CLI.

### Important points or questions:

1.	**Config files contrast_params**

`config_classification.json` and `config_sctTesting.json` use `contrast_params` like `acq-ax00012_T2w`. I think this is due to the old parsing of filename in `bids_neuropoly` where the code assumed that the filename was `sub-XX_modality`. In the new loader with `pybids`, the last part of the filename (after the last underscore) is recognized as the modality. The current config files would not work unless `acq-XXXX_T2w` is changed for `T2w`, `acq-XXXX_T1w` for `T1w` and `acq-XXXX_T2star` for `T2star`. Is there another reason than the filename parsing to have multiple modalities in these fields?

There is also a file `config_dct.json` that seems to be some king of lookup table for those “special” modalities. I could not find where it is used in the code and why. Is the file still in use?

For `config_vertebral_labeling.json`, there is an issue with BIDS compliance. The `contrast_params` are `T2w_straight_mid`. Because we now rely on `pybids`, this config file and the dataset would have to be changed to be BIDS compliant i.e. modality should be T2w after the last underscore in the filenames. @lrouhier, is this something that can/should be corrected at this point?

2.	**Config files split dataset params**

The keys `method` and `center_test` are changed to `split_method` and `data_testing`. Do we want to maintain retrocompatibility with the old keys?

In config manager, there is a [KEY_CHANGE_DICT](https://github.com/ivadomed/ivadomed/blob/86da257c5f1c7d5841e693fbeeb8077fc3f33ff0/ivadomed/config_manager.py#L69) to do that but in that case, I would like to change the key and the value. I don’t know how to best do this. @ahill187, @gisellemartel?

For reference, the changes would be:
- `"method": "per_patient"` --> `"split_method" : "participant_id"`
- `"method": "per_center"` --> `"split_method" : "participant_id" and "data_testing": {"data_type": null, "data_value":[]}`
- `"method": "per_center" and "center_test": [LIST_CONTENT]` --> `"split_method" : "participant_id" and "data_testing": {"data_type": "institution_id", "data_value":[LIST_CONTENT]}`

3.	**run_segment_command in main.py**

This function was a bit tricky to refactor, I would appreciate if it could be tested by the reviewers with previous experiment to make sure everything is alright, especially with `multichannel` and `film_layers`.

4.	 **test_split_dataset.py**

Previously, when a `center_list` was provided, the `train_frac` was applied to the remainder of participant after isolating the test set, instead of the total number of participants. We decided in the generalization of the function in #591 to correct this behavior. 
In the new loader, the `train_frac` is always applied to the total number. For that reason, I had to change the the assert comparison value in `test_per_center_testcenter_0` and `test_per_center_balance`.

5.	**Split retrocompatibility and Tutorial 1**

Because the split dataset is now done with the list of filenames instead of participants, the split is not the same as before with a same `random_seed`. Is this going to be a problem?
Note that as of PR #691, old `joblib` files with `participant_id` instead of `filenames` are retrocompatible and should work.
A related minor issue is the tutorial 1 in RTD, where the examples given don’t match the new split. this should be address in another PR.

gisellemartel:
> 1. **Config files split dataset params**
> 
> The keys `method` and `center_test` are changed to `split_method` and `data_testing`. Do we want to maintain retrocompatibility with the old keys?
> 
> In config manager, there is a [KEY_CHANGE_DICT](https://github.com/ivadomed/ivadomed/blob/86da257c5f1c7d5841e693fbeeb8077fc3f33ff0/ivadomed/config_manager.py#L69) to do that but in that case, I would like to change the key and the value. I don’t know how to best do this. @ahill187, @gisellemartel?
> 
> For reference, the changes would be:
> 
> * `"method": "per_patient"` --> `"split_method" : "participant_id"`
> * `"method": "per_center"` --> `"split_method" : "participant_id" and "data_testing": {"data_type": null, "data_value":[]}`
> * `"method": "per_center" and "center_test": [LIST_CONTENT]` --> `"split_method" : "participant_id" and "data_testing": {"data_type": "institution_id", "data_value":[LIST_CONTENT]}`
> 


One suggestion I would have for `KEY_CHANGE_DICT` is to use have both key and value be associated to each old key. So in the case of "method" we could have something like `"method": {"split_method" : "participant_id"}`. In the code you can just check if the old key is found in the config file, and if so you can swap both the key and associated value using the `KEY_CHANGE_DICT` lookup table

ivadomed/loader/adaptative.py
```diff
@@ -482,8 +472,9 @@ class HDF5Dataset:
     """HDF5 dataset object.
 
     Args:
+        bids_df (BidsDataframe): Object containing dataframe with all BIDS image files and their metadata.
         path_data (list) or (str): Path(s) to BIDS datasets
-        subject_lst (list of str): List of subjects.
+        subject_lst (list of str): List of subject filenames.
```
gisellemartel, line 477: might be better to rename `subject_lst` to `subject_file_lst` for code cohesion

ivadomed/loader/loader.py
```diff
@@ -708,8 +712,9 @@ class Bids3DDataset(MRI3DSubVolumeSegmentationDataset):
     """BIDS specific dataset loader for 3D dataset.
 
     Args:
+        bids_df (BidsDataframe): Object containing dataframe with all BIDS image files and their metadata.
         path_data (list) or (str): List of Paths to the BIDS datasets.
-        subject_lst (list): Subject names list.
+        subject_lst (list): Subject filenames list.
```
gisellemartel, line 717: could be renamed `subject_file_lst`

ivadomed/loader/loader.py
```diff
@@ -817,104 +825,94 @@ def __init__(self, path_data, subject_lst, target_suffix, contrast_params, slice
...
+            metadata['contrast'] = contrast
+
+            if len(bounding_box_dict):
+                # Take only one bounding box for cropping
+                metadata['bounding_box'] = bounding_box_dict[str(df_sub['path'].values[0])][0]
+
+            if metadata_choice == 'mri_params':
+                if not all([imed_film.check_isMRIparam(m, metadata, subject, self.metadata) for m in
+                            self.metadata.keys()]):
+                    continue
```
gisellemartel, line 888: there is a lot of code nested within the for loop from line 844. Perhaps there is a way to extract some of the logic into other functions for better readability

gisellemartel (commented):
Great work! Went through the code and left a few minor suggestions.

ivadomed/loader/adaptative.py
mariehbourget: Good catch, thanks!
I fixed it in 4e7ce79.
I made similar changes to split functions names for code cohesion in 991c3b6.

ivadomed/loader/loader.py
mariehbourget: Fixed in 4e7ce79.

ivadomed/loader/loader.py
mariehbourget: I agree with you. However I think this is out of scope for this already big PR, I suggest we open a separate issue for this.

andreanne-lemay (approved):
It looks great! I tested all combinations of 2D, 3D, multi-class, multi-channel, single-channel models, and everything worked smoothly! 

We probably want to delete the `df_new_loader.py` file or at least update it. Currently, it tries to import code from a file that doesn't exist anymore.

mariehbourget:
> We probably want to delete the `df_new_loader.py` file or at least update it. Currently, it tries to import code from a file that doesn't exist anymore.

Good point! I updated it in 050064a. I will still use it to update the reference df in the data testing repo.

ivadomed/config/config_default.json
```diff
@@ -33,8 +33,8 @@
     "split_dataset": {
         "fname_split": null,
         "random_seed": 6,
-        "center_test": [],
```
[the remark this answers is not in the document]
andreanne-lemay: Good to know! Most BIDS dataset only have one file extension so I think it's fine to let it by default to an empty list.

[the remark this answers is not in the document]
mariehbourget: Fixed in f18dc9e. I found a way to force index files under a `microscopy` folder only.
So now, a non-bids-compliant file like a `.png` that would be under a folder `anat` won't be indexed by pybids.

mariehbourget:
> One suggestion I would have for `KEY_CHANGE_DICT` is to use have both key and value be associated to each old key. So in the case of "method" we could have something like `"method": {"split_method" : "participant_id"}`. In the code you can just check if the old key is found in the config file, and if so you can swap both the key and associated value using the `KEY_CHANGE_DICT` lookup table.
> 
> There is also this [issue](https://github.com/ivadomed/ivadomed/issues/670) which indicates that changes will be need to be made in order for the look-up table to work as-intended. Perhaps suggested changes from this PR could be added to the linked issue?

@gisellemartel I did something in b1208b7 to ensure retrocompatibility with old config. Because it was not a 1 to 1 change between and old and new key/value pairs, I was not able to use the existing `KEY_CHANGE_DICT`. My solution works but there is likely a better way to do it. As you suggested, I will add a comment about this in related issue #670.

gisellemartel (approved):
Great work!

charleygros:
> `config_classification.json` and `config_sctTesting.json` use `contrast_params` like `acq-ax00012_T2w`. I think this is due to the old parsing of filename in `bids_neuropoly` where the code assumed that the filename was `sub-XX_modality`. In the new loader with `pybids`, the last part of the filename (after the last underscore) is recognized as the modality. The current config files would not work unless `acq-XXXX_T2w` is changed for `T2w`, `acq-XXXX_T1w` for `T1w` and `acq-XXXX_T2star` for `T2star`. Is there another reason than the filename parsing to have multiple modalities in these fields?

This is a relic of ol' times mate! :-) I think these config files should be purified, as explained in this newly opened issue #706.
To explain: in this experiment we wanted to use only "axial" or "isotropic" images, hence this long list --> we wanted the "acq-axXX_T2w" but not the "acq-sagXX_T2w".
This is a very very specific user case and not easily generalisable --> I would suggest not to bother trying to fix this config file!
The same behaviour could be achieved by... eg "manually" creating a .jobfile containing the images of interest or filtering images based on nifti caracteristics. The latter is not implemented yet: we will think about it before crossing the bridge, leave it for now.

> There is also a file `config_dct.json` that seems to be some king of lookup table for those “special” modalities. I could not find where it is used in the code and why. Is the file still in use?

Oh yes... another relic!! It was to map the contrasts found in `sct_testing/large` with the tree main MRI contrasts (T1w, T2w, T2star). Again, I think it should be removed or moved to `dev` as only used by neuropoly-users, once a blue moon.

> Previously, when a `center_list` was provided, the `train_frac` was applied to the remainder of participant after isolating the test set, instead of the total number of participants. We decided in the generalization of the function in #591 to correct this behavior.
> In the new loader, the `train_frac` is always applied to the total number. For that reason, I had to change the the assert comparison value in `test_per_center_testcenter_0` and `test_per_center_balance`.

Fantastic, crystal clear.

> Because the split dataset is now done with the list of filenames instead of participants, the split is not the same as before with a same `random_seed`. Is this going to be a problem?

All good for me! Very thoughtful asking!!

charleygros (approved):
Everything is running smoothly on my side! 🥇 

No doubt that this is a major piece of work... corner stone of `ivadomed`... along with the XXX previous related PRs. Many many thanks @mariehbourget ! And always a pleasure to read your PR thanks of the efforts you put in the documentation!!

lrouhier:
>For config_vertebral_labeling.json, there is an issue with BIDS compliance. The contrast_params are T2w_straight_mid. Because we now rely on pybids, this config file and the dataset would have to be changed to be BIDS compliant i.e. modality should be T2w after the last underscore in the filenames. @lrouhier, is this something that can/should be corrected at this point?

I think it would be better to fix that in another PR. There is a quick way to fix that. When I am extracting the mid (with `ivadomed/script/prepare_dataset_vertebral_labeling`), I can change the name of the saved file. I opened an issue (#707), I'll check that out today. 

By the way: Great work !

mariehbourget:
Thank you @charleygros and @lrouhier for looking into config files.
So for this PR, I will leave `classification.json`, `sctTesting.json` and `config_vertebral_labeling.json` as they are, and we can deal with them with issue #706 and #707.

For `contrast_dct.json`, I finally found where it is used. It is a lookup table for the option "contrasts" in `film.py` [here](https://github.com/ivadomed/ivadomed/blob/321a91c7e3c82e6296764895e39695b04a80c8af/ivadomed/loader/film.py#L13) and [here](https://github.com/ivadomed/ivadomed/blob/321a91c7e3c82e6296764895e39695b04a80c8af/ivadomed/loader/film.py#L67).
I'll leave it in `config` for now. However, with the new loader, the config files associated with this option of film probably won't work (for the same reasons as classification and sctTesting). Also, the following [contrast categories](https://github.com/ivadomed/ivadomed/blob/321a91c7e3c82e6296764895e39695b04a80c8af/ivadomed/loader/film.py#L16) won't work either: `"acq-MToff_MTS": 3, "acq-MTon_MTS": 4, "acq-T1w_MTS": 5`. @charleygros Who should I talk to clarify the intended behavior and make a plan to fix it?

charleygros:
> However, with the new loader, the config files associated with this option of film probably won't work (for the same reasons as classification and sctTesting). Also, the following [contrast categories](https://github.com/ivadomed/ivadomed/blob/321a91c7e3c82e6296764895e39695b04a80c8af/ivadomed/loader/film.py#L16) won't work either: `"acq-MToff_MTS": 3, "acq-MTon_MTS": 4, "acq-T1w_MTS": 5`. @charleygros Who should I talk to to clarify the intended behavior and make a plan to fix it?

Yes. Again, a relic of old internal project. It needs to change, be more generalizable.

> @charleygros Who should I talk to to clarify the intended behavior and make a plan to fix it?

eheh, probably mee 🙈 ! I won't be around a keyboard next week, so feel free to comment these lines if they prevent you from merging this PR, open an issue, assign me and I will be happy to correct this when back :-)

mariehbourget:
> feel free to comment these lines if they prevent you from merging this PR, open an issue, assign me and I will be happy to correct this when back :-)

Ok thanks! I opened the issue #710 for this.
I won't be commenting any lines but the metadata option in FiLM won't work when merging the PR.

Issue #686, opened 2021-02-15
Image segmentation (of the spinal cord/CSF) with missing modalities


leandervaneekelen:
**Context**
We often train multi-modal deep learning models (in our case, various contrasts of MRI, e.g. T1(ce), T2, T2*, FLAIR, etc) to take advantage of the complementary information that these modalities give about the underlying anatomy. However, in a realistic setting, each subject in a dataset might have a variable number of available modalities - after all, sometimes data gets corrupted, a patient misses a scan, data goes missing. Traditional deep learning models cannot handle this scenario of missing modalities well. 

Developing models that would be able to segment structures in images with an arbitrary combination of modalities would be a great boon to the overall applicability of deep learning to neural imaging and clinical practice.

**Existing work**
Ivadomed currently has an implementation of the HeMIS framework (Hetero-modal image segmentation, originally proposed by [Havaei et al](https://arxiv.org/abs/1607.05194)). In brief, HeMIS is a convolutional neural network (CNN) architecture that works by having an independent convolutional branch per modality; all branches then feed into a single latent space by calculating statistics over each branch. The latent space contains 'information' from all modalities and can then be further processed to get the desired segmentation. If a modality is missing, that branch is simply skipped and won't contribute to the latent space. See the figure below.

![image](https://user-images.githubusercontent.com/47320151/107994263-96e4d700-6fdc-11eb-82bf-93fb2506850a.png)

I (Leander van Eekelen) have done a short internship in the group on applying HeMIS for spinal cord and CSF segmentation in MRI (various contrasts) and have gotten slightly lackluster results. The details about the done work can be found in the repository dedicated to the project [here](https://github.com/ivadomed/pipeline-hemis).

For the continuation of the project, we want to consider other (more state of the art) alternatives. 

**Potential future avenues**
One promising avenue of research is teacher-student networks for knowledge distillation. We train a teacher network for each available modality and have all of these teachers integrate their knowledge into a single student network. Recommended reading material is, among others:
* https://dl.acm.org/doi/abs/10.1145/3394486.3403234
* https://ieeexplore.ieee.org/document/8979396
* https://arxiv.org/abs/1908.06683

arbalest2:
This paper looks promising :

[SMIL: Multimodal Learning with Severely Missing Modality](https://www.aaai.org/AAAI21Papers/AAAI-437.MaM.pdf)
[paper talk](https://slideslive.com/38947797/smil-multimodal-learning-with-severely-missing-modality)

[...]

PR #705, opened 2021-03-10
Input-level dropout feature

andreanne-lemay:
## Description
This PR implements the input-level dropout presented in https://www.nature.com/articles/s41746-021-00398-4. The purpose of this feature is to train models robust to missing modalities. When the option is enabled, 0 to all channels - 1 are randomly chosen to be set to zeros simulating missing modalities. If a channel is already empty due to an actual missing modality, it will be considered by the function as a dropped modality.

To apply the input-level dropout, the following optional parameter can be set to `True`:

```
"loader_parameters": {
    "is_input_dropout": true
}
```

## Linked issues
This method is a simpler alternative to HeMIS which presented poor results #686. It can also a good training strategy for the Karolinska model which has partial missing modalities (see [issue](https://github.com/ivadomed/pipeline-ms-lesion/issues/3) )

## PR content
- [x] Addition of `dropout_input` function
- [x] PEP8 correction in the `main.py` file 
- [x] Update RDT
- [x] Add unit test

## Design questions
Do we want to allow more than one modality to be set to zero?
Do we want an option to decide which modality we want to drop (e.g., always train with T2w or T1w, but always keep T2star)?
Do we want to allow input dropout at test time (currently it's impossible)?

Update: These questions have been addressed further in this PR.


charleygros:
First and foremost, this is 1,000 cool.

> ## Design questions
> 
> Do we want to allow more than one modality to be set to zero?

In real life datasets, that's what we experiment... so I would say yes! (ofc: make sure at least one is available).

> Do we want an option to decide which modality we want to drop (e.g., always train with T2w or T1w, but always keep T2star)?

Yes! And I was thinking of a suggestion that we could couple with your question: What about adding a parameter that controls the frequency of this dropping? AND this dropping could be modality dependent. For instance, having in the config file: `[0.5, 0.7, 1]` for "T2w or T1w can drop but always keep T2star".

> Do we want to allow input dropout at test time (currently it's impossible)?

That would be good to validate (eg with SpineGeneric) this method before running it on our real life dataset, so I would say yes!

Thinking out loud: Let's say you have a dataset with 3 images for each subjects, same size, same resolution, co registered. But, as they have different coverage along (IS), some sections (eg top of T2star) are filled with zeros --> so basically we will have the same behaviour that you are forcing here while having `is_input_dropout: false` and `"filter_empty_input": false`. So for this dataset we will not use this feature per say (except if we want to do some transfer learning etc). But this paper (and this implementation) kinda validates that this approach is promising for our user case.

charleygros:
Thinking out loud: Let's say you have a dataset with 3 images for each subjects, same size, same resolution, co registered. But, as they have different coverage along (IS), some sections (eg top of T2star) are filled with zeros --> so basically we will have the same behaviour that you are forcing here while having is_input_dropout: false and "filter_empty_input": false. So for this dataset we will not use this feature per say (except if we want to do some transfer learning etc). But this paper (and this implementation) kinda validates that this approach is promising for our user case.

charleygros:
Again just thinking, as the intensity normalisation is done per channel (with the @ thingy) we will not have issues with dropped channels. All goodie.

andreanne-lemay:
> > ## Design questions
> > Do we want to allow more than one modality to be set to zero?
> 
> In real life datasets, that's what we experiment... so I would say yes! (ofc: make sure at least one is available).

Great, so I updated the code to fit this behavior (40f1fc1). From 0 to all channels - 1 will be dropped.



> > Do we want to allow input dropout at test time (currently it's impossible)?
> 
> That would be good to validate (eg with SpineGeneric) this method before running it on our real life dataset, so I would say yes!

Goop point! I enabled the option to dropout inputs during testing but I added a warning to inform the user if the option is on during test time (70d984d). 




> > Do we want an option to decide which modality we want to drop (e.g., always train with T2w or T1w, but always keep T2star)?
> 
> Yes! And I was thinking of a suggestion that we could couple with your question: What about adding a parameter that controls the frequency of this dropping? AND this dropping could be modality dependent. For instance, having in the config file: `[0.5, 0.7, 1]` for "T2w or T1w can drop but always keep T2star".

I didn't implement this option yet, but it would be very interesting! I think the first step would be to test the current version, and I believe the model will benefit to be trained when all modalities can be dropped since it teaches the model to predict in harder conditions.

charleygros (approved):
Looks ready to be merged :-)

I opened an issue, #730, re the dropping parameter, so that we remember about it.

Issue #713, opened 2021-03-14
Verify qform, sform, qform_code and sform_code on the output prediction


jcohenadad:
Context: https://github.com/ivadomed/ivadomed/issues/711

This should be done after fixing the q/sform on the testing data: https://github.com/ivadomed/data-testing/issues/12

PR #714, opened 2021-03-14
Copy nibabel header when creating output prediction

### LINKED ISSUE #711: Output segmentation produces inconsistent qform/sform codes
### Description

When running `inference.pred_to_nib()` on an input image which has qform_code=1 and sform_code=1, the output segmentation has the fields qform_code=0 (incorrect) and sform_code=2. 

For more details on q/sform conventions, see https://github.com/neuropoly/spinalcordtoolbox/issues/3283
For more details on how to reproduce this issue, see https://github.com/neuropoly/spinalcordtoolbox/issues/3284

### Suggestion

Change the way the nibabel object is created for the output segmentation (also see [nibabel doc](https://nipy.org/nibabel/nifti_images.html#default-sform-and-qform-codes)).

jcohenadad:
#### GitHub

#### PR contents

## Description
Before, the output prediction was created by instantiating a new nibabel object, via the copy of the original affine matrix. However, doing so does *not* ensure the same qform_code and sform_code are preserved. This PR fixes the issue.
More context: https://nipy.org/nibabel/nifti_images.html#default-sform-and-qform-codes

Testing for the proposed change will be implemented in a subsequent PR: https://github.com/ivadomed/ivadomed/issues/713

This PR also includes minor fixes to the unit testing documentation.

## Linked issues
Fixes #711
Related to https://github.com/neuropoly/spinalcordtoolbox/issues/3284, https://github.com/ivadomed/ivadomed/issues/720


ivadomed/inference.py
```diff
@@ -86,7 +86,16 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
                                               nib_ref.header['pixdim'][1:4],
                                               fname_prefix)
         arr_pred_ref_space = postpro.apply()
-    nib_pred = nib.Nifti1Image(arr_pred_ref_space, nib_ref.affine)
+    if arr_pred_ref_space.shape == nib_ref.get_fdata().shape:
+        # Here we prefer to copy the header (rather than just the affine matrix), in order to preserve the qform_code.
+        # See: https://github.com/ivadomed/ivadomed/issues/711
```
charleygros, line 91: Sounds great! and well documented, thanks

charleygros (commented):
Mmmm thank you for solving this... very specific issue!!

Less critical, but we may also want to apply this change to other instances in ivadomed?
- [uncertainty](https://github.com/ivadomed/ivadomed/blob/87d6a63d0bb0114bb3d3d8e13584089bf2ce0142/ivadomed/uncertainty.py#L76)
- [preprocessing](https://github.com/ivadomed/ivadomed/blob/3c9d32c41ef4222848f0dfb024e89c11f7264046/ivadomed/preprocessing.py#L36)
- [postprocessing](https://github.com/ivadomed/ivadomed/blob/3004dfe53afcc04529380ac419370ccb76f3a128/ivadomed/postprocessing.py#L26)
- [visualize](https://github.com/ivadomed/ivadomed/blob/a8f1839ace712347f046a4131cdd75a8c6cc43da/ivadomed/visualize.py#L226)
- [loader/adaptive](https://github.com/ivadomed/ivadomed/blob/fb3c913ddc3fb1942884bdbb16a7796e94d20192/ivadomed/loader/adaptative.py#L753)
- [evaluation](https://github.com/ivadomed/ivadomed/blob/87156e33be17eb96edb124e01e98d57619ea8a59/ivadomed/evaluation.py#L84)

jcohenadad:
> Mmmm thank you for solving this... very specific issue!!
> 
> Less critical, but we may also want to apply this change to other instances in ivadomed?
> 
> * [uncertainty](https://github.com/ivadomed/ivadomed/blob/87d6a63d0bb0114bb3d3d8e13584089bf2ce0142/ivadomed/uncertainty.py#L76)
> * [preprocessing](https://github.com/ivadomed/ivadomed/blob/3c9d32c41ef4222848f0dfb024e89c11f7264046/ivadomed/preprocessing.py#L36)
> * [postprocessing](https://github.com/ivadomed/ivadomed/blob/3004dfe53afcc04529380ac419370ccb76f3a128/ivadomed/postprocessing.py#L26)
> * [visualize](https://github.com/ivadomed/ivadomed/blob/a8f1839ace712347f046a4131cdd75a8c6cc43da/ivadomed/visualize.py#L226)
> * [loader/adaptive](https://github.com/ivadomed/ivadomed/blob/fb3c913ddc3fb1942884bdbb16a7796e94d20192/ivadomed/loader/adaptative.py#L753)
> * [evaluation](https://github.com/ivadomed/ivadomed/blob/87156e33be17eb96edb124e01e98d57619ea8a59/ivadomed/evaluation.py#L84)

Excellent point-- issue opened here https://github.com/ivadomed/ivadomed/issues/715

ivadomed/inference.py
```diff
@@ -86,7 +86,23 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
...
         arr_pred_ref_space = postpro.apply()
-    nib_pred = nib.Nifti1Image(arr_pred_ref_space, nib_ref.affine)
+    # TODO: remove the "[0:3]" in the test if test below. The reason I needed to add it is because in some cases, the
+    #  prediction has the shape (nx, ny, nz, 1) while the inpute image is (nx, ny, nz), even for a one-channel
+    #  prediction. Consequently, the if check below goes to the "else" case, and
+    #  issue https://github.com/ivadomed/ivadomed/issues/711 is still happening.
+    #  I'm afraid this [0:3] hack will break for 2D data. Also, we might want to make sure the output arr_pred_ref_space
+    #  always has the same shape as the input image. For multiple class prediction, maybe we can find another way
+    #  to store the predictions (e.g. list of np arrays?).
+    if arr_pred_ref_space.shape[0:3] == nib_ref.get_fdata().shape[0:3]:
```
andreanne-lemay, line 96: From what I understand in the description above, if the dimensions of the prediction is (nx, ny, nz, 1) while the input image is (nx, ny, nz) the code should go to the "else" statement (correct me if I'm wrong). It seems to me that in all cases the code under `if arr_pred_ref_space.shape[0:3] == nib_ref.get_fdata().shape[0:3]:` will be ran since `[nx, ny, nz, 1][0:3] == [nx, ny, nz][[0:3] `. Maybe this bit of code is in the case of a prediction with the dimensions (1, nx, ny, nz)? 

From this error https://github.com/ivadomed/ivadomed/pull/714/checks?check_run_id=2107966525, the problem seems to come from the `save_color_labels` function which generates nifti files with non-int or float data. Rather than introducing this condition that might cause errors with 2D images, I'd simply comment the lines related to this function (which simply creates an extra file that we barely use) and adjust the function `save_color_labels`  in a future PR. If this plans work for you, I can take care of it.

jcohenadad: > if the dimensions of the prediction is (nx, ny, nz, 1) while the input image is (nx, ny, nz) the code should go to the "else" statement (correct me if I'm wrong). 

sorry i wasn't clear-- let me know if this is clearer-- if it is i will update the code with this comment below:
~~~
    # TODO: remove the "[0:3]" in "if test" below. The reason I added it is because in some cases, the
    #  prediction has the shape (nx, ny, nz, 1) while the input image is (nx, ny, nz), even for a one-channel
    #  prediction. Consequently, the "if check" below goes to the "else" case, and
    #  issue https://github.com/ivadomed/ivadomed/issues/711 is still happening.
    #  What we want is: if the output prediction has the same (nx, ny, nz) as the reference image, the we need
    #  to copy the header (not just the affine), in order to fix #711.
    #  However, I am afraid this "[0:3] hack" will break for 2D data. Also, we might want to make sure the output arr_pred_ref_space
    #  always has the same shape as the input image. For multiple class prediction, maybe we can find another way
    #  to store the predictions (e.g. list of np arrays?).
~~~

ivadomed/inference.py
```diff
@@ -314,7 +330,7 @@ def split_classes(nib_prediction):
     pred = nib_prediction.get_fdata()
     pred_list = []
     for c in range(pred.shape[-1]):
-        class_pred = nib.Nifti1Image(pred[..., c].astype('float32'), nib_prediction.affine)
+        class_pred = nib.Nifti1Image(pred[..., c].astype('float32'), None, nib_prediction.header)
```
jcohenadad, line 333: need to copy to another object (to avoid pointer conflicts)
```suggestion
        class_pred = nib.Nifti1Image(pred[..., c].astype('float32'), None, nib_prediction.header.copy())
```

ivadomed/inference.py
jcohenadad: > From this error https://github.com/ivadomed/ivadomed/pull/714/checks?check_run_id=2107966525, the problem seems to come from the save_color_labels function which generates nifti files with non-int or float data. Rather than introducing this condition that might cause errors with 2D images, I'd simply comment the lines related to this function (which simply creates an extra file that we barely use) and adjust the function save_color_labels in a future PR. If this plans work for you, I can take care of it.

that would be awesome! yes-- i am completely swamped with other tasks now

jcohenadad: so IIUC: what you are suggesting is that after the fix you suggest, `arr_pred_ref_space.shape` will ALWAYS be the same as `nib_ref.get_fdata().shape`?  glancing at the code, my understanding was that in some cases the 4th dimension is used for different channels (so you suggest to not do that anymore?)

andreanne-lemay: Yes, exactly! The prediction and the original image will always have different dimensions when using ivadomed as we store the different classes in the 4th dimension. Changing this would require an important refactoring. However, in SCT we always store the data in different files (one file per class) so this shape discrepancy should never happen when calling segment_volume from SCT. I'll do some test to verify if this is true. If so, I think only this line is required `nib_pred = nib.Nifti1Image(arr_pred_ref_space, None, nib_ref.header.copy())` since this code still works when the shapes are different (even though it might generate the qform and form codes, but with the current form of the code I think it is inevitable).

jcohenadad: > since this code still works when the shapes are different

when i tested it, the following code:
~~~
nib_pred = nib.Nifti1Image(arr_pred_ref_space, None, nib_ref.header.copy())
~~~

will produce an error if e.g. one is (nx, ny, nz) and the other is (nx, ny, nz, 1). Hence my "hack"

ahill187: If there is a concern that this "[0:3] hack" will break for 2D data, should we add a test for 2D data?

andreanne-lemay: I was wrong, the predictions generated from SCT initially have a shape of (nx, ny, nz, X) and are converted to a list afterward: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/inference.py#L300. There will always be a shape discrepancy in the `pred_to_nib` function. 

I tested the code in its actual form (with the header copied even if the shapes are different), and I got the following qform/sform codes: 
image

The expected behavior is that qform and sform should always match the original image?

ahill187:
What's the reasoning for having the tests in a separate pull request?

jcohenadad:
> What's the reasoning for having the tests in a separate pull request?

because i don't have the time to do it 😊   but if pls feel free to chip in1

ivadomed/inference.py
jcohenadad: > The expected behavior is that qform and sform should always match the original image?

that is correct.

ahill187:
> > What's the reasoning for having the tests in a separate pull request?
> 
> because i don't have the time to do it 😊 but if pls feel free to chip in1

Haha okay, I'll look into adding some tests then! :)

andreanne-lemay:
@jcohenadad the easiest patch I can think of is doing something like this in the `segment_volume` function ([here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/inference.py#L300)), instead of changing the code in `pred_to_nib` where the prediction will never have the same dimensions as the original image.

```               
                pred_list = split_classes(pred_nib)

                def change_header(pred_list, fname_ref):
                    """Change header to keep qform and sform codes
                    Args:
                        pred_list (list): List of predictions (with shape H,W,D)
                        fname_ref (str): Path to the reference image used to extract the header.
                    
                    Return:
                        list: List of prediction with updated header.                      
                    
                    """
                    nib_ref = nib.load(fname_ref)
                    pred_updated_list = []
                    print("CHANGE HEADER")
                    for pred in pred_list:
                        nib_pred = nib.Nifti1Image(pred.get_fdata(), None, nib_ref.header.copy())
                        pred_updated_list.append(nib_pred)
                    return pred_updated_list

                pred_list = change_header(pred_list, fname_images[0])
                target_list = context['loader_parameters']['target_suffix']
```

If you agree, with this change I can update the code in the branch. I tested it locally and it fixes the issue of the qform and sform codes.

ivadomed/inference.py
andreanne-lemay: The current version of the code doesn't fix the issue regarding the qform and sform codes I believe. See suggestion below.

andreanne-lemay:
> @jcohenadad the easiest patch I can think of is doing something like this in the `segment_volume` function ([here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/inference.py#L300)), instead of changing the code in `pred_to_nib` where the prediction will never have the same dimensions as the original image.
> 
> ```
>                 pred_list = split_classes(pred_nib)
> 
>                 def change_header(pred_list, fname_ref):
>                     """Change header to keep qform and sform codes
>                     Args:
>                         pred_list (list): List of predictions (with shape H,W,D)
>                         fname_ref (str): Path to the reference image used to extract the header.
>                     
>                     Return:
>                         list: List of prediction with updated header.                      
>                     
>                     """
>                     nib_ref = nib.load(fname_ref)
>                     pred_updated_list = []
>                     print("CHANGE HEADER")
>                     for pred in pred_list:
>                         nib_pred = nib.Nifti1Image(pred.get_fdata(), None, nib_ref.header.copy())
>                         pred_updated_list.append(nib_pred)
>                     return pred_updated_list
> 
>                 pred_list = change_header(pred_list, fname_images[0])
>                 target_list = context['loader_parameters']['target_suffix']
> ```
> 
> If you agree, with this change I can update the code in the branch. I tested it locally and it fixes the issue of the qform and sform codes.

Ah 3290f85 this does the same as I what I suggested. I think only the code in `split_classes` should be change. I don't think we need to change the code in `pred_to_nib`.

jcohenadad:
> I think only the code in `split_classes` should be changed. 

`split_classes` takes as input `nib_prediction`:
https://github.com/ivadomed/ivadomed/blob/0a4b7bb3fad8092b31b347332b16f5308be09a54/ivadomed/inference.py#L306

Which means, that if the header of `nib_prediction` is wrong, then the wrong header will be propagated inside `split_classes` as well, even if we introduce the change in https://github.com/ivadomed/ivadomed/commit/3290f85b0f417af451fbfd3d8f07152ec3ea44f2.

I think we need to change only one line in `pred_to_nib`:

If you are suggesting to only change this line:
https://github.com/ivadomed/ivadomed/blob/0a4b7bb3fad8092b31b347332b16f5308be09a54/ivadomed/inference.py#L89
into this:
> `nib_pred = nib.Nifti1Image( dataobj=arr_pred_ref_space, affine=None, header=nib_ref.header.copy() )`

we will end up with the situation i am describing in https://github.com/ivadomed/ivadomed/pull/714/files#r594479677 (hence the hack).

I feel we are going in circles-- maybe a quick chat would be more efficient

andreanne-lemay (approved):
LGTM!

ivadomed/testing.py
```diff
@@ -217,11 +220,13 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
-                        imed_visualize.save_color_labels(np.stack(pred_tmp_lst, -1),
-                                                     False,
-                                                     fname_tmp,
-                                                     fname_pred.split(".nii.gz")[0] + '_color.nii.gz',
-                                                     imed_utils.AXIS_DCT[testing_params['slice_axis']])
+                        logger.warning('No color labels saved due to a temporary issue. For more details see:'
+                                       'https://github.com/ivadomed/ivadomed/issues/720')
+                        # imed_visualize.save_color_labels(np.stack(pred_tmp_lst, -1),
+                        #                              False,
+                        #                              fname_tmp,
```
gisellemartel, line 227: Is there a reason we are leaving this commented out code here?

ivadomed/testing.py
```diff
@@ -270,11 +275,13 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
...
-                                                     batch['input_metadata'][smp_idx][0]['input_filenames'],
-                                                     fname_pred.split(".nii.gz")[0] + '_color.nii.gz',
-                                                     slice_axis)
+                        logger.warning('No color labels saved due to a temporary issue. For more details see:'
+                                       'https://github.com/ivadomed/ivadomed/issues/720')
+                        # imed_visualize.save_color_labels(pred_undo,
+                        #                              False,
+                        #                              batch['input_metadata'][smp_idx][0]['input_filenames'],
+                        #                              fname_pred.split(".nii.gz")[0] + '_color.nii.gz',
+                        #                              slice_axis)
```
gisellemartel, line 284: same as comment above

testing/unit_tests/test_rgb.py
```diff
@@ -9,23 +9,27 @@ def setup_function():
...
-    imed_visualize.save_color_labels(
-        gt_data=image_n,
-        binarize=False,
-        gt_filename=os.path.join(
-            __data_testing_dir__,
-            "derivatives/labels/sub-unf01/anat/sub-unf01_T2w_lesion-manual.nii.gz"),
-        output_filename=os.path.join(__tmp_dir__, "rgb_test.nii.gz"),
-        slice_axis=0
-    )
+# TODO: un-silence this test. This test was silenced because it was failing after introducing the copy of the header in
```
gisellemartel, line 12: If there is an issue ticket for this, it would be good practice to reference the issue # here

gisellemartel (approved):
Look's good! Just some minor suggestions for code quality

testing/unit_tests/test_rgb.py
jcohenadad: thanks, done in https://github.com/ivadomed/ivadomed/pull/714/commits/a00a54fc5c701ba6f717168c2ed43a972230b051

ivadomed/testing.py
jcohenadad: justified in eca1c5f089e1a3f5a5fefe248a7746af10ef7fff

ivadomed/testing.py
jcohenadad: justified in justified in eca1c5f089e1a3f5a5fefe248a7746af10ef7fff

PR #734, opened 2021-03-23
New loader: Load PNG/TIF/JPG microscopy files as numpy array

### LINKED ISSUE #733: Load PNG/TIF/JPG microscopy files as numpy array in new loader
## Context

This issue is part of the loader refactoring project to include microscopy data (#304).
The first 3 steps of the project were integrated in the current pipeline in PR #699.

## To do in this issue

The step 4 consist of adding support to load microscopy png file in the `SegmentationPair` class called by `load_filenames`, and modify `get_pair_shape`, `get_pair_metadata` and `get_pair_data` methods to load PNG microscopy data as numpy array.

## Next steps

For microscopy, we will need a feature to do 2D sub-patches of the images, similar to [_prepare_indices](https://github.com/ivadomed/ivadomed/blob/b977a20375c2e00bafb94b9a1bebe6bad8e2a44a/ivadomed/loader/loader.py#L659) in MRI3DSubVolumeSegmentationDataset class, which is not implemented in 2D.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR enables the loading of microsocpy PNG files as numpy arrays for training, as well as PNG/TIF/JPG files for inference.
I added 5 methods to the `SegmentationPair` class to replace functions from nibabel library specific to nifti files.

Nibabel functions `nib.load`, `nib.as_closest_canonical`, `handle.header.get_data_shape`, `handle.header.get_zooms`, `handle.get_fdata` are replaced respectively by `read_file`, `apply_canonical`, `get_shape`, `get_voxel_size` and `get_data` throughout the class.

These new methods include a "per file format" behavior based on file extension. For nifti files, the nibabel functions are used as they were before. For PNG/TIF/JPG files, a corresponding behavior is applied. Adding a new modalities of file formats in the future would require to update those 5 functions.

I chose this "minimally invasive" approach because of the complexity of the SegmentationPair class where each nibabel functions is called several times, which I think would require a major refactoring to avoid.
@Drulex, @joshuacwnewton, @ahill187, @gisellemartel, @dyt811 Let us know if you have advice for us on how to approach this refactoring. Thanks!

## To test the PR:
- Any job with nifti files should work as usual
- To test with PNG, use `config_new_loader.json` and the `ivadomed/data_example_microscopy_sem` dataset.

## Linked issues
Fixes #733

### DIFF
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -204,16 +204,17 @@ class SegmentationPair(object):
         self.input_handle = []
 
         # Ordered list of supported file extensions
-        ext_lst = [".nii", ".nii.gz", ".tif", ".tiff", ".png", ".jpg", ".jpeg"]
-
-        # TODO: add the following items between ".nii" and ".tif" in ext_lst when implementing OMETIFF support (#739)
+        # TODO: Implement support of the following OMETIFF formats (#739):
         # [".ome.tif", ".ome.tiff", ".ome.tf2", ".ome.tf8", ".ome.btf"]
+        # They are included in the list to avoid a ".ome.tif" or ".ome.tiff" following the ".tif" or ".tiff" pipeline
+        ext_lst = [".nii", ".nii.gz", ".ome.tif", ".ome.tiff", ".ome.tf2", ".ome.tf8", ".ome.btf", ".tif",
+                   ".tiff", ".png", ".jpg", ".jpeg"]
 
         # Returns the first match from the list
         self.extension = next((ext for ext in ext_lst if input_filenames[0].endswith(ext)), None)
         # TODO: remove "ome" from condition when implementing OMETIFF support (#739)
         if (not self.extension) or ("ome" in self.extension):
-            raise RuntimeError("Input file type of '{}' not supported".format(input_filenames[0]))
+            raise RuntimeError("The input file type of '{}' is not supported".format(input_filenames[0]))
 
         # loop over the filenames (list)
         for input_file in self.input_filenames:
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -200,9 +200,19 @@ class SegmentationPair(object):
         self.slice_axis = slice_axis
         self.soft_gt = soft_gt
         self.prepro_transforms = prepro_transforms
-        self.extension = input_filenames[0].split('.', 1)[1]
         # list of the images
         self.input_handle = []
+
+        # Ordered list of supported file extensions
+        ext_lst = [".nii", ".nii.gz", ".tif", ".tiff", ".png", ".jpg", ".jpeg"]
+
+        # TODO: add the following items between ".nii" and ".tif" in ext_lst when implementing OME-TIFF support:
+        # [".ome.tif", ".ome.tiff", ".ome.tf2", ".ome.tf8", ".ome.btf"]
+
+        # Returns the first match from the list
+        self.extension = next((ext for ext in ext_lst if input_filenames[0].endswith(ext)), None)
+        if not self.extension:
+            raise RuntimeError("Input file type of '{}' not supported".format(input_filenames[0]))
 
         # loop over the filenames (list)
         for input_file in self.input_filenames:
@@ -439,9 +449,6 @@ class SegmentationPair(object):
             # Returns data from file as a 3D numpy array
             # Behavior for grayscale png only, behavior TBD for RGB or RBGA png
             return np.expand_dims(imageio.imread(filename, as_gray=True), axis=-1)
-
-        else:
-            raise RuntimeError("Input file type '{}' not supported".format(self.extension))
 
     def apply_canonical(self, data):
         """Apply nibabel as_closest_canonical function to nifti data only.
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -234,16 +234,16 @@ class SegmentationPair(object):
                 raise RuntimeError('Input and ground truth with different dimensions.')
 
         for idx, handle in enumerate(self.input_handle):
-            self.input_handle[idx] = nib.as_closest_canonical(handle)
+            self.input_handle[idx] = self.apply_canonical(handle)
 
         # Labeled data (ie not inference time)
         if self.gt_filenames is not None:
             for idx, gt in enumerate(self.gt_handle):
                 if gt is not None:
                     if not isinstance(gt, list):  # this tissue has annotation from only one rater
-                        self.gt_handle[idx] = nib.as_closest_canonical(gt)
+                        self.gt_handle[idx] = self.apply_canonical(gt)
                     else:  # this tissue has annotation from several raters
-                        self.gt_handle[idx] = [nib.as_closest_canonical(gt_rater) for gt_rater in gt]
+                        self.gt_handle[idx] = [self.apply_canonical(gt_rater) for gt_rater in gt]
 
         # If binary classification, then extract labels from GT mask
 
@@ -442,6 +442,19 @@ class SegmentationPair(object):
             return np.expand_dims(imageio.imread(filename, as_gray=True), axis=-1)
         else:
             raise RuntimeError('Input file type not supported')
+
+    def apply_canonical(self, data):
+        """Apply nibabel as_closest_canonical function to nifti data only.
+        Args:
+            data ('nibabel.nifti1.Nifti1Image' object or 'ndarray'):
+                for nifti or png file respectively.
+        """
+        if "nii" in self.extension:
+            # Returns 'nibabel.nifti1.Nifti1Image' object
+            return nib.as_closest_canonical(data)
+        if self.extension == "png":
+            # Returns data as is in numpy array
+            return data
 
 
 class MRI2DSegmentationDataset(Dataset):
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -6,6 +6,7 @@ import numpy as np
 import torch
 import pandas as pd
 import os
+import imageio
 from torch.utils.data import Dataset
 from tqdm import tqdm
 
@@ -204,7 +205,7 @@ class SegmentationPair(object):
 
         # loop over the filenames (list)
         for input_file in self.input_filenames:
-            input_img = nib.load(input_file)
+            input_img = self.read_file(input_file)
             self.input_handle.append(input_img)
             if len(input_img.shape) > 3:
                 raise RuntimeError("4-dimensional volumes not supported.")
@@ -219,9 +220,9 @@ class SegmentationPair(object):
             for gt in self.gt_filenames:
                 if gt is not None:
                     if isinstance(gt, str):  # this tissue has annotation from only one rater
-                        self.gt_handle.append(nib.load(gt))
+                        self.gt_handle.append(self.read_file(gt))
                     else:  # this tissue has annotation from several raters
-                        self.gt_handle.append([nib.load(gt_rater) for gt_rater in gt])
+                        self.gt_handle.append([self.read_file(gt_rater) for gt_rater in gt])
                 else:
                     self.gt_handle.append(None)
 
@@ -425,6 +426,22 @@ class SegmentationPair(object):
         }
 
         return dreturn
+
+    def read_file(self, filename):
+        """Read file according to file type.
+        Args:
+            filename (str): Subject filename.
+        """
+        if "nii" in self.extension:
+            # Returns 'nibabel.nifti1.Nifti1Image' object
+            # For 'nii' and 'nii.gz' extentions
+            return nib.load(filename)
+        elif self.extension == "png":
+            # Returns data from file as a 3D numpy array
+            # Behavior for grayscale png only, behavior TBD for RGB or RBGA png
+            return np.expand_dims(imageio.imread(filename, as_gray=True), axis=-1)
+        else:
+            raise RuntimeError('Input file type not supported')
 
 
 class MRI2DSegmentationDataset(Dataset):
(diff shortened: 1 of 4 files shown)


ivadomed/loader/loader.py
```diff
@@ -197,12 +199,13 @@ def __init__(self, input_filenames, gt_filenames, metadata=None, slice_axis=2, c
         self.slice_axis = slice_axis
         self.soft_gt = soft_gt
         self.prepro_transforms = prepro_transforms
+        self.extension = input_filenames[0].split('.', 1)[1]
```
charleygros, line 202: Simply curious: why not using this info from the config file?

charleygros (commented):
Looks all good to me! Now testing the code on my pipelines, keep you posted

ivadomed/loader/loader.py
```diff
@@ -424,6 +424,83 @@ def get_pair_slice(self, slice_index, gt_type="segmentation"):
...
+        Args:
+            data ('nibabel.nifti1.Nifti1Image' object or 'ndarray'):
+                for nifti or png file respectively.
+        """
+        if "nii" in self.extension:
+            # Returns 'nibabel.nifti1.Nifti1Image' object
+            return nib.as_closest_canonical(data)
+        if self.extension == "png":
+            # Returns data as is in numpy array
+            return data
```
joshuacwnewton, line 454: Here, we have two `if` statements. What about the `if .. elif .. else raise` pattern like was done in `read_file`?

joshuacwnewton: Also, two `if` statements are used. What about the `if .. elif .. else raise` pattern like was done in `read_file`?

joshuacwnewton, line 454: Instead of checking `self.extension`, it might be more reliable to check the type of the `data` variable, since the behavior is specific to `Nifti1Image` and `ndarray` types.

(This also applies to other functions here.)

joshuacwnewton: Also, two `if` statements are used. What about the `if .. elif .. else raise` pattern like was done in `read_file`?

ivadomed/loader/loader.py
```diff
@@ -424,6 +424,83 @@ def get_pair_slice(self, slice_index, gt_type="segmentation"):
...
+            cache_mode (str): cache mode for nifti files
+        Returns:
+            ndarray: File data.
+        """
+        if "nii" in self.extension:
+            # Load data from file as numpy array
+            return data.get_fdata(cache_mode, dtype=np.float32)
+        if self.extension == "png":
+            # Returns data as is in numpy array
+            return data
```
joshuacwnewton, line 502: I notice we cast to `float32` in the `Nifti1Image` case, but not in the `ndarray` case. Why is this?

ivadomed/loader/loader.py
```diff
@@ -424,6 +424,83 @@ def get_pair_slice(self, slice_index, gt_type="segmentation"):
...
+        if "nii" in self.extension:
+            # Read zooms metadata from nifti file header
+            return data.header.get_zooms()
+        if self.extension == "png":
+            # Voxel size for PNG is extracted from PixelSize metadata (from BIDS JSON sidecar)
+            # PixelSize definition in example dataset is a scalar in micrometers (BIDS BEP031 v 0.0.2)
+            # PixelSize definition will change, 2D (XY) or 3D (XYZ) array in micrometers (BIDS BEP031 v 0.0.3)
+            # Behavior will have to be ajusted here to follow BEP development
+            pixel_size_in_mm = self.metadata[0]['PixelSize'] / 1000
+            return (pixel_size_in_mm, pixel_size_in_mm, 0)
```
joshuacwnewton, line 486: This worries me a little, because we rely on `PixelSize` to be defined the JSON sidecar. Is there a possibility that this info could ever be missing? 

If so, then perhaps we could provide a more informative error message.

ivadomed/loader/loader.py
joshuacwnewton: > Simply curious: why not using this info from the config file?

Alternatively, we could just not store the extension at all. 

* `read_file` could parse the `filename` string.
* The other functions could check the type of the `data` variable. (`Nifti1Image` or `ndarray`)

joshuacwnewton (commented):
> @Drulex, @joshuacwnewton, @ahill187, @gisellemartel, @dyt811 Let us know if you have advice for us on how to approach this refactoring. Thanks!

One thing to keep in mind with the "minimally invasive" approach is that it adds ambiguity to the existing variables. For example:

```python
for input_file in self.input_filenames:
    input_img = self.read_file(input_file)  # Q: What is `input_img`? ndarray or a nibabel.Nifti1Image? 
    self.input_handle.append(input_img)
    if len(input_img.shape) > 3:  # Q: Is `input_img` guaranteed to have `shape`? Hard to say by looking at it.
                                  # ANS: They both have `shape`. But what about other attributes/methods?
        raise RuntimeError("4-dimensional volumes not supported.")   
```

This might be an acceptable cost considering the refactoring effort this saves, though.

At the very least, we should make sure to update the docstrings and etc., too. For example:

https://github.com/ivadomed/ivadomed/blob/d578aead456c4c572b909ae5ae650c6c586afc55/ivadomed/loader/loader.py#L187-L188

Not just NifTi images anymore. :sweat_smile:

ivadomed/loader/loader.py
joshuacwnewton: Side question: Does the BIDS JSON sidecar have a field for PixelSize units (e.g. `um` for micrometers)? That would make our life a lot easier, hehe.

ivadomed/loader/loader.py
mariehbourget: This is a very good question!
There are 3 places to fetch the extension:

1- From the config file:
As implemented right now, the extension param accepts multiple extensions as a list of string and not just single one. With multiple formats, we won't be able to tell which file has which extension. Also, the param is optional. If it is empty or not present, all files are indexed.

2 - From the metadata argument: 
That seems to be the best solution. It comes from the BIDSDataframe. However, it would not work for `segment_volume` in `inference.py`. This is use by `sct_deepseg` (if I'm not mistaken) and bypass the BIDS part of the loader entirely.

3 - From the filename:
That is what I implemented at the moment. I realize now, that it may not always work as intended with `segment_volume` if there are other dots in the filenames.

There may be a way to recognize the file type by its content without relying on the filename?

ivadomed/loader/loader.py
mariehbourget: I kept the condition on the file extension for now. Because we use BIDS convention, the accepted file formats are limited. For microscopy, we will also want a loader for OMETIFF (not implemented yet), and the behavior with ndarray may differ in some cases (ex: voxel size could be extract from file metadata instead of bids metadata).

I corrected the `if/elif` statements but I did not include an `else` in the functions because `read_file` is the first to be called with the same `if/elif` conditions and would raise the error. Is it bad practice? It would be necessary to add an `else` if we change de `if/elif` condition though.

ivadomed/loader/loader.py
mariehbourget: As it is defined now, the `PixelSize` in BIDS JSON sidecar is always in `micrometers`, so I just converted to `mm`. However,  microscopy is not integrated in the BIDS specification yet. We are working on an BIDS extension proposal (BEP) in parallel, so microscopy data/metadata may change.

I updated the function to add an error message and support PixelSize as a float or as a list.

mariehbourget: As defined in BIDS right now 'PixelSize' is always in `micrometer `so I just converted to `mm`. However, microscopy is not integrated in the BIDS specification yet. We are working in parallel on a BIDS extension proposal (BEP) for it and microscopy data/metadata may change.

I added an error message if the field is not present and support for `PixelSize` as a float or as a list.

mariehbourget:
> At the very least, we should make sure to update the docstrings and etc.,

Doctrings are updated in 8693214 along with documentation in 3a72f09.

Thank you @joshuacwnewton for your help!

ivadomed/loader/loader.py
joshuacwnewton: > For microscopy, we will also want a loader for OMETIFF (not implemented yet), and the behavior with ndarray may differ in some cases (ex: voxel size could be extract from file metadata instead of bids metadata).

Thank you for the clarification. Apologies for making that suggestion without having the full context. :sheep: 

> I corrected the if/elif statements (in d9124b2) but I did not include an else in the functions because read_file is the first to be called with the same if/elif conditions and would raise the error. Is it bad practice?

Good point. I was thinking "what if `read_file` isn't called first?" but I guess there's not much of a risk of that happening, is there...

Still, perhaps we could raise that error in `__init__` when `self.extension` is first set. That way we guarantee the error is thrown before anything else happens (when `loader` is first created).

ivadomed/loader/loader.py
joshuacwnewton: Thank you for the clarification. :slightly_smiling_face:

ivadomed/loader/loader.py
```diff
@@ -424,14 +425,112 @@ def get_pair_slice(self, slice_index, gt_type="segmentation"):
...
+        elif self.extension == "png":
+            # Voxel size for PNG is extracted from PixelSize metadata (from BIDS JSON sidecar)
+            # PixelSize definition in example dataset is a scalar in micrometers (BIDS BEP031 v 0.0.2)
+            # PixelSize definition may change for 2D [X, Y] and 3D [X, Y, Z] arrays in micrometers (BIDS BEP031 v 0.0.3)
+            # This method supports both behaviors.
+            # TODO: Update behavior to follow BEP microscopy development
+            # TODO: Review beahvior when implementing inference with PNG files (no BIDS metadata)
+            if 'PixelSize' in self.metadata[0]:
+                ps_in_um = self.metadata[0]['PixelSize']
+                if isinstance(ps_in_um, list) and (len(ps_in_um) == 2 or len(ps_in_um) == 3):
```
gisellemartel, line 495: For readability I would suggest to store numerical values in a variable. It would help the maintainer understand why we are checking for in this if statement

ivadomed/loader/loader.py
```diff
@@ -424,14 +425,112 @@ def get_pair_slice(self, slice_index, gt_type="segmentation"):
...
+                    ps_in_um = np.asarray(ps_in_um)
+                    ps_in_um.resize(3)
+                elif isinstance(ps_in_um, float):
+                    ps_in_um = np.asarray([ps_in_um, ps_in_um, 0])
+                else:
+                    raise RuntimeError("'PixelSize' metadata type is not supported. Format must be 2D [X, Y] list, "
+                                       "3D [X, Y, Z] list or float.")
+                ps_in_mm = tuple(ps_in_um / 1000)
```
gisellemartel, line 503: Same concern as comment above

ivadomed/loader/loader.py
```diff
@@ -424,14 +425,112 @@ def get_pair_slice(self, slice_index, gt_type="segmentation"):
...
+            # Read zooms metadata from nifti file header
+            return data.header.get_zooms()
+
+        elif self.extension == "png":
+            # Voxel size for PNG is extracted from PixelSize metadata (from BIDS JSON sidecar)
+            # PixelSize definition in example dataset is a scalar in micrometers (BIDS BEP031 v 0.0.2)
+            # PixelSize definition may change for 2D [X, Y] and 3D [X, Y, Z] arrays in micrometers (BIDS BEP031 v 0.0.3)
+            # This method supports both behaviors.
+            # TODO: Update behavior to follow BEP microscopy development
+            # TODO: Review beahvior when implementing inference with PNG files (no BIDS metadata)
```
gisellemartel, line 492: Are these TODOs documented in an issue? If so I would suggest maybe mentioning the issue number here

ivadomed/loader/loader.py
jcohenadad: @mariehbourget to follow-up on our meeting today, this is how we check file type in SCT, with additional checks based on the header (eg: we had bad surprises in the past, where files were NIfTI, but not images):

https://github.com/neuropoly/spinalcordtoolbox/blob/af2d4de531f66cd062af6fefa87c3f84185875ff/spinalcordtoolbox/scripts/sct_apply_transfo.py#L164-L169

ivadomed/loader/loader.py
mariehbourget: Fixed in 253566b

ivadomed/loader/loader.py
mariehbourget: > Still, perhaps we could raise that error in `__init__` when `self.extension` is first set. That way we guarantee the error is thrown before anything else happens (i.e. when `loader` is first created).

Fixed in 66b2e90

ivadomed/loader/loader.py
mariehbourget: Fixed in 66b2e90

ivadomed/loader/loader.py
mariehbourget: Fixed in dfcdee2

ivadomed/loader/loader.py
mariehbourget: Fixed in dfcdee2

charleygros:
> Now testing the code on my pipelines, keep you posted

As said during last meeting: It worked well (just forgot to message you, oopsie)

Let me know if you would like me to re-run tests after your recent changes. Cheers

mariehbourget:
> Let me know if you would like me to re-run tests after your recent changes. Cheers

@charleygros, the PR is ready for re-review.
I've made corrections based on previous comments and sorted out the "extensions" issue.
I also added a test for the loading of 2d PNG data.

Note: the loader now accept PNG, TIF and JPG 2d files for microscopy because we will train with BIDS compliant datasets in PNG but we want to be able to run segmentation from ADS with other formats as well.

Any job with nifti files should work as usual.
For PNG data, only "training" will work at the moment. This is because "testing" and "segment" functions rely heavily on nifti files.  I added comments and issue numbers in the code about that and I will make the necessary changes in future PRs.

charleygros:
> @charleygros, the PR is ready for re-review.

I tested on few scenarii of nifiti trainings and... all as expected 🥇 !!

charleygros (approved):

PR #753, opened 2021-04-06
Autopublishing improvements

kousu:
Via https://github.com/neuropoly/spinalcordtoolbox/issues/1526

- add a manual trigger from the Actions tab; it's behaves the same as making a new release, it's just a bit nicer to have both options
- run on *publishing* not on *creating* a release, to avoid glitches with semi-published packages
- copy the sdist and .whl to Github Releases, as an alternate mirror
- use pypa's twine wrapper Action (maybe this is silly; this might actually slow *down* the release slightly; but also in theory pypa are the ones who know best)
- use `build` because it's faster and better maintained than bdist_wheel: https://stackoverflow.com/a/58756491

#### GitHub

#### PR contents


dyt811:
Hey, @kousu  

Just took a look at the code base. I think everything looks solid. The questions I have are mostly higher level:

1. This only builds and release for 3.6 on Ubuntu right? Should we setup multi-Python version and multi OS build? Would that require matrix build? 
2. Should we also update setuptools/wheel after ensuring PIP is up to date with something like
``` python -m pip install --upgrade setuptools wheel ```
3. I know this will probably require its separate PRs but what are your thoughts about Conda packaing process? Should we do that as well as the PyPI process as well? Do you have any additional examples from SCT or others about it?
4. I am not sure if there is a way to test this process, short of making a release? 

My only recommendation code-wise is perhaps adding line breaks or additional comments to break up the steps further for ease of readability

**Review by dyt811 — APPROVED** [review]
No major issues observed. Minor suggestion for optional code visibility improvement.

File: .github/workflows/publish_package.yml [hunk 1]
```diff
@@ -1,11 +1,18 @@
...
+# This workflows will upload a Python Package
 # For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
 
 name: Publish Package
 
 on:
+  # publish from the Releases page:
   release:
-    types: [created]
+    types: [published]
```
jcohenadad, line 9: out of curiosity, concretely speaking, when will this action be triggered? should the [release workflow](https://github.com/ivadomed/ivadomed/wiki/Releases#checklist-for-creating-a-new-release) be updated if there is a change in behaviour?

kousu: It looks like those instructions are either implying or missing "Click Publish".

kousu:
> Hey, @kousu
> 
> Just took a look at the code base. I think everything looks solid. The questions I have are mostly higher level:
> 
>     1. This only builds and release for 3.6 on Ubuntu right? Should we setup multi-Python version and multi OS build? Would that require matrix build?

It looks like we've been publishing platform-independent packages ("universal wheels"), e.g. notice the '-py3-none-any' in https://pypi.org/project/ivadomed/2.0.2/#files

A bunch of ivadomed's dependencies need compiling and platform-specificity but not ivadomed itself.

>     2. Should we also update setuptools/wheel after ensuring PIP is up to date with something like
>        `python -m pip install --upgrade setuptools wheel`

Well it does

```
pip install build wheel
```

`build` is the hot-off-the-presses python packager, so I think that should imply 'latest build' which should imply the latest setuptools and wheel. Probably.

If we're doing something that really needs the absolute latest wheel we're probably doing something fragile.

>     3. I know this will probably require its separate PRs but what are your thoughts about Conda packaing process? Should we do that as well as the PyPI process as well? Do you have any additional examples from SCT or others about it?

I don't really know yet. But I think we can probably wrap pip into conda, so getting the pip releases really solid will make that easy.

I just looked up how scikit-learn is packaged into conda. This looks like the key lines:

https://github.com/conda-forge/scikit-learn-feedstock/blob/65262c6f8c0a0c066b921c23b34deb35dbc2bf1e/recipe/meta.yaml#L19-L20

So, basically, yes, they just fall back on `pip`.

We can use the same strategy for other distros, e.g. https://pypi.org/project/pip2nix/ for  https://github.com/NixOS/nixpkgs/

>     4. I am not sure if there is a way to test this process, short of making a release?
> 

The best we can do is:

- use testpypi -- which requires temporarily editing
- append release candidate numbers ('rc1' or 'rc2' or 'rc3') to the tag you use

Pypi (and testpypi!) don't allow releases to ever be erased. You can take them down but the tag can never be reused, so we should get in the habit of using 

I thought it would be cool if we could make a flag in the manual trigger section that chooses between PyPI and TestPyPI, but I didn't get around to it.

.github/workflows/publish_package.yml
jcohenadad: ahhhh! ok i got it now. Yes, this change totally makes sense.

kousu: I've updated the wiki

kousu:
@dyt811 was right about the whitespace, so I fixed that. Ready to merge?

dyt811:
Looking great! Thank you for all the in-depth explanation. Much learning. :D.  I just update merged the master to your branch. I think everything looks good to me. (Although I am very green when it comes to pypi releases).

PR #755, opened 2021-04-12
Addressing Windows Path issue causing unit test_automated_training.py unit tests to fail

dyt811:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [X] I've assigned a reviewer

#### PR contents

- [X] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [X] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

TLDR: fixed Windows pathing related bug that causes prediction files not to be generated at the correct location (stuck in the old tmp location)

Identified the root cause of #604 for unit test output issues which prevented adding Windows CI to GitHub Action (that is coming soon). 
- Applied Pathlib approach to avoid manual str parsing. 
- This will no longer result in malformed file path on Windows
- _**Prior to this fix, no Windows users can generate test prediction in the proper location**_ (it will be left at the old source input folder... surprising we have not heard a single complaint from end users about this yet. 

## How to test this PR:
- Be on windows. 
- Pull down this PR branch and ensure all other Python ENV is setup. 
- Making SURE to `pip uninstall ivadomed` and then reinstall it via ` pip install -e .`
- Ensuring the test files have been downloaded
- Run pytest from repo root folder. 

## Linked issues

Partially resolve #604 by unlocking GitHub Actions to be setup against Windows Environment as well going forward for all PRs and future automated CIs, first step towards ~world domination~ Windows 10 cross compatibility

Issue #633, opened 2021-01-12
Issue with test_adaptative.py


ahill187:
### Current behavior
I have noticed when running tests locally, `test_adaptative.py` always fails, while the tests on GitHub are running. @gisellemartel I believe you had this issue as well?

When the tests are run in `GitHub`, they seem to be running fine. I think the issue might be to do with `hdf5` files and read/write issues. 

### Steps to reproduce
```
pytest testing/unit_tests/test_adaptative.py
```

<details>
<summary> An example summary message </summary>

```
======================================= FAILURES =======================================
______________________________________ test_hdf5 _______________________________________

    def test_hdf5():
        print('[INFO]: Starting test ... \n')
        train_lst = ['sub-unf01']
    
        training_transform_dict = {
            "Resample":
                {
                    "wspace": 0.75,
                    "hspace": 0.75
                },
            "CenterCrop":
                {
                    "size": [48, 48]
                },
            "NumpyToTensor": {}
        }
        transform_lst, _ = imed_transforms.prepare_transforms(training_transform_dict)
    
        roi_params = {"suffix": "_seg-manual", "slice_filter_roi": None}
    
        bids_to_hdf5 = imed_adaptative.BIDStoHDF5(PATH_BIDS,
                                                    subject_lst=train_lst,
                                                    path_hdf5='testing_data/mytestfile.hdf5',
                                                    target_suffix=["_lesion-manual"],
                                                    roi_params=roi_params,
                                                    contrast_lst=['T1w', 'T2w', 'T2star'],
                                                    metadata_choice="contrast",
                                                    transform=transform_lst,
                                                    contrast_balance={},
                                                    slice_axis=2,
                                                    slice_filter_fn=imed_loader_utils.SliceFilter(filter_empty_input=True,
                                                                                        filter_empty_mask=True))
    
        # Checking architecture
        def print_attrs(name, obj):
            print("\nName of the object: {}".format(name))
            print("Type: {}".format(type(obj)))
            print("Including the following attributes:")
            for key, val in obj.attrs.items():
                print("    %s: %s" % (key, val))
    
        print('\n[INFO]: HDF5 architecture:')
        with h5py.File(bids_to_hdf5.path_hdf5, "a") as hdf5_file:
            hdf5_file.visititems(print_attrs)
            print('\n[INFO]: HDF5 file successfully generated.')
            print('[INFO]: Generating dataframe ...\n')
    
[...]

Issue #604, opened 2021-01-07
Add compatibility for Windows


ahill187:
We would like to add compatibility for `Windows` in `ivadomed`. Currently we are only `Linux` and `Mac` compatible.

### Motivation for the feature
In this review, some of the reviewers mentioned that they had issues trying to install `ivadomed` on `Windows`: https://github.com/openjournals/joss-reviews/issues/2868#issuecomment-752687437

### Description of the feature
To address the issue in the review, we will add to our documentation that we do not currently support Windows. However, it would be nice to support Windows in the future.

dyt811:
So, a quick recap on the potential process to start tackling this as it may become a multi-headed beasts soon if we start unearth more issues. 

Per IvadoMed Meeting agenda discussion item [here](https://docs.google.com/document/d/19mEHKOZKFdprb3UPQYxmfO46mn0bPQfOgE-WPdeF7AQ/edit#heading=h.3ndjkvlk9ktj) at the time, there does not appear to be any underlying libraries that fundamentally prevents the adoption of IvadoMed as library or usable binary on Windows platform. Further extensive and comprehensive validation and testing are required to ensure Windows compatibility and  accompanying concerns. 

So, in light of that, what I propose as simple and rather naïve starting steps to attempt to resolve this: 

1. Examine existing unit tests performance on Windows (as preliminary results already indicating a few unit tests are failing). 
2. Evaluate the cases where they are failing (with GPU or with CPU, as failed test cases appear to VARY). More issues might needs to be unearthed, identified/opened soon based on these.
3. Issue PR fixes and supplement existing unit tests to address above concern. 
4. Documentation update to address limitations or potential issues once complete. 
5. Issue PR to conclude Documentation update. 

More details to be updated.

ssinad:
> So, a quick recap on the potential process to start tackling this as it may become a multi-headed beasts soon if we start unearth more issues.
> 
> Per IvadoMed Meeting agenda discussion item [here](https://docs.google.com/document/d/19mEHKOZKFdprb3UPQYxmfO46mn0bPQfOgE-WPdeF7AQ/edit#heading=h.3ndjkvlk9ktj) at the time, there does not appear to be any underlying libraries that fundamentally prevents the adoption of IvadoMed as library or usable binary on Windows platform. Further extensive and comprehensive validation and testing are required to ensure Windows compatibility and accompanying concerns.
> 
> So, in light of that, what I propose as simple and rather naïve starting steps to attempt to resolve this:
[...]

PR #757, opened 2021-04-15
Improve current GitHub Action CI with multi OS support

### LINKED ISSUE #740: python3.9 incompatibility due to onnxruntime
ivadomed doesn't install on python3.9 because onnxruntime is pinned to an old version that 

a. doesn't provide packages for 3.9: https://pypi.org/project/onnxruntime/1.4.0/#files
b. doesn't even provide an `sdist` for anything

```
(venv) [kousu@requiem ivadomed]$ pip install -e .
Obtaining file:///home/kousu/src/neuropoly/ivadomed
Collecting bids-neuropoly>=0.2
  Using cached bids_neuropoly-0.2-py3-none-any.whl (3.1 kB)
Collecting csv-diff>=1.0
  Using cached csv_diff-1.1-py3-none-any.whl (12 kB)
Collecting h5py==2.10.0
  Using cached h5py-2.10.0.tar.gz (301 kB)
Collecting joblib
  Using cached joblib-1.0.1-py3-none-any.whl (303 kB)
Collecting matplotlib>=3.3.0
  Using cached matplotlib-3.4.0-cp39-cp39-manylinux1_x86_64.whl (10.3 MB)
Collecting nibabel
  Using cached nibabel-3.2.1-py3-none-any.whl (3.3 MB)
ERROR: Could not find a version that satisfies the requirement onnxruntime==1.4.0 (from ivadomed)
ERROR: No matching distribution found for onnxruntime==1.4.0
```

Their latest version https://pypi.org/project/onnxruntime/1.7.0/#files does provide 3.9 packages.

### LINKED ISSUE #778: CI Test gets stuck in an infinite loop
## Issue description
testing/functional_tests/test_automate_training.py does not terminate on macOS and Windows CI environments.

### Current behavior

The test gets stuck.

### Expected behavior
 
The test should terminate within minutes.

### Steps to reproduce

https://github.com/ssinad/ivadomed/actions/runs/814749769

### System description

macOS 10.15, windows-latest

### Installed packages

<details open>
<summary> Output of <code>pip freeze</code> </summary>

```

```

</details>

dyt811:
#### GitHub

#### PR contents

### Why: 
- We currently have no CI for macOS/Windows
- Pip install -r requirement.txt will fail on Windows side because pytorch 1.5.0 does not have wheel on Windows
- We don’t have a Conda script to deploy environment even if we want to. 
- Dev don’t have quick way to deploy Dev environment (that includes pytest etc) via pip or conda 
- User/Dev don’t have quick way to deploy GPU related environment easily. 

### What:

- Current GitHub Action CI does not fully address Windows OR even macOS platform testing due to earlier bug which has been recently resolved. See #604 #755. Now we can properly robustly test Windows integration. I have incorporated many ADS CI tips and tricks learned from there (https://github.com/neuropoly/axondeepseg/pull/479) to improve our existing matrix testing to be more comprehensive and thorough while adding additional comments to make it easier for subsequent future updates. 

### How: 
Applying cascading `requirements.txt` without changing anything major it now allows us to accommodate various deployment scenarios while maintaining single source of truth
- `Requirements_common.txt` shared with `Conda` and `setup.py` cascade to various different higher level requirements. 
- `Conda` now supported via `environment.yml` which refer to the `requirements.txt` to respect SINGLE source of truth
- User deploys `requirements.txt` with CPU support, standard deployment scenario. 
- Advanced user deploys `requirements_gpu.txt` for GPU support. 
- Dev/CI/RTD uses `requirements_dev.txt` which cascades off the main requirements.txt

### How to test 
- Create new venv/conda env and validate using whichever conda/venv/pip installation setup approach you are most comfortable playing with. Report back issues and concerns. 
- `requirements_dev.txt` might require some extra attention as not all sphinx build/wheel are available for Windows? 
- Could also use more feedback on the documentation side

## Linked issues

Resolves #778, also provide CI for its precursor #633 #768


dyt811:
Took a while to realize that macOS does not support the Wheel +CPU tag... sigh. 
image

No wonder it fails on macOS only. FML.

charleygros:
@dyt811 : Thanks for all this! I'd be happy to review, but I'm scared I don't have a relevant skill set for this. Maybe someone else in the team would be more appropriate? Let me know :-)

dyt811:
Closer examination of the output showed that automated_training_process was skipped accidentally. Once properly restored, it suggested that we are still running into issue with Mac/Windows hanging and it is on top of the earlier mentioned problem with HDF5. Root cause still not fully identified...

.github/workflows/run_tests.yml
```diff
@@ -1,62 +1,100 @@
...
 
+      # Step 1: Setup python version
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # Step 2: Install requirement dependencies.
+      - name: Install dependencies
+        uses: py-actions/py-dependency-install@v2
```
ssinad, line 60: @dyt811 Which requirements file does this one install?

environment.yml
```diff
@@ -0,0 +1,10 @@
+name: IvadoMedEnv
```
ssinad, line 1: @dyt811  Could you explain what the purpose of this file is?

setup.py
```diff
@@ -2,7 +2,9 @@
 from codecs import open
 from os import path
 
-with open('requirements.txt') as f:
+# Manually specified, more generic version of the software.
+# See: https://stackoverflow.com/a/49684835
+with open('requirements_common.txt') as f:
     requirements = f.read().splitlines()
```
ssinad, line 8:
```python
requirements = f.readlines()
```
would also work.

ssinad (changes requested):
@dyt811
I would recommend that you get another approval from someone else on the team before you merge.

.github/workflows/run_tests.yml
ssinad: Maybe you could add
```yaml
path: requirements_dev.txt
```

requirements_dev.txt
```diff
@@ -0,0 +1,13 @@
...
+-r requirements.txt
+
+# Dev:
+# Development, documentation, testing related dependencies.
+pytest
+pytest-ordering
+sphinx
+pypandoc
+sphinx_rtd_theme
+sphinx-jsonschema
```
cakester, line 11: no version number? in the context that you just deleted this has a version number https://github.com/ivadomed/ivadomed/pull/757/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552L19

dyt811: Good catch. Yeah, I think this PR started before your PR so this needs to be manually added in this new file. I will added it.

.github/workflows/run_tests.yml
```diff
@@ -1,62 +1,100 @@
...
+
+      # Step 3: install additional pytest/dev related dependencies
+      - name: Install dev dependencies
+        run: |
+          pip install flake8 pytest coverage coveralls pytest-cov pytest-console-scripts pytest-ordering
+          pip install torch==1.5.0+cpu torchvision==0.6.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```
ssinad, line 66: @dyt811 Consider using the previous step instead of hardcoding package versions.

docs/source/installation.rst
```diff
@@ -27,8 +27,40 @@ Write in your .bashrc/.zshrc file with following line.
 
 Save this change and restart your terminal to apply the change.
 
-Python
-------
+There are fundamentally two different approaches to install IvadoMed:
```
jcohenadad, line 30:
```suggestion
There are fundamentally two different approaches to install ``ivadomed``:
```

.github/workflows/run_tests.yml
ssinad, line 60:
```suggestion
        uses: py-actions/py-dependency-install@v2
        path: requirements_dev.txt
```

setup.py
ssinad, line 8:
```suggestion
    requirements = f.readlines()
```

.github/workflows/pre-commit.yml
```diff
@@ -27,13 +36,17 @@ jobs:
           key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
           restore-keys: |
             ${{ runner.os }}-pip-
+
       - name: Install dependencies
         run: |
               python -m pip install --upgrade pip
-              if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+              if [ -f requirements_dev.txt ]; then pip install -r requirements_dev.txt; fi
```
ssinad, line 43:
```suggestion
              if [[ -f requirements_dev.txt ]]; then pip install -r requirements_dev.txt; fi
```
See [here](https://github.com/ivadomed/ivadomed/wiki/variables#if-statements).

.github/workflows/run_tests.yml
dyt811: Good point!

.github/workflows/run_tests.yml
```diff
@@ -1,62 +1,100 @@
...
+    name: Matrix Test of Python ${{ matrix.python-version }} on ${{ matrix.os }}
 
-    runs-on: ubuntu-latest
+    # Matrix driven OS
+    runs-on: ${{ matrix.os }}
+
+    # Default shell for ALL subsequent steps.
+    defaults:
+      run:
+        shell: bash {0}
```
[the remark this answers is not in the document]
dyt811: I believe this [disables fast fail in GitHub CI](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference) which result in more likelihood of revealing bugs in CI. I have actually [copied this bit from the template GitHub Action from ADS](https://github.com/neuropoly/axondeepseg/blob/master/.github/workflows/run_tests.yaml#L44).

environment.yml
dyt811: This environment YML is provided such that conda users can easily instantiate a conda environment while still be able to source the configure file from a single source of truth which is shared across multiple dev/user environment setups such as: 1) CPU Dev, 2) GPU users. 

Conda user can certainly just instantiate their own environment and pip install the requirement as well, just save them a few key strokes.

setup.py
dyt811: TIL. Cool. Definitely didn't know that before. :)

.github/workflows/pre-commit.yml
dyt811: COOL. TIL. Thanks for the info!

ssinad (approved):
Great job. I would still recommend you get another approval before you merge, though.

jcohenadad:
> Great job. I would still recommend you get another approval before you merge, though.

we are a very small team and there are *many* other urgent priorities so unfortunately we need to live with PRs not being reviewed as much as everyone would like--

ssinad:
> > Great job. I would still recommend you get another approval before you merge, though.
> 
> we are a very small team and there are _many_ other urgent priorities so unfortunately we need to live with PRs not being reviewed as much as everyone would like--

I understand.

requirements_dev.txt
```diff
@@ -0,0 +1,18 @@
+# CPU Compatible Requirements
+-r requirements.txt
+
+# Dev:
+# Development, documentation, testing related dependencies.
+pytest
```
ssinad, line 6:
```suggestion
pytest~=6.2
```

requirements_dev.txt
```diff
@@ -0,0 +1,18 @@
...
+pytest-cov
+pytest-console-scripts
+pytest-ordering
```
ssinad, line 9:
```suggestion
pytest-ordering~=0.6
```

requirements_dev.txt
dyt811: Good catch. Yup. Better to version these. Gonna add it in.

requirements_dev.txt
dyt811: Yup. Will add this in as well. I was REALLY hoping that we will get CI to a state where daily CI runs will catch any subversion induced problems but I think there is a whole bunch of tests to enhance before that happens.

.github/workflows/pre-commit.yml
```diff
@@ -8,15 +8,24 @@ on:
       - master
 
 jobs:
+
   pre-commit-checks:
-    runs-on: 'ubuntu-latest'
+    strategy:
+      matrix:
+        os: [ "macos-11", "windows-latest", "ubuntu-18.04", "ubuntu-20.04",  "macos-10.15"]
```
konantian, line 15: We use "macos-11" here is because "macos-latest" is equals to "macos-10.15" instead of "macos-11".

.github/workflows/run_tests.yml
```diff
@@ -1,62 +1,100 @@
...
+
+    # Defining matrix for OS and Python
     strategy:
+
+      # Ensure exhaustive testing of all scenarios to gather maximum amount of information
+      fail-fast: false
+
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        os: [ "macos-11", "windows-latest", "ubuntu-18.04", "ubuntu-20.04",  "macos-10.15" ]
```
konantian, line 42: We use "macos-11" here is because "macos-latest" is equals to "macos-10.15" instead of "macos-11".

konantian (commented):
Looks good to me, I've tested them on both Windows and macOs locally, no issues found.

.github/workflows/run_tests.yml
ssinad: @dyt811 
macOS 11 is not available in GitHub actions yet. That is why the tests are failing. See [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners).

> Note: The macOS 11 virtual environment is currently provided as a private preview only. Any users or organizations that are already using this runner can continue using it, but we're not accepting any further users or organizations at this time. The macos-latest YAML workflow label still uses the macOS 10.15 virtual environment.

.github/workflows/run_tests.yml
dyt811: Yup. Change included.

setup.py
dyt811: Updated.

.github/workflows/run_tests.yml
dyt811: Good point. I thought it was available but didn't read the fine print about it being "private preview only". Gonna revert that and merge PR then.

environment.yml
```diff
@@ -0,0 +1,10 @@
...
+channels:
+  - conda-forge
+  - pytorch
+dependencies:
+  - python>=3.6, < 3.9
+  - anaconda
```
kousu, line 7: @dyt811 can you explain why you depend on `anaconda` explicitly?

This doesn't seem to be compatible with https://github.com/conda-forge/miniforge, which is what I picked to use in https://github.com/neuropoly/computers/pull/120/ on the theory that it would be smaller and better supported.

Removing this one line seems to make ivadomed fully compatible with miniforge.

dyt811: Yup. No need to use anaconda. Miniconda is good enough. 👍

Issue #3, opened 2019-03-01
Optimize data preprocessing before training


jcohenadad:
A couple of preprocessing could be done to improve performance of the network. Including:
- Bias field correction

dyt811:
@jcohenadad Is N4 good enough for this or do you need more flexibility/options etc?

jcohenadad:
N4 is ok in general, but for for small FOV and highly heterogeneous tissues it is often not appropriate. But it's a good start. And AntsPy might offer it now...?

dyt811:
Okay, Yup, [various forms of N3 and N4](https://github.com/ANTsX/ANTsPy/blob/9c2a3624f23f47297c8edb1ab3909519a8d42fe4/ants/utils/bias_correction.py).  I will see how best to incorporate this as a start to play with preprocessing and if more interesting methods come along. I will build it general enough to allow flexibility of various bias correction approaches with clearly defined interfaces. 

Functional Requirement added:
- Must be able to accommodate alternative bias field correction algorithms

PR #761, opened 2021-04-15
Add n4 correction script

cakester:
#### GitHub

#### PR contents

## Linked issues

resolve https://github.com/ivadomed/ivadomed/issues/3


cakester:
note the piece of code uses the library older than mentioned in https://github.com/ivadomed/ivadomed/issues/3, so the syntax is a little bit different

dyt811:
@cakester could you add relevant unit test call (i.e. low level non cli call to the basic Python function) using data from data_testing, specifically these ones, but for all subjects: 
image

requirements.txt
```diff
@@ -19,3 +19,4 @@ pytest-ordering
 sphinx-jsonschema
 pytest-console-scripts
 pre-commit==2.10.1
+SimpleITK
```
dyt811, line 22: To be discussed in the upcoming IvadoMed meetings to ensure we are okay adding the additional dependencies..

setup.py
```diff
@@ -52,7 +52,8 @@
             'ivadomed_convert_to_onnx=ivadomed.scripts.convert_to_onnx:main',
             'ivadomed_extract_small_dataset=ivadomed.scripts.extract_small_dataset:main',
             'ivadomed_download_data=ivadomed.scripts.download_data:main',
-            'ivadomed_training_curve=ivadomed.scripts.training_curve:main'
+            'ivadomed_training_curve=ivadomed.scripts.training_curve:main',
+            'ivadomed_n4_correction=ivadomed.scripts.n4_correction:main'
```
dyt811, line 56: Again, to be discussed a the meeting to ensure 1) whether we want to generalize the magnetic field inhomogeneity correction beyond N4, 2) we okay with adding additional CLI command like this for magnetic field inhomogeneity correction.

dyt811 (changes requested):
Some changes to be discussed within the wider IvadoMed dev group before proceeding.

@cakester could you add relevant unit test call (i.e. low level non cli call to the basic Python function) using data from data_testing, specifically these ones, but for all subjects: 
image

If the group okays the CLI approach, we will then add the functional script_caller test as well.

setup.py
jcohenadad: can you pls add it to next meeting agenda?

cakester:
> Some changes to be discussed within the wider IvadoMed dev group before proceeding.
> 
> @cakester could you add relevant unit test call (i.e. low level non cli call to the basic Python function) using data from data_testing, specifically these ones, but for all subjects:
> image
> 
> If the group okays the CLI approach, we will then add the functional script_caller test as well.

what else do I have to check in the unit test call except to check if the output file is there?

charleygros:
> what else do I have to check in the unit test call except to check if the output file is there?

Somehow to check whether the intensity profile has less variance along a given axis than prior to N4?

jcohenadad:
> > what else do I have to check in the unit test call except to check if the output file is there?
> 
> Somehow to check whether the intensity profile has less variance along a given axis than prior to N4?

yes, that or comparing the output with a "ground truth" image, which would go through N4 and be visually checked, and be located in the testing data. Just a suggestion

dyt811:
I brought up the possibility to maybe MD5 hash check against a properly corrected N4 output during our last meeting. That can be a viable approach to unit test as well. 

N4 process should be deterministic and barring any underlying changes in the dependency/calculation level, the bit output should be identical (to be 100% validated using methods below). 

Might best to run a local quick check (maybe after locking down SimpleITK version), i.e. 

Platform/Environment/OS Combo 1:
- Data A -> N4 Call -> N4 Corrected Data A -> MD5 Hash 

Platform/NEW Reinstalled VENV/CONDA Environment/OS Combo 2:
- Data A -> N4 Call -> N4 Corrected Data A -> MD5 Hash

NEW VM/Docker/Platform/VENV/CONDA Environment/OS Combo 3:
- Data A -> N4 Call -> N4 Corrected Data A -> MD5 Hash

With Data A being one of the test imaging data shown from the picture (those nii.gz files, can start with just one before doing more).  

In theory, MD5 should be the same... Please let us know if you are able to experimentally confirm that, ideally across more OS. Try to keep ITK version, python etc, as close as you can.

cakester:
im trying to generate md5 for the n4 corrected files but the test fails and the md5 hash is different from ci environment to my own environment and I downloaded linux (and made sure I get the same simpleitk version) and got another set of hashes......md5 should be the same for all platforms in theory, which means simpleitk generates different files across different platforms? or maybe there is multiple release of the library for the same version? redirecting to @dyt811

PR #763, opened 2021-04-20
New loader: Generate 2D patches

### LINKED ISSUE #747: Generate 2D patches for microscopy loader
## Context

This issue is part of the loader refactoring project to include microscopy data (#304).
The first 3 steps of the project were integrated in the current pipeline in PR #699.
Support to load microscopy 2D PNG files is done in PR #734.

## To do in this issue

For microscopy, we will need a feature to do 2D patches of the images, similar to [_prepare_indices](https://github.com/ivadomed/ivadomed/blob/b977a20375c2e00bafb94b9a1bebe6bad8e2a44a/ivadomed/loader/loader.py#L659) in MRI3DSubVolumeSegmentationDataset class, which is not implemented in 2D.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR adds a feature to `MRI2DSegmentationDataset` to use 2D patches for training.
I used the same approach as in the MRI3DSubVolumeSegmentationDataset class with `handlers` and `prepare_indices`. In the 2D case, the class can accomodate a training with or without patches. I also added 2 optional parameters to the `default_model` in the config file (`length_2D` and `stride_2D`).

When run with the `config_new_loader.json` file and the `data_example_microscopy_sem` dataset, it should load 28 slices for validation and 59 slices for training.

@andreanne-lemay, @charleygros, it would help me to have your feedback on this draft. Particularly on "how" the patch feature is integrated within the `MRI2DSegmentationDataset` and the 2 new params in `default_model`. Does that make sense to you or would you rather have another model and another class only for "sub-images"? Thanks!

## Linked issues
Fixes #747


docs/source/configuration_file.rst
```diff
@@ -981,6 +981,20 @@ being used for the segmentation task).
                    "Final activation layer. Options: ``sigmoid`` (default), ``relu``(normalized ReLU), or ``softmax``."
                ]
            },
+           "length_2D": {
```
charleygros, line 984: I'm opening the question of "How to define the default patch size, ie no patch creation, ie full slice without patching".

For now, from what I understand, if these new params are in the config file, then we do patching, otherwise no patching.

Question: Should we have these parameters in the `default_model`? I was going to say no, and recommend a "modified dict" for this "patchy unet".. but maybe, instead, we could have these params in the default model BUT have default values for "no patching" eg `"length_2D": []` (or similar).

ivadomed/loader/loader.py
```diff
@@ -564,6 +565,8 @@ class MRI2DSegmentationDataset(Dataset):
     Args:
         filename_pairs (list): a list of tuples in the format (input filename list containing all modalities,ground \
             truth filename, ROI filename, metadata).
+        length (tuple): Size of each dimensions of the subimages, length equals 2.
```
charleygros, line 568:
```suggestion
        length (list): Size of each dimensions of the subimages, length equals 2.
```

ivadomed/loader/loader.py
```diff
@@ -661,7 +706,13 @@ def __getitem__(self, index):
         Args:
             index (int): Slice index.
         """
-        seg_pair_slice, roi_pair_slice = self.indexes[index]
+
+        # If self.length in model_params, extract pair slices subimages from handlers
+        if self.length:
```
charleygros, line 711: Related to the question regarding the "default values" for no patching.
Also, here we are testing length, but what happen in no striding in model_params? to be discussed?

charleygros:
> Does that make sense to you or would you rather have another model and another class only for "sub-images"? Thanks!

Makes sense to me, I like that you used the same approach as in the MRI3DSubVolumeSegmentationDataset.
Only thing is how / when is triggered the patching --> cf above messages. Would be good to have a chat about it.

ivadomed/loader/loader.py
andreanne-lemay: I think we should support both (no param or length_2d=[] as Charley suggested). So maybe add a condition and len(self.length)

andreanne-lemay:
> > Does that make sense to you or would you rather have another model and another class only for "sub-images"? Thanks!
> 
> Makes sense to me, I like that you used the same approach as in the MRI3DSubVolumeSegmentationDataset.
> Only thing is how / when is triggered the patching --> cf above messages. Would be good to have a chat about it.

I totally agree with Charley! Your current approach seems the most logical to me. Nothing to flag on my side!

docs/source/configuration_file.rst
mariehbourget: > For now, from what I understand (new L1211), if these new params are in the config file, then we do patching, otherwise no patching.

Yes, that's correct, `length_2D` and `stride_2D` are optional in the default_model params, and `length_2D` triggers the patching.

I am not against creating a modified dict for "patchy unet" if we want to keep the "default_model" as simple as possible, or having `"length_2D": []` and `"stride_2D": []` as default values for "no patching".

ivadomed/loader/loader.py
mariehbourget: I think `if self.length` will return false for both `None` and `[]`. However, I use the condition a few times, maybe I should put it in a bool variable ex: `is_2d_patch`, at the beginning.

For testing the dimensions and values of length and stride, I was planning on doing it in `prepare_indices` (similar to the 3D version). If I use a bool variable, maybe I should test it at the same time.

docs/source/configuration_file.rst
charleygros: I'm leaning towards this option I reckon:
```
 having "length_2D": [] and "stride_2D": [] as default values for "no patching"
```

ivadomed/loader/loader.py
charleygros: I like this:
```
put it in a bool variable ex: is_2d_patch, at the beginning
```

mariehbourget:
This PR is ready for review.

I changed the default for `length_2D` and `stride_2D` from `None` to `[]`.
I added a `is_2d_patch` attribute to `MRI2DSegmentationDataset` and verification of stride, length and shape in `prepare_indices`.
I added a test.

Note: the 2d patching works only for the training right now, I will implement the reconstruction of the image and testing/inference process separately.

docs/source/configuration_file.rst
```diff
@@ -981,10 +981,24 @@ being used for the segmentation task).
...
+                "type": "[int, int]",
+                "description": "(Optional) Size of the 2D patches used as model's input tensors.",
+                "required": "false"
+            },
+            "stride_2D": {
+                "type": "[int, int]",
+                "$$description": [
+                    "(Optional) Pixels' shift over the input matrix to create 2D patches. Ex: Stride of [1, 2]\n",
```
charleygros, line 992: Should we mention "strictly positive integer"? I have not thought much when inputting `[0, 0]` (which is stupid I know) --> gave me an error.

charleygros:
@mariehbourget: I tried to do a patching with no stride, ie:
```
        "length_2D": [50, 50],
        "stride_2D": []
```

I have the following error:
```
  File "/home/GRAMES.POLYMTL.CA/chgroc/venv-ivadomed/bin/ivadomed", line 33, in <module>
    sys.exit(load_entry_point('ivadomed', 'console_scripts', 'ivadomed')())
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/main.py", line 551, in run_main
    resume_training=bool(args.resume_training))
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/main.py", line 369, in run_command
    'validation')
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/main.py", line 120, in get_dataset
    cuda_available=cuda_available)
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/loader/loader.py", line 115, in load_dataset
    dataset.load_filenames()
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/loader/loader.py", line 675, in load_filenames
    self.prepare_indices()
  File "/home/GRAMES.POLYMTL.CA/chgroc/ivadomed/ivadomed/loader/loader.py", line 685, in prepare_indices
    raise RuntimeError('"length_2D" and "stride_2D" must be of length 2.')
RuntimeError: "length_2D" and "stride_2D" must be of length 2.
```

Would it be possible to: when `length_2D` is specified and `stride_2D` is not, to enforce `stride_2D` = `length_2D` or something like this?

charleygros:
When using:
```
        "length_2D": [50, 50],
        "stride_2D": [5, 5]
```

The patching works well! Yeeeehaa

Small comment: Could we adapt the terminal output, from:
```
Loaded 79764 axial slices for the training set.
```
to something like:
```
Loaded 79764 axial patches for the training set.
```

charleygros:
I did multiple tests and everything else worked beautifully 🏅 !

docs/source/configuration_file.rst
mariehbourget: Fixed in e42d9dc and e4dacd5.
I'm wondering if I should add length_2D and stride_2D in the doc in this PR or if I should wait for the inference part to be dealt with. With only this PR, the length and stride works only for the training part but the reconstruction of the image will not work yet. What do you think?

mariehbourget:
> Would it be possible to: when `length_2D` is specified and `stride_2D` is not, to enforce `stride_2D` = `length_2D`? or something like this?

I'm not sure about this, as a user, I would prefer to know a parameter is not right and fix it in the config file than having the software fix it for me and maybe not seeing the warning in the terminal output.

mariehbourget:
> Small comment: Could we adapt the terminal output, from:
> 
> ```
> Loaded 79764 axial slices for the training set.
> ```
> 
> to something like:
> 
> ```
> Loaded 79764 axial patches for the training set.
> ```

Fixed in 935bf2e

charleygros (approved):

docs/source/configuration_file.rst
charleygros: I think it should be added from now, ie in this PR, as it will be present in the config file etc.

PR #765, opened 2021-04-23
Refactoring: BidsDataset __init__ reduce complexity 

### LINKED ISSUE #719: Refactoring: reduce complexity of BidsDataset __init__ for loop
As commented [here](https://github.com/ivadomed/ivadomed/pull/699#discussion_r590564613), there is a lot of code nested within the for loop in BidsDataset [here](https://github.com/ivadomed/ivadomed/blob/0a4b7bb3fad8092b31b347332b16f5308be09a54/ivadomed/loader/loader.py#L837).
We should check if there is a way to extract some of the logic into other functions for better readability.

gisellemartel:
#### GitHub

#### PR contents

## Description
The readability of the __init__ function of the BidsDataset class can be improved by reducing the complexity of the function, specifically by abstracting some of the logic nested within for-loops into other functions.

## Linked issues
Resolves #719


ivadomed/loader/loader.py
```diff
@@ -1037,77 +1037,9 @@ def __init__(self, bids_df, subject_file_lst, target_suffix, contrast_params, sl
...
-                subj_id = subject.split('.')[0].split('_')[0]
-                multichannel_subjects[subj_id]["absolute_paths"][idx] = df_sub['path'].values[0]
-                multichannel_subjects[subj_id]["deriv_path"] = target_filename
-                multichannel_subjects[subj_id]["metadata"][idx] = metadata
-                if roi_filename:
-                    multichannel_subjects[subj_id]["roi_filename"] = roi_filename
-            else:
-                self.filename_pairs.append(([df_sub['path'].values[0]],
-                                            target_filename, roi_filename, [metadata]))
+            multichannel_subjects = self.create_filename_pair(multichannel_subjects, subject, c, tot, multichannel,
```
mariehbourget, line 1040: I find the return of `multichannel_subjects` here hard to read/understand because `create_filename_pair` does 2 different things. 1- Returns `multichannel_subjects` when multichannel is true. 2- Fills `self.filename_pairs` when multichannel is false.

ivadomed/loader/loader.py
```diff
@@ -1121,3 +1053,91 @@ def __init__(self, bids_df, subject_file_lst, target_suffix, contrast_params, sl
...
+        
+        # Fill multichannel dictionary
+        # subj_id is the filename without modality suffix and extension
+        if multichannel:
+            multichannel_subjects = self.fill_multichannel_dict(multichannel_subjects, subject, idx_dict, df_sub, 
+                                                                roi_filename, target_filename, metadata)
+        else:
+            self.filename_pairs.append(([df_sub['path'].values[0]],
+                                        target_filename, roi_filename, [metadata]))
+        return multichannel_subjects
```
mariehbourget, line 1143: Maybe the function should not have a return, and these lines could remain in the for loop outside of `create_filename_pair`. It would be clearer that the behavior is different when multichannel or not.

ivadomed/loader/loader.py
```diff
@@ -1121,3 +1053,91 @@ def __init__(self, bids_df, subject_file_lst, target_suffix, contrast_params, sl
...
+            # If suffix_list is a string, then only one rater annotation per class is available.
+            # Otherwise, multiple raters segmented the same class.
+            if isinstance(suffix_list, list):
+                for suffix in suffix_list:
+                    if suffix in derivative:
+                        target_filename[idx].append(derivative)
+            elif suffix_list in derivative:
+                target_filename[idx] = derivative
+            
+            return target_filename
```
mariehbourget, line 1068: If I'm not mistaken, the return should be outside of the for loop.

This was not picked up by our tests, maybe we should add a test to make sure all target_suffix are considered, both when provided as a list and as a list of list.

ak-gmartel: @mariehbourget  I am in the process of writing a unit test for this now, however I am bit confused what the intention is for `target_filename`? I noticed that we appear to be setting it as a list of filenames, but looking elsewhere in the code it seems like `target_filename` is supposed to be a string and not a list. Was the intention to return a list of target_filenames or just one? If we want to just return one file name then I think in the `elif` block we should probably return the filename as soon as we find a matching derivative in the `suffix_list`? In the `if` block it appears we are trying to return a list of filenames? I just want to make sure my assumptions here are correct before pushing the change with the unit test

gisellemartel: @mariehbourget  I am in the process of writing a unit test for this now, however I am bit confused what the intention is for `target_filename`? I noticed that we appear to be setting it as a list of filenames, but looking elsewhere in the code it seems like `target_filename` is supposed to be a string and not a list. Was the intention to return a list of target_filenames or just one? If we want to just return one file name then I think in the `elif` block we should probably return the filename as soon as we find a matching derivative in the `suffix_list`? In the `if` block it appears we are trying to return a list of filenames? I just want to make sure my assumptions here are correct before pushing the change with the unit test

mariehbourget: @gisellemartel 
The intention is to return a list of `target_filename` corresponding to the list of `target_suffix` of the config file.

There are 2 cases in the `elif` block:
1) When `target_suffix` is a list containing 1 string to train a single class model.
2) When `target_suffix` is a list containing more than 1 string to train a multi class model.

For example (from the data_testing dataset), we have the image `sub-unf1_T1w.nii.gz`. In the derivatives, we have 2 different labels for this image: `sub-unf01_T1w_lesion-manual.nii.gz` and  `sub-unf01_T1w_seg-manual.nii.gz`.

1) In case 1: the `target_suffix` list is `["_lesion-manual"]` to train a model the class `lesion` only. Then the `target_filename` should be `["sub-unf01_T1w_lesion-manual.nii.gz"]`.
2) In case 2: the `target_suffix` list is `["_lesion-manual", "_seg-manual"]` to train both the class `lesion` and the class `seg`. Then the `target_filename` should be `["sub-unf01_T1w_lesion-manual.nii.gz", "sub-unf01_T1w_seg-manual.nii.gz"]`. This is the case that was not detected by the unit testing earlier when the function was only returning the first match.

The `if isinstance(suffix_list, list):` block deals with a 3rd case where we have `target_suffix` as a list of list of string.
I don't think we have an example for this in the testing dataset or a unit test for that case.
Here the `target_suffix` would be something like `[["_lesion-manual-rater1", "_lesion-manual-rater2"]]` for the single class case and `[["_lesion-manual-rater1", "_lesion-manual-rater2"], ["_seg-manual-rater1", "_seg-manual-rater2"]]` for the multi class case (meaning than we have 2 different target for the class `lesion` and 2 different target for the class `seg`), and the `target_filename` would also be a list of list of string with the same form.

gisellemartel: Thank you for the very detailed explanation!! This was quite helpful. I have added a unit test for now that tests the simpler case when we have `target_suffix` that is a list. In the case of having a list of lists, this seems to break the code elsewhere before the modifications to the loader I made. I would suggest perhaps putting this in a separate issue. What do you think @mariehbourget?

log.txt
```diff
@@ -0,0 +1,66 @@
+INFO:ivadomed.utils:
```
mariehbourget, line 1: I don't think this file is supposed to be track in the repo?

tmp/bids_dataframe.csv
```diff
@@ -0,0 +1,10 @@
+path,filename,AcquisitionMatrixPE,AcquisitionNumber,AcquisitionTime,BaseResolution,BodyPartExamined,ConsistencyInfo,ConversionSoftware,ConversionSoftwareVersion,DeviceSerialNumber,DwellTime,EchoTime,EchoTrainLength,FlipAngle,ImageOrientationPatientDICOM,ImageType,InPlanePhaseEncodingDirectionDICOM,InstitutionAddress,InstitutionName,InstitutionalDepartmentName,InversionTime,MRAcquisitionType,MagneticFieldStrength,Manufacturer,ManufacturersModelName,Modality,ParallelReductionFactorInPlane,PartialFourier,PatientPosition,PercentPhaseFOV,PhaseEncodingDirection,PhaseEncodingSteps,PhaseOversampling,PhaseResolution,PixelBandwidth,ProcedureStepDescription,ProtocolName,PulseSequenceDetails,ReceiveCoilActiveElements,ReceiveCoilName,ReconMatrixPE,RepetitionTime,SAR,ScanOptions,ScanningSequence,SequenceName,SequenceVariant,SeriesDescription,SeriesNumber,ShimSetting,SliceThickness,SoftwareVersions,SpacingBetweenSlices,StationName,TxRefAmp,datatype,extension,subject,suffix,participant_id,age,data_id,date_of_birth,date_of_scan,first_name,institution_id,last_name,pathology,sex
```
mariehbourget, line 1: I don't think this file is supposed to be track in the repo?

testing/unit_tests/test_loader.py
```diff
@@ -216,5 +216,51 @@ def test_2d_patches(download_data_testing_test_files,
     assert len(ds) == 16
 
 
+@pytest.mark.parametrize('loader_parameters', [{
+    "path_data": [os.path.join(__data_testing_dir__, "microscopy_png")],
+    "bids_config": f"{path_repo_root}/ivadomed/config/config_bids.json",
+    "target_suffix": ["_seg-myelin-manual"],
```
mariehbourget, line 222: maybe we should test with a list of 2 target suffix, ex: with ["_seg_myelin_manual", "_seg-axon-manual"]. I tried it but it does not work, I'm not sure but I think `target_filename` may be overwritten in the for loop.

testing/unit_tests/test_loader.py
```diff
@@ -216,5 +216,51 @@ def test_2d_patches(download_data_testing_test_files,
...
+    """
+    loader_parameters.update({"model_params": model_parameters})
+    bids_df = imed_loader_utils.BidsDataframe(loader_parameters, __tmp_dir__, derivatives=True)
+    data_lst = ['sub-rat3_ses-01_sample-data9_SEM.png']
+    test_ds = imed_loader.load_dataset(bids_df,
+                                  **{**loader_parameters, **{'data_list': data_lst,
+                                                             'transforms_params': transform_parameters,
+                                                             'dataset_type': 'training'}})
+
+    test_subject = 'sub-unf02_T2star.nii.gz'
```
mariehbourget, line 251: The subject should be `sub-rat3_ses-01_sample-data9_SEM.png` from the data_lst above (so it is the same as the files in the dataframe)

ivadomed/loader/loader.py
```diff
@@ -1222,3 +1164,83 @@ def __init__(self, bids_df, subject_file_lst, target_suffix, contrast_params, mo
...
+        if isinstance(target_suffix[0], str):
+            target_filename, roi_filename = [None] * len(target_suffix), None
+        else:
+            target_filename, roi_filename = [[] for _ in range(len(target_suffix))], None
+
+        derivatives = bids_df.df[bids_df.df['filename']
+                        .str.contains('|'.join(bids_df.get_derivatives(subject, all_deriv)))]['path'].to_list()
+
+        for derivative in derivatives:
+            target_filename = self.get_target_filename(target_suffix, target_filename, derivative)
```
mariehbourget, line 1224: I'm not sure but I think `target_filename` is overwritten in the for loop (see comment in test)

mariehbourget (commented):
Hi @gisellemartel, I tried the test with 2 `target_suffix` (list of strings) and it doesn't work. I think there is an issue with the `for derivative in derivatives:` loop.

> In the case of having a list of lists (setting target_suffix as [['_seg-manual']], this seems to break the code elsewhere before the modifications to the loader I made (or perhaps my mocked test parameters are missing something?). I would suggest perhaps putting this in a separate issue. What do you think @mariehbourget?

Yes, I agree, if the list of list breaks before the loader, we should open a separate issue.

log.txt
gisellemartel: I'll revert it, sorry about that! For some reason I didn't see this file indexed after staging my changes. Perhaps we should also open an issue to add and log files to the gitignore file

tmp/bids_dataframe.csv
gisellemartel: Same issue as above, seems like the tmp folder is not tracked by the .gitignore file

ivadomed/loader/loader.py
gisellemartel: So target_filename is copied and returned by the function after its contents is modified. What makes the test fail when changing the parameters is when we initialize `target_filename = [None]` on line 257. Since the function `get_target_filename` accesses the index of the target_filename list by index instead of appending to it, it causes the program to crash. I will make the initialization more generalized to avoid future problems

gisellemartel: I removed the setting of `target_filename` as I realize it is not really necessary since `get_target_filename` does not create `target_filename`, but only alters its contents.

gisellemartel:
> Hi @gisellemartel, I tried the test with 2 `target_suffix` (list of strings) and it doesn't work. I think there is an issue with the `for derivative in derivatives:` loop.
> 
> > In the case of having a list of lists (setting target_suffix as [['_seg-manual']], this seems to break the code elsewhere before the modifications to the loader I made (or perhaps my mocked test parameters are missing something?). I would suggest perhaps putting this in a separate issue. What do you think @mariehbourget?
> 
> Yes, I agree, if the list of list breaks before the loader, we should open a separate issue.

I have created [this](https://github.com/ivadomed/ivadomed/issues/789) issue related to this discussion

mariehbourget (commented):
Thanks Giselle!
I found a duplication in the test (we were loading the `target_filename` twice. I made a suggestion to correct it. Also, I think we should remove the return in `get_target_filename` (it doesn't seem to affect the behavior but is not necessary)

I also tested the loader with different configs (single/multichannel, single/multiclass and with/ without ROI, everything seems to work as expected.

testing/unit_tests/test_loader.py
```diff
@@ -216,5 +216,51 @@ def test_2d_patches(download_data_testing_test_files,
     assert len(ds) == 16
 
 
+@pytest.mark.parametrize('loader_parameters', [{
+    "path_data": [os.path.join(__data_testing_dir__, "microscopy_png")],
+    "bids_config": f"{path_repo_root}/ivadomed/config/config_bids.json",
+    "target_suffix": ["_seg_myelin_manual", "_seg-axon-manual"],
```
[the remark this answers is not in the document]
gisellemartel: right good catch!

testing/unit_tests/test_loader.py
```diff
@@ -216,5 +216,51 @@ def test_2d_patches(download_data_testing_test_files,
...
+    test_subject = 'sub-rat3_ses-01_sample-data9_SEM.png'
+
+    all_deriv = bids_df.get_deriv_fnames()
+    test_derivatives = bids_df.df[bids_df.df['filename']
+                .str.contains('|'.join(bids_df.get_derivatives(test_subject, all_deriv)))]['path'].to_list()
+
+    target_filename = [None] * len(loader_parameters["target_suffix"])
+
+    for test_deriv in test_derivatives:
+        test_ds.get_target_filename(loader_parameters["target_suffix"], target_filename, test_deriv)
```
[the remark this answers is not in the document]
gisellemartel: Great suggestion! This is much simpler than what I had, so thank you 😅

ivadomed/loader/loader.py
```diff
@@ -1135,77 +1135,19 @@ def __init__(self, bids_df, subject_file_lst, target_suffix, contrast_params, mo
...
-                metadata[metadata_choice] = df_sub[metadata_choice].values[0]
-                # Create metadata dict for OHE
-                data_lst = sorted(set(bids_df.df[metadata_choice].dropna().values))
-                metadata_dict = {}
-                for idx, data in enumerate(data_lst):
-                    metadata_dict[data] = idx
-                metadata['metadata_dict'] = metadata_dict
-
+            df_sub, roi_filename, target_filename, metadata = self.create_filename_pair(multichannel_subjects, subject, 
+                                                                                c, tot, multichannel, df_subjects,
```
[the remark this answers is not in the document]
gisellemartel: Good catch! Just pushed a fix for the alignment issues

mariehbourget (approved):
LGTM, thanks @gisellemartel!

Issue #306, opened 2020-06-15
Adapt the evaluation procedure for histology data


mariehbourget:
- Deal with patch overlap for inference

mariehbourget:
As discussed in meeting:

- We will try to keep the NifTI/Nibabel workflow to avoid a major refactoring
- We will refactor again the loader part to convert images + ground truth to Nibabel objects instead of numpy arrays, this only impact the work done in the previous PR #734.
- All the processing for inference/testing will be done with those NifTI files (as it is right now)
- We will need a function to convert the predictions back to the original file format at the end
- This does not impact the image patching done in #747/#763 nor the reconstruction from 2D patches started in #781/#782.

mariehbourget:
Update:
- 2D reconstruction is implemented in #782.
- Loading PNG/TIF as Nibabel objects/NifTI files is implemented in #813.
- Bug fix for testing of 2D files in #835.
- Remaning inference steps are implemented in #834, including the conversions from Nibabel/NifTI to PNG of predictions.

TODO:
- test `uncertainty` functions
- test `threshold_analysis` function
- implement inference steps for OMETIFF

mariehbourget:
Update:
- Multiclass evaluation bug fixed in #837.
- `threshold_analysis` function is tested and works without any changes.

TODO:
- Implement inference steps for OMETIFF
- Tutorial in progress in PR #1036
- Update “[Data](https://ivadomed.org/data.html)” in documentation after the release of BIDS 1.7.0.
- Issue #912
- Issue #972
    - Related issue #977
    - Partially addressed in PR #982
    - Investigate `deepcopy` (partially revert changes from PR #806), see this [comment](https://github.com/ivadomed/ivadomed/issues/972#issuecomment-960022021)
    - Additional investigation for regarding image resampling in this [comment](https://github.com/ivadomed/ivadomed/issues/977#issuecomment-960057398)
- Issue #975:
    - Adapt evaluation metrics to microscopy: “get_ltp_lfn” and “get_lfp” don’t make sense for axon/myelin segmentation. Evaluation is unusable with very large images/datasets (ex: with BF model).

PR #782, opened 2021-05-09
Add image reconstruction from 2D patches

### LINKED ISSUE #781: Implement image reconstruction from 2D patches for MRI data
### Context
This issue is part of the inference refactoring project to include microscopy data (#306).
As a first step, we want to implement the image reconstruction from the 2D patches for MRI data when running inference.

### TODO
- Add an `image_reconstruction` function in `inference.py` (similar to `volume_reconstruction` for 3D subvolumes).
- Implement how to manage patches’ borders and overlap. This may require some modification on the 2D patching feature implemented previously for training in #747.
- Trigger the image reconstruction from `testing.py/run_inference` and from `inference.py/segment_volume`.

mariehbourget:
#### GitHub

#### PR contents

## Description
- This PR is part of the inference refactoring project to include microscopy data (#306).
- It implements an `image_reconstruction` function in `inference.py` to reconstruct images from 2d patches.
- The function is triggered both in `testing.py/run_inference` and `inference.py/segment_volume` at inference time with command `test` and `segment` from the config file respectively.
- As discussed in #306, the feature is working only for MRI data at the moment due to dependencies with NifTI/Nibabel in the inference process.
- The current reconstruction is done by averaging the predictions in the overlap between patches (as it is done in volume_reconstruction for 3D subvolumes).
- The PR includes a functional test.

## Linked issues
Fixes #781


mariehbourget:
This PR is ready for review.

To test the PR:
- You can run a 1 epoch training on 1 subject with the config_new_loader.json file
- Then use the same config file with the command "segment".
You should have something like this on `sub-cardiff03_T2w`.

test_pr782_patch

Note 1:
The "patch pattern" is visible because it's only a 1 epoch and 1 subject training. After many tests on my part, this remains the best way to see if the reconstruction is working on MRI data.

Note 2: 
The 2d reconstruction of the image from the patches is done in the same way than sub-volumes in 3D i.e. by averaging the prediction on the overlaps.  I first wanted to do a proper long training and assess the quality of the reconstruction (no borders visible, good predictions in the overlap, etc) on MRI data, and optimize the reconstruction algorithm based on that. But I changed my mind and decided to wait until the inference process is completed for microscopy to do that. The quality assessment will be 100x easier to do on axon/myelin segmentations than on spinal cord segmentation.

ivadomed/inference.py
```diff
@@ -345,6 +345,10 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
 
     kernel_3D = bool('Modified3DUNet' in context and context['Modified3DUNet']['applied']) or \
                 not context['default_model']['is_2d']
+    length_2D = context["default_model"]["length_2D"] if "length_2D" in context["default_model"] else []
+    stride_2D = context["default_model"]["stride_2D"] if "stride_2D" in context["default_model"] else []
```
charleygros, line 349: Would it work if the selected model is not the "default_model"? Or are we enforcing that these params are only defined in the `default_model`?

charleygros:
> and optimize the reconstruction algorithm based on that.

Just curious: what did you have in mind here?

charleygros:
I've done different tests: Mouhahah that's so good! 🥇 💯 👍

charleygros (approved):

ivadomed/inference.py
mariehbourget: We had this discussion previously [here](https://github.com/ivadomed/ivadomed/pull/763#discussion_r616367317), and decided to go with `length_2D` and `stride_2D` in the `default_model`.
From the [architecture settings](https://ivadomed.org/en/latest/configuration_file.html#architecture) in the doc, there are 4 architectures: default, Filmed, Hemis and 3D. Hemis and 3D do not apply in this case. And if I understand correctly, the FilmedUnet settings are applied "on top" of the default model?
Maybe I'm wrong, I feel like I'm missing something here. Can you give me an example of model where we would want to use the patch that is not the default_model? Thanks!

mariehbourget:
> > and optimize the reconstruction algorithm based on that.
> 
> Just curious: what did you have in mind here?

On microscopy data in ADS, we sometimes see the "patch pattern" or loss of continuity at the borders of patches in the segmentations. In ADS, the overlap is entirely cropped to remove predictions at the borders of patches (and the stride is adjusted to cover all the images). In the current implementation of ivadomed, the overlap is averaged between patches (but include the pixels at the borders). So, the borders we see on my "1 epoch, 1 subject" training may remain on microscopy segmentations even after proper training.

One possibility we discussed a while ago was to do "the best of both world", which would be to crop part of the overlap (pixels at the borders of patches) and average the rest of the overlap.

ivadomed/inference.py
charleygros: >

eheh you are very right :-)

> And if I understand correctly, the FilmedUnet settings are applied "on top" of the default model?

Yes absolutely, it is done [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/main.py#L156).

All good then!!

PR #791, opened 2021-05-14
Refactor segment_volume to reduce complexity

### LINKED ISSUE #693: Refactoring: Reduce Cognitive complexity of `segment_volume` funciton
After running sonarlint, I have noticed that the `segment_volume` function in `inference.py` has a [cognitive complexity ](https://docs.codeclimate.com/docs/cognitive-complexity) of 92, which is very high indicating that the code is difficult to understand and maintain (normally values around 15 or less are ideal). The function should be refactored to reduce its length and complexity to improve the code design.

gisellemartel:
#### GitHub

#### PR contents

## Description
The `segment_volume` function in inference.py was measured to have a very high cognitive complexity, rendering it difficult to maintain and understand. This PR aims to simplify and abstract some of the logic into other functions to allow the maintainer to have a better high-level understanding of the code.

## Linked issues
Resolves #693


ivadomed/inference.py
```diff
@@ -16,6 +16,31 @@
 from ivadomed import training as imed_training
 
 
+def get_preds(context, model, fname_model, onnx_inference, model_params, cuda_available, batch):
```
dyt811, line 19: I would recommend adding comments for the refactored/extracted function here, in addition to start introduce type hints. @gisellemartel I can take over the PR and the branch if you don't have to wrap this up. 

While the numerous parameter is a bit concern, it does at least alleviate the original cognitive burdens compare to before so I think this is a step in the right direction and should be further continued to simplify the inference process. 

I can help continue this effort in future PRs.

ivadomed/inference.py
```diff
@@ -100,6 +125,52 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
 
     return nib_pred
 
+def process_transformations(context, fname_roi, fname_prior, metadata, slice_axis, fname_images):
```
dyt811, line 128: Ditto above.

ivadomed/inference.py
```diff
@@ -100,6 +125,52 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
...
+
+    if 'object_detection_params' in context and \
+            context['object_detection_params']['object_detection_path'] is not None:
+        imed_obj_detect.bounding_box_prior(fname_prior, metadata, slice_axis,
+                                            context['object_detection_params']['safety_factor'])
+        metadata = [metadata] * len(fname_images)
+    
+    return metadata
+
+def set_option(options, postpro, context, key):
```
dyt811, line 147: Thank you for generalizing this to make things easier in the future. Ditto above.

ivadomed/inference.py
```diff
@@ -100,6 +125,52 @@ def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False,
...
+    if options[key]:
+        postpro[key] = {}
+    # Remove key in context if value set to 0
+    elif key in context['postprocessing']:
+        del context['postprocessing'][key]
+    return postpro
+
+def set_postprocessing_options(options, context):
```
dyt811, line 155: Ditto.

ivadomed/inference.py
```diff
@@ -324,6 +307,47 @@ def split_classes(nib_prediction):
         pred_list.append(class_pred)
     return pred_list
 
+def reconstruct_3d_object(context, batch, undo_transforms, preds, preds_list, kernel_3D, volume,
```
dyt811, line 310: I recommend future PR to start thinking of reducing the parameter complexities, either via setting classes or configuration modules/dict etc to avoid having large number of parameters past around such as shown. I can work on those PRs drawing some inspirations from Detectron2's approach but I believe everything in this PR is a very important first step before we worry about those. The original segmentation function was way complex for trying to do too many things in a single function that can be further streamlined after this PR.

dyt811 (changes requested):
I verified almost all refactoring changes are verbatim in terms of the key underlying ML operations and should not impact the inference performances whatsoever but should improve the readability and debug burden substantially for the devs who will be working on it in the near future. 

While we can merge the PR as it is in the current stage, I strongly recommend minor documentation changes below, namely need function signature comments and elaborations for functions: 
- get_preds
- get onhotencoder
- process transformations
- set_option
- set_postprocessing_options
- reconstruct_3d_object

ivadomed/inference.py
gisellemartel: I agree with everything mentioned here! I think given the large number of changes needed in reducing parameter complexities it would have been best placed in another PR. Taking advantage of class settings would definitely help with this process

ivadomed/inference.py
gisellemartel: Thank you @dyt811! I can take care of adding some comments to the functions.

ivadomed/inference.py
```diff
@@ -16,6 +18,73 @@
 from ivadomed import training as imed_training
 
 
+def onnx_inference(model_path, inputs):
```
dyt811, line 21: Could you add the missing type hints here?

dyt811 (approved):
Great work! Thanks for the awesome update to the segment_volume class which has been a bit monolithic before. Now it is a lot easier to digest for new devs and existing devs looking to make changes!

PR #794, opened 2021-05-17
Add tmp and log file to gitignore

### LINKED ISSUE #790: Certain temporary files not detected by .gitignore
I have noticed that certain files are not detected by the .gitignore. For instance, any left over temporary files that were not cleaned up or log files. Here are some examples of files/folders that ideally should be ignored by git:

`log.txt`
`tmp/`

Also, as another suggestion I would also add a mechanism so that the temporary output directory should also be ignored. Perhaps one of way of doing this is eventually placing the folders currently in the root in another folder named something like `src`, so that all temp folders created at the root can always be ignored, even dynamically

cakester:
#### GitHub

#### PR contents

## Description

Adding log.txt and tmp folder in .gitignore per issue #790.

## Linked issues

resolves #790

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,12 @@ bin/
 # All __pycache__
 __pycache__/
 
+# All test output folders
+tmp/
+
+# Logs
+log.txt
+
 # Pycharm stuff
 .idea
 


.gitignore
```diff
@@ -12,6 +12,12 @@ bin/
 # All __pycache__
 __pycache__/
 
+# All test output folders
+tmp/
+
+# Logs
+log.txt
```
dyt811, line 19: So if I recall correctly, these log.txt/tmp folder show up not necessarily in the repo root directory but could be in one or more of the subdirectory. Could you revise it accordingly? 

Secondly, could you investigate the root cause and see if it is possible to prevent this behavior in the first place? i.e. generate these files in the unit test temporary folder if the above file/folder were generated during unit testing scenarios.

dyt811 (changes requested):
I think we need to investigate the behavior further.

.gitignore
cakester: I just tested it if you put log.txt in .gitignore it ignores this file doesn't matter which subdirectory it is in

dyt811: TIL. Good call. I need to brush up on my gitignore syntax. :)

dyt811 (approved):

PR #804, opened 2021-05-25
Hot patch to address Inference issue #803

### LINKED ISSUE #803: segment_volume issues after refactoring
## Issue description

While working on #782, I encountered a two issues with the `segment_volume` function since the refactoring in #791 (tagging @dyt811 so he is aware). The issues described below are present on `master` branch.

#### 1) local variable 'model' referenced before assignment
In the call for `get_preds` [here](https://github.com/ivadomed/ivadomed/blob/05dbd217cc9b4c06a0bef77ce06cc57937dd45ec/ivadomed/inference.py#L358), the argument `model` is not defined when the model in a onnx model.
<details>
<summary> Error </summary>

```
File "/home/mhbourget/ivadomed/ivadomed/inference.py", line 361, in segment_volume
preds = get_preds(context, model, fname_model, model_params, cuda_available, batch)
UnboundLocalError: local variable 'model' referenced before assignment
```
</details>

#### 2) last_sample_bool, weight_matrix and volume are re-initialized at each iteration
`reconstruct_3d_object` is in a the "batch" for loop  [here](https://github.com/ivadomed/ivadomed/blob/05dbd217cc9b4c06a0bef77ce06cc57937dd45ec/ivadomed/inference.py#L366). `last_sample_bool, weight_matrix, volume` are initialized in `reconstruct_3d_object` [here](https://github.com/ivadomed/ivadomed/blob/05dbd217cc9b4c06a0bef77ce06cc57937dd45ec/ivadomed/inference.py#L409), whereas they were initialized outside of the for loop before. They are used to reconstruct a 3D volume from several sub-volumes and should not be re-initialized at each batch iteration. This result in the following error, because volume is None at the second iteration and should not be.
<details>
<summary> Error </summary>

```
File "/home/mhbourget/ivadomed/ivadomed/inference.py", line 479, in volume_reconstruction
volume[:, x_min:x_max, y_min:y_max, z_min:z_max] += pred[smp_idx]
TypeError: 'NoneType' object is not subscriptable
```
</details>

From my understanding, `last_sample_bool, weight_matrix, volume` should be initialized before the for loop, they should be passed in arguments to `reconstruct_3d_object` and also return by the function.

### Expected behavior and test
I found this because I used the exact same principle in my PR #782 to reconstruct images from 2d patches and it does not work anymore because of the re-initialization of `last_sample_bool, weight_matrix, volume`. I confirmed that the issue is on my branch with 2D patches as well as on `master` branch with 3D sub-volumes. We should fix this issue on `master` first before I go on with PR.

We have a test for the second issue: `test_segment_volume_3d` in `functional_tests/test_segment_volume.py`.

The test did not catch the error because the `CenterCrop` transform (size of the volume being segmented) is the same as the `length_3D` of the `Modified3DUNet` model (size of 1 sub-volume). Because of that, there is only 1 sub-volume in the full volume, hence the for loop does only 1 iteration and the issue is not caught.

To test properly, we should change the `CenterCrop` parameter of this test from `[96, 96, 16]` to `[192, 192, 16]`, which would give us 4 sub-volume of `[96, 96, 16]` to reconstruct the full volume of `[192, 192, 16]`.

### Steps to reproduce

- I tested on `master` branch.
- I trained a dummy 3D model for a few epochs on the `data_example_spinegeneric` dataset with the `config_tumorSeg.json` file, changing the `length_3D` and `stride_3D` parameters of `Modified3DUnet` by `[256, 128, 16]` (for an entire volume i.e. `CenterCrop` of `[512, 256, 16]`) (there may be an already trained 3D model somewhere to test this @andreanne-lemay, @charleygros?)
- I switched the command to `segment` in the config file, segment uses `segment_volume` and the segmentation fails with the errors describe above.
- Note: the same model with the command `test` works, because `test` does not use the `segment_volume` function.

dyt811:
#### GitHub

#### PR contents

## Description

Attempting to hot patch Issue 803. 
- Bug 1 has been tentatively patched. 
- Bug 2 still needs proper investigation
- Unit test updated to help detect Bug 2, currently failing. 

## Linked issues

Hoping to resolve #803. Still needs work.

### DIFF
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -55,9 +55,7 @@ def get_preds(context: dict, path_model_file: str, model_params: dict, gpu_id: i
         tensor: predictions from the model.
     """
     # Define device
-    cuda_available = torch.cuda.is_available()
-    logger.debug(f"CUDA Status: {cuda_available}")
-    device = torch.device("cpu") if not cuda_available else torch.device("cuda:" + str(gpu_id))
+    cuda_available, device = imed_utils.define_device(gpu_id)
 
     with torch.no_grad():
 
@@ -87,7 +85,7 @@ def get_preds(context: dict, path_model_file: str, model_params: dict, gpu_id: i
             logger.debug(f"Conduct ONNX model inference... ")
             preds = onnx_inference(path_model_file, img)
 
-        logger.debug("Sending ONNX model to CPU")
+        logger.debug("Sending predictions to CPU")
         # Move prediction to CPU
         preds = preds.cpu()
 
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -116,7 +116,7 @@ def get_onehotencoder(context: dict, folder_model: str, options: dict, ds: Datas
 
 
 def pred_to_nib(data_lst: List[np.ndarray], z_lst: List[int], fname_ref: str, fname_out: str, slice_axis: int,
-                debug: bool = False, kernel_dim='2d', bin_thr=0.5, discard_noise: bool = True,
+                debug: bool = False, kernel_dim: str='2d', bin_thr: float=0.5, discard_noise: bool = True,
                 postprocessing: dict = None) -> nib.Nifti1Image:
     """Save the network predictions as nibabel object.
 
@@ -310,12 +310,12 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
     """
 
     # Check if model folder exists and get filenames to be stored as string
-    path_model_file: str
-    path_model_metadata_file: str
-    path_model_file, path_model_metadata_file = imed_models.get_model_filenames(folder_model)
+    fname_model: str
+    fname_model_metadata: str
+    fname_model, fname_model_metadata = imed_models.get_model_filenames(folder_model)
 
     # Load model training config
-    context = imed_config_manager.ConfigurationManager(path_model_metadata_file).get_config()
+    context = imed_config_manager.ConfigurationManager(fname_model_metadata).get_config()
 
     postpro_list = ['binarize_prediction', 'keep_largest', ' fill_holes', 'remove_small']
     if options is not None and any(pp in options for pp in postpro_list):
@@ -381,7 +381,7 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
     preds_list, slice_idx_list = [], []
     last_sample_bool, weight_matrix, volume = False, None, None
     for i_batch, batch in enumerate(data_loader):
-        preds = get_preds(context, path_model_file, model_params, gpu_id, batch)
+        preds = get_preds(context, fname_model, model_params, gpu_id, batch)
 
         # Set datatype to gt since prediction should be processed the same way as gt
         for b in batch['input_metadata']:
@@ -434,7 +434,7 @@ def reconstruct_3d_object(context: dict, batch: dict, undo_transforms: UndoCompo
         fname_images (list): list of image filenames (e.g. .nii.gz) to segment.
         i_batch (int): index of current batch.
 
-        last_sample_bool:
+        last_sample_bool: : flag to indicate whether this is the last sample in the 3D volume
         weight_matrix (tensor): the weight matrix
         volume (tensor): the volume tensor that is being partially reconstructed through the loop
 
@@ -500,11 +500,11 @@ def volume_reconstruction(batch: dict, pred: tensor, undo_transforms: UndoCompos
         weight_matrix (tensor): Weights containing the number of predictions for each voxel
 
     Returns:
-        pred_undo (tensor), undone subvolume,
-        metadata (dict), metadata,
-        last_sample_bool (bool), boolean representing if its the last sample of the volume
-        volume (tensor), representing the volume reconstructed
-        weight_matrix (tensor), weight matrix
+        pred_undo (tensor): undone subvolume,
+        metadata (dict): metadata,
+        last_sample_bool (bool): boolean representing if its the last sample of the volume
+        volume (tensor): representing the volume reconstructed
+        weight_matrix (tensor): weight matrix
     """
     x_min, x_max, y_min, y_max, z_min, z_max = batch['input_metadata'][smp_idx][0]['coord']
     num_pred = pred[smp_idx].shape[0]
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -4,6 +4,7 @@ import numpy as np
 import onnxruntime
 import torch
 import joblib
+from typing import List
 
 from torch.utils.data import Dataset
 from torch.utils.data import DataLoader
@@ -17,10 +18,13 @@ from ivadomed.loader import utils as imed_loader_utils, loader as imed_loader, f
 from ivadomed.object_detection import utils as imed_obj_detect
 from ivadomed import utils as imed_utils
 from ivadomed import training as imed_training
-from loguru import logger
+
+# To be replaced in upcoming PR by loguru update
+import logging
+logger = logging.getLogger()
 
 
-def onnx_inference(model_path, inputs):
+def onnx_inference(model_path: str, inputs: tensor) -> tensor:
     """Run ONNX inference
 
     Args:
@@ -37,7 +41,7 @@ def onnx_inference(model_path, inputs):
     return torch.tensor(ort_outs[0])
 
 
-def get_preds(context: dict, path_model_file: str, model_params: dict, gpu_id: int, batch: dict):
+def get_preds(context: dict, path_model_file: str, model_params: dict, gpu_id: int, batch: dict) -> tensor:
     """Returns the predictions from the given model.
 
     Args:
@@ -48,11 +52,11 @@ def get_preds(context: dict, path_model_file: str, model_params: dict, gpu_id: i
         batch (dict): dictionary containing input, gt and metadata
 
     Returns:
-        Tensor: predictions from the model.
+        tensor: predictions from the model.
     """
     # Define device
     cuda_available = torch.cuda.is_available()
-    logger.debug(f"Cuda Status: {cuda_available}")
+    logger.debug(f"CUDA Status: {cuda_available}")
     device = torch.device("cpu") if not cuda_available else torch.device("cuda:" + str(gpu_id))
 
     with torch.no_grad():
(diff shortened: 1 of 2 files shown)


dyt811:
@mariehbourget could you try what I have so far and see if that ameliorate the issue? I have identified the issue to be most likely related to Master branch Line 409: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/inference.py#L409 and moved it to [L377 on my branch](https://github.com/ivadomed/ivadomed/blob/yd/803-hotpatch_segment_volume/ivadomed/inference.py#L377).

ivadomed/inference.py
```diff
@@ -336,35 +364,30 @@ def segment_volume(folder_model, fname_images, gpu_id=0, options=None):
...
 
         # Set datatype to gt since prediction should be processed the same way as gt
         for b in batch['input_metadata']:
             for modality in b:
                 modality['data_type'] = 'gt'
 
         # Reconstruct 3D object
-        pred_list, target_list = reconstruct_3d_object(context, batch, undo_transforms, preds, preds_list, kernel_3D, 
-                                                        slice_axis, slice_idx_list, data_loader, fname_images, i_batch)
+        pred_list, target_list, last_sample_bool = reconstruct_3d_object(context, batch, undo_transforms, preds, preds_list, kernel_3D,
```
mariehbourget, line 388: If I try to print `volume` and `weight_matrix` before and after this function, the result is always `None` after, even though they should be passed by reference. If I add a return of `volume` and `weight_matrix`, then they are not None and the test pass.

testing/functional_tests/test_segment_volume.py
```diff
@@ -79,7 +80,8 @@ def test_segment_volume_2d(download_functional_test_files):
     shutil.rmtree(PATH_MODEL)
 
 
-def test_segment_volume_3d(download_functional_test_files):
+@pytest.mark.parametrize("length_3d", [[192, 192, 16], LENGTH_3D, ])
```
mariehbourget, line 83:
```suggestion
@pytest.mark.parametrize("center_crop", [[192, 192, 16], LENGTH_3D])
```
I think we could only test for [192, 192, 16] here, if it works for 4 sub-volumes, then it will work for 1.

testing/functional_tests/test_segment_volume.py
```diff
@@ -79,7 +80,8 @@ def test_segment_volume_2d(download_functional_test_files):
...
+def test_segment_volume_3d(download_functional_test_files, length_3d):
```
mariehbourget, line 84:
```suggestion
def test_segment_volume_3d(download_functional_test_files, center_crop):
```

testing/functional_tests/test_segment_volume.py
```diff
@@ -114,7 +116,7 @@ def test_segment_volume_3d(download_functional_test_files):
                     "dspace": 2
                 },
             "CenterCrop": {
-                "size": LENGTH_3D
+                "size": length_3d
```
mariehbourget, line 119:
```suggestion
                "size": center_crop
```

mariehbourget (commented):
Thank you @dyt811 for looking into this.
I only looked at the bug 2 for now. I'm only able to make the test pass by adding the return of `volume` and `weight_matrix` to `reconstruct_3d_object`, I don't know why the references don't work here.
I also suggested some minor changes to the test for clarity.
I'll make a more detailed review of the whole PR once we fix bug 2.

dyt811:
Understood. That is quite helpful. I will make that change accordingly and see if it helps with the new test scenario once objects are returned properly.

ivadomed/inference.py
dyt811: Great point. I will implement that fix tonight and try again. I suspect it might be related to [the issue shown here](https://discuss.pytorch.org/t/what-is-the-recommended-way-to-re-assign-update-values-in-a-variable-or-tensor/6125): i.e. in place update vs object re-assignment of PyTorch tensor objects. Super interesting, super gotcha.

testing/functional_tests/test_segment_volume.py
dyt811: Great point. I will just simplify it then to make it more readable. I kept the old one as I wasn't sure if there was anything special/unique about the old value.

ivadomed/inference.py
```diff
@@ -34,36 +40,61 @@ def onnx_inference(model_path, inputs):
...
 
-
     Returns:
-        Tensor: predictions from the model.
+        tensor: predictions from the model.
     """
+    # Define device
+    cuda_available = torch.cuda.is_available()
+    logger.debug(f"CUDA Status: {cuda_available}")
+    device = torch.device("cpu") if not cuda_available else torch.device("cuda:" + str(gpu_id))
```
charleygros, line 60: Is it to avoid the import of `utils` here? cf [this function](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/utils.py#L117)

ivadomed/inference.py
```diff
@@ -34,36 +40,61 @@ def onnx_inference(model_path, inputs):
...
+        # Otherwise, Onnex Inference (PyTorch can't load .onnx)
         else:
-            preds = model(img) if fname_model.endswith('.pt') else onnx_inference(fname_model, img)
+            logger.debug(f"Likely ONNX model detected at: {path_model_file}")
+            logger.debug(f"Conduct ONNX model inference... ")
+            preds = onnx_inference(path_model_file, img)
 
+        logger.debug("Sending ONNX model to CPU")
+        # Move prediction to CPU
         preds = preds.cpu()
```
charleygros, line 92:
```suggestion
            logger.debug("Sending ONNX model to CPU")
            # Move prediction to CPU
            preds = preds.cpu()
```

ivadomed/inference.py
```diff
@@ -34,36 +40,61 @@ def onnx_inference(model_path, inputs):
...
+            else:
+                preds = model(img)
+        # Otherwise, Onnex Inference (PyTorch can't load .onnx)
         else:
-            preds = model(img) if fname_model.endswith('.pt') else onnx_inference(fname_model, img)
+            logger.debug(f"Likely ONNX model detected at: {path_model_file}")
+            logger.debug(f"Conduct ONNX model inference... ")
+            preds = onnx_inference(path_model_file, img)
 
+        logger.debug("Sending ONNX model to CPU")
```
charleygros, line 90:
```suggestion
        logger.debug("Sending predictions to CPU")
```

ivadomed/inference.py
dyt811: Good spot! I totally didn't know about that function, we can totally import this. I had to refactor it from the earlier code as it was a bit confusing but if we have dedicated function. I will use that instead. Excellent!

ivadomed/inference.py
```diff
@@ -275,15 +308,14 @@ def segment_volume(folder_model, fname_images, gpu_id=0, options=None):
...
 
     """
-    # Define device
-    cuda_available = torch.cuda.is_available()
-    device = torch.device("cpu") if not cuda_available else torch.device("cuda:" + str(gpu_id))
 
-    # Check if model folder exists and get filenames
-    fname_model, fname_model_metadata = imed_models.get_model_filenames(folder_model)
+    # Check if model folder exists and get filenames to be stored as string
+    path_model_file: str
```
mariehbourget, line 313: Is `fname_model` replaced by `path_model_file`? I'm curious why the name changed?

mariehbourget (commented):
I reviewed the code (with minors suggestions), and tested the volume reconstruction which works as expected. Onnx models work as well.
However, I did not test the `postpro/set_option` part as I'm very unfamiliar with this portion of the code.
Thanks @dyt811!

ivadomed/inference.py
jcohenadad: just so @dyt811 is aware of [our naming conventions](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Naming-conventions-(variables,-files)#variables)

ivadomed/inference.py
```diff
@@ -85,8 +114,10 @@ def get_onehotencoder(context: dict, folder_model: str, options: dict, ds: Datas
 
     return joblib.load(os.path.join(folder_model, 'one_hot_encoder.joblib'))
 
-def pred_to_nib(data_lst, z_lst, fname_ref, fname_out, slice_axis, debug=False, kernel_dim='2d', bin_thr=0.5,
-                discard_noise=True, postprocessing=None):
+
+def pred_to_nib(data_lst: List[np.ndarray], z_lst: List[int], fname_ref: str, fname_out: str, slice_axis: int,
+                debug: bool = False, kernel_dim='2d', bin_thr=0.5, discard_noise: bool = True,
```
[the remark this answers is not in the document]
dyt811: Good eyes. That is absolutely not normal. Not sure how I missed that. Correcting it now.

ivadomed/inference.py
```diff
@@ -403,10 +434,19 @@ def reconstruct_3d_object(context: dict, batch: dict, undo_transforms: UndoCompo
         fname_images (list): list of image filenames (e.g. .nii.gz) to segment.
         i_batch (int): index of current batch.
 
+        last_sample_bool:
```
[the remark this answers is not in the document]
dyt811: Thank you for catching this. I really need to review my own PR first. Really hard to spot these looking at the entire code. I will makes sure to do that first before requesting reviews going forward.

testing/functional_tests/test_segment_volume.py
```diff
@@ -79,7 +80,8 @@ def test_segment_volume_2d(download_functional_test_files):
     shutil.rmtree(PATH_MODEL)
 
 
-def test_segment_volume_3d(download_functional_test_files):
+@pytest.mark.parametrize("center_crop", [[192, 192, 16], LENGTH_3D, ])
```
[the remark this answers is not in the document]
dyt811: Redundancy removed.

ivadomed/inference.py
dyt811: Cool. TIL. I reverted it back and will be watching out for that going forward. Definitely shorter than my typical path_to_file_reason_x_y_z style. I didn't see that in IvadoMed wiki so I must have missed a meeting on the issue. Should I copy SCT's guideline over the IvadoMed on these conventions?

mariehbourget (approved):
LGTM! Thanks @dyt811!

ivadomed/inference.py
jcohenadad: > Should I copy SCT's guideline over the IvadoMed on these conventions?

i'm always reluctant to duplicate information (which implies: more than one place to maintain). We might want to centralize our conventions and redirect instead. @joshuacwnewton had thought about it a lot so i'm tagging him here

joshuacwnewton: > (which implies: more than one place to maintain)

FYI: My opinions have shifted a little, ever since trying to coordinate the contributing guidelines.

I found it takes just as much effort (perhaps even more) to have to maintain a centralized guide. Someone has to coordinate projects to agree on shared guidelines, make sure the recommendations are general enough to suit every project, create links to point at the centralized guide, and ensure developers know to go look at this external guide (that alone is a big ask). 

So, for smaller things like variable naming conventions, I've started to feel that it's OK (even preferable) to duplicate the info for each project-specific Wiki, just to keep the info close by for each individual project.

jcohenadad: thank you for your insights @joshuacwnewton ! it makes perfect sense.

jcohenadad: so it means we should probably copy some of the parts of the SCT's wiki into ivadomed, @dyt811 if you want to help with that

dyt811: Will do. Absolutely makes sense. Will copy it over to clarify this point and review the SCT guideline in depth as well.

PR #805, opened 2021-05-25
Update test_segment_volume_2d to use ROI

### LINKED ISSUE #802: Verify intended behavior or test_segment_volume_2d with ROI
## Issue description
In `funtional_tests/test_segment_volume.py`, I think `test_segment_volume_2d` was intended to test the segmentation with a ROI because of the use of `ROI_PATH` in the `segment_volume`.

### Current behavior
The `ROI_PATH` param passed to the command in the test [here](https://github.com/ivadomed/ivadomed/blob/05dbd217cc9b4c06a0bef77ce06cc57937dd45ec/testing/functional_tests/test_segment_volume.py#L73) is not assigned to the `options` argument, and end up in the third argument `gpu_id` instead.

So, the segmentation test works, but it doesn't test with the ROI, as confirmed by the terminal output:
`WARNING: fname_roi has not been specified, then the entire volume is processed. `

The test would also fail on GPU because of the wrong `gpu_id`.

In addition, in order for the test to use the ROI, a `suffix` ("_seg-manual") should be specified in `roi_params` of the the `loader parameters`, otherwise the `ROICrop` transform will do a crop around the center of the image instead of a crop around the ROI.

### Expected behavior
I could take care of the PR to fix this but I would need someone to confirm if the test should or should not test with ROI, @andreanne-lemay, @charleygros?
Thanks!

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR updates `test_segment_volume_2d` to use the `ROI_PATH` and `ROICrop` specified in the parameters as it was first intended. The ROI file was previously ignored and all the volume was segmented. The ROI is now taken into account.

The test would also previously fail on GPU because the ROI_PATH was assigned to `gpu_id`. The test now pass on both CPU and GPU, and uses the GPU 0 as intended.

## Linked issues
Fixes #802

### DIFF
--- a/testing/functional_tests/test_segment_volume.py
+++ b/testing/functional_tests/test_segment_volume.py
@@ -44,7 +44,7 @@ def test_segment_volume_2d(download_functional_test_files):
                 "filter_empty_input": False
             },
             "roi_params": {
-                "suffix": None,
+                "suffix": "_seg-manual",
                 "slice_filter_roi": 10
             },
             "slice_axis": "axial"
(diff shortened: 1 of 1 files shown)


charleygros (approved):
Teste et approuve! Thanks heaps for fixing this crucial feature @mariehbourget

PR #806, opened 2021-05-25
Fix bug in loader for multiple raters

### LINKED ISSUE #789: Loader does not support target_suffix when containing nested lists
### Current behavior
As discussed [here](https://github.com/ivadomed/ivadomed/pull/765#pullrequestreview-659363753), the current loader seems to break when we set the `target_suffix` param of loader_params in the config file as a list of lists. For example, having it set like this: `"target_suffix: [["_seg_myelin_manual"], ["_seg_axon_manual"]]"` does not seem to be fully supported. I believe the loader may need to be modified slightly, in addition some unit tests should be added to ensure this functionality is fully supported. 

### Expected behavior
Setting `target_suffix` as a list of lists should be supported by the loader

### Steps to reproduce
Modify the `target_suffix` param of `loader_params` in the config file to be a list of lists rather than a list. Run the training. An error will likely occur and crash the program.

For example, in `config_small.json` initially we have:

`        "target_suffix": ["_seg-manual"],`

Changing it to the following should be supported:

`        "target_suffix": [["_seg-manual"]],`

However, it results in the following error:

```
Annotations from multiple raters will be used during model training, one annotation from one rater randomly selected at each iteration.


ERROR: Please provide only one annotation per class in 'target_suffix' when not training a model.

```

### System description
macOS Catalina 10.15.7 Python 3.8

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR fixes a bug when providing multiple raters annotations for training. For some reason, the flag `is_train` passed to the command `check_multiple_raters` was wrong in `main.py`. This was resulting in a error when providing `target_suffix` as a list of list with the command `train` instead of an error with the command `test`.

In addition, in `loader.py`, the GT was cast as type `data_type` for multiple raters but `data_type` was not defined. I removed it since the casting to `data_type = np.float32 if self.soft_gt else np.uint8` was removed in an earlier PR (#624) for preserving soft GT.

The error was first spotted while writing a test for PR #765. Now that the behavior is fixed, I also added a test to make sure the `get_target_filename` returns `target_filename` with the same dimensions as `target_suffix`.

## Linked issues
Fixes #789

### DIFF
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -735,9 +735,9 @@ class MRI2DSegmentationDataset(Dataset):
                 seg_pair_slice['gt'][idx_class] = seg_pair_slice['gt'][idx_class][idx_rater]
                 seg_pair_slice['gt_metadata'][idx_class] = seg_pair_slice['gt_metadata'][idx_class][idx_rater]
 
-        metadata_input = seg_pair_slice['input_metadata']
-        metadata_roi = roi_pair_slice['gt_metadata']
-        metadata_gt = seg_pair_slice['gt_metadata']
+        metadata_input = seg_pair_slice['input_metadata'] if seg_pair_slice['input_metadata'] is not None else []
+        metadata_roi = roi_pair_slice['gt_metadata'] if roi_pair_slice['gt_metadata'] is not None else []
+        metadata_gt = seg_pair_slice['gt_metadata'] if seg_pair_slice['gt_metadata'] is not None else []
 
         # Run transforms on ROI
         # ROI goes first because params of ROICrop are needed for the followings
@@ -941,8 +941,8 @@ class MRI3DSubVolumeSegmentationDataset(Dataset):
                 seg_pair['gt'][idx_class] = seg_pair['gt'][idx_class][idx_rater]
                 seg_pair['gt_metadata'][idx_class] = seg_pair['gt_metadata'][idx_class][idx_rater]
 
-        metadata_input = seg_pair['input_metadata']
-        metadata_gt = seg_pair['gt_metadata']
+        metadata_input = seg_pair['input_metadata'] if seg_pair['input_metadata'] is not None else []
+        metadata_gt = seg_pair['gt_metadata'] if seg_pair['gt_metadata'] is not None else []
 
         # Run transforms on images
         stack_input, metadata_input = self.transform(sample=seg_pair['input'],
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -923,6 +923,10 @@ class MRI3DSubVolumeSegmentationDataset(Dataset):
         Args:
             index (int): Subvolume index.
         """
+
+        # copy.deepcopy is used to have different coordinates for reconstruction for a given handler,
+        # to allow a different rater at each iteration of training, and to clean transforms params from previous
+        # transforms i.e. remove params from previous iterations so that the coming transforms are different
         coord = self.indexes[index]
         seg_pair, _ = copy.deepcopy(self.handlers[coord['handler_index']])
 
@@ -936,11 +940,8 @@ class MRI3DSubVolumeSegmentationDataset(Dataset):
                 seg_pair['gt'][idx_class] = seg_pair['gt'][idx_class][idx_rater]
                 seg_pair['gt_metadata'][idx_class] = seg_pair['gt_metadata'][idx_class][idx_rater]
 
-        # Clean transforms params from previous transforms
-        # i.e. remove params from previous iterations so that the coming transforms are different
-        # Use copy to have different coordinates for reconstruction for a given handler
-        metadata_input = imed_loader_utils.clean_metadata(seg_pair['input_metadata'])
-        metadata_gt = imed_loader_utils.clean_metadata(seg_pair['gt_metadata'])
+        metadata_input = seg_pair['input_metadata']
+        metadata_gt = seg_pair['gt_metadata']
 
         # Run transforms on images
         stack_input, metadata_input = self.transform(sample=seg_pair['input'],
--- a/ivadomed/loader/loader.py
+++ b/ivadomed/loader/loader.py
@@ -719,9 +719,9 @@ class MRI2DSegmentationDataset(Dataset):
 
         if self.is_2d_patch:
             coord = self.indexes[index]
-            seg_pair_slice, roi_pair_slice = self.handlers[coord['handler_index']]
+            seg_pair_slice, roi_pair_slice = copy.deepcopy(self.handlers[coord['handler_index']])
         else:
-            seg_pair_slice, roi_pair_slice = self.indexes[index]
+            seg_pair_slice, roi_pair_slice = copy.deepcopy(self.indexes[index])
 
         # In case multiple raters
         if seg_pair_slice['gt'] is not None and isinstance(seg_pair_slice['gt'][0], list):
@@ -736,9 +736,9 @@ class MRI2DSegmentationDataset(Dataset):
         # Clean transforms params from previous transforms
         # i.e. remove params from previous iterations so that the coming transforms are different
         # Use copy to have different coordinates for reconstruction for a given handler with patch
-        metadata_input = imed_loader_utils.clean_metadata(copy.deepcopy(seg_pair_slice['input_metadata']))
-        metadata_roi = imed_loader_utils.clean_metadata(copy.deepcopy(roi_pair_slice['gt_metadata']))
-        metadata_gt = imed_loader_utils.clean_metadata(copy.deepcopy(seg_pair_slice['gt_metadata']))
+        metadata_input = imed_loader_utils.clean_metadata(seg_pair_slice['input_metadata'])
+        metadata_roi = imed_loader_utils.clean_metadata(roi_pair_slice['gt_metadata'])
+        metadata_gt = imed_loader_utils.clean_metadata(seg_pair_slice['gt_metadata'])
 
         # Run transforms on ROI
         # ROI goes first because params of ROICrop are needed for the followings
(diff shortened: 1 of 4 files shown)


ivadomed/main.py
```diff
@@ -364,7 +364,7 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
         imed_utils.display_selected_transfoms(transformation_dict, dataset_type=["testing"])
 
     # Check if multiple raters
-    check_multiple_raters(command != "train", loader_params)
+    check_multiple_raters(command == "train", loader_params)
```
charleygros, line 367: Ahah.. .. yes thanks!! .. !!

testing/unit_tests/test_loader.py
```diff
@@ -244,14 +243,51 @@ def test_get_target_filename_list(loader_parameters, model_parameters, transform
...
-
+                                       **{**loader_parameters, **{'data_list': data_lst,
+                                                                  'transforms_params': transform_parameters,
+                                                                  'dataset_type': 'training'}})
     target_filename = test_ds.filename_pairs[0][1]
     
     assert len(target_filename) == len(loader_parameters["target_suffix"])
 
 
+@pytest.mark.parametrize('loader_parameters', [{
```
charleygros, line 254: Reviewed but not tested, I trust you on this one @mariehbourget, except if you think important that someone else test it, 'd be happy to: lemme know :-)

ivadomed/loader/loader.py
```diff
@@ -317,7 +317,7 @@ def get_pair_data(self):
                     hwd_oriented_list = [
                         imed_loader_utils.orient_img_hwd(self.get_data(gt_rater, cache_mode),
                                                          self.slice_axis) for gt_rater in gt]
-                    gt_data.append([hwd_oriented.astype(data_type) for hwd_oriented in hwd_oriented_list])
+                    gt_data.append([hwd_oriented for hwd_oriented in hwd_oriented_list])
```
charleygros, line 320: Good call!!

charleygros (approved):

charleygros:
> I did not change the feature, only the "trigger" (when command `train` instead of `test`), so the feature should work as it was before the refactoring of the main.

Fantastic

> From my understanding, a random rater should be chosen at each epoch of the training, is that correct?

Perfectly correct

> I tested it and [here](https://github.com/ivadomed/ivadomed/blob/fa4215246844c11af86a7c5035c2c829c65034e7/ivadomed/loader/loader.py#L727), the `if block` is only executed on the first epoch. On other epochs, the `seg_pair_slice['gt'][0]` is a `numpy array` and not a `list`.

Mmm .. yeah..! Lemme come back to you shortly

charleygros:
@mariehbourget: you were very right, as pretty much always it seems like!

I've pushed a quick fix, which shooooould address the problem. Lemme know your thoughts on that if you can please :-)

It has quite an impact on the ongoing study we are having on multiple raters: means that we have to re run all the experiments associated with this method. Maybe a good surprise down the road? Let's see

Thanks again!!

charleygros:
Now testing if this fix is not damaging the `clean_metadata`

charleygros:
> 
> 
> Now testing if this fix is not damaging the `clean_metadata`

I thiiiink it alright. But double check from others would be great!

mariehbourget:
> I've pushed a quick fix, which shooooould address the problem. Lemme know your thoughts on that if you can please :-)

Thanks @charleygros! I'll test it again when able. This may take a moment, as I know nothing about `clean_metadata` yet.

> It has quite an impact on the ongoing study we are having on multiple raters: means that we have to re run all the experiments associated with this method.

I'm curious, how were you able to use the multiple raters feature on the ongoing study with the wrong flag in `check_mulitple_raters`?

charleygros:
> I'm curious, how were you able to use the multiple raters feature on the ongoing study with the wrong flag in `check_mulitple_raters`?

To make our experiments reproducible etc. we are running all our experiemnts with the same ivadomed version --> an older version before the main refactoring --> this error wasn't present :-)

> I'll test it again when able.

Thank you so much!!

> This may take a moment, as I know nothing about clean_metadata yet.

No worries. In short: during the DA we are adding some keys to the metadata, eg translation of 13% along X direction, so that when we loop across the modalities we are applying the same transformation. This function, `clean_medata` is here to flush these DA metadata at the start of each new epoch, so that we will apply a new random DA different from the previous epoch, see eg [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/transforms.py#L699).

It should be fine.. but an extra pair of eye would be great before merging this PR :-)

ivadomed/loader/loader.py
```diff
@@ -736,9 +736,9 @@ def __getitem__(self, index):
         # Clean transforms params from previous transforms
         # i.e. remove params from previous iterations so that the coming transforms are different
         # Use copy to have different coordinates for reconstruction for a given handler with patch
-        metadata_input = imed_loader_utils.clean_metadata(copy.deepcopy(seg_pair_slice['input_metadata']))
-        metadata_roi = imed_loader_utils.clean_metadata(copy.deepcopy(roi_pair_slice['gt_metadata']))
-        metadata_gt = imed_loader_utils.clean_metadata(copy.deepcopy(seg_pair_slice['gt_metadata']))
```
mariehbourget, line 741: I had trouble understanding what `clean_metadata` was supposed to do because with the `copy.deepcopy`, it seems that "old" transforms from a previous epoch are not there anyway. My understanding is that if we keep `deepcopy` (here or at line 724) `clean_metadata` is not necessary, but this would need to be confirm.

As a reference, I was the one who add the `copy.deepcopy` in 2D in PR #763  to be able to keep the patch coordinates (same as what was already in 3D for sub-volumes).

mariehbourget (commented):
Thanks for the explanation, that helped at lot!

So 2 things:
- I think maybe `clean_metadata` is not necessary because of `deepcopy` (see comments for details)
- The rater is now randomly selected at each epoch, but also for each patch or each sub-volumes, this may not be what we want

charleygros:
>     * The rater is now randomly selected at each epoch, but also for each patch or each sub-volumes, this may not be what we want

I'm okay with this, maxi randomisation :-)

For the clean_metadata: That's an excellent point. I'll think about it and come back to you by Monday if that s okay

ivadomed/loader/loader.py
charleygros: I agree with you @mariehbourget : I think we can get ride of `clean_metadata`!

ivadomed/loader/loader.py
```diff
@@ -926,7 +926,7 @@ def __getitem__(self, index):
             index (int): Subvolume index.
         """
         coord = self.indexes[index]
-        seg_pair, _ = self.handlers[coord['handler_index']]
+        seg_pair, _ = copy.deepcopy(self.handlers[coord['handler_index']])
```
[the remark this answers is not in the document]
charleygros: Can you please help me to further understand your point here? Cheers

ivadomed/loader/loader.py
```diff
@@ -736,9 +736,9 @@ def __getitem__(self, index):
         # Clean transforms params from previous transforms
         # i.e. remove params from previous iterations so that the coming transforms are different
         # Use copy to have different coordinates for reconstruction for a given handler with patch
```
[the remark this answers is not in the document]
mariehbourget: Comment updated in d04f21a

ivadomed/loader/loader.py
```diff
@@ -941,8 +941,8 @@ def __getitem__(self, index):
         # Clean transforms params from previous transforms
         # i.e. remove params from previous iterations so that the coming transforms are different
         # Use copy to have different coordinates for reconstruction for a given handler
```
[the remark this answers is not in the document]
mariehbourget: Comment updated in 11ed8e5

ivadomed/loader/loader.py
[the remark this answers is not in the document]
mariehbourget: I thought we wanted a single rater per image/volume and not per patch/sub-volumes. Thinking about it, that is not really a problem, as you said, maximum randomization. I'm ok with that.

mariehbourget:
@charleygros
Ok so I removed `clean_metadata`, which works for training but fails for `test_segment_volume` with this error:
```
ivadomed/loader/loader.py:760: in __getitem__
    stack_gt, metadata_gt = self.transform(sample=seg_pair_slice["gt"],
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ivadomed.transforms.Compose object at 0x7f1f9af3fdf0>, sample = None, metadata = None, data_type = 'gt'

    def __call__(self, sample, metadata, data_type='im'):
        print("print metadata:", metadata)
>       if self.transform[data_type] is None or len(metadata) == 0:
E       TypeError: object of type 'NoneType' has no len()

ivadomed/transforms.py:147: TypeError
```

I think this is because `clean_metadata` was returning an empty list, and now the metadata is just `None`.

We could either assign an empty list to the metadata like this in the loader [here](https://github.com/ivadomed/ivadomed/blob/32b911b335fcb36219be90c44c6700ccfa3989f4/ivadomed/loader/loader.py#L737):
`metadata_input = seg_pair_slice['input_metadata'] if seg_pair_slice['input_metadata'] is not None else []`

or change the condition in the call of Compose in transforms.py [here](https://github.com/ivadomed/ivadomed/blob/32b911b335fcb36219be90c44c6700ccfa3989f4/ivadomed/transforms.py#L146):
from `if self.transform[data_type] is None or len(metadata) == 0:` to `if self.transform[data_type] is None or not metadata:`

I'm not sure what are the implications of this and which solutions would be best. Let me know what you think.

charleygros:
Thanks heaps for looking into this @mariehbourget !

Mmm probably a consequence of the refactoring we did some time ago between `None` and `[]`.

I'd definitely recommend using the first option your suggested, ie in the loader! Cheers

mariehbourget:
> I'd definitely recommend using the first option your suggested, ie in the loader! Cheers

I made the change as discussed. All tests are passing. Let me know if this is ok to merge.

I found another issue with the feature with the command segment. I'll open a separate issue since it is not directly related to the loader and this PR.

charleygros:
> I made the change as discussed. All tests are passing. Let me know if this is ok to merge.

I tried it and works beautifully! Thanks heaps!!

> I found another issue with the feature with the command segment. I'll open a separate issue since it is not directly related to the loader and this PR (issue #816)

:crown:

PR #830, opened 2021-06-23
Skip all NumpyToTensor transformation for retrocompatibility

### LINKED ISSUE #829: segment_volume fails with past models containing NumpyToTensor key
## Issue description

The function `segment_volume` fails after the merge of #818 when segmenting with a model containing the `"NumpyToTensor": {}` field in transformation parameters. (tagging @dyt811 and @cakester so they are aware)

**I found this https://github.com/ivadomed/ivadomed/pull/758#discussion_r654584695 on the opened PR #758 with what seems to be a fix for this. If that is the case, I think the fix should be done in a dedicated PR as soon as possible.**

### Current behavior
Config file of models trained prior to the merge of #818 contains the `"NumpyToTensor": {}` field. This config file is used by the function `segment_volume` for segmentation.

The segmentation fails with the following error when undoing transforms:

```
  File "/home/mhbourget/ivadomed/ivadomed/main.py", line 560, in run_main
    run_command(context=context,
  File "/home/mhbourget/ivadomed/ivadomed/main.py", line 340, in run_command
    run_segment_command(context, model_params)
  File "/home/mhbourget/ivadomed/ivadomed/main.py", line 274, in run_segment_command
    pred_list, target_list = imed_inference.segment_volume(path_model,
  File "/home/mhbourget/ivadomed/ivadomed/inference.py", line 407, in segment_volume
    pred_list, target_list, last_sample_bool, weight_matrix, volume, image = reconstruct_3d_object(
  File "/home/mhbourget/ivadomed/ivadomed/inference.py", line 483, in reconstruct_3d_object
    image_reconstruction(batch, preds, undo_transforms, i_slice, image, weight_matrix)
  File "/home/mhbourget/ivadomed/ivadomed/inference.py", line 609, in image_reconstruction
    pred_undo, metadata = undo_transforms(image, batch['gt_metadata'][smp_idx], data_type='gt')
  File "/home/mhbourget/ivadomed/ivadomed/transforms.py", line 184, in __call__
    sample, metadata = tr.undo_transform(sample, metadata)
  File "/home/mhbourget/ivadomed/ivadomed/transforms.py", line 210, in undo_transform
    return list(sample.numpy()), metadata
AttributeError: 'list' object has no attribute 'numpy'
```

Note: a newly trained model without the `NumpyToTensor` key does not produced this error.

### Expected behavior

- `"NumpyToTensor": {}` field from old config files should be ignored somehow to ensure retrocompatibility and avoid this error.
- `functional_tests/test_segment_volume.py` should test for this. This is not the case right now and tests are passing because they don't use old config files.

cakester:
#### GitHub

#### PR contents

## Description

https://github.com/ivadomed/ivadomed/pull/830/files#diff-d44e0c652f133445ff5578a458ec6d510cea45e4d4038d9ecfd06c1a700ba3f0R121 this line fixes another bug that adds NumpyToTensor in the list of transforms. The rest resolves #829 

## Linked issues

resolves #829

### DIFF
--- a/ivadomed/transforms.py
+++ b/ivadomed/transforms.py
@@ -150,8 +150,6 @@ class Compose(object):
             return None, None
         else:
             for tr in self.transform[data_type].transforms:
-                if isinstance(tr, NumpyToTensor):
-                    continue
                 sample, metadata = tr(sample, metadata)
 
             if not preprocessing:
@@ -183,8 +181,6 @@ class UndoCompose(object):
             numpy_to_tensor = NumpyToTensor()
             sample, metadata = numpy_to_tensor.undo_transform(sample, metadata)
             for tr in self.transforms.transform[data_type].transforms[::-1]:
-                if isinstance(tr, NumpyToTensor):
-                    continue
                 sample, metadata = tr.undo_transform(sample, metadata)
             return sample, metadata
 
--- a/ivadomed/transforms.py
+++ b/ivadomed/transforms.py
@@ -118,6 +118,8 @@ class Compose(object):
 
             # call transform
             if transform in globals():
+                if transform == "NumpyToTensor":
+                    continue
                 params_cur = {k: parameters[k] for k in parameters if k != "applied_to" and k != "preprocessing"}
                 transform_obj = globals()[transform](**params_cur)
             else:
@@ -148,6 +150,8 @@ class Compose(object):
             return None, None
         else:
             for tr in self.transform[data_type].transforms:
+                if isinstance(tr, NumpyToTensor):
+                    continue
                 sample, metadata = tr(sample, metadata)
 
             if not preprocessing:
@@ -179,6 +183,8 @@ class UndoCompose(object):
             numpy_to_tensor = NumpyToTensor()
             sample, metadata = numpy_to_tensor.undo_transform(sample, metadata)
             for tr in self.transforms.transform[data_type].transforms[::-1]:
+                if isinstance(tr, NumpyToTensor):
+                    continue
                 sample, metadata = tr.undo_transform(sample, metadata)
             return sample, metadata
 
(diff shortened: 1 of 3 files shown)


testing/functional_tests/test_segment_volume.py
```diff
@@ -28,6 +28,59 @@ def setup_function():
...
+
+    torch.save(model, os.path.join(PATH_MODEL, "model_test.pt"))
+    config = {
+        "loader_parameters": {
+            "slice_filter_params": {
+                "filter_empty_mask": False,
+                "filter_empty_input": False
+            },
+            "roi_params": {
+                "suffix": "_seg-manual",
```
mariehbourget, line 49:
```suggestion
                "suffix": None,
```

mariehbourget (commented):
Thank you @cakester for looking into this. I tested the PR and `segment_volume` is working as expected with an old model.

I would recommend to add comments at the 3 places you apply the fix, so we have a trace for future developers.

But I may have another suggestion. Would it be possible/advisable to remove the key when we load the context [here](https://github.com/ivadomed/ivadomed/blob/3e3989d408e5cc85cbc50354dc13e1b460648dbc/ivadomed/inference.py#L320) at the beginning of `segment_volume`? We've done something similar in the past for retrocompatibility with old config files by replacing certain key and/or values. If it's possible, I think we would not need to apply the fix in transforms.

ivadomed/transforms.py
```diff
@@ -118,6 +118,8 @@ def __init__(self, dict_transforms, requires_undo=False):
 
             # call transform
             if transform in globals():
+                if transform == "NumpyToTensor":
```
cakester, line 121: note this fixes another bug but it needs to go in

mariehbourget: Could you perhaps add information on the bug it fixes in the PR description if it is not the same as #829. Thanks!

mariehbourget (commented):
Tested again with the change in config manager, works as expected, thanks!

ivadomed/transforms.py
cakester: done

mariehbourget (approved):

PR #832, opened 2021-06-25
Unsilence test_rbg

### LINKED ISSUE #720: Un-silence a unit test: test_rgb.py
### Context

While working on https://github.com/ivadomed/ivadomed/pull/714, I had to silence the function `save_color_labels` because the way it was implemented produced an error when the header of the reference image was copied to the output prediction. Namely, the field "dtype" was not compatible, yielding the following error when saving the  NIFTI file:
~~~
>           raise WriterError('Cannot cast to or from non-numeric types')
E           nibabel.arraywriters.WriterError: Cannot cast to or from non-numeric types
~~~

### TODO

- Fix the implementation of `save_color_labels`
- Reestablish the calls to this function, which were removed in https://github.com/ivadomed/ivadomed/pull/714/commits/52079bd339b6e0cdffa61b778da2bae9a3560e55
- I had to silence a test (see this commit https://github.com/ivadomed/ivadomed/pull/714/commits/0f946fc3fe4b083d33783fe96a4be5e9ed207558), which was lower priority compared to the PR. This silenced test should be revisited.

cakester:
#### GitHub

#### PR contents

## Description

note: need to get rgb_test_file.nii.gz inside data_testing repo first

## Linked issues

resolves #720

### DIFF
--- a/ivadomed/scripts/download_data.py
+++ b/ivadomed/scripts/download_data.py
@@ -22,7 +22,7 @@ DICT_URL = {
             `Spine Generic <https://github.com/spine-generic/data-multi-subject>`_.
             Used for Tutorial and example in Ivadomed."""},
     "data_testing": {
-        "url": ["https://github.com/ivadomed/data-testing/archive/r20210322.zip"],
+        "url": ["https://github.com/ivadomed/data-testing/archive/r20210628.zip"],
         "description": "Data Used for integration/unit test in Ivadomed."},
     "t2_tumor": {
         "url": ["https://github.com/ivadomed/t2_tumor/archive/r20200621.zip"],
--- a/testing/unit_tests/test_rgb.py
+++ b/testing/unit_tests/test_rgb.py
@@ -2,7 +2,7 @@ from ivadomed import visualize as imed_visualize
 import numpy as np
 import torch
 import os
-from testing.unit_tests.t_utils import create_tmp_dir,  __data_testing_dir__, __tmp_dir__
+from testing.unit_tests.t_utils import create_tmp_dir,  __data_testing_dir__, __tmp_dir__, download_data_testing_test_files
 from testing.common_testing_util import remove_tmp_dir
 
 def setup_function():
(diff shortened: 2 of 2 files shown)


cakester:
@jcohenadad can you approve this https://github.com/ivadomed/data-testing/pull/14 for this pr? thanks

jcohenadad:
> @jcohenadad can you approve this [ivadomed/data-testing#14](https://github.com/ivadomed/data-testing/pull/14) for this pr? thanks

done 😊

dyt811:
The RGB_test.nii.gz file looks good. @cakester, FYI, you can use MRICRON.exe to check. 

image

Also, make sure to turn off smoothing... to look at real voxel data
image

testing/unit_tests/test_rgb.py
```diff
@@ -2,33 +2,30 @@
...
+        image[0][2][5][i] = 1
+        image[1][2][5][i+3] = 1
+        image[2][2][5][i + 6] = 1
+    image_n = np.array(image)
+    imed_visualize.save_color_labels(
+        gt_data=image_n,
+        binarize=False,
+        gt_filename=os.path.join(
+            __data_testing_dir__,
+            "rgb_test_file.nii.gz"),
```
dyt811, line 25: This input reference image I presume already has the correct header. Could you maybe briefly elaborate a bit about how that was created initially for record purposes? For the purpose of the test of "saving" functionalities, it is more than sufficient. From a test perspective, we achieved the goal if restoring the earlier test and ensure the `save_color_labels` are correctly saving the files inputs are correct.

ivadomed/scripts/download_data.py
```diff
@@ -22,7 +22,7 @@
             `Spine Generic <https://github.com/spine-generic/data-multi-subject>`_.
             Used for Tutorial and example in Ivadomed."""},
     "data_testing": {
-        "url": ["https://github.com/ivadomed/data-testing/archive/r20210322.zip"],
+        "url": ["https://github.com/ivadomed/data-testing/archive/r20210628.zip"],
```
dyt811, line 25: 👍

dyt811 (approved):
While I am not 100% familiar with RGB labelling in nifti and ground truth labelling, from a unit test perspective, this restored the intended goals of checking the `saving` functionality. However, the edge cases for incompatible input (such as the old ground truth file referenced), will not work and would likely require additional unit test cases to catch.  

I was a bit surprised that this function is not called anywhere else it seems. 
image except some alternative commented out path during testing phase. So we might need to revaluate its important down the road. For now, it is good enough.

testing/unit_tests/test_rgb.py
cakester: this new image file is created by using save_color_labels function without providing a header for the nifti image, I am just using it again to copy the header

Issue #823, opened 2021-06-09
Prediction volume missing one slice with command test


mariehbourget:
## Issue description
While working on PR #813 and more generally on inference for microscopy #306, I found something odd in the code in the `run_inference` function of `testing.py`, and I would like a second opinion about that before making changes (@andreanne-lemay @charleygros)

In brief, the prediction volume created and saved into a nifti file is missing the last slice.

### Current behavior
When running the command `test`, in the prediction [for loop](https://github.com/ivadomed/ivadomed/blob/1d81b32c61891e4b5b4f2ffbbb7923e367111dc9/ivadomed/testing.py#L178), predictions for each slices are added to `pred_tmp_lst` at line [253](https://github.com/ivadomed/ivadomed/blob/1d81b32c61891e4b5b4f2ffbbb7923e367111dc9/ivadomed/testing.py#L253). 

However, when running the loop for the last sample of a volume (`last_sample_bool=True`), the "new complete volume" is treated before that at line [213](https://github.com/ivadomed/ivadomed/blob/1d81b32c61891e4b5b4f2ffbbb7923e367111dc9/ivadomed/testing.py#L213). So for example, with a volume of 16 slices, the `pred_tmp_lst` at this stage has a length of 15 (missing the last slice).
**EDIT**: for MRI, if there are more than one volume to segment this only happens for the last volume of the batch because of `last_sample_bool=True`.

The resulting nifti file saved in `pred_to_nib` and the `output_data` [here](https://github.com/ivadomed/ivadomed/blob/1d81b32c61891e4b5b4f2ffbbb7923e367111dc9/ivadomed/testing.py#L225) have 16 slices, but the last one is empty (seems to be filled by zeros in pred_to_nib [here](https://github.com/ivadomed/ivadomed/blob/1d81b32c61891e4b5b4f2ffbbb7923e367111dc9/ivadomed/inference.py#L146))

### Expected behavior
I would have expected the append to `pred_tmp_lst` to happen before "new complete volume" to include the last slice predictions.

For example in 2D microscopy, I have only 1 slice, so 1 iteration of the for loop and my `pred_tmp_lst` is empty as it gets to "new complete volume" before the append.
**EDIT**: for microscopy, there is only one slice per volume. So the behavior above happens if I have only one volume to segment. If I try segmenting 2 volumes, the first one is OK, but the second one doens't work because of `last_sample_bool=True`. 

[...]

PR #834, opened 2021-06-25
Add the inference steps for PNG/TIF microscopy data

### LINKED ISSUE #833: Implement inference steps for microscopy PNG/TIF data
### Context
This issue is part of the inference refactoring project to include microscopy data (#306).
It aims to implement the inference step for microscopy PNG/TIF data.

### TODO
- Retrieve `PixelSize` metadata from bids_dataframe for command `segment`.
- Add the `PixelSize` metadata field to `options` in `segment_volume`.
- Add an `overlap_2D` metadata field to `options` in `segment_volume`.
- Implement a function to write PNG of predictions when the source file is not NifTI:
    - with command `test` in `testing.py/run_inference`
    - with command `segment` in `main.py/run_segment_command`
- Implement a function to write PNG of "painted" predictions when the source file is not NifTI with command `test` in `evaluation.py/evaluate`.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR implements the last steps for inference on microscopy PNG/TIFF file.
- Addition of `PixelSize` metadata to `run_segment_command` and `segment_volume`
- Addition of `overlap_2D` metadata to `segment_volume`
- New function `pred_to_png` in `inference.py` to write PNG from predictions when the source files are not NifTI
- This new function in called in `main,py/run_segment_command`, `testing.py/run_inference` and `evaluation.py/evaluate` covering both the commands `test` and `segment`.
- `test_testing.py` runs inference on 2d microscopy data and checks if PNG are written properly. ~~**Note: This test won't pass until the bug fix in PR #835 is merged.**~~ EDIT: #835 is now merged, the test should pass.

## Linked issues
Fixes #833


mariehbourget:
This PR is ready for review.

**To test the PR:**
- Run a short training (CPU friendly) with `config_new_loader.json` and the [data_example_microscopy_sem](https://github.com/ivadomed/data_example_microscopy_sem) dataset.
- Modify the command in config to `test`and run the test.
- In `output_path/pred_masks`, there should be 1 `_pred.nii.gz` and 1 `_pred_painted.nii.gz` file for the axon+myelin prediction of subject `sub-rat3_sample-data10_SEM`, 2 `_pred.png` files (1 for axon and 1 for myelin) and `2 _pred_painted.png` files. PNG files should look similar to the following:

**sub-rat3_sample-data10_SEM_seg-axon-manual_pred.png**
sub-rat3_sample-data10_SEM_seg-axon-manual_pred

**sub-rat3_sample-data10_SEM_seg-axon-manual_pred_painted.png**
sub-rat3_sample-data10_SEM_seg-axon-manual_pred_painted

**sub-rat3_sample-data10_SEM_seg-myelin-manual_pred.png**
sub-rat3_sample-data10_SEM_seg-myelin-manual_pred

**sub-rat3_sample-data10_SEM_seg-myelin-manual_pred_painted.png**
sub-rat3_sample-data10_SEM_seg-myelin-manual_pred_painted

- Note: the `test_fraction` in the config file is 0.2, there should be `preds_mask` for 2 subjects after the test but there will be only one because of another issue (#823 ) with inference, this is normal until PR #835 is merged. Once it is merged, the `pred_masks` folder will also contains NifTI and PNG predictions for subject `sub-rat3_sample-data9_SEM`. This is also why the test `test_inference_2d_microscopy` for `test_testing.py` is currently failing.

- Finally, modify the command in config to `segment`. This will create 1 `_pred.nii.gz` and 1 `_pred.png` for each subject and `target_suffix` (axon and myelin) of the dataset.

ivadomed/main.py
```diff
@@ -276,10 +280,18 @@ def run_segment_command(context, model_params):
                 os.makedirs(pred_path)
 
             for pred, target in zip(pred_list, target_list):
-                filename = subject.split('.')[0] + target + "_pred" + \
-                           ".nii.gz"
+                filename = subject.split('.')[0] + target + "_pred" + ".nii.gz"
                 nib.save(pred, os.path.join(pred_path, filename))
 
+            # Get subject file extension and write PNG if not NifTI
+            # For Microscopy PNG/TIF files
```
joshuacwnewton, line 287: Very minor nitpick: These commends might be a little self-explanatory (the code itself describes almost exactly this).

ivadomed/inference.py
```diff
@@ -299,9 +319,9 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
...
             images to segment, e.i., len(fname_images) > 1.
         gpu_id (int): Number representing gpu number if available. Currently does NOT support multiple GPU segmentation.
-        options (dict): Contains postprocessing steps and prior filename (fname_prior) which is an image filename
-            (e.g., .nii.gz) containing processing information (e.i., spinal cord segmentation, spinal location or MS
-            lesion classification)
+        options (dict): Contains postprocessing steps, film metadata, microscopy PixelSize and overlap_2D metadata, and
+            prior filename (fname_prior) which is an image filename (e.g., .nii.gz) containing processing information
+            (e.g., spinal cord segmentation, spinal location or MS lesion classification).
             e.g., spinal cord centerline, used to crop the image prior to segment it if provided.
             The segmentation is not performed on the slices that are empty in this image.
```
joshuacwnewton, line 326: I did a quick search for `options (dict)` and it looks like we use the same argument in multiple functions in `inference.py`.

Do we need to keep this description consistent across all of the functions?

joshuacwnewton: Also, this description is starting to feel a little unclear to me...

We have a text description (`"Contains postprocessing steps, film metadata, microscopy PixelSize and overlap_2D metadata, and prior filename (fname_prior) which is an image filename"`), but it doesn't tell me explicitly which dict keys are supported (or required, even). (`PixelSize`, `fill_holes`, `keep_largest`, etc.)

I'm wondering if we could rewrite the text description into a list (just a rough sketch, probably inaccurate):

```python
        options (dict): This can optionally contain any of the following key-value pairs:
            * 'fill_holes': (bool) <description for this postprocessing step>
            * 'keep_largest': (bool) <and so on, etc.>
            * 'overlap_2D': (bool)
            * 'PixelSize': (float) <microscopy PixelSize>
            * 'fname_prior': An image filename (e.g., .nii.gz) containing processing information
                (e.g., spinal cord segmentation, spinal location or MS lesion classification).
                e.g., spinal cord centerline, used to crop the image prior to segment it if provided.
                The segmentation is not performed on the slices that are empty in this image.
```

Alternatively, is `options` documented anywhere else? I guess my main thinking here is, how does the user discover which options are supported?

joshuacwnewton (commented):
We're finally near the end!

For this first pass, most of my comments are just nitpicks and small documentation suggestions (because there's not much else to comment on! It mostly looks good. :smile:)

Note: I do still need to set up the actual test procedure from https://github.com/ivadomed/ivadomed/pull/834#issuecomment-870064078 on my local machine, but I wanted to get a first review in today at the very least.

ivadomed/evaluation.py
```diff
@@ -88,6 +93,16 @@ def evaluate(bids_df, path_output, target_suffix, eval_params):
...
+        # Write painted PNG if subject file is not NifTI
+        # For Microscopy PNG/TIF files
+        # TODO: implement OMETIFF behavior (elif "ome" in extension)
+        if "nii" not in extension:
+            painted_list = imed_inference.split_classes(nib_painted)
+            imed_inference.pred_to_png(painted_list,
+                                       target_suffix,
+                                       os.path.join(path_preds, subj_acq),
+                                       painted=True)
+
```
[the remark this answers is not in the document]
joshuacwnewton: I also noticed that for the test suite, only `test_testing.py` was modified, so that also leads me to believe that `testing.py` is the right place for the png-generating code.

ivadomed/inference.py
```diff
@@ -203,6 +204,25 @@ def pred_to_nib(data_lst: List[np.ndarray], z_lst: List[int], fname_ref: str, fn
     return nib_pred
 
 
+def pred_to_png(pred_list: list, target_list: list, subj_path: str, painted: bool = False):
+    """Save the network predictions as PNG files with suffix "_target_pred".
+
+    Args:
+        pred_list (list of np arrays): list of 2D predictions.
+        target_list (list of str): list of target suffixes.
```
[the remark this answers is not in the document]
mariehbourget: There is kind of a reason behind this in the current codebase. From what I gathered, the list is called `target_suffix` when it comes directly from the `target_suffix` key in the config file (this is the case in `evaluation.py/evaluate` and `testing.py/run_inference`). And the list is called `target_list` when it comes from the return of `inference.py/segment_volume` (this is the case in `main.py/run_segment_command`).

So for the `pred_to_png` function, I'm dealing with both and I choose `target_list` alongside `pred_list`. This makes more sene to me, knowing that the `segment_volume` function and the `pred_to_png` functions will be used together for the future integration with ADS.

ivadomed/evaluation.py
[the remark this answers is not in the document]
mariehbourget: The 2 calls for the function write different files. The first one in `test` writes the prediction from the network to PNG (`_pred`). In evaluate, the prediction is compared to the ground-truth and "painted" according to true positives, false positives and false negatives. So the second call is to write PNG of these painted images (`_pred_painted`).

As for the test, I only wanted to test the function once, and it was better suited with the already existing 2D test in `test_testing.py`.

ivadomed/inference.py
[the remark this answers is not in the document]
joshuacwnewton: Thank you for the thorough explanation. :smile: 

I guess what caught me off guard is that I saw "target_list" and on first glance, I thought it was a list of target images (rather than just the suffixes). Hm... would `target_suffix_list` work, by chance?

If not, this is a nitpicky thing anyway, so NBD to just leave it as-is, too.

ivadomed/evaluation.py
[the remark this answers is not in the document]
joshuacwnewton: > The 2 calls for the function write different files. 

Oh! My apologies, this is exactly what I was missing. :facepalm:

ivadomed/inference.py
```diff
@@ -203,6 +204,25 @@ def pred_to_nib(data_lst: List[np.ndarray], z_lst: List[int], fname_ref: str, fn
...
+        subj_path (str): Path of the subject filename in output folder without extension
+            (e.g. "path_output/pred_masks/sub-01_sample-01_SEM").
+        painted (bool): If True, the suffix "_painted" is appended to the filename.
```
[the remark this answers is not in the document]
mariehbourget: Good idea, I fixed it in 7821f6e.

ivadomed/main.py
mariehbourget: You are right, I updated the comment to only 1 line in 64fd62a. I only kept the info about which microscopy formats are supported and which are not (in relation to the microscopy BEP), this will help when implementing OMETIFF at some point.

ivadomed/inference.py
joshuacwnewton: Oh, 5010993 looks fantastic! Thank you for taking the time to write such thorough documentation. :heart:

mariehbourget: Very good point! Some of those options go beyond my knowledge of ivadomed at this point, but I tried something.
In 5010993, I took your suggestions and completed the descriptions.

For the 4 postprocessing options, the definitions are a mix of the definitions in ivadomed doc [here](https://ivadomed.org/en/latest/configuration_file.html#postprocessing) and sct_deepseg [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b97e0a58c47dafef2cb74a07eff1df6681422b94/spinalcordtoolbox/scripts/sct_deepseg.py#L75).
@charleygros, I would need your help to go over the definitions and following concerns/questions:
1. `binarize_prediction`: SCT definition says 0 for soft-seg, isn't it -1?
2. `fill_holes`: is an int (0 or 1) in SCT, so essentially a bool. Should we describe it as a bool or the same as SCT?
3. `keep_largest`: definitions in SCT and ivadomed are not consistent: keep only largest in ivadomed with no argument vs. keep the number specified in SCT?
4. `remove_small`: I was not able to make it work with the current definition.
5. `metadata`: I have no idea how to describe the film metadata, is it the same as `metadata` from [FiLMedUnet params](https://ivadomed.org/en/latest/configuration_file.html#filmedunet)?

> Do we need to keep this description consistent across all of the function docstrings?

I don't think it's necessary, I updated the docstrings other than segment_volume in d3605b5, I think that is enough. Let me know your thoughts.

> Alternatively, is options documented anywhere else? 

I don't think so. But these options are not visible "from outside" except in the arguments from SCT and the future similar implementation with ADS. I think the updated API reference [here](https://ivadomed.org/en/latest/api_ref.html#ivadomed.inference.segment_volume) will be enough.

mariehbourget: @charleygros, @joshuacwnewton 
I made some more tests and updated the definitions of `options` in `segment_volume`. See updated comment above.
There are some discrepancies between the definition of SCT and ivadomed. It does not affect directly this PR but we may want to address them in the future.
I would be comfortable to merge the PR as is, but I would like to have your approval on the definitions. Thanks!

ivadomed/inference.py
```diff
@@ -299,11 +319,26 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
...
-            e.g., spinal cord centerline, used to crop the image prior to segment it if provided.
-            The segmentation is not performed on the slices that are empty in this image.
+        options (dict): This can optionally contain any of the following key-value pairs:
+            * 'binarize_prediction': (float) Binarize segmentation with specified threshold.
+                                     Predictions below the threshold become 0, and predictions above or equal to
+                                     threshold become 1. Set to -1 for no thresholding (i.e., soft segmentation).
+            * 'binarize_maxpooling': (bool) Binarize by setting to 1 the voxel having the maximum prediction across
+                                     all classes. Useful for multiclass models.
+            * 'fill_holes': (bool) Fill small holes in the segmentation.
+            * 'keep_largest': (bool) Keep the largest connected-object from the output segmentation.
```
charleygros, line 329:
```suggestion
            * 'keep_largest': (bool) Keep the largest connected-object for each class from the output segmentation.
```

ivadomed/inference.py
```diff
@@ -299,11 +319,26 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
...
+                              or one value per prediction class. Single value example: ["1mm3"], ["5vox"]. Multiple values
+                              example: ["10", "20", "10vox"] (remove objects smaller than 10 voxels for class 1 and 3,
+                              and smaller than 20 voxels for class 2).
+            * 'PixelSize': (list of float) List of microscopy PixelSize in micrometers.
+                           Length equals 2 [X, Y] for 2D or 3 [X, Y, Z] for 3D.
+            * 'overlap_2D': (list of int) List of overlaps in pixels for 2D patching. Length equals 2 [X, Y].
+            * 'metadata': (str) Film metadata.
+            * 'fname_prior': (str) An image filename (e.g., .nii.gz) containing processing information
+                (e.g., spinal cord segmentation, spinal location or MS lesion classification).
+                e.g., spinal cord centerline, used to crop the image prior to segment it if provided.
```
charleygros, line 340:
```suggestion
                (e.g., spinal cord segmentation, spinal location or MS lesion classification, spinal cord centerline), 
                used to crop the image prior to segment it if provided.
```

ivadomed/inference.py
charleygros: Looks very good to me @mariehbourget !!

charleygros: And I also agree with.. well All your comments regarding discrepancies between the definition of SCT and ivadomed, will need to be addressed in the future yup.

joshuacwnewton: Apologies for the delay! 

> For the 4 postprocessing options, the definitions are a mix of the definitions in ivadomed doc [here](https://ivadomed.org/en/latest/configuration_file.html#postprocessing) and sct_deepseg [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b97e0a58c47dafef2cb74a07eff1df6681422b94/spinalcordtoolbox/scripts/sct_deepseg.py#L75).

Just to make sure I'm understanding correctly, all of the post-processing related code in both projects will ultimately end up at the following places:

* [`inference.py :: set_postprocessing_options()`](https://github.com/ivadomed/ivadomed/blob/e5e3d83d03b0ef9af180448a0ecd72522bed8511/ivadomed/inference.py#L278-L305)
* [`postprocessing.py :: Postprocessing()`](https://github.com/ivadomed/ivadomed/blob/50216fd5a5b573c99a40714280dc84364f1c5e1e/ivadomed/postprocessing.py#L242-L371)

In that case, I definitely agree with your choice to disregard the SCT definitions. It very much looks like SCT will need to update its documentation to properly match what ivadomed is doing. (Especially since `sct_deepseg` just [passes its arguments as-is](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b97e0a58c47dafef2cb74a07eff1df6681422b94/spinalcordtoolbox/scripts/sct_deepseg.py#L206) to the ivadomed call.)

Other than that, LGTM! :slightly_smiling_face:

mariehbourget: > Just to make sure I'm understanding correctly, all of the post-processing related code in both projects will ultimately end up at the following places:
> 
> * [`inference.py :: set_postprocessing_options()`](https://github.com/ivadomed/ivadomed/blob/e5e3d83d03b0ef9af180448a0ecd72522bed8511/ivadomed/inference.py#L278-L305)
> * [`postprocessing.py :: Postprocessing()`](https://github.com/ivadomed/ivadomed/blob/50216fd5a5b573c99a40714280dc84364f1c5e1e/ivadomed/postprocessing.py#L242-L371)

Yes, that is my understanding as well. Also, as far as I know, outside of `ivadomed` only `sct_deepseg` is using `segment_volume` at the moment. Soon it will also be used by ADS for inference with ivadomed trained models (https://github.com/neuropoly/axondeepseg/issues/523).

mariehbourget:
@charleygros, @joshuacwnewton, thank you for your reviews, I committed the suggested changes and resolved the conflict with master. Let me know if this is good to go. Thanks!

ivadomed/testing.py
```diff
@@ -254,6 +254,13 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
                         #                              fname_pred.split(".nii.gz")[0] + '_color.nii.gz',
                         #                              imed_utils.AXIS_DCT[testing_params['slice_axis']])
 
+                    # For Microscopy PNG/TIF files (TODO: implement OMETIFF behavior)
+                    extension = imed_loader_utils.get_file_extension(fname_ref)
+                    if "nii" not in extension and fname_pred:
+                        output_list = imed_inference.split_classes(output_nii)
+                        imed_inference.pred_to_png(output_list,
+                                                   testing_params['target_suffix'],
+                                                   fname_pred.split("_pred.nii.gz")[0])
```
joshuacwnewton, line 263: One curious thing I noticed in the `test` output was this, though:

```
2021-07-06 13:54:16.147 | INFO     | ivadomed.testing:test:51 - Loading model: new_loader/best_model.pt
Inference - Iteration 0:   0%|          | 0/1 [00:00<?, ?it/s]Lossy conversion from float64 to uint8. Range [0, 1]. Convert image to uint8 prior to saving to suppress this warning.
2021-07-06 13:54:23.388 | WARNING  | ivadomed.testing:run_inference:248 - No color labels saved due to a temporary issue. For more details see:https://github.com/ivadomed/ivadomed/issues/720
Lossy conversion from float64 to uint8. Range [0, 1]. Convert image to uint8 prior to saving to suppress this warning.
2021-07-06 13:54:23.646 | WARNING  | ivadomed.testing:run_inference:248 - No color labels saved due to a temporary issue. For more details see:https://github.com/ivadomed/ivadomed/issues/720
```

(The error message line of code is just a few lines above this, but GH won't let me highlight.)

It's out of scope for this PR, but I realize the linked issue https://github.com/ivadomed/ivadomed/issues/720 was apparently closed a few days ago -- can we update this code?

joshuacwnewton (approved):
Sorry for the delay! I finally got around to testing the procedures as mentioned, and I was able to reproduce the output png files. [sub-rat3_sample-data.zip](https://github.com/ivadomed/ivadomed/files/6771879/sub-rat3_sample-data.zip) 

So, I think we're all good. :slightly_smiling_face:

ivadomed/testing.py
mariehbourget: Good catch! I had not realized that #720 had been addressed. It is indeed out of scope but I realized that the function was using the variable `fname_tmp` which no longer exists and was replaced by `fname_ref`. I made the change in d861fe9.

However, I read a little bit the PR #832 and I'm not sure if the functionality is fixed based on this [comment](https://github.com/ivadomed/ivadomed/pull/832#pullrequestreview-695691581), I tested it by uncommenting `imed_visualize.save_color_labels` and got the error blow so I think it is not fixed (tagging @dyt811 so he can follow-up).
Thanks again!

<details>
<summary> Error  </summary>

```
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 566, in run_main
    run_command(context=context,
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 470, in run_command
    pred_metrics = imed_testing.test(model_params=model_params,
  File "/home/mhbourget/code/ivadomed/ivadomed/testing.py", line 76, in test
    preds_npy, gt_npy = run_inference(test_loader, model, model_params, testing_params, path_3Dpred,
  File "/home/mhbourget/code/ivadomed/ivadomed/testing.py", line 251, in run_inference
    imed_visualize.save_color_labels(np.stack(pred_tmp_lst, -1),
  File "/home/mhbourget/code/ivadomed/ivadomed/visualize.py", line 101, in save_color_labels
    imed_inference.pred_to_nib([multi_labeled_pred], [], gt_filename,
  File "/home/mhbourget/code/ivadomed/ivadomed/inference.py", line 205, in pred_to_nib
    nib.save(nib_pred, fname_out)
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/loadsave.py", line 99, in save
    img.to_filename(filename)
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/filebasedimages.py", line 333, in to_filename
    self.to_file_map()
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/analyze.py", line 1019, in to_file_map
    arr_writer = make_array_writer(data,
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/arraywriters.py", line 790, in make_array_writer
    return SlopeInterArrayWriter(data, out_type, **kwargs)
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/arraywriters.py", line 521, in __init__
    super(SlopeInterArrayWriter, self).__init__(array,
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/arraywriters.py", line 303, in __init__
    self.calc_scale()
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/arraywriters.py", line 351, in calc_scale
    if not self.scaling_needed():
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/arraywriters.py", line 325, in scaling_needed
    if not super(SlopeArrayWriter, self).scaling_needed():
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/nibabel/arraywriters.py", line 125, in scaling_needed
    raise WriterError('Cannot cast to or from non-numeric types')
nibabel.arraywriters.WriterError: Cannot cast to or from non-numeric types
```
</details>

cakester: > Good catch! I had not realized that #720 had been addressed. It is indeed out of scope but I realized that the function was using the variable `fname_tmp` which no longer exists and was replaced by `fname_ref`. I made the change in [d861fe9](https://github.com/ivadomed/ivadomed/commit/d861fe9c25fe6ef2b35f45822ee71e904bbd810a).
> 
> However, I read a little bit the PR #832 and I'm not sure if the functionality is fixed based on this [comment](https://github.com/ivadomed/ivadomed/pull/832#pullrequestreview-695691581), I tested it by uncommenting `imed_visualize.save_color_labels` and got the error below so I think it is not fixed (tagging @dyt811 so he can follow-up).
> Thanks again!
> 
> Error

how did you test save_color_labels? you have to provide a gt_file with the right header (ie rgb data type and etc)

mariehbourget: I was not able to test the function because the gt_file doesn't have the right header, but I changed the variable name from `fname_tmp` to `fname_ref` anyway because this is part of a refactoring I did in #835 (but I had forgotten to change it in the commented code).
So the code for `imed_visualize.save_color_labels` is still commented, the function would need to be adjusted to match the header of our gt_file.

PR #835, opened 2021-06-27
Fix last slice missing in testing bug

### LINKED ISSUE #823: Prediction volume missing one slice with command test

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR fixes a bug in testing where the predictions of the last slice of the last volume of a batch were not appended in the `_pred` NifTI file, and therefore not taken into account in evaluation metrics.

For 2d microscopy, that meant that the last prediction for the last subject was entirely missing in the predictions.

### Changes made in this PR:
- The predictions for each slice are now appended to `pred_tmp_lst` before entering the "new complete volume" block where the reconstruction and saving of the NifTI takes place.
- I removed `fname_tmp` from the loop and the previous trigger for "new complete volume" which was a `fname_ref != fname_tmp` . Because `fname_tmp` had to be assigned after "new complete volume", it would not work with 2d microscopy with only 1 slice per volume.
- I replaced the trigger with `last_slice_bool`, which compare the slice index with the number of slice of each volume. Hence, "new complete volume" is triggered when the prediction is for the last patch of the last slice of each volume, instead of when the filename changes. **My understanding is that it is going to work with all axis (axial, sagittal, coronal) but I only tested with axial, I would appreciate some insights on this.**
- I added a test in `test_testing.py` for 2d microscopy. This test ensures the testing is working with volumes of only one slice and would fail without the fix.

### To test the PR:
**On 3D MRI without patches:**
1. On master branch: Run a 2D model with command `test` and check the last volume resulting nifti file (last slice (top) is all zeros).
2. On this branch: Run the same command, check the last volume resulting nifti file (no missing slice in the prediction).

**On 2D Microscopy with patches:**
1. On master branch, run a short training (CPU friendly) with `config_new_loader.json` and the [data_example_microscopy_sem](https://github.com/ivadomed/data_example_microscopy_sem) dataset.
2. Modify the command in config to `test` and run the test. The test fraction is 0.2 and there should be 2 predictions in the `output_path/preds_mask`. However, there is only 1 because of the bug.
3. On this branch, re-run the command `test`. This time, 2 predictions should be in `preds_mask`.

## Linked issues
Fixes #823

### DIFF
--- a/ivadomed/testing.py
+++ b/ivadomed/testing.py
@@ -182,6 +182,9 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
 
             if model_params["is_2d"]:
                 preds_idx_arr = None
+                idx_slice = batch['input_metadata'][smp_idx][0]['slice_index']
+                n_slices = batch['input_metadata'][smp_idx][0]['data_shape'][-1]
+                last_slice_bool = (idx_slice + 1 == n_slices)
                 last_sample_bool = (last_batch_bool and smp_idx == len(preds_cpu) - 1)
 
                 length_2D = model_params["length_2D"] if "length_2D" in model_params else []
@@ -191,30 +194,34 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
                     preds_idx_undo, metadata_idx, last_patch_bool, image, weight_matrix = \
                         imed_inference.image_reconstruction(batch, preds_cpu, testing_params['undo_transforms'],
                                                             smp_idx, image, weight_matrix)
-                    # If last patch of the slice
-                    if last_patch_bool:
-                        # preds_idx_undo is a list n_label arrays
-                        preds_idx_arr = np.array(preds_idx_undo)
-
-                        # TODO: gt_filenames should not be a list
-                        fname_ref = list(filter(None, metadata_idx[0]['gt_filenames']))[0]
-
                 else:
+                    # Set last_patch_bool to True (only one patch per slice)
+                    last_patch_bool = True
                     # undo transformations for slice
                     preds_idx_undo, metadata_idx = testing_params["undo_transforms"](preds_cpu[smp_idx],
                                                                                      batch['gt_metadata'][smp_idx],
                                                                                      data_type='gt')
+                if last_patch_bool:
                     # preds_idx_undo is a list n_label arrays
                     preds_idx_arr = np.array(preds_idx_undo)
 
                     # TODO: gt_filenames should not be a list
                     fname_ref = list(filter(None, metadata_idx[0]['gt_filenames']))[0]
 
+                if preds_idx_arr is not None:
+                    # add new sample to pred_tmp_lst, of size n_label X h X w ...
+                    pred_tmp_lst.append(preds_idx_arr)
+
+                    # TODO: slice_index should be stored in gt_metadata as well
+                    z_tmp_lst.append(int(idx_slice))
+                    filenames = metadata_idx[0]['gt_filenames']
+
                 # NEW COMPLETE VOLUME
-                if pred_tmp_lst and (fname_ref != fname_tmp or last_sample_bool) and task != "classification":
+                if (pred_tmp_lst and ((last_patch_bool and last_slice_bool) or last_sample_bool)
+                    and task != "classification"):
                     # save the completely processed file as a NifTI file
                     if ofolder:
-                        fname_pred = os.path.join(ofolder, Path(fname_tmp).name)
+                        fname_pred = os.path.join(ofolder, Path(fname_ref).name)
                         fname_pred = fname_pred.rsplit("_", 1)[0] + '_pred.nii.gz'
                         # If Uncertainty running, then we save each simulation result
                         if testing_params['uncertainty']['applied']:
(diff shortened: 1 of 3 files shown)


charleygros:
Changes in code look sensible. Now testing on data.

charleygros (approved):

mariehbourget:
Additional information for the reviewer to reproduce the error on master branch.

- With this config file: [config_test_pr835.zip](https://github.com/ivadomed/ivadomed/files/6726801/config_test_pr835.zip)
- Run the training on `data_example_spinegeneric` (short training, 1 epoch, 1 subject)
- Then change the command to `test` and run the test
- The "last" volume to be processed in testing should be `sub-unf01_T2w`.
- When visualizing this volume in FSLeyes, the last slice (at the top) is empty (all zeros)

**Example, second to last slice (not empty):**
avant_derniere_slice

**Example, last slice at the top empty:**
derniere_slice

- This only happens with `test` and the volume is complete with the command `segment`.

ivadomed/testing.py
```diff
@@ -182,6 +182,9 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
 
             if model_params["is_2d"]:
                 preds_idx_arr = None
+                idx_slice = batch['input_metadata'][smp_idx][0]['slice_index']
+                n_slices = batch['input_metadata'][smp_idx][0]['data_shape'][-1]
```
mariehbourget, line 186: @charleygros, before merging the PR, I would like to confirm with you that this will work for coronal and sagittal axis as well. Here I extract the number of slices based on the last dimension of the `data_shape`. I think it works because the dimension are swapped in orient_shapes_hwd [here](https://github.com/ivadomed/ivadomed/blob/cc16227f1fca40ec4619027563f1f2be513944f7/ivadomed/loader/loader.py#L262) according to the `slice_axis` param. However, I do not have any dataset/model to test it. Let me know your thoughts, thanks!

charleygros: Fully agree with your reasoning! Testing it now

charleygros:
> This only happens with `test` and the volume is complete with the command `segment`.

That would be good to share this chunk of code at some point between the two commands :-(

ivadomed/testing.py
charleygros: All green

charleygros:
> Additional information for the reviewer to reproduce the error on master branch.

🏅

PR #787, opened 2021-05-13
Intervertebral disc labeling using pose estimation

rezazad68:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [ ] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [x] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
In this pull, we aim to reformulate the semantic intervertebral disc labeling using the pose estimation technique. To this end, we utilize the stacked hourglass network. The proposed deep model takes into account the strength of semantic segmentation and pose estimation technique to handle the missing and falsely additional disc areas.

## Linked issues

resolves #539 
resolves #766 

## Run Demo: 
1. Download [Spine Generic Public Database (Multi-Subject)](https://github.com/spine-generic/data-multi-subject) dataset
2. Run (update path to proper location):
   ```
   python prepare_dataset_vertebral_labeling.py --path ~/data-multi-subject/ -s _T2w
   ```

After running the above command, it will generate the mid files. While running it will print the list of subjects that the mid file is generating for them. For example: 
image

3- Run ivadomed --train -c `config_pose.json` --gif 2

PR #852, opened 2021-07-06
Intervertebral disc labeling using pose estimation

### LINKED ISSUE #539: Implement intervertebral disc detection using pose estimation
After implementing the first deep learning approach for detecting discs https://github.com/neuropoly/spinalcordtoolbox/pull/2679, we could investigate the post [estimation method](https://doi.org/10.1038/s41593-018-0209-y):
[mathis2018.pdf](https://github.com/ivadomed/ivadomed/files/5593248/mathis2018.pdf)

### LINKED ISSUE #766: Adding a new method to ivadomed, intervertebral disc labeling
Hi,

Would appreciate your comments and guidance regarding the process for adding a new method to the ivadomed library. I am following the below steps:

1- Add model to the [model.py](https://github.com/ivadomed/ivadomed/blob/a245bc4b1a4e2b8c34311d331b12cb93fa8de77b/ivadomed/models.py) 
2- Add loss function to [losses.py](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/losses.py)
3- Add metric to [metrics.py](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/metrics.py)
3- For training and evaluation, the [main.py](https://github.com/ivadomed/ivadomed/blob/a245bc4b1a4e2b8c34311d331b12cb93fa8de77b/ivadomed/main.py) should revise? 
4- weight files should be stored in which address? 
Any further comments are welcome. Thanks in advance.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

> **Note:** This PR is a continuation of @rezazad68's excellent work in https://github.com/ivadomed/ivadomed/pull/787. _(The commits from the `rezazad68:master` branch were transferred from Reza's fork over to ivadomed in the [`ra/intervertebral-disc-labeling`](https://github.com/ivadomed/ivadomed/commits/ra/intervertebral-disc-labeling) branch, in order to better preserve them. Then, the `jn/539-intervertebral-disc-labeling` branch was created from the same point, and the commits were squashed for clarity. After that, the pending feedback from https://github.com/ivadomed/ivadomed/pull/787 was addressed. Finally, a few bugs were fixed to ensure that the testing steps worked OK.)_

In this pull, we aim to reformulate the semantic intervertebral disc labeling using the pose estimation technique. To this end, we utilize the stacked hourglass network. The proposed deep model takes into account the strength of semantic segmentation and pose estimation technique to handle the missing and falsely additional disc areas.

## Using this PR (`--train`)

1. Download a dataset (e.g. [`data-multi-subject`](https://github.com/spine-generic/data-multi-subject).
2. Apply preprocessing steps to your dataset to generate 2 images per subject (an image with a single slice (mid-sagittal), as well as an image with the intervertebral disc heatmap groundtruth):
    ```
    python ivadomed/scripts/prepare_dataset_vertebral_labeling.py --path ~/data-multi-subject/ -s _T1w
    ```
   <details>
   <summary> Sample output log... </summary>

    **Note:** This was only tested using 7 subjects for debugging purposes,
   
    ```
        Outdated. TODO, JN: Paste up-to-date logs.
    ```

3. Train the model using the preprocessed data:
    ```
    ivadomed --train -c ivadomed/config/config_pose.json -pd ~/data-multi-subject/ --gif 2
    ```

   <details>
   <summary> Sample output log... </summary>

    **Note:** This was only tested using 7 subjects for debugging purposes, so the loss values are not representative.
   
    ```
	Outdated. TODO, JN: Paste up-to-date logs.
    ```

## Next steps

I am currently following this rough outline:

- [x] Run `--train` using T1w/T2w images from `data-multi-subject`. 
    - **Note:** `data-multi-subject` contains curved data. However, in hindsight, this data probably should have been straightened for the reasons in [this comment](https://github.com/ivadomed/ivadomed/pull/852#discussion_r710455668). So, the training steps will need to be modified accordingly.
- [X] Preliminary evaluation of the trained models using `--test`. (See [this comment](https://github.com/ivadomed/ivadomed/pull/852#discussion_r708534669).)
    - **Note:** Evaluation has been done for curved models only. This will need to be repeated for straightened models. 
- [ ] Upload the trained models to repos in the @ivadomed org, and create releases for each.
- [ ] In a new SCT branch, modify `sct_label_vertebrae` to try to perform inference with these new models.
- [ ] Update [`vertebral-labeling-validation`](https://github.com/sct-pipeline/vertebral-labeling-validation) to be able to compare all three approaches.
- [ ] Come up with a plan to compare the vertebral labeling approaches to determine the current baseline performance of each approach (i.e. without hyperparameter tweaking). 
    - [ ] TBD: Test dataset? 
        - Test subjects from data-multi-subject? (This would require that the two models were trained on the same subjects, and thus would have the same test subjects as well.)
        - A different dataset? (e.g. sct-testing-large)
    - [ ] TBD: Metrics for comparison? [(Missing discs? MSE? etc.)](https://github.com/ivadomed/ivadomed/pull/852#discussion_r708534726)
- [ ] Tweak hyperparameters to improve HourglassNet performance against existing approaches (if needed).

## Linked issues

Resolves #539.
Resolves #766.
Supersedes #787.

## Additional context

- [SCT PR #2679](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2679): Integrating Lucas' labeling approach (Countception) into `sct_label_vertebrae`.
- [`vertebral-labeling-validation`](https://github.com/sct-pipeline/vertebral-labeling-validation): Repository to compare the Countception model vs. `sct_label_vertebrae`. (This should be adapted to also compare the Hourglass Network, as well.)
- TODO: add a link to a documentation that explains the strategy used to generate the disc labels. @lrouhier do you have that information?
   - JN: Disc labels were created by manually labeling single-voxel points, then expanding these into Gaussian heatmaps.
   - Source: [vertebral-labeling-deep-learning](https://github.com/neuropoly/vertebral-labeling-deep-learning) repo and specifically [`train_utils.py :: label2MaskMap_GT()`](https://github.com/neuropoly/vertebral-labeling-deep-learning/blob/c9aaf4e3191ccd4e158fcffce3c556143374f8dd/scripts_vertebral_labeling/train_utils.py#L21-L91)


joshuacwnewton:
Ah right, there were failing tests. My apologies -- I'll address this shortly, then mark it as ready for review ASAP.

ivadomed/config/config_pose.json
```diff
@@ -0,0 +1,94 @@
+{
```
joshuacwnewton, line 1: I've cleaned up this PR a little, and fixed a few issues that were preventing the testing commands from running. My hope is that this will provide a fresh base for us to work from. :slightly_smiling_face: 

I'm also happy to help out with whatever the next steps will be (e.g. training/evaluating the model using `spine-generic`).

jcohenadad (commented):
i've got this error message when running the help on the preprocessing function:

<details>

~~~
(venv) julien-macbook:~/code/ivadomed $ python ivadomed/scripts/prepare_dataset_vertebral_labeling.py -h
2021-07-07 10:56:06.667 | INFO     | ivadomed.utils:init_ivadomed:401 - 
ivadomed (git-jn/539-intervertebral-disc-labeling-pose-estimation-b5bed10c5760b50b1e1df2e4bb4742ecbd71030a)

usage: prepare_dataset_vertebral_labeling.py [-h] -p <file> -s <str> [-a <int>]

This function takes as input a path to a dataset and generates a set of images:
    (i) mid-sagittal image and
    (ii) heatmap of disc labels associated with the mid-sagittal image. Please refer to for more details. 

Example:
    python3 prepare_dataset_vertebral_labeling.py --path /data/data-multi-subject/ -s _T1w

Reference:
    https://github.com/neuropoly/vertebral-labeling-deep-learning

optional arguments:
  -p <file>, --path <file>
                        Path to bids folder
  -s <str>, --suffix <str>
                        Suffix of the input file as in sub-xxxSUFFIX.nii.gz (E.g., _T2w)
  -a <int>, --aim <int>
                        0 or positive int. If set to any positive int, only label with this value will be taken into account
  File "/Users/julien/code/ivadomed/ivadomed/utils.py", line 265, in get_arguments
    args = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 1768, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 1800, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 2006, in _parse_known_args
    start_index = consume_optional(start_index)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 1946, in consume_optional
    take_action(action, args, option_string)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 1874, in take_action
    action(self, namespace, argument_values, option_string)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 1045, in __call__
    parser.exit()
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 2508, in exit
    _sys.exit(status)
SystemExit: 0

During handling of the above exception, another exception occurred:

  File "ivadomed/scripts/prepare_dataset_vertebral_labeling.py", line 154, in <module>
    main()
  File "ivadomed/scripts/prepare_dataset_vertebral_labeling.py", line 148, in main
    args = imed_utils.get_arguments(parser, args)
  File "/Users/julien/code/ivadomed/ivadomed/utils.py", line 267, in get_arguments
    raise ArgParseException('Error parsing args')
ivadomed.utils.ArgParseException: Error parsing args
~~~
</details>

EDIT: Resolved by #854.

jcohenadad:
When running `prepare_dataset_vertebral_labeling.py`, new images are created (eg: sub-amu01_rec-mid_T2w.nii.gz), however they are wrongly oriented:
Screen Shot 2021-07-07 at 11 23 29 AM

While this is not a fundamental issue for training a model (because the associated `sub-amu01_rec-mid_T2w_heatmap0.nii.gz` is _also_ inverted), this is wrong and if other people use chunk of that code they might end up with bugs. I suggest we fix this issue (shouldn't be too difficult).

joshuacwnewton:
> i've got this error message when running the help on the preprocessing function:

The error message seems to be a false positive due to an existing section of ivadomed's codebase:

https://github.com/ivadomed/ivadomed/blob/1fb27341473bda40ade221f5e4e06e16ba2a0f20/ivadomed/utils.py#L261-L267

Argparse is exiting with status code 0 (no error):

> ```
>   File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/argparse.py", line 2508, in exit
>     _sys.exit(status)
> SystemExit: 0
> ```

Yet, ivadomed is interpreting it as an argparse failure. This is a quick fix that I can open a separate PR for. :slightly_smiling_face:

joshuacwnewton:
> When running `prepare_dataset_vertebral_labeling.py`, new images are created (eg: sub-amu01_rec-mid_T2w.nii.gz), however they are wrongly oriented:

Good catch! I also notice that the existing test for this script is quite bare (just checking file existence):

https://github.com/ivadomed/ivadomed/blob/1fb27341473bda40ade221f5e4e06e16ba2a0f20/testing/functional_tests/test_prepare_dataset_vertebral_labelling.py#L13-L22

So this might be a good opportunity to check some basic properties of the generated images.

joshuacwnewton:
> When running `prepare_dataset_vertebral_labeling.py`, new images are created (eg: sub-amu01_rec-mid_T2w.nii.gz), however they are wrongly oriented:

I rewrote the test as follows to try and catch this issue:

```python
def test_prepare_dataset_vertebral_labeling(download_functional_test_files):
    prepare_dataset_vertebral_labeling.main(args=['--path', __data_testing_dir__,
                                                  '--suffix', '_T2w',
                                                  '--aim', '3'])
    for subj in ['sub-unf01', 'sub-unf02', 'sub-unf03']:
        im_in = nib.load(f"{__data_testing_dir__}/{subj}/anat/{subj}_T2w.nii.gz")
        im_mid = nib.load(f"{__data_testing_dir__}/derivatives/labels/{subj}/anat/{subj}_rec-mid_T2w_heatmap3.nii.gz")
        assert (nib.orientations.aff2axcodes(im_in.header.get_best_affine()) ==
                nib.orientations.aff2axcodes(im_mid.header.get_best_affine()) == ('R', 'A', 'S'))
```

But, this test passes (in other words, the script seems to preserve the orientations of the input images.)

Thinking it might be a data-specific issue, I tried taking a look at `sub-amu01` specifically, but I'm seeing LPI (or RAS, depending on convention) orientation for both the input and output images:

<details>
<summary> Output of fslhd </summary>

```console
> sct_image -i sub-amu01_T1w.nii.gz -header

--
Spinal Cord Toolbox (git-jn/2857-template-registration-tutorial-3bf9783b8a7eb982b4fba7c324c057b680bb1719)

sct_image -i sub-amu01_T1w.nii.gz -header
--

sizeof_hdr	348
data_type	FLOAT64
dim		[3, 192, 260, 320, 1, 1, 1, 1]
vox_units	mm
time_units	s
datatype	64
nbyper		8
bitpix		64
pixdim		[1.0, 1.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0]
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	1
freq_dim	2
slice_dim	3
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Unknown
intent_code	0
intent_name	
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	1.000000 0.000000 0.000000 -100.092461 
qto_xyz:2	0.000000 1.000000 -0.000349 -109.344482 
qto_xyz:3	0.000000 0.000349 1.000000 -94.816086 
qto_xyz:4	0.000000 0.000000 0.000000 1.000000 
qform_xorient	Left-to-Right
qform_yorient	Posterior-to-Anterior
qform_zorient	Inferior-to-Superior
sform_name	Scanner Anat
sform_code	1
sto_xyz:1	1.000000 0.000000 0.000000 -100.092461 
sto_xyz:2	0.000000 1.000000 -0.000349 -109.344482 
sto_xyz:3	0.000000 0.000349 1.000000 -94.816086 
sto_xyz:4	0.000000 0.000000 0.000000 1.000000 
sform_xorient	Left-to-Right
sform_yorient	Posterior-to-Anterior
sform_zorient	Inferior-to-Superior
file_type	NIFTI-1+
file_code	1
descrip		TE=3.2;Time=115007.980;phase=1
aux_file
	
> sct_image -i sub-amu01_rec-mid_T1w.nii.gz -header

--
Spinal Cord Toolbox (git-jn/2857-template-registration-tutorial-3bf9783b8a7eb982b4fba7c324c057b680bb1719)

sct_image -i sub-amu01_rec-mid_T1w.nii.gz -header
--

sizeof_hdr	348
data_type	FLOAT64
dim		[3, 1, 260, 320, 1, 1, 1, 1]
vox_units	Unknown
time_units	Unknown
datatype	64
nbyper		8
bitpix		64
pixdim		[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	0
freq_dim	0
slice_dim	0
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Unknown
intent_code	0
intent_name	
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Unknown
qform_code	0
qto_xyz:1	1.000000 0.000000 0.000000 0.000000 
qto_xyz:2	0.000000 1.000000 0.000000 0.000000 
qto_xyz:3	0.000000 0.000000 1.000000 0.000000 
qto_xyz:4	0.000000 0.000000 0.000000 1.000000 
qform_xorient	Left-to-Right
qform_yorient	Posterior-to-Anterior
qform_zorient	Inferior-to-Superior
sform_name	Aligned Anat
sform_code	2
sto_xyz:1	1.000000 0.000000 0.000000 0.000000 
sto_xyz:2	0.000000 1.000000 0.000000 0.000000 
sto_xyz:3	0.000000 0.000000 1.000000 0.000000 
sto_xyz:4	0.000000 0.000000 0.000000 1.000000 
sform_xorient	Left-to-Right
sform_yorient	Posterior-to-Anterior
sform_zorient	Inferior-to-Superior
file_type	NIFTI-1+
file_code	1
descrip		
aux_file
```
</details>

What tool are you using to display the images? It seems new to me. :sweat_smile:

jcohenadad:
> But, this test passes (in other words, the script seems to preserve the original orientations of the input images.)

yeah, it's a tricky one. The problem is not the header (which you are testing), but the actual data (ie: the content of the nd.array, which is up/down flipped). Testing it pretty much implies copy/pasting the code that is used in the script in the first place so i'm not sure it is a good investment of our time. I suggest making sure the output looks good (ie: head is at the top) and it _should_ stay good forever.

joshuacwnewton:
> The problem is not the header (which you are testing), but the actual data (ie: the content of the nd.array, which is up/down flipped).

Ah, of course! 

It looks like the rotation was a new addition from @rezazad68 in https://github.com/ivadomed/ivadomed/pull/787/commits/eab08a3604a4c0a7212e2caf5238c3618197a6ea, but I'm not sure what the justification was, because the commit message is bare. I'll ask. :slightly_smiling_face:

jcohenadad:
> It looks like the rotation was a new addition from @rezazad68 in eab08a3, but I'm not sure what the justification was, because the commit message is bare. I'll ask. 🙂

My gut feeling is that this was done for displaying images in the correct orientation for generating figures for the paper, but this is not necessary for the actual function.

ivadomed/transforms.py
```diff
@@ -287,6 +294,50 @@ def __call__(self, sample, metadata=None):
...
+
+    def undo_transform(self, sample, metadata=None):
+        return sample, metadata
+
+    def get_posedata(self, msk, num_ch):
+        ys = msk.shape
+        ys_ch = [np.zeros([ys[0], ys[1]])] * num_ch
+        msk_uint = np.uint8(np.where(msk>0.2, 1, 0))
+        
+        labels_im, num_labels = measure.label(msk_uint, background=0, return_num=True)
```
joshuacwnewton, line 314: > My gut feeling is that this was done for displaying images in the correct orientation for generating figures for the paper, but this is not necessary for the actual function.

I've spoken with Reza, and the reason the images were rotated was because of the [`measure.label()`](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.measurements.label.html) connected components labeling algorithm used to split the ground truth image. This algorithm uses the Y coordinate to number the segments, so images like this become problematic, since the Y coordinates are not monotonic:

image

But, as you've pointed out @jcohenadad  the _data array itself_ was flipped during preprocessing. Instead, if I'm not mistaken, the proper way to make sure the images are in the correct orientation would have been to use the header orientation metadata (and leave the array itself alone).

joshuacwnewton, line 314: I've removed the rotation code. Now, when I open up the preprocessed images in FSLeyes, they appear properly oriented in the S-I direction.

Screenshot from 2021-07-09 10-35-19

When loading these images during training, the resulting numpy array seems to be upside-down.

tmpm5yhyxia

Is this because numpy arrays have a [top-left origin](https://stackoverflow.com/questions/51849547/why-is-numpys-origin-at-the-upper-left-unlike-mathematical-convention/51849585), while MRI convention is to use a bottom-left origin? (My apologies if this is a really basic question.)

jcohenadad: > I notice when loading these images during training, the resulting numpy array seems to be upside-down.

are you sure this is the corresponding image? The number of points differ (8 vs. 9) and the pattern seem slightly different. But anyways, I think the key is to have the MRI image and the ground truth (ie: heatmap) treated the same way, and it won't import how the matrix is ordered with numpy

joshuacwnewton: > are you sure this is the corresponding image? The number of points differ (8 vs. 9) and the pattern seem slightly different.

It's not, sorry -- I just grabbed a random sample image to demonstrate the orientation.

> But anyways, I think the key is to have the MRI image and the ground truth (ie: heatmap) treated the same way, and it won't import how the matrix is ordered with numpy

:+1:

ivadomed/models.py
```diff
@@ -1429,6 +1429,214 @@ def forward(self, x):
...
+        up1 = self.hg[n-1][0](x)
+        low1 = F.max_pool2d(x, 2, stride=2)
+        low1 = self.hg[n-1][1](low1)
+
+        if n > 1:
+            low2 = self._hour_glass_forward(n-1, low1)
+        else:
+            low2 = self.hg[n-1][3](low1)
+        low3 = self.hg[n-1][2](low2)
+        up2 = F.interpolate(low3, scale_factor=2)
```
joshuacwnewton, line 1533: After training the `HourglassNet`, I encountered an error during conversion from `.pt` to `.onnx`:

<details>
<summary> Full traceback </summary>

```python
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/onnx/utils.py", line 483, in _export
    graph, params_dict, torch_out = _model_to_graph(model, args, verbose,
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/onnx/utils.py", line 334, in _model_to_graph
    graph, torch_out = _trace_and_get_graph_from_model(model, args, training)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/onnx/utils.py", line 291, in _trace_and_get_graph_from_model
    torch.jit._get_trace_graph(model, args, _force_outplace=False, _return_inputs_states=True)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/jit/__init__.py", line 279, in _get_trace_graph
    outs = module(*args, **kwargs)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/jit/__init__.py", line 358, in forward
    graph, out = torch._C._create_graph_by_tracing(
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/jit/__init__.py", line 349, in wrapper
    outs.append(self.inner(*trace_inputs))
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 548, in __call__
    result = self._slow_forward(*input, **kwargs)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 534, in _slow_forward
    result = self.forward(*input, **kwargs)
  File "/home/joshua/repos/ivadomed/ivadomed/models.py", line 1641, in forward
    y = self.hg[i](x)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 548, in __call__
    result = self._slow_forward(*input, **kwargs)
  File "/home/joshua/repos/ivadomed/venv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 534, in _slow_forward
    result = self.forward(*input, **kwargs)
  File "/home/joshua/repos/ivadomed/ivadomed/models.py", line 1551, in forward
    return self._hour_glass_forward(self.depth, x)
  File "/home/joshua/repos/ivadomed/ivadomed/models.py", line 1540, in _hour_glass_forward
    low2 = self._hour_glass_forward(n-1, low1)
  File "/home/joshua/repos/ivadomed/ivadomed/models.py", line 1540, in _hour_glass_forward
    low2 = self._hour_glass_forward(n-1, low1)
  File "/home/joshua/repos/ivadomed/ivadomed/models.py", line 1540, in _hour_glass_forward
    low2 = self._hour_glass_forward(n-1, low1)
  File "/home/joshua/repos/ivadomed/ivadomed/models.py", line 1547, in _hour_glass_forward
    out = up1 + up2
RuntimeError: The size of tensor a (3) must match the size of tensor b (2) at non-singleton dimension 3
```

</details>

After investigating, I think this is a bug in the original model architecture, so I have reported the bug (plus a potential fix) upstream to https://github.com/bearpaw/pytorch-pose/issues/124.

I am going to try retraining the model using this fix, and see if it lets me convert to `.onnx` without failure.

ivadomed/config/config_pose.json
```diff
@@ -0,0 +1,94 @@
...
+	    "soft_gt": true
+    },
+    "split_dataset": {
+        "fname_split": null,
+        "random_seed": 8,
+        "split_method" : "participant_id",
+        "data_testing": {"data_type": null, "data_value":[]},
+        "balance": null,
+        "train_fraction": 0.6,
+        "test_fraction": 0.2
```
joshuacwnewton, line 35: I've run `--test` using the T1w model (mostly because I'm trying to learn how to use `ivadomed`, and I was curious what the output was). Here are the metrics for the 52-subject test dataset: [`evaluation_3Dmetrics.csv`](https://github.com/ivadomed/ivadomed/files/7164215/evaluation_3Dmetrics.csv)

Skimming the results, however, I notice that ivadomed is calculating metrics as though this were a segmentation problem. For example, take the subject `sub-balgrist05`:

Screenshot from 2021-09-14 14-15-38

Here are the resulting metrics for this subject:

| image_id                   | avd_class0        | dice_class0       | lfdr_0-20vox_class0 | lfdr_21-100vox_class0 | lfdr_class0 | ltpr_101-INFvox_class0 | ltpr_class0 | mse_class0        | n_101-INFvox | n_gt_class0 | n_pred_class0 | precision_class0  | recall_class0    | rvd_class0        | specificity_class0 | vol_gt_class0    | vol_pred_class0 | lfdr_101-INFvox_class0 |
|----------------------------|-------------------|-------------------|---------------------|-----------------------|-------------|------------------------|-------------|-------------------|--------------|-------------|---------------|-------------------|------------------|-------------------|--------------------|------------------|-----------------|------------------------|
| sub-balgrist05_rec-mid_T1w | 0.418087197212938 | 0.537764707940283 |                     | 1                     | 0.5         | 1                      | 1           | 0.564196995428872 | 1            | 1           | 6             | 0.730948754094583 | 0.42534843818889 | 0.418087197212938 | 0.999298371317197  | 371.189633507755 | 216             |                        |

Since SCT is more concerned with single-voxel labels, rather than disc heatmaps, the metrics above don't exactly tell the whole picture.

joshuacwnewton: I notice that in the [`vertebral-labeling-validation`](https://github.com/sct-pipeline/vertebral-labeling-validation/blob/800d9db0810a9cba53b9d85a157366e582c1bf48/run_prediction_deep.sh#L59-L60) repository, @lrouhier chose to look for [missing discs](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/spinalcordtoolbox/labels.py#L278-L304) and [MSE](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/spinalcordtoolbox/labels.py#L308-L333) instead. I will keep this in mind when evaluating the performance of the various methods.

ivadomed/config/config_pose.json
```diff
@@ -0,0 +1,94 @@
...
+        {
+            "wspace": 256,
+            "hspace": 256,
+            "dspace": 1,
+            "flag_pixel": true
+        },
+      "VertebralSplitting": {"max_joint": 11, "applied_to": ["gt"]},
+      "NumpyToTensor": {},
+      "NormalizeInstance": {"applied_to": ["im"]}
+    }
```
joshuacwnewton, line 93: ## Comparing HourglassNet (Reza) to Countception (Lucas)

I'm currently trying to figure out how to use `vertebral-labeling-validation` to compare the two approaches, however I've run into an issue:

* Lucas' model was trained on [straightened data](https://github.com/ivadomed/ivadomed/blob/b15d7cfd4b271261835b4c93864d92411c45c3d6/ivadomed/config/config_vertebral_labeling.json#L16-L17) (because `sct_label_vertebrae` uses straightened data for template-matching with the PAM50 template).
* Reza's model, however, is currently trained on [unstraightened data](https://github.com/ivadomed/ivadomed/blob/a2929e689ad5a3a55a66e39f2c54f98af36b6cff/ivadomed/config/config_pose.json#L16-L17) from `data-multi-subject`.

This makes Reza's model much trickier to integrate into the [current structure of `sct_label_vertebrae`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2679/files#r710454872), which in turn makes it much trickier to evaluate Reza's approach using the `vertebral-labeling-validation` repo.

To fix this, I'm thinking that training the HourglassNet with straightened data would make our lives much easier down the line when trying to evaluate the approach +  integrate it into SCT.

jcohenadad: > To fix this, I'm thinking that training the HourglassNet with straightened data would make our lives much easier down the line when trying to evaluate the approach + integrate it into SCT

good idea.

joshuacwnewton: Small progress update: I'm still currently working on getting [`vertebral-labeling-validation`](https://github.com/sct-pipeline/vertebral-labeling-validation) set up and running. There are [some existing issues](https://github.com/sct-pipeline/vertebral-labeling-validation/issues) with the scripts, but I'm trying to sort those out right now, and I will open a PR in that repo when I have a clear set of working instructions for running the test procedure.

joshuacwnewton: Small update 2: I have a model trained on straightened data now!

sub-hamburg05__rec-mid_T1w__0

(Arriving here took a little longer than I hoped, because the precise preprocessing commands used for the Countception model weren't documented beyond a [short paragraph description](https://github.com/neuropoly/vertebral-labeling-deep-learning#preprocessing). This time, I'll make sure to share a script that will help with reproduction.)

But, now that the pieces are in place, I should be able to now work towards comparing the two models using [`vertebral-labeling-validation`](https://github.com/sct-pipeline/vertebral-labeling-validation).

jcohenadad: This is fantastic @joshuacwnewton ! few comments:
- the image quantization seems low-- i am just wondering if this is caused by the GIF generation, or if the low quality was used during training
- the straightened cord seems squeezed in the superior-inferior direction. I am wondering if this is just something that happened during the GIF creation or also before, during training.
- did you see changes in performance between the straightened and the non-straightened scenario?
- can you please remind me again what was the rationale for doing the inference in the straightened scenario: was it to:
  - be more compatible with the current workflow of sct_label_vertebrae
  - allow for easier comparison between the current method and the method you are implementing
  - supposedly better performance on the straightened cord because less variability (and cropping around spinal cord made easier)?
- (minor comment: the head/feet is upside down)

joshuacwnewton: > the image quantization seems low-- i am just wondering if this is caused by the GIF generation, or if the low quality was used during training

I believe the low quantization is caused by the GIF generation. It is present in Reza's original gif, too: https://github.com/ivadomed/ivadomed/issues/766#issuecomment-835714961

<details>
<summary> Here is the (higher-quality) input image corresponding to the gif </summary>

Screenshot from 2021-09-30 12-07-33
</details>

> the straightened cord seems squeezed in the superior-inferior direction. I am wondering if this is just something that happened during the GIF creation or also before, during training.

The SI squeezing occurs during training. This is because the model configuration currently resamples images to 256x256, but `sct_straightened_spinalcord` crops images around the spinal cord, so the rectangular image is reshaped to a square image.

The resampling is present because I had been re-using the config file from Reza's initial PR as-is, just to get things working. 

> did you see changes in performance between the straightened and the non-straightened scenario?

This is something I want to evaluate more thoroughly, and so I don't want to comment prematurely, but I'll get back to you on this. :slightly_smiling_face: 

> can you please remind me again what was the rationale for doing the inference in the straightened scenario: was it to: 

Initially, the motivation was for the first two reasons that you describe (`sct_label_vertebrae` compatibility, ease of comparison). 

Specifically, working with curved data would require a much bigger refactor on SCT's end. But, working on straightened data means that we could swap the new model very easily into the [existing SCT PR](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2679), since it would work with the same input data as Lucas' model.

However, after I learned that `sct_straighted_spinalcord` includes cropping around the spinal cord, I began to think about the other reasons you describe -- i.e. that straightening would reduce variability between subjects, and cropping would reduce the # of non-SC pixels and thus reduce the class imbalance further. So, I agree that these are secondary motivations to preprocess the data in this way.

> (minor comment: the head/feet is upside down)

I believe this is due to differences in the location of the origin between numpy arrays and NIfTI images (top-left vs bottom-left). When I open the images in FSLeyes, they're correctly oriented. So, I was thinking that it might be an issue with the gif functionality in ivadomed.

Still, if it's a visual bug with `--gif`, I feel like it should have been noticed already... So, maybe it _is_ a problem on my end. (Insight from the ivadomed team would be appreciated, e.g. if others have used the `--gif` functionality without this issue.)

jcohenadad: > This is because the model configuration currently resamples images to 256x256, but sct_straighten_spinalcord crops images around the spinal cord (resulting in a rectangular image), so it gets squished during resampling to the square dimensions.

I see. That might be a red flag, as we want to conserve the image proportion during the resampling. I remember there was a conversation about that between Reza, Charley and me (probably buried somewhere in the PR). But long story short: what we want is an algorithm that _resamples_ to a 256x256 square (ie: keeping the physical proportion of the image in case the input image is not isotropic resolution), versus _resizing_ (ie: re-defining the physical size of each pixel). If I'm not clear, I can explain better in a meeting

> Still, if it's a visual bug with --gif, I feel like it should have been noticed already... 

Not necessarily, because so far ivadomed has mostly been used with axial images, where swapping top/bottom doesn't look "weird".

joshuacwnewton: > > This is because the model configuration currently resamples images to 256x256, but sct_straighten_spinalcord crops images around the spinal cord (resulting in a rectangular image), so it gets squished during resampling to the square dimensions.
> 
> I see. That might be a red flag, as we want to conserve the image proportion during the resampling. I remember there was a conversation about that between Reza, Charley and me (probably buried somewhere in the PR). But long story short: what we want is an algorithm that _resamples_ to a 256x256 square (ie: keeping the physical proportion of the image in case the input image is not isotropic resolution), versus _resizing_ (ie: re-defining the physical size of each pixel). If I'm not clear, I can explain better in a meeting

Ahhh, I see! Thank you for bringing this up again, because it actually turns out that in [the thread you've referenced](https://github.com/ivadomed/ivadomed/pull/787#discussion_r645318419) containing the discussion from Charley, the ultimate conclusion was [this](https://github.com/ivadomed/ivadomed/pull/787#discussion_r645318419):

> Ok, I will update the config file with `Center Crop` instead of `resize`.

It looks like this was never completed in Reza's PR, and I missed it as well, so the resizing was still being applied. I'll make that change ASAP.

joshuacwnewton:
I'm going to temporarily close this PR (**while preserving the underlying branch**), as I am no longer involved in this effort. 

Further progress can be tracked in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3793. If this effort is resumed,  feel free to re-open, or duplicate the work from the underlying branch.

PR #854, opened 2021-07-07
`utils.py`: Only raise ArgParseException for non-zero SystemExits

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Currently, for the CLI script argparser in `utils.py`, we try and catch `SystemExit`, then swap in our own custom exception type. However, `SystemExit` doesn't always mean error. If `-h` is passed, the parser will exit successfully with code 0, yet we interpret this as a failure. (See https://github.com/ivadomed/ivadomed/pull/852#pullrequestreview-701116341.)

So, this PR changes the try/except block to only raise the custom exception if the error code was non-zero.

(Tagging @ahill187 as it was PR #580 that introduced this, so maybe there was a good reason for the custom exception type that I'm missing. :sweat_smile: )

## Linked issues

Needed for https://github.com/ivadomed/ivadomed/pull/852#issuecomment-875764329.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -81,25 +81,29 @@ working under a virtual environment, which could be set as follows:
 Step 2: Clone the `ivadomed <https://github.com/ivadomed/ivadomed>`_ repository.
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
+
 ::
 
     git clone git@github.com:ivadomed/ivadomed.git
     cd ivadomed
+ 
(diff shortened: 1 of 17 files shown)


ahill187 (commented):
I think it's cleaner having a custom exception, especially because it throws a more distinct error message. Is there a way to change the try/except clause in ArgParseException so that it raises an exception on the correct error? I.e. maybe we could catch the `-h` in the ArgParseException class and not raise an error?

ahill187 (approved):

PR #758, opened 2021-04-15
Standardize Header Handling when creating new nibabel object

cakester:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

In PR https://github.com/ivadomed/ivadomed/pull/714 the way nibabel objects are created was changed, so we want to update the object creation in the following places as well

* uncertainty
* preprocessing
* postprocessing
* visualize
* loader/adaptive
* evaluation

## Linked issues

Timeline of the related issues: 
- Original Issue raised #711 and also summarized at https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3283  
- Subsequent PR fix #714. 
- (which created issue: #720 and that test needs to be eventually re-enabled)
- Recent issue #775 discovered when using Affine=None when creating new NifTI image, meaning [upstream issue](https://github.com/nipy/nibabel/issues/1012) resolution needs to reach consensus among IvadoMed/Neuropoly devs. 
- The likely proposal is to either implement: 
   - Nifti1Image(data, img.header.get_best_affine(), img.header)  OR
   - Nifti1Image(data, img.affine, img.header)

Resolve https://github.com/ivadomed/ivadomed/issues/715
Resolve https://github.com/ivadomed/ivadomed/issues/775

Issue #715, opened 2021-03-15
Copy header (not just affine matrix) when creating new nibabel objects


jcohenadad:
### Context

In #714 we changed the way new nibabel objects were created: copying the header (not just the affine matrix) from the reference nibabel image, in order to preserve the qform_code. 

We probably want to apply this change to other instances in ivadomed, namely:
- [uncertainty](https://github.com/ivadomed/ivadomed/blob/87d6a63d0bb0114bb3d3d8e13584089bf2ce0142/ivadomed/uncertainty.py#L76)
- [preprocessing](https://github.com/ivadomed/ivadomed/blob/3c9d32c41ef4222848f0dfb024e89c11f7264046/ivadomed/preprocessing.py#L36)
- [postprocessing](https://github.com/ivadomed/ivadomed/blob/3004dfe53afcc04529380ac419370ccb76f3a128/ivadomed/postprocessing.py#L26)
- [visualize](https://github.com/ivadomed/ivadomed/blob/a8f1839ace712347f046a4131cdd75a8c6cc43da/ivadomed/visualize.py#L226)
- [loader/adaptive](https://github.com/ivadomed/ivadomed/blob/fb3c913ddc3fb1942884bdbb16a7796e94d20192/ivadomed/loader/adaptative.py#L753)
- [evaluation](https://github.com/ivadomed/ivadomed/blob/87156e33be17eb96edb124e01e98d57619ea8a59/ivadomed/evaluation.py#L84)

PR #859, opened 2021-07-12
Hm/fix color label

### LINKED ISSUE #720: Un-silence a unit test: test_rgb.py

cakester:
#### GitHub

#### PR contents

## Description

this should resolve https://github.com/ivadomed/ivadomed/pull/834#discussion_r668044624

## Linked issues

resolves #720


cakester:
and I need to remove the test_rgb_file from the data_functional_testing after this is merged

mariehbourget (commented):
I tested the PR and realize I may not be the best person to do this review. I am not familiar with the `save_color_labels` function (I never used it before it was commented). On a small training on the `data_example_spinegeneric` with `"target_suffix": ["_seg-manual", "_csfseg-manual"]`. I got this `_color.nii.gz` image as a result (in black and white).

color_label

Moreover, because the function now writes a NifTI file without header, the orientation is not known (? on the side of the image).

From the issue #715 and PR #758, my understanding is that we want every Nibabel object to be created with the complete header, which is now not the case in this PR for the `save_color_labels` function in `visualize.py`.

The function and associated test were first silenced in issue #720 and PR #714.
It seems that the test was fixed in #832, but the function still expects a different header than the one that are in our `gt_files`.

Question: it is possible to modify the implementation of the function itself to accept the right kind of datatype from the header of our `gt_files`, instead of changing how the Nifti file is written?

cakester:
> I tested the PR and realize I may not be the best person to do this review. I am not familiar with the `save_color_labels` function (I never used it before it was commented). On a small training on the `data_example_spinegeneric` with `"target_suffix": ["_seg-manual", "_csfseg-manual"]`. I got this `_color.nii.gz` image as a result (in black and white).
> 
> color_label
> 
> Moreover, because the function now writes a NifTI file without header, the orientation is not known (? on the side of the image).
> 
> From the issue #715 and PR #758, my understanding is that we want every Nibabel object to be created with the complete header, which is now not the case in this PR for the `save_color_labels` function in `visualize.py`.
> 
> The function and associated test were first silenced in issue #720 and PR #714.
> It seems that the test was fixed in #832, but the function still expects a different header than the one that are in our `gt_files`.
> 
> Question: it is possible to modify the implementation of the function itself to accept the right kind of datatype from the header of our `gt_files`, instead of changing how the Nifti file is written?

it is colored label...so the datatype would be rgb, so does the header

mariehbourget:
> it is colored label...so the datatype would be rgb, so does the header

You are right, I checked the header and the `data_type` is indeed `RGB` for the `_pred_color.nii.gz` file. I think I found how to see it in colors in `fsleyes` with the `3D/4D RGB(A) Volume` display option. This is what I got:

color_label_2

However, I think preserving the rest of the header is important to preserve the shape and orientation of the image. For example, the same image without the color (`_pred.nii.gz`) gives me this:

color_label_3

As I mentioned earlier, I don't know enough to comment further on this and other inputs are needed.

cakester:
change strategy...right now I copy the all of the header and change the datatype to rgb when save to color label

ivadomed/inference.py
```diff
@@ -195,11 +195,15 @@ def pred_to_nib(data_lst: List[np.ndarray], z_lst: List[int], fname_ref: str, fn
 
     # Here we prefer to copy the header (rather than just the affine matrix), in order to preserve the qform_code.
     # See: https://github.com/ivadomed/ivadomed/issues/711
+    if color_label:
+        super(nib.Nifti1Header, nib_ref.header).set_data_dtype(128)
```
cakester, line 199: this sets the data type to rgb inside header

Issue #468, opened 2020-10-07
Dice > 1


olix86:
On GM dataset, hard training : 
![image](https://user-images.githubusercontent.com/5721601/95394642-ea2a7880-08ca-11eb-8f4d-7c848e9be097.png)

charleygros:
eheh this model is bloody good hum

As mentioned, I would recommend checking the values of the input arrays, [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/metrics.py#L92).
If that's the source of the error --> what about throwing an error if values are out of the range [0, 1] to avoid such error?

If that's not the source of the error... well... that's concerning eheh... so let's hope that's not the case (-:

olix86:
```
im1 : max, min  2.0 0.0
im2 : max, min  1.0 0.0
````

andreanne-lemay:
I am relieved! Great news :')

PR #870, opened 2021-07-21
Fix microscopy ground-truth range of values

### LINKED ISSUE #868: Assess range of array values during metrics computation

While training some microscopy models, I encountered dice scores superior to 1 during the testing phase. Investigating this further, I noticed that the dice score computation [here](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/metrics.py#L91) used a "float" representation for one operand (MIN 0.0 - MAX 1.0) and an 8-bit integer value (stored as a float64) for the other operand (MIN 0.0 - MAX 255.0). @mariehbourget's hypothesis for this is that data read from the NiFTI files is converted during the training pipeline but not during the microscopy testing pipeline.
Another user previously encountered something similar in #468.

Besides fixing this by adding the right conversion during the testing phase, I would suggest adding some form of check for this because this kind of bug invalidates the eval metrics. Something like this:
```python
if im1.max() != im2.max():
    raise ValueError("Range mismatch: im1 and im2 must be on the same scale.")
```

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR fix an issue where the Dice score would be >1 when training with microscopy data. In the loader, the PNG/TIF ground-truth values of 0 and 255 are now converted to 0 and 1 before converting to Nibabel/NifTI (as it is assumed for MRI data). These Nibabel/NifTI are then used for training, testing and evaluation and the values will have the same 0-1 range everywhere.

Additionally, the PNG/TIF values are now read and written as uint8 in the NifTI file to save space, both for the images and ground-truths.

## Questions
- In #868 and #468, we talked about checking the range of the input values for the metrics. Wouldn't it apply to the other metrics as well? If so, where should we do the verification? Feedback welcome!
- I also ran into an issue while trying to check the range in `dice_score`. The prediction files (`_pred.nii.gz`) sometimes have values slightly over 1 (!), ex: 1.0000000591389835. It happens both with microscopy and MRI. It may be a precision thing but we can't directly check for the [0-1] range.

## Linked issues
Fixes #868


hermancollin:
While testing this PR I still encountered dice scores superior to 1 (1.30 for class0 and 1.85 for class1). Adding a debut print inside the `ivadomed.metrics.py:dice_score()` function still gives me this when running a test:
```
dice_score() - im1.max(): 255.0
dice_score() - im2.max(): 1.0
dice_score() - im1.max(): 255.0
dice_score() - im2.max(): 1.0
```
To reproduce the issue with dice scores >1, I put an example on `duke/temp/arcol/to_test_pr_870`.

hermancollin:
Ok I found out why it wasn't working on my side. I hadn't deleted the nii.gz files so although the conversion to the [0,1] range was working, the nifti files weren't saved by the `ivadomed.loader.segmentation_pair.py:convert_file_to_nifti()` function. Now, I get the expected 0.7002 and 0.7971 dice scores for classes 0 and 1. Great work :tada: 

I did however encounter the problem you mentionned with max values in predictions slightly over 1 (1.0000000591389835... exactly the same value @mariehbourget reported)

mariehbourget:
Thanks @hermancollin for the debugging!

> I did however encounter the problem you mentionned with max values in predictions slightly over 1 (1.0000000591389835... exactly the same value @mariehbourget reported). This happened in evaluation but not in inference.

The evaluation step reads the NifTI file written previously at the testing step. I checked the `_pred.nii.gz` file and the value is indeed 1.0000000591389835, even though it prints as 1.0 just before saving the NifTI here 
https://github.com/ivadomed/ivadomed/blob/7a595ae39e94085837582f7176b8ec59bd9a2bc9/ivadomed/inference.py#L198

I ran some more tests:

1.  With single-class model on microscopy or MRI `pred_masks`:
- The max value is 1 (int) when using `binarize_prediction` with threshold 0.5
- The max value is 1.0000000591389835 without binarization

2. With multi-class model on microscopy or MRI `pred_masks`:
- The max value is 1.0000000591389835 with or without `binarize_maxpooling`

So it seems that `binarize_maxpooling` doesn't return the value as an integer and when values are in float, 1.0 becomes 1.0000000591389835 in the NifTI file. Interesting...

hermancollin:
> 1. With multi-class model on microscopy or MRI `pred_masks`:
> 
> * The max value is 1.0000000591389835 with or without `binarize_maxpooling`
> 
> So it seems that `binarize_maxpooling` doesn't return the value as an integer and when values are in float, 1.0 becomes 1.0000000591389835 in the NifTI file. Interesting...

@mariehbourget correct me if I'm wrong but shouldn't the `binarize_maxpooling` postprocess return integers instead of floats?

mariehbourget:
> @mariehbourget correct me if I'm wrong but shouldn't the `binarize_maxpooling` postprocess return integers instead of floats?

I think so to, right now it is returned as a float that we could convert as an int [here](https://github.com/ivadomed/ivadomed/blob/7a595ae39e94085837582f7176b8ec59bd9a2bc9/ivadomed/postprocessing.py#L300) but it would be good to know if there was a reason for the float in the first place @andreanne-lemay, @charleygros?

> Although this would fix the current problem, the issue would persist when no postprocessing is done on the predictions...

Yes, this would require more investigation, the value is 1.0 before the NifTI file is saved. Maybe there is a change in data type that affects the precision?

testing/unit_tests/test_loader.py
```diff
@@ -159,7 +160,8 @@ def test_dropout_input(seg_pair):
 def test_load_dataset_2d_png(download_data_testing_test_files,
                              loader_parameters, model_parameters, transform_parameters):
     """
-    Test to make sure load_dataset runs with 2D PNG file and write corresponding NifTI file.
+    Test to make sure load_dataset runs with 2D PNG files, writes corresponding NifTI files,
```
jcohenadad, line 163:
```suggestion
    Test to make sure load_dataset runs with 2D PNG files, write corresponding NIfTI files,
```

mariehbourget:
As discussed in meeting, I may have found an issue with this PR where I have "not as good" result with training on this branch as I had on July 5th on master (dice went down from ~0.85 to 0.75 with similar config). I'm investigating now if it's due to the 8 bit conversion or the GT values of 0 and 1, and I'll keep you posted when ready for review.

In the meantime, I would like to have some feedback on the value of `1.0000000591389835` found in prediction files that are not binarized or binarized with maxpooling, see this [comment](https://github.com/ivadomed/ivadomed/pull/870#issuecomment-885155963).

charleygros:
> I think so to, right now it is returned as a float that we could convert as an int [here](https://github.com/ivadomed/ivadomed/blob/7a595ae39e94085837582f7176b8ec59bd9a2bc9/ivadomed/postprocessing.py#L300) but it would be good to know if there was a reason for the float in the first place @andreanne-lemay, @charleygros?

I do think it would be ok to have a int-conversion here.

hermancollin:
> As discussed in meeting, I may have found an issue with this PR where I have "not as good" result with training on this branch as I had on July 5th on master (dice went down from ~0.85 to 0.75 with similar config). I'm investigating now if it's due to the 8 bit conversion or the GT values of 0 and 1, and I'll keep you posted when ready for review.

I may have found one cause of the problem here. Using @mariehbourget's config file on the SEM microscopy dataset, the following problem occurs:  the SEM dataset contains some values other than 0 or 255 in the ground-truth (most likely a compression artifact as they were intended to be "hard" segs). The `img = img // 255` may be too simple to fix the problem because values such as 130, 218 and 233 in the dataset turn to 0s with this floor division. 

In general, I don't think this is a behavior we want for the loader. However, looking at the color profiles of the troublesome masks, the % of pixel anomaly is ridiculously low so this definitely doesn't explain the training performance difference @mariehbourget reported.

mariehbourget:
>  The `img = img // 255` may be inappropriate to fix the problem because values such as 130, 218 and 233 in the dataset turn to 0s with this floor division.
In general, I don't think this is a behavior we want for the loader. However, looking at the color profiles of the troublesome masks, the % of pixel anomaly is ridiculously low so this definitely doesn't explain the training performance difference

Thank you @hermancollin for checking this. I agree with you and changed the floor division by a thresholding at 0.5. In theory, ground-truth should already be 0 and 255 with no intermediate values. But if that is not the case, a threshold makes more sense than a floor division.

> Hypothesis: There seems to be conversion loss somewhere. On my side, everytime a prediction is saved, I have a lossy conversion warning (float64 to uint8). When forcing GT values to 0 or 1 this early in the training pipeline, maybe we suffer from more conversion loss in the validation/testing step.
This conversion from float64 to uint8 of each prediction is normal, it happens when the final Nibabel predictions (in float64) is written to PNG (uint8).

Also, I have found the source of the discrepancies between training on master and training on this branch and will document it here in a separate comment as soon as possible.

mariehbourget:
This PR is ready for review.
There was a few questions along the way with discrepancies between training on master and training on this branch. I have now found the sources for these discrepancies and will document them here.
It's a long story, please bear with me.

### What the PR does:
- The PNG/TIF values (8 bit between 0 and 255) that were previously stored as `float32`, are now read and written as `uint8` in the NifTI files to save space, both for the images and ground-truths.
- The PNG/TIF ground-truth values of range 0-255 are now converted to 0s and 1s to fix an issue where the Dice score would be >1 when training with microscopy data. This is exactly the same as MRI data using 0s and 1s in `uint8` for their ground-truth. Additionally, the range 0-255 for the ground-truth was incompatible with the softseg pipeline.
- I added a test in test_loader.py to check the conversion of 0s and 1s values.

### First issue with training:
- I reported previously "not as good" result with training on this branch as I had on master (Dice went down from ~0.85 to 0.75 with similar config).
- After investigation, this was due to a poor management of the split dataset on my part. The split was vastly different between my two batches of tests causing the big differences in Dice.

### Second issue with training:
- When using the exact same split and config, I still encountered some minor systematic differences in Dice. I ran each test 4 times with always the same results.
    - On master (2cdc40ad and 0a6fffd): `Training Dice ~0.86` and `Validation Dice ~0.85`
    - On this branch (4523a0f): `Training Dice ~ 0.85` and `Validation Dice ~0.83-0.84`
- I thought it could be due to bad conversion between `uint8` and `float` data somewhere in the pipeline, but it's not the case. Results  on this branch are similar with either formats and I checked the values throughout the code. I can confirm that my issue had nothing to do with the datatype and that no bad conversion happens with either `imageio` (to read PNG) or `Nibabel` (to read/write Nifti).
- So the differences had to be because of the 0-1 range for ground-truths. I finally found the difference in the thresholding for the ground-truth that happens after the pre-processing and data augmentation [here](https://github.com/ivadomed/ivadomed/blob/4460d1379b77ffd114d4b416fda0924b40a51e3e/ivadomed/loader/mri2d_segmentation_dataset.py#L216).
- At this point, all data (regardless of how it is stored in the NIfTI file) is in `float32` in ivadomed.
- But the threshold for binarizing the intermediate values of ground-truth after the data augmentation is hardcoded to 0.5 (which makes sense when the ground-truths are initially 0s and 1s).
- So the differences between training on master and training on this branch came from the threshold being applied at a different "place" in the values range i.e. `0.5/255` vs. `0.5/1`.
- I confirmed this by running trainings with a threshold of `0.00196/1` (equivalent of `0.5/255`), and I obtain the same results as I had on master.

### Conclusion:
- Converting ground-truth to 0s and 1s when reading the PNG files and converting to NIfti files is essential to ensure the rest of the pipeline is applied correctly and metrics computed properly.
- Writing the NIfTI files in uint8 of float32 has no incidence on the rest of the pipeline.
- Differences were due to the threshold after data augmentation of 0.5 being wrongly applied on 0-255 values on master instead of  0-1 values on this branch.
- Even though the training/validation Dice are slight lower with 0-1 values, testing Dice on my test image are similar if not better with with 0-1.
- We could discuss about adding a parameter to the config and optimizing the threshold for binarization that is currently hardcoded at 0.5, but this is out of scope for this PR.

### About the `1.0000000591389835` value in NIftI predictions:
- As discussed previously in this PR [here](https://github.com/ivadomed/ivadomed/pull/870#issuecomment-885155963), we found a value >1 on some NIfTI predictions files.
- This seems to be a precision error when the files are written.
- I also decided that this subject is out of scope for this PR and will open a separate issue on the matter.

charleygros (approved):
Changes look good to me. Not tested tho.

hermancollin (approved):
Haven't tried the training, but evaluation works as expected. Thank you for fixing this!

Issue #872, opened 2021-07-24
Make IVADOMED Python API compatible with Path() objects (from pathlib)


mathieuboudreau:
### Motivation for the feature
Provides better compatibility across platforms, and is the more modern way to handle paths/files/suffixes/etc

https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

### Description of the feature
Make it so the IVADOMED API (e.g. https://github.com/ivadomed/ivadomed/blob/7a595ae39e94085837582f7176b8ec59bd9a2bc9/ivadomed/inference.py#L311 path to model and image) accepts objects of the class `from pathlib import Path` type.

dyt811:
Great point. This is an issue that has been raised before and numerous new functions or refactored modules are started using the pathlib already but there are still many legacy. I will assign someone to get this started piece meal wise.

dyt811:
We will update this ticket to include link to the PRs and list of files that are undergoing conversion process.

mathieuboudreau:
Should this issue have been closed? That PR only resolved this issue in one file.

jcohenadad:
@mathieuboudreau is right the issue shouldn’t have been closed

dyt811:
My apologies. Yes, it was indeed only the beginning of a series micro PRs to address the overhaul migration. Accidentally used the keywords in the PR description result in the closure. Will ensure this is addressed in future subsequent PRs.

PR #879, opened 2021-08-03
pathlib for loader/utils.py

cakester:
#### GitHub

#### PR contents

- [X] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

Phased refactor adoption of loader/utils.py as detailed in #872

## Linked issues

Step wise remediation of #872


ivadomed/loader/utils.py
```diff
@@ -178,7 +179,7 @@ def get_new_subject_file_split(df, split_method, data_testing, random_seed,
 
     # save the subject distribution
     split_dct = {'train': train_lst, 'valid': valid_lst, 'test': test_lst}
-    split_path = os.path.join(path_output, "split_datasets.joblib")
+    split_path = Path(path_output).joinpath("split_datasets.joblib")
```
dyt811, line 182: Use `/` here.

ivadomed/loader/utils.py
```diff
@@ -623,7 +624,7 @@ def __init__(self, loader_params, path_output, derivatives):
         self.create_bids_dataframe()
 
         # Save dataframe as csv file
-        self.save(os.path.join(path_output, "bids_dataframe.csv"))
+        self.save(str(Path(path_output).joinpath("bids_dataframe.csv")))
```
dyt811, line 627: Use `/` here.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
             ext_ct = ('.nii.gz', '.nii')
             suffix_ct = ('ct', 'CT')
             force_index = []
-            for root, dirs, files in os.walk(path_data):
-                for file in files:
+            for path_object in path_data.glob('**/*'):
```
dyt811, line 650: TIL. `path.glob` and return all path objects. Very cool. 👍 
Cool stuff like this is the reason why we moved away from string handling.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
+                if path_object.is_file():
                     # Microscopy
-                    if file == "samples.tsv" or file == "samples.json":
-                        force_index.append(file)
-                    if (file.endswith(ext_microscopy) and os.path.basename(root) == "microscopy" and
-                            (root.replace(path_data, '').startswith("sub"))):
-                        force_index.append(os.path.join(root.replace(path_data, '')))
+                    if path_object.name == "samples.tsv" or path_object.name == "samples.json":
+                        force_index.append(path_object.name)
```
dyt811, line 654: Do you know if force_index here needs to be simply file names or full path strings?

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
+                    if (path_object.name.endswith(ext_microscopy) and path_object.parent.name == "microscopy" and
```
dyt811, line 655: WOW. This is interesting. I never knew `str.endswith()` can accept a tuple (but not a list). Had to look it up: https://stackoverflow.com/a/18351977 . TIL.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
+                            (str(path_object.parent).replace(str(path_data) + os.sep, '').startswith("sub"))):
```
dyt811, line 656: This expression SHOULD be greatly simplified. I think all it is trying to do was to 1) get folder path, 2) remove everything before path_data (via the str replace call) including the OS separator? and check that the folder starts with "sub". I believe this can be vastly simplified with Path object through parent and stem etc. Please change that here.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
+                        force_index.append(str(path_object.parent).replace(str(path_data) + os.sep, ''))
```
dyt811, line 657: This line suggests.. for force index, we do not need full path information. Again,we can easily probably easily take out the str/replacement part. Please further refactor.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
                     # CT-scan
-                    if (file.endswith(ext_ct) and file.split('.')[0].endswith(suffix_ct) and
-                            (os.path.basename(root) == "anat" or os.path.basename(root) == "ct") and
-                            (root.replace(path_data, '').startswith("sub"))):
-                        force_index.append(os.path.join(root.replace(path_data, '')))
+                    if (path_object.name.endswith(ext_ct) and path_object.name.split('.')[0].endswith(suffix_ct) and
+                            (path_object.parent.name == "anat" or path_object.parent.name == "ct") and
+                            (str(path_object.parent).replace(str(path_data) + os.sep, '').startswith("sub"))):
```
dyt811, line 661: Refactor this logic as before with regarding to simplify the `str.replace` portion. There shouldn't be the need to stringify path object just take a portion out SO that we can check for them starting with `sub` string.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
+                        force_index.append(path_object.name)
+                    if (path_object.name.endswith(ext_microscopy) and path_object.parent.name == "microscopy" and
+                            (str(path_object.parent).replace(str(path_data) + os.sep, '').startswith("sub"))):
+                        force_index.append(str(path_object.parent).replace(str(path_data) + os.sep, ''))
                     # CT-scan
-                    if (file.endswith(ext_ct) and file.split('.')[0].endswith(suffix_ct) and
-                            (os.path.basename(root) == "anat" or os.path.basename(root) == "ct") and
-                            (root.replace(path_data, '').startswith("sub"))):
-                        force_index.append(os.path.join(root.replace(path_data, '')))
+                    if (path_object.name.endswith(ext_ct) and path_object.name.split('.')[0].endswith(suffix_ct) and
```
dyt811, line 659: Please use https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffixes to handle multiple extensions.

ivadomed/loader/utils.py
```diff
@@ -646,25 +647,25 @@ def create_bids_dataframe(self):
...
+                            (path_object.parent.name == "anat" or path_object.parent.name == "ct") and
+                            (str(path_object.parent).replace(str(path_data) + os.sep, '').startswith("sub"))):
+                        force_index.append(str(path_object.parent).replace(str(path_data) + os.sep, ''))
```
dyt811, line 662: 🔁 ☝️

ivadomed/loader/utils.py
```diff
@@ -634,7 +635,7 @@ def create_bids_dataframe(self):
         pybids.config.set_option('extension_initial_dot', True)
 
         for path_data in self.paths_data:
-            path_data = os.path.join(path_data, '')
+            path_data = Path(path_data).joinpath('')
```
dyt811, line 638: Please change to `path_data: Path` so IDE/MyPy can auto check type congruency.

ivadomed/loader/utils.py
```diff
@@ -689,10 +690,10 @@ def create_bids_dataframe(self):
 
             # Warning if no subject files are found in path_data
             if df_next[~df_next['path'].str.contains('derivatives')].empty:
-                logger.warning("No subject files were found in '{}' dataset. Skipping dataset.".format(path_data))
+                logger.warning("No subject files were found in '{}' dataset. Skipping dataset.".format(str(path_data)))
```
dyt811, line 693: f string

ivadomed/loader/utils.py
```diff
@@ -766,11 +767,12 @@ def add_tsv_metadata(self, df, path_data, layout):
         # TODO: use pybids function after BEP microscopy is merged in BIDS
         # TODO: verify merge behavior with EEG and DWI scans files, tested with anat and microscopy only
         df_scans = pd.DataFrame()
-        for root, dirs, files in os.walk(path_data):
-            for file in files:
-                if file.endswith("scans.tsv"):
-                    df_temp = pd.read_csv(os.path.join(root, file), sep='\t')
+        for path_object in Path(path_data).glob("**/*"):
+            if path_object.is_file():
+                if path_object.name.endswith("scans.tsv"):
```
dyt811, line 772: NICE. I like how the changes made it so much easier for human readability.

ivadomed/loader/utils.py
```diff
@@ -749,9 +750,9 @@ def add_tsv_metadata(self, df, path_data, layout):
         # TODO: use pybids function after BEP microscopy is merged in BIDS
         if 'sample' in df:
             df['sample_id'] = "sample-" + df['sample']
-        fname_samples = os.path.join(path_data, "samples.tsv")
-        if os.path.exists(fname_samples):
-            df_samples = pd.read_csv(fname_samples, sep='\t')
+        fname_samples = Path(path_data).joinpath("samples.tsv")
```
dyt811, line 753: `/` to simplify.

dyt811 (changes requested):
Please change these areas for substantially improved readability. The purpose of changing to pathlib is not brute force conversion but to leverage some of their existing cool features (like the `.suffixes` and `.glob` to really make the code base easier to follow and more intuitive and compatible across OSes. Thanks for taking the time to make these changes.

ivadomed/loader/utils.py
cakester: filename

ivadomed/loader/utils.py
cakester: suffixes is a list

ivadomed/loader/utils.py
```diff
@@ -178,7 +179,7 @@ def get_new_subject_file_split(df, split_method, data_testing, random_seed,
 
     # save the subject distribution
     split_dct = {'train': train_lst, 'valid': valid_lst, 'test': test_lst}
-    split_path = os.path.join(path_output, "split_datasets.joblib")
+    split_path = Path(path_output, "split_datasets.joblib")
```
dyt811, line 182: Good use of flexible constructor that support both string and path. TIL. 👍

ivadomed/loader/utils.py
dyt811: So I meant more for the later part. ` and path_object.name.split('.')[0].endswith(suffix_ct)` where path_object.name.split('.') is more for inferring the suffixes.

cakester: in this case it is not getting suffixes, path_object.name.split('.')[0] is stem, but if you were to get stem for a.tar.gz you would get a.tar, basically you are getting the part in front of the first dot

dyt811 (approved):
Great to see that this class has since been drastically simplified and even review process is going to be a lot faster.

Issue #910, opened 2021-08-11
Batch size affects the dice loss 


gbazad93:
First, I would like to thank you for the excellent implementation. The [dice loss function](https://github.com/ivadomed/ivadomed/blob/92b97bb93e947c67ddc21d6bf6b23d5bfc905241/ivadomed/losses.py#L41) implemented in the library seems to be prone to the batch size. When the batch contains a sample with a tiny target, its contribution to the loss function is comparatively less than a sample with a big object, while each instance should have the same weight on the loss function. Is there a way to control such behavior? Or does the loss function need to be revised with a loop on instances?

charleygros:
Hi @gbazad93, Many thanks for your message 👍 

You are correct: the current implementation considers the entire batch to compute the dice loss.

In order to consider each sample separately, a basic work-around would be to have a batch size = 1.
Otherwise, we could consider adjusting the current implementation by, as you suggested, adding a for loop on instances and returning the averaged dice across the samples of the batch.

Would you like to create a pull request with this feature? Otherwise, let us know, we would be happy to help with this. Cheers.

gbazad93:
@charleygros thanks for the response. I have submitted the pull request. Would appreciate your comments.

charleygros:
> 
> 
> @charleygros thanks for the response. I have submitted the pull request. Would appreciate your comments.

Fantastic, thanks @gbazad93, working on this today 👍

PR #911, opened 2021-08-16
Dice loss modified to eliminate the batch size affects (issue #910)

gbazad93:
#### GitHub

#### PR contents

## Description
The dice loss function was modified to eliminate the batch size effect. When the batch contains a sample with a tiny target, its contribution to the loss function is comparatively less than a sample with a big object, while each instance should have the same weight on the loss function. Thus, the current pull request aims to address this issue. 

## Linked issues
[##910](https://github.com/ivadomed/ivadomed/issues/910)


hermancollin:
I think this is an important feature. I am currently re-running a training I did a few weeks ago using this PR and I am very curious to see if it will affect the model's performance. I predict it will, because my dataset contains a few near-empty samples.

gbazad93:
> 
> 
> I think this is an important feature. I am currently re-running a training I did a few weeks ago using this PR and I am very curious to see if it will affect the model's performance. I suspect it will, because my dataset contains a few near-empty samples.
> 
> update: unfortunately, my model does not perform as well as before. FWI I used MultiClassDiceLoss (wrapper over the dice loss class)

Hi, @hermancollin thanks for your experiment. Could you please specify your batch size? And by performance, do you mean results on the validation or test set? In the case of data set with relatively small or empty target samples, I wonder if the modification might be responsible for not complete convergence. Training and validation loss diagrams might be a helpful metric to observe this behavior.

charleygros:
Implementation looks good to me, many thanks @gbazad93, great contribution!

I wonder if we should keep this (eg average per sample or per batch) as a parameter of the Loss function, ie let the user choose how the loss is computed. I feel it's really task-specific. We could have a default behviour of course.
What do you think @gbazad93 / @andreanne-lemay / @mariehbourget / @uzaymacar ? Cheers

gbazad93:
@charleygros, thanks for the update. In a medical case, the ultimate objective is to produce the segmentation mask for an individual case. Thus, sample-wise averaging might provide more realistic results than batch-wise. On the other hand, averaging the batch can provide a smooth convergence and is useful for training. With further scrutinized the [Pytorch loss implementations](https://github.com/pytorch/pytorch/blob/22f78144c70d9d215bc390805aa8d0657e548b85/torch/nn/functional.py) I found that Pytorch uses average over batch. Have a look at the following experiments:
image

So, as @charleygros suggested it can be optional for specific dataset and necessary for the test set.

hermancollin:
> Hi, @hermancollin thanks for your experiment. Could you please specify your batch size? And by performance, do you mean results on the validation or test set? In the case of data set with relatively small or empty target samples, I wonder if the modification might be responsible for not complete convergence. Training and validation loss diagrams might be a helpful metric to observe this behavior.

I used a batch size of 6. The results were worst on both validation and testing sets. To be fair, I suspect the hyperparameters were optimized for batch-wise averaging. Obtaining the same results as before with this PR should be possible but might require more optimization.

I am also in favor of making this feature optional. Batch-wise averaging seems to work well currently with default architectures.

gbazad93:
@hermancollin Thanks for the update. My initial issue was for the performance on the validation set. For the training purpose I think the evidence shows that the  batch-wise normalization can provide a better generalization performance. Thanks again for your experiment.

hermancollin:
@gbazad93 Ah I understand the issue now. In that case, I agree that computing sample-wise averaged Dice on the test set makes sense.

charleygros:
> So, as @charleygros suggested it can be optional for specific dataset and necessary for the test set.

Agree with this suggestion, let's do it :-)

Note that we have something similar already implemented at test time: the [evaluation pipeline on the 3D reconstructed images](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/evaluation.py#L114) --> which computes the evaluation metrics on the full 3D MRI image, instead of "per batch" or "per sample".

gbazad93:
> 
> 
> > So, as @charleygros suggested it can be optional for specific dataset and necessary for the test set.
> 
> Agree with this suggestion, let's do it :-)
> 
> Note that we have something similar already implemented at test time: the [evaluation pipeline on the 3D reconstructed images](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/evaluation.py#L114) --> which computes the evaluation metrics on the full 3D MRI image, instead of "per batch" or "per sample".

@charleygros now it is optional. Yes just noticed the evaluation pipeline is already considered this case. Thanks

ivadomed/losses.py
```diff
@@ -51,16 +51,19 @@ class DiceLoss(nn.Module):
     Attributes:
         smooth (float): Value to avoid division by zero when images and predictions are empty.
```
hermancollin, line 52:
```suggestion
        smooth (float): Value to avoid division by zero when images and predictions are empty.
        sample_wise (bool): If set, dice loss is averaged over sample instead of over batch
```

charleygros:
> @charleygros now it is optional. Yes just noticed the evaluation pipeline is already considered this case. Thanks

Looks good to me!

Would be great to have someone else review before merging as it is quite a crucial part of the ivadomed creature, @uzaymacar @mariehbourget @andreanne-lemay ? :-)

rezazad68:
@gbazad93 implementation looks good. I only concern about the extra computation burden. Since the sample-wise loss only applies in rare cases, the current implementation adds `two unsqueeze and one average operation`, which is unnecessary in the batch cases (consider that the loss needs to be calculated for each batch). Simple conditions and grouping into two separate parts might be computationally efficient.

mariehbourget:
Thank you @gbazad93 for this contribution.
I tested it and everything looks good.
As @rezazad68 suggested, it may be worth it to test the computation burden.
Regarding the new option, @charleygros is sample_wise an optional parameter that we would want to add to the config file and the entire pipeline (with default behavior sample_wise=False)?

charleygros:
>  Regarding the new option, @charleygros is `sample_wise` an optional parameter that we would want to add to the config file and the entire pipeline (with default behavior `sample_wise=False`)?

Happy either ways.
I think we don't need to make it appear in the config files etc. since that's quite an advanced parameter, and will appear in the online doc anyway.
But please have the conversation during meeting if unsure :-)

mariehbourget:
> I think we don't need to make it appear in the config files etc. since that's quite an advanced parameter, and will appear in the online doc anyway.

My only concern about not having it in the config file is that the parameter becomes "invisible" and could affect reproducibility of the experiments.

gbazad93:
Hi @charleygros and @uzaymacar, I hope you're doing well! Just a friendly reminder about this pull request. Please let me know if there’s anything else I can address to move it forward. Thanks!

jcohenadad:
@gbazad93 ivadomed is not maintained anymore, sorry for the trouble.

PR #925, opened 2021-08-26
Release v2.8.0 

dyt811:
Added 2.8.0 PR summary from Changelog-Neuropoly and updated version.txt


mariehbourget (commented):
Thank you @dyt811 for taking care of this.
I have a question though regarding the changelog, it is normal that PRs from labels like "feature" or "installation" are not included?

jcohenadad:
> I rolled back 2.8.0 release milestone until we get this resolved so the changelog would be more inclusive (we have been missing tons of stuff from prior release in release note I believe).

is that a problem if the tag is still present? (can it be deleted as well in the git history?):

image

dyt811:
> is that a problem if the tag is still present? (can it be deleted as well in the git history?):

Yes, apparently it can. I just deleted the tag and last remnant of the troubled release has been erased for now.

dyt811:
Alright. I used a locally patched version of `changelog` to generate these updated change logs and I think they are a lot more comprehensive than before.

mariehbourget (approved):
LGTM! Thank you @dyt811!

PR #937, opened 2021-09-02
Fix dropout_rate key in models.py

### LINKED ISSUE #926: Wrong key for dropout in config file
## Issue description
The dropout key in configuration file does not correspond to the expected key in models' definitions.

### Current behavior
In config files (including `config_default.json`) and in the documentation, the key for dropout in the model parameter is `dropout_rate`. This key is passed to the model in training in `model_params` here: https://github.com/ivadomed/ivadomed/blob/bc1b543fd3df448a18b8ed87f99f18ca19af2fe7/ivadomed/training.py#L100
However, all models are expecting the key to be `drop_rate`.
Ex:ample here in Unet model: https://github.com/ivadomed/ivadomed/blob/bc1b543fd3df448a18b8ed87f99f18ca19af2fe7/ivadomed/models.py#L489
So `dropout_rate` from the config is never used and the default `drop_rate` of each model applies.

### Expected behavior
For the models to use the correct dropout from the config files.
I see 3 options for fixing it.

**Option 1:**
- Update all the config files in `ivadomed/config` with the key `drop_rate` instead of `dropout_rate`.
- Update the documentation accordingly.

**Option 2:**
- Update all the config files in `ivadomed/config` with the key `drop_rate` instead of `dropout_rate` (same as option 1).
- Update the documentation accordingly (same as option 1).
- Make a key change in `config_manager.py` to change `dropout_rate` to `drop_rate`, so both are supported.

**Option 3:**
- Change the key in all models' definitions to `dropout_rate`.

I'm more in favor of option 1 for retrocompatibility purposes i.e. old training with the wrong key will remain the same and still be wrong if the training is run again.
Option 2 and 3 would change the behavior of old config files.

Let me know what you think.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR fixes issue #926 where the key for dropout was different between the config files and the expected keys in classes of `models.py`.
- As discussed, I updated the key in all models and classes of `models.py` from `drop_rate` to `dropout_rate`.
- I updated the documentation of the default Unet model with default values for `dropout`, `bn_momentum` and `depth`. Those values are applied by default by the `config_manager` from the `config_default.json` file when the key is missing from the config file.

I tested the following architectures: `default_model` (Unet), the `FiLMedUnet`, the `Modified3DUnet` and `HeMISUnet`.

### Questions:
@charleygros, the default value from the `config_default.json` for dropout is `0.3`, whereas the default values in `models.py` are `0`, `0.4` or `0.6`. I don't think that is an issue i.e if the model is called via the config file, the default will be 0.3 but if the model is called directly, the default value will be `0`, `0.4` or `0.6` depending on the model as documented in the docstrings. Let me know what you think.

Also the initialization of `FiLMedUnet(Unet)` looks a bit strange to me:
https://github.com/ivadomed/ivadomed/blob/d71a1c4711d03da7c61a473f571134b262684073/ivadomed/models.py#L528-L530
The `super().__init__` with Unet uses default values, I'm not sure but it would make more sense to me to use to values define by `def __init__`.
ex:
```
def __init__(self, in_channel=1, out_channel=1, depth=3, dropout_rate=0.4,
             bn_momentum=0.1, n_metadata=None, film_layers=None, is_2d=True, n_filters=64, **kwargs):
    super().__init__(in_channel=in_channel, out_channel=out_channel, depth=depth, dropout_rate=dropout_rate, bn_momentum=bn_momentum)
```
At the end of the day, it does not change the behavior and the right values from the config file are applied in the model, so maybe I'm being picky here.

### To test the PR:
- Run a training with a dropout value different from 0.3 in the config_file.
- print the model [after this line](https://github.com/ivadomed/ivadomed/blob/d71a1c4711d03da7c61a473f571134b262684073/ivadomed/training.py#L101) and check the value of dropout in the model.

## Linked issues
Fixes #926


charleygros (commented):
Looks all good to me.

Few more instances of `drop_rate` in the repo:
https://github.com/ivadomed/ivadomed/blob/4460d1379b77ffd114d4b416fda0924b40a51e3e/testing/functional_tests/test_segment_volume.py
https://github.com/ivadomed/ivadomed/blob/959b54d7f0094479b64182a7a386978d0147618e/testing/unit_tests/test_HeMIS.py

Otherwise, all good to go :)

cakester (approved):

mariehbourget:
> Few more instances of `drop_rate` in the repo:
> https://github.com/ivadomed/ivadomed/blob/4460d1379b77ffd114d4b416fda0924b40a51e3e/testing/functional_tests/test_segment_volume.py
> https://github.com/ivadomed/ivadomed/blob/959b54d7f0094479b64182a7a386978d0147618e/testing/unit_tests/test_HeMIS.py

Thanks @charleygros, I totally missed them, they are correct now.
Also, just to be sure, can you let me know you thoughts about the 2 questions in the first comment before I merge the PR. Thank you!

charleygros:
> @charleygros, the default value from the `config_default.json` for dropout is `0.3`, whereas the default values in `models.py` are `0`, `0.4` or `0.6`. I don't think that is an issue i.e if the model is called via the config file, the default will be 0.3 but if the model is called directly, the default value will be `0`, `0.4` or `0.6` depending on the model as documented in the docstrings. Let me know what you think.

I agree. Let's put `0.3` everywhere: config files and Model default values.

charleygros:
> The `super().__init__` with Unet uses default values, I'm not sure but it would make more sense to me to use to values define by `def __init__`.
> ex:
> 
> ```
> def __init__(self, in_channel=1, out_channel=1, depth=3, dropout_rate=0.4,
>              bn_momentum=0.1, n_metadata=None, film_layers=None, is_2d=True, n_filters=64, **kwargs):
>     super().__init__(in_channel=in_channel, out_channel=out_channel, depth=depth, dropout_rate=dropout_rate, bn_momentum=bn_momentum)
> ```
> 
> At the end of the day, it does not change the behavior and the right values from the config file are applied in the model, so maybe I'm being picky here.

Fully agree. I dont think you are being picky: it would prevent from mistakes (eg changing at one place without modifying the other one).

mariehbourget:
Thank you @charleygros, corrections are done.

charleygros (approved):

PR #952, opened 2021-09-22
add conda environment and documentation for GPU support

taowa:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] (n/a) I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [x] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
A user that wishes to install ivadomed using conda will install a version that does not include GPU support. This updates the documentation to make GPU support more obvious to enable.

PR #955, opened 2021-09-23
Test GPU code on `romane`

kousu:
#### GitHub

#### PR contents

## Description

**DO NOT MERGE AS IS** please. I need to have a PR open to be able to test the self-hosted runner.

## Linked issues

* https://github.com/neuropoly/computers/issues/202
* #951 
* #952

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        os: [ "macos-latest", "windows-latest", "ubuntu-18.04", "ubuntu-20.04",  "macos-10.15" ]
+        os: [ "gpu" ]
         python-version: [ 3.6, 3.7, 3.8 ]
         test-name:
           - integration-test
@@ -59,7 +59,7 @@ jobs:
       - name: Install dependencies
         uses: py-actions/py-dependency-install@v2
         with:
-          path: "requirements_dev.txt"
+          path: "requirements_gpu.txt"
 
       # Step 3: install additional pytest/dev related dependencies
       - name: Install IvadoMed Main Modules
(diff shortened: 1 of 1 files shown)


kousu:
I tried to get this to work in https://github.com/neuropoly/computers/issues/202 but I haven't figured it out: I can use `romane.neuro.polymtl.ca` as a runner but it doesn't behave much at all like the Github runners, because Github left the allocate and cleanup parts as manual tasks. I think they aren't really expecting our kind of use case; or, if they are, they expect it's only for private projects where the hacking potential is limited (but not zero!), or that you are renting GPUs from The Cloud and are dynamically creating and destroying those as part of your workflows (which begs the question: doesn't that just mean that, instead of potentially getting `romane` hacked by strangers, it potentially gets your bank account emptied by letting strangers spawn hundreds of GPU servers on your behalf?).

Anyway I don't know how to get this working. Sorry @dyt811; I tried. Maybe you can take a shot at it? You said you were more comfortable with GitlabCI right? GitlabCI is I guess a bit safer by default because it uses containers by default, unlike Actions which leaves containerizing up to us. It might be instructive to see what setting up GitlabCI with GPUs looks like in comparison; does it hit any of the same pain points I hit?

dyt811:
@kousu thanks for looking into this. Yeah, making GPU available yet preventing abuse (even CPU got "hijacked" for mining sigh... ) I was thinking a private mirror repo on GitLab with runner setup on a GPU PC/Romane etc so at least we can pull the master from GitHub Public repo, send it to GitLab runner with GPU and manually trigger this type of tests once in a while to check for things/bugs/common problems. Some of these processes can be triggered by webhook/GitHub APIs so definitely doable but having issues raised by GitLab probably won't auto raise issues/warnings to GitHub so that might get complicated real soon.

kousu:
Sure. Let's close this for now and re-open it when we can devote some time to getting it solid.

PR #964, opened 2021-10-05
Implement CrossEntropyLoss

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR aims to implement `CrossEntropyLoss`.
I implemented it and it works great on axon/myelin segmentation task. However, while reviewing my work, I think there is a mistake in the implementation and I would like to discuss it further, hence the draft PR.

So [CrossEntropyLoss](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html) is a combination of `LogSoftmax` and `NLLLoss` in one single class.
This brings 3 points:

1) I think it means that the final activation of the model should be a `LogSoftmax` (instead of `sigmoid` I am using), and the output of the model in `LogSoftmax` should be fed to the `NLLLoss` directly instead of `CrossEntropyLoss`.

Other options would be to use `Softmax` as the final activation of the model and a `log()` before `NLLLoss`, or to skip the final activation and go directly to `CrossEntropyLoss`.

2) In my current implementation, I have an option to add the background class in the same way as it is done in `GeneralizedDiceLoss`. However, I don't think it makes sense for `CrossEntropyLoss` to not have the background included.

3) Because of the previous point, I think we should keep the background class from the output predictions of the model for `CrossEntropyLoss`. Currently, the background class is removed from the model predictions in the decoder [here](https://github.com/ivadomed/ivadomed/blob/ac29c7cbe299e6478854823ed6bd818c0e5f9642/ivadomed/models.py#L465).

Please let me know if you have more insights on this. Thanks!

### DIFF
--- a/ivadomed/losses.py
+++ b/ivadomed/losses.py
@@ -78,6 +78,44 @@ class BinaryCrossEntropyLoss(nn.Module):
         return self.loss_fct(prediction, target.float())
 
 
+class CrossEntropyLoss(nn.Module):
+    """(`CrossEntropyLoss <https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#CrossEntropyLoss`__).
+
+    Args:
+        include_background (bool): If True, then an extra channel is added, which represents the background class.
+        weight (list of floats): List containing the weight given to each class. Default: None.
+
+    Attributes:
+        include_background (bool): If True, then an extra channel is added, which represents the background class.
+        device (str): Indicate "cuda" if GPU is available, else "cpu".
+        weight (Tensor): 1D Tensor containing the weight given to each class.
+        loss_fct (CrossEntropyLoss): Cross entropy loss function from torch library.
+    """
+
+    def __init__(self, include_background=True, weight=None):
+        super(CrossEntropyLoss, self).__init__()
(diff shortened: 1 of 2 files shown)


naga-karthik:
I have some things to say about those 3 points:
> or to skip the final activation and go directly to CrossEntropyLoss.
1. Out of all the possibilities you mentioned, I think the one quoted above would be the simplest as you don't need to worry about what final activation to use or think about `NLLLoss`. See also [here](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#:~:text=The%20input%20is%20expected%20to%20contain%20raw%2C%20unnormalized%20scores%20for%20each%20class.) that `nn.CrossEntropyLoss` expects a raw, unnormalized input i.e. the logits (not even sigmoid).
2. Yes, since you mentioned that you had 3 classes, if the background is removed, then `CrossEntropyLoss` doesn't make sense anymore, you could use `BCELoss` which is more suitable.
3. I see that the background class is removed in that implementation linked. Are you overriding that to use CrossEntropy for your use case?

Also, [here's](https://cs230.stanford.edu/blog/pytorch/#loss-function) an alternative manual implementation I found, in case it helps. Let me know how it goes!

PR #968, opened 2021-10-08
Fix pytest problem

### LINKED ISSUE #963: Unit test failure only when running testing\functional_tests\test_download_data.py in conjunction with other unit tests
## Issue description

`test_download_data_no_dataset_specified` fails when running with other tests, but not when just launching itself.

### Current behavior

- Oddly, this particular unit test will fail when we run into in combination with various other tests but will fail when running only by itself. 
- We originally suspect it was something carried over from running multiple tests but the cause is yet to be determined. 
- Reproducible at least on Windows/Ubuntu systems. 

### Expected behavior
 

- It should pass all the time. 

### Steps to reproduce

- Run test_download_data_no_dataset_specified.py and it will pass. 
- Run pytest on root folder, it will fail. 

<details>
<summary>Pytest SystemExit:0 will break pytest during system wide runs but not during individual pytest runs</summary>

See detailed GitLab CI Log here: 
[outputlog.txt](https://github.com/ivadomed/ivadomed/files/7288893/outputlog.txt)

</details>

## Environment

- Ubuntu 20 as well as Windows 10

### System description

- Typically VENV setup or Conda ENV setup, except with GPU. Noticed this issue while running GPU tests. CI so far does not appear to suffer from this problem. 

### Installed packages

<details open>
<summary> Output of <code>pip freeze</code> </summary>

```

(ivadomed) C:\git\ivadomed>pip list
Package                       Version             Location
----------------------------- ------------------- ----------------
absl-py                       0.12.0
alabaster                     0.7.12
appdirs                       1.4.4
atomicwrites                  1.4.0
attrs                         20.3.0
autobahn                      21.3.1
Automat                       20.2.0
Babel                         2.9.1
bids-validator                1.7.1
cachetools                    4.2.1
certifi                       2020.12.5
cffi                          1.14.5
cfgv                          3.2.0
changelog-neuropoly           1.1.0               c:\git\changelog
chardet                       4.0.0
click                         7.1.2
colorama                      0.4.4
constantly                    15.1.0
cryptography                  3.4.7
csv-diff                      1.1
cycler                        0.10.0
decopatch                     1.4.8
decorator                     4.4.2
dictdiffer                    0.8.1
distlib                       0.3.1
docopt                        0.6.2
docutils                      0.16
filelock                      3.0.12
future                        0.18.2
google-auth                   1.29.0
google-auth-oauthlib          0.4.4
grpcio                        1.37.0
h5py                          2.10.0
hyperlink                     21.0.0
identify                      2.2.4
idna                          2.10
imageio                       2.9.0
imagesize                     1.2.0
incremental                   21.3.0
iniconfig                     1.1.1
ivadomed                      2.8.0               c:\git\ivadomed
Jinja2                        2.11.3
joblib                        1.0.1
jsonpointer                   2.1
kiwisolver                    1.3.1
loguru                        0.5.3
makefun                       1.11.3
Markdown                      3.3.4
MarkupSafe                    1.1.1
matplotlib                    3.4.1
meshio                        4.4.6
mock                          4.0.3
networkx                      2.5.1
nibabel                       3.2.1
nodeenv                       1.6.0
num2words                     0.5.10
numpy                         1.20.2
oauthlib                      3.1.0
onnxruntime                   1.4.0
packaging                     20.9
pandas                        1.2.4
patsy                         0.5.1
Pillow                        8.2.0
pip                           21.0.1
pluggy                        0.13.1
pre-commit                    2.10.1
protobuf                      3.15.8
py                            1.10.0
pyasn1                        0.4.8
pyasn1-modules                0.2.8
pybids                        0.13
pycparser                     2.20
Pygments                      2.9.0
pypandoc                      1.5
pyparsing                     2.4.7
pytest                        6.2.3
pytest-cases                  3.6.1
pytest-console-scripts        1.1.0
pytest-ordering               0.6
pytest-repeat                 0.9.1
python-dateutil               2.8.1
pytz                          2021.1
pyvista                       0.31.3
pyvistaqt                     0.5.0
PyWavelets                    1.1.1
PyYAML                        5.4.1
QtPy                          1.9.0
requests                      2.25.1
requests-oauthlib             1.3.0
rsa                           4.7.2
scikit-image                  0.18.1
scikit-learn                  0.24.1
scipy                         1.6.2
scooby                        0.5.7
seaborn                       0.11.1
setuptools                    52.0.0.post20210125
six                           1.15.0
snowballstemmer               2.1.0
Sphinx                        3.5.4
sphinx-jsonschema             1.16.8
sphinxcontrib-applehelp       1.0.2
sphinxcontrib-devhelp         1.0.2
sphinxcontrib-htmlhelp        1.0.3
sphinxcontrib-jsmath          1.0.1
sphinxcontrib-qthelp          1.0.3
sphinxcontrib-serializinghtml 1.1.4
SQLAlchemy                    1.3.24
tensorboard                   2.5.0
tensorboard-data-server       0.6.0
tensorboard-plugin-wit        1.8.0
threadpoolctl                 2.1.0
tifffile                      2021.4.8
toml                          0.10.2
torch                         1.5.0+cpu
torchvision                   0.6.0+cpu
tqdm                          4.60.0
transforms3d                  0.3.1
tuna                          0.5.6
Twisted                       21.2.0
twisted-iocpsupport           1.0.1
txaio                         21.2.1
urllib3                       1.26.4
virtualenv                    20.4.4
vtk                           9.0.2
Werkzeug                      1.0.1
wheel                         0.36.2
win32-setctime                1.0.3
wincertstore                  0.2
wslink                        0.2.0
zope.interface                5.4.0

```

</details>

cakester:
#### GitHub

#### PR contents

## Description

fix test_download_data_no_dataset_specified fails when running from root with command only 'pytest'

## Linked issues

resolves #963

### DIFF
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -265,10 +265,7 @@ def get_arguments(parser, args):
             ["-d", "SOME_ARG", "--model", "SOME_ARG"]
     """
     try:
-        if args:
-            args = parser.parse_args(args)
-        else:
-            args = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
+        args = parser.parse_args(args)
     except SystemExit as e:
         if e.code != 0:  # Calling `--help` raises SystemExit with 0 exit code (i.e. not an ArgParseException)
             raise ArgParseException('Error parsing args')
--- a/testing/functional_tests/test_download_data.py
+++ b/testing/functional_tests/test_download_data.py
@@ -23,12 +23,8 @@ def test_download_data():
 
 
 def test_download_data_no_dataset_specified():
-    if sys.argv[1:]:
-        with pytest.raises(ArgParseException, match=r"Error parsing args"):
-            download_data.main()
-    else:
-        with pytest.raises(SystemExit, match='0'):
-            download_data.main()
+    with pytest.raises(ArgParseException, match=r"Error parsing args"):
+        download_data.main()
 
 
 def teardown_function():
--- a/testing/functional_tests/test_download_data.py
+++ b/testing/functional_tests/test_download_data.py
@@ -1,6 +1,7 @@
 import logging
 import pytest
 import os
+import sys
 from testing.functional_tests.t_utils import __tmp_dir__, create_tmp_dir
 from testing.common_testing_util import remove_dataset, remove_tmp_dir
 from ivadomed.scripts import download_data
@@ -22,8 +23,12 @@ def test_download_data():
 
 
 def test_download_data_no_dataset_specified():
-    with pytest.raises(ArgParseException, match=r"Error parsing args"):
-        download_data.main()
+    if sys.argv[1:]:
+        with pytest.raises(ArgParseException, match=r"Error parsing args"):
+            download_data.main()
+    else:
+        with pytest.raises(SystemExit, match='0'):
+            download_data.main()
 
 
 def teardown_function():


dyt811:
This has been a rather annoying and very peculiar issue observed. We are still not 100% sure its cause but the fact when no parameter is past to the `pyest` call and it will result in failure where as any other way to call pytest on the same unit test will pass sounds like an extreme corner case.

dyt811:
@joshuacwnewton no worry man. I think Hanna probably resolved this. I am testing it now and if it all goes well we should be good. It is just a really really odd corner case that trip up manual testing scenarios. 

I think the reason we didn't see this in CI is because we typically ran `pytest . -v --cov ivadomed/ --cov-report term-missing`  instead of `pytest`. The fact that the lack of additional parameters trips it up is rather odd.

dyt811:
Looking all good.

dyt811 (approved):

joshuacwnewton:
Wait, sorry, I don't think we should merge this...

joshuacwnewton:
It turns out that the `SystemExit(0)` is actually a bug in the code. 

What happens is this:

1. Pytest is called with no args. (therefore, `sys.argv[1:] == None`)

2. The test `test_download_data_no_dataset_specified` is run:

    https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/testing/functional_tests/test_download_data.py#L24-L26

2. `download.main()` is called, and it will have `args=None` by default:

    https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/ivadomed/scripts/download_data.py#L266

3. `get_arguments()` is called, also with `args=None`:

    https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/ivadomed/scripts/download_data.py#L274

4. Inside `get_arguments()`, the `else:` case is triggered, since `args=None`:

   https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/ivadomed/utils.py#L267-L271

5. Since `sys.argv[1:] == None` (no Pytest args), the second `else:` case is run, and `download_data`'s  `--help` will be run!! This is what causes the `SystemExit(0)`.

So, IMO what we actually need to do:

* Leave the test as-is (since `SystemExit(0)` is a bug, we don't want to include that in the test)
* Modify the code to prevent the Pytest args from affecting whether or not `--help` is called for `download_data`

dyt811:
@cakester could you also expand out Line 271 of utils.py and make it much more clear? It should also help out future 

So if I am following correctly, the get_arguments() really shouldn't have used `sys.argv` call? That caused the mixing of `sys.argv` when calling higher script that calls `download_data` vs calling `download_data` directly causing mixing parameters from the two scripts? I guess the Line 271 in Utils.py should really be `args[1:]`? Also... if it is None, it will probably throw index error so all in all. 271 should be rehauled majorly?

cakester:
@joshuacwnewton I have a question regarding to the suggested change, right now the tests are failing for test_automated_training, the reason is that when we call ivadomed_automated_training, we are actually calling the main function inside script/automated_training.py, check the setup.py below
` 47     entry_points={
 48         'console_scripts': [
 49             'ivadomed=ivadomed.main:run_main',
 50             'ivadomed_prepare_dataset_vertebral_labeling=ivadomed.scripts.prepare_dataset_vertebral_labeling:main',
 51             'ivadomed_automate_training=ivadomed.scripts.automate_training:main',
 52             'ivadomed_compare_models=ivadomed.scripts.compare_models:main',
 53             'ivadomed_visualize_transforms=ivadomed.scripts.visualize_transforms:main',
 54             'ivadomed_convert_to_onnx=ivadomed.scripts.convert_to_onnx:main',
 55             'ivadomed_extract_small_dataset=ivadomed.scripts.extract_small_dataset:main',
 56             'ivadomed_download_data=ivadomed.scripts.download_data:main',
 57             'ivadomed_training_curve=ivadomed.scripts.training_curve:main',
 58             'ivadomed_visualize_and_compare_testing_models=ivadomed.scripts.visualize_and_compare_testing_models:main'
 59         ],
 60     },
`
by default, the main function uses args=None, but inside get_arguments we originally have 
` try: 
     if args: 
         args = parser.parse_args(args) 
     else: 
         args = parser.parse_args(args=None if sys.argv[1:] else ['--help']) `
and if you check the argparse library, if you pass args=None, it reads from sys.argv[1:],
`    def parse_known_args(self, args=None, namespace=None):
        if args is None:
            # args default to the system args
            args = _sys.argv[1:]
        else:
            # make sure that args are mutable
            args = list(args)`
but if you change get_arguments to
`args = parser.parse_args(args if args else ['--help'])`
then you would get args=['--help'] which would throw sys.exit(0) instead of fetching from commandline if args=None, that would change the behavior for get_arguments function. right now, i fixed test_automated_training by adding args=sys.argv[1:] inside the main function to force it to pick up args or else get_arguments won't give me args anymore, let me know if you find better solutions

joshuacwnewton:
>  right now the tests are failing for test_automated_training, the reason is that when we call ivadomed_automated_training, we are actually calling the main function inside script/automated_training.py

In that case, would it be possible to update the failing tests to use `.main()` as well? For example in [this test](https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/testing/functional_tests/test_automate_training.py#L16-L30), what if we made the following change:

```diff
-     ret = script_runner.run('ivadomed_automate_training', '--config', f'{file_config}', 
-                             '--config-hyper', f'{file_config_hyper}', 
-                             '--path-data', f'{__data_testing_dir__}', 
-                             '--output_dir', f'{__output_dir__}') 
+     from ivadomed.scripts import automate_training
+
+     automate_training.main(args=['--config', f'{file_config}', 
+                                  '--config-hyper', f'{file_config_hyper}', 
+                                  '--path-data', f'{__data_testing_dir__}', 
+                                  '--output_dir', f'{__output_dir__}']) 
```

Then `args` would no longer be `None`, and the scripts would not display the `--help`.

cakester:
> > right now the tests are failing for test_automated_training, the reason is that when we call ivadomed_automated_training, we are actually calling the main function inside script/automated_training.py
> 
> In that case, would it be possible to update the failing tests to use `.main()` as well?
> 
> For example in [this test](https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/testing/functional_tests/test_automate_training.py#L16-L30), what if we made the following change:
> 
> ```diff
> -     @pytest.mark.script_launch_mode('subprocess')
> - 
> -     ret = script_runner.run('ivadomed_automate_training', '--config', f'{file_config}', 
> -                             '--config-hyper', f'{file_config_hyper}', 
> -                             '--path-data', f'{__data_testing_dir__}', 
> -                             '--output_dir', f'{__output_dir__}') 
> 
> +     from ivadomed.scripts import automate_training
> +
> +     automate_training.main(args=['--config', f'{file_config}', 
> +                                  '--config-hyper', f'{file_config_hyper}', 
> +                                  '--path-data', f'{__data_testing_dir__}', 
> +                                  '--output_dir', f'{__output_dir__}']) 
> ```
> 
> Then `args` would no longer be `None`, and the scripts would not display the `--help`.

yes we can do that to make the tests pass, but when you actually run ivadomed_automate_training manually from command line, it runs main function with args=None

joshuacwnewton:
> nvm, I think you meant using automate_training.main in testcases on top of the change, sure no problem

:+1: 

One other thing... Earlier you suggested `args=sys.argv[:1]`... 

> right now, i fixed test_automated_training by adding args=sys.argv[1:] inside the main function to force it to pick up args or else get_arguments won't give me args anymore

That was because `entry_points` points to `main()` right, so `sys.argv[:1]` needed to be the default for `main()`?

I'm wondering why `entry_points` points to `main()` in the first place... Because, we do already have these blocks in each script:

https://github.com/ivadomed/ivadomed/blob/4b0b72c067976c016bba28cb051b644d319e62b0/ivadomed/scripts/automate_training.py#L815-L816

So, I'm wondering why `ivadomed` doesn't just point to the scripts themselves in `entry_points`. (Wouldn't it have the same result?)

If they did, then we could leave `main(args=None`), and instead do:

```python
 if __name__ == '__main__': 
     main(sys.argv[1:]) 
```

Which would mean in the `main(args=None)` case (tests, etc.), `sys.argv[1:]` wouldn't be picked up, which is probably a good thing...

----- 

Honestly though, the scope of the changes are getting a bit large... so you might be right about going for a simpler solution, rather than overhauling how args are parsed. 

> the solution I would propose to this problem is to always use args=None for parser.parse_args if args=None and forget about [--help]

joshuacwnewton:
Ah, I see... so there was in fact a good rationale for using subprocesses. Sorry about that... :sweat_smile:

Issue #861, opened 2021-07-13
New `torch` instructions will make it harder for downstream packages to depend on ivadomed's Python API


joshuacwnewton:
In #757, `ivadomed` stopped listing `torch`+`torchvision` as default dependencies, instead opting to ask users to choose between the GPU/CPU versions and [install them themselves](https://ivadomed.org/en/latest/installation.html#step-3-install-pytorch-1-5-and-torchvision). But, I think this change might make it harder for any downstream packages to depend on `ivadomed`:

*  If another Python package lists `ivadomed` in their package requirements (e.g. via `setup.py` -> `install_requires=`), then `ivadomed` will be broken out of the box (since `torch` is missing).
* The [replacement instructions](https://ivadomed.org/en/latest/installation.html#approach-2-venv) aren't really possible for a downstream package to follow (because another package's `setup.py` can't clone a repository to install `torch` via `requirements_gpu.txt`/`requirements.txt`).
* So, a downstream package will have to list `torch` as one of their dependencies _on behalf_ of `ivadomed`. 
    * This is what [`spinalcordtoolbox` is currently doing](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0bef2991f0c05251fbf43df8d1fc0360960565c2/requirements.txt#L28-L36), which is easy enough for us, because SCT's development has been somewhat intertwined with ivadomed's development.
    * But, this might discourage other packages from using ivadomed's Python API, since it might not be clear which functionalities are accessible and which aren't.

mathieuboudreau:
Hi from AxonDeepSeg 👋

I was been debugging for like an hour right now my ADS installation using IVADOMED as a (new) requirements, wondering why it was throwing me a missing torch package error when I was loading IVADOMED in python.

I'd assume you guys had included it in your setup, if it was a requirement to run IVADOMED.

So... yeah, +1 for this issue getting resolved. I saw no instructions before I had the idea to search "requirements" in your issues and found this.

mathieuboudreau:
(I know I was being a bad user by not reading your README thoroughly, sorry about that. But... not every user will 🤷)

mathieuboudreau:
> (I know I was being a bad user by not reading your ReadTheDocs thoroughly, sorry about that. But... not every user will 🤷)

ALTHOUGH...... in my defence.....

I just noticed that your ReadTheDocs button links to your stable release: https://ivadomed.org/en/stable/?badge=stable

[...]

PR #969, opened 2021-10-12
Update installation instruction to fit recent CUDA11 and torch 1.8+ push

dyt811:
…based approach

#### GitHub

#### PR contents

- [X] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

Current documentation has many branching path. It is best to use tabs to separate out these optional parallel contents. 
This also introduce the additional sphinx tab for documentation building that can be used elsewhere as well.
RTD.org has been updated as well to use requirement_dev.txt to not necessitate end user to install sphinx related dependencies.


dyt811:
All feedback incorporated. As for validate installation success, I will probably have to do that in a separate PR. We currently have a GitHub Project to semi-automate the tutorials testings as well so it will likely be a collaborative effort from that front to allow users and devs to validate installation end results as well as the actual training performance.

docs/source/installation.rst
```diff
@@ -2,147 +2,123 @@ Installation
...
+        This is the easiest way for personal computers.
 
-::
+        1. Create new conda environment using ``environment.yml`` file
 
-    export HDF5_USE_FILE_LOCKING='FALSE'
+        ::
 
-Save this change and restart your terminal to apply the change.
+            conda env create --file environment.yml
```
joshuacwnewton, line 24: Now that I think about it, I'm a bit confused about the purpose of this `environment.yml` file. 

https://github.com/ivadomed/ivadomed/blob/c226c49c384a8c322b99f9529f10b408b6687637/environment.yml#L1-L9

Since all of the dependencies are installed via `pip` + `requirements.txt`, it seems redundant with Step 3 of this page, where `requirements.txt` will be installed once more.

(Also, I notice the name is still hardcoded as `IvadoMedEnv` in this file, so it would probably need to be updated to `ivadomed_env` as well.)

docs/source/installation.rst
```diff
@@ -2,147 +2,123 @@ Installation
...
 
+.. tabs::
 
-.. warning::
-   If the default Python version installed in your system does not fit the version requirements, you might need to specify a version of Python associated with your virtual environment:
+    .. tab:: Pypi Installation
 
-   ::
+        Install ``ivadomed`` and its requirements from
+        `Pypi <https://pypi.org/project/ivadomed/>`__:
```
joshuacwnewton, line 73: Nitpick: `Pypi` -> [`PyPI`](https://pypi.org/)

docs/source/installation.rst
```diff
@@ -2,147 +2,123 @@ Installation
...
+        If you have a compatible NVIDIA GPU that supports CUDA11 and with the right driver installed, try run the following command:
 
-::
+        ::
 
-    cd ivadomed
-    pip install -e .
+           pip install -r requirements_gpu.txt
 
+        Please note that this must happens after the previous ``ivadomed`` installation step.
```
joshuacwnewton, line 118: I think that this line might be a little unnecessary, just because I think it's a fair expectation that people should follow steps in order, hehe.

docs/source/installation.rst
```diff
@@ -2,147 +2,123 @@ Installation
...
 
-(Optional) Step 5 for Developers Install pre-commit hooks
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+Additional Steps Required For ``ivadomed`` Developers: Additional Dependencies and ``pre-commit``
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
joshuacwnewton, line 121: This is a nitpicky thing, but what if we had the headers for this section be:

```rest
Developer-only installation steps
+++++++++++++++++++++++++++++++++

The additional steps below are only necessary for contributors to the ``ivadomed`` project.

Additional dependencies and ``pre-commit``
------------------------------------------

Blah blah blah...

``macOS``-specific development steps
------------------------------------

Blah blah blah...
```

That way, the titles can be a little shorter/easier to read, especially in the table-of-contents.

Screenshot from 2021-11-05 12-25-30

docs/source/installation.rst
```diff
@@ -2,147 +2,123 @@ Installation
...
-    conda activate IvadoMedEnv
+        .. warning::
+           If the default Python version installed in your system does not fit the version requirements, you might need to specify a version of Python associated with your virtual environment:
 
-Step 4 : Install from source
-++++++++++++++++++++++++++++++++++++++
+           ::
 
-::
+             python -m venv ivadomed_env --python=python3.7
```
joshuacwnewton, line 55: I've never seen this `-python=` syntax... and when I try it, I get an error:

```
joshua@tadpole:~$ python -m venv ivadomed_env --python=python3.7
usage: venv [-h] [--system-site-packages] [--symlinks | --copies] [--clear]
            [--upgrade] [--without-pip] [--prompt PROMPT] [--upgrade-deps]
            ENV_DIR [ENV_DIR ...]
venv: error: unrecognized arguments: --python=python3.7
```

Just curious, where did you find this? o:

docs/source/installation.rst
```diff
@@ -2,147 +2,123 @@ Installation
...
-------------------
+    .. tab:: Install via ``venv``
 
-Step 1: Clone the `ivadomed <https://github.com/ivadomed/ivadomed>`_ repository.
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+        1. Setup Python Venv Virtual Environment.
 
-::
+        ``ivadomed`` requires Python >= 3.6 and <3.9 (If you are using `Compute Canada <https://www.computecanada.ca/>`_, you can load modules (e.g. python 3.9) as `mentioned here <https://intranet.neuro.polymtl.ca/computing-resources/compute-canada#modules>`_ and `also here <https://docs.computecanada.ca/wiki/Utiliser_des_modules/en#Loading_modules_automatically>`_ ). We recommend
+        working under a virtual environment, which could be set as follows:
```
joshuacwnewton, line 37: I think this "Compute Canada" bracketed text might be better off in the Compute Canada tab.

joshuacwnewton (changes requested):
Thank you for taking care of the feedback!!

A few more things stood out to me on my second pass.

docs/source/installation.rst
dyt811: Good point. I think the original rationale was to actually have a one stop installation for all dependencies including torch etc (as those distribute via conda rather well) but then the issue with that is, GPU are not always supported and also we won't know at installation time which CUDA to use etc, which significantly complicate the matter. 

I like your suggestion to keep it simple so I will just keep the steps simple here to create an empty environment and leave the rest alone (removing pip installation part)

docs/source/installation.rst
dyt811: Nice. TIL. Totally didn't know that before.

docs/source/installation.rst
dyt811: Good point. I was actually thinking about this too.

docs/source/installation.rst
dyt811: Love it! I just couldn't recall the Restructure text heading hierarchy.

dyt811:
Deprecated the `environment.yml` from installation instructions. 

Also deprecated all the `HDF5` instructions as PR#984 are deprecating them from repo imminently.

docs/source/installation.rst
dyt811: Oh man... VENV doesn't support this :(. Okay. Will have to remove this section then and hope they use the right python version to begin with.

docs/source/installation.rst
dyt811: Good point. They were originally sort of together. Now separated. Moving there.

docs/source/installation.rst
joshuacwnewton: > Love it! I just couldn't recall the Restructure text heading hierarchy.

The interesting thing is that RST doesn't actually have a heading hierarchy! You can use the symbols in any order, and `docutils`/`sphinx` will interpret each new symbol as the next-smallest header. ([source](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections))

> Normally, there are no heading levels assigned to certain characters as the structure is determined from the succession of headings.

This can get a bit weird, though, because header symbols will often end up inconsistent between pages. :sweat_smile:

docs/source/installation.rst
joshuacwnewton: Thankfully, the range of versions that ivadomed supports is quite wide, so I think in the vast majority of cases, the default Python will be OK.

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
 
-::
+    .. tab:: Compute Canada HPC
 
-    pip install -e .
+        There are numerous constraints and limited package availabilities with ComputeCanada cluster environment.
 
-Approach 2: Venv
------------------
+        It is best to attempt ``venv`` based installations and follow up with ComputeCanada technicall support as MANY specially compiled packages (e.g. numpy) are exclusively available for Compute Canada HPC environment.
```
joshuacwnewton, line 53: Typo: `technicall` -> `technical`

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
 
-Step 1: Setup Python Virtual Environment.
-+++++++++++++++++++++++++++++++++++++++++
+        If you are using `Compute Canada <https://www.computecanada.ca/>`_, you can load modules as `mentioned here <https://intranet.neuro.polymtl.ca/computing-resources/compute-canada#modules>`_ and `also here <https://docs.computecanada.ca/wiki/Utiliser_des_modules/en#Loading_modules_automatically>`_ .
```
joshuacwnewton, line 55: Tiny typo: There's an extra space between `also here` and the `.`.

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
+++++++++++++++++++++++++++++
 
-    virtualenv venv-ivadomed
-    source venv-ivadomed/bin/activate
 
+.. tabs::
 
-.. warning::
-   If the default Python version installed in your system does not fit the version requirements, you might need to specify a version of Python associated with your virtual environment:
+    .. tab:: Pypi Installation
```
joshuacwnewton, line 64: Nitpick: This still reads as `Pypi`.

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
 
+The additional steps below are only necessary for contributors to the ``ivadomed`` project.
 
-(Optional) Step 5 for Developers Install pre-commit hooks
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+Additional Dependencies and ``pre-commit``
+-------------------------------------------
 
 We use ``pre-commit`` to enforce a limit on file size.
 After you've installed ``ivadomed``, install the hooks:
```
joshuacwnewton, line 121: Before, when I suggested the two headings (`+++` and `---`), it was when we had two "developer-only" subsections. But, since the HDF5 warning is removed here, and we have only 1 subsection, then we probably don't need the two different headings anymore. :sweat_smile: 

We could probably just reduce this down to:

```rest
Developer-only dependencies
+++++++++++++++++++++++++++

For contributors to the ``ivadomed`` project, the ``pre-commit`` package is used to enforce a size limit on committed files. 

After you've installed ``ivadomed``, install the ``pre-commit`` hooks by running:

File: docs/source/installation.rst [hunk 11]
```diff
@@ -2,152 +2,125 @@ Installation
...
 
-::
+        ::
 
-    conda env create --file environment.yml
+            python -m venv ivadomed_env
 
-Step 3 : Activate environment and use
-++++++++++++++++++++++++++++++++++++++
+        2. Activate the new virtual environment (default named `ivadomed_env`)
```
joshuacwnewton, line 42:
```suggestion
        2. Activate the new virtual environment (default named ``ivadomed_env``)
```

(Needs a second set of ` to make it code-formatted.)

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
+You can setup ``ivadomed`` using either Conda or Venv:
 
-Write in your .bashrc/.zshrc file with following line.
+.. tabs::
 
-::
+    .. tab:: Install via ``conda``
 
-    export HDF5_USE_FILE_LOCKING='FALSE'
+        This is the easiest way for personal computers.
```
joshuacwnewton, line 18: Hmm... I'm wondering, is `conda` really the "easiest way for personal computers"?

Most people won't have conda installed on their machines by default, while a lot of people (e.g. Linux users) will have access to a system Python which they can use to create a virtual environment.

So, for the order of the tabs, I'm wondering if `venv` should be the default tab, while `conda` should be the second tab.

joshuacwnewton (changes requested):
Almost there! It's looking really good. I mostly have very nitpicky comments left. :slightly_smiling_face:

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
+            pip install ivadomed
 
-::
+    .. tab:: Repo Installation (Advanced or Developer)
 
-    git clone https://github.com/ivadomed/ivadomed.git
-    cd ivadomed
- 
+        Bleeding-edge developments are available on the project's master branch
+        on Github. Installation ``ivadomed`` from source:
```
mariehbourget, line 78:
```suggestion
        on Github. Install ``ivadomed`` from source:
```

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
-::
-   
-   pip install -r requirements_gpu.txt
+            cd ivadomed
 
+            pip install -e .
 
-If you do not have a compatible GPU, run the following installer to use ``ivadomed`` with CPU.
 
+Step 3: Install ``ivadomed`` with CPU or GPU Support, Install ``torch`` and ``torchvision``
```
mariehbourget, line 89:
```suggestion
Step 3: Install ``torch`` and ``torchvision`` with CPU or GPU Support. 
```

I'm not sure about this, but if I understand correctly "ivadomed" is already instaled at step 2 and step 3 is only for torch and torchvision. Feel free to disregard if I'm mistaken.
The length of the title underline below also has to be adjusted accordingly.

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
 
-Step 2: Create new Conda Env called IvadoMedEnv (may take 5 to 15 minutes)
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+        ``ivadomed`` requires Python >= 3.6 and <3.10. We recommend working under a virtual environment, which could be set as follows:
 
-::
+        ::
 
-    conda env create --file environment.yml
+            python -m venv ivadomed_env
```
mariehbourget, line 40: This did not work for me, I had to `sudo apt install python3.8-venv` first, and then `python3.8 -m venv ivadomed_env`.

mariehbourget: Also, there was a previous discussion about being unable to set the python is `venv`. I think this is problematic, and as a user, I want to be able to choose my version. I think the way to do this is maybe with:
`python3.X -m venv ivadomed_env`

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+Additional Dependencies and ``pre-commit``
+-------------------------------------------
 
 We use ``pre-commit`` to enforce a limit on file size.
 After you've installed ``ivadomed``, install the hooks:
 
 ::
 
     pip install -r requirements_dev.txt
```
mariehbourget, line 125: I get this error when installing:
`ERROR: sphinx-tabs 3.2.0 has requirement docutils~=0.16.0, but you'll have docutils 0.17.1 which is incompatible.`

I'm not sure what the impact is, I was able to build the doc locally without errors.

docs/source/installation.rst
```diff
@@ -2,152 +2,125 @@ Installation
...
 
-If you have a compatible NVIDIA GPU that supports CUDA, run the following command:
+            git clone https://github.com/ivadomed/ivadomed.git
 
-::
-   
-   pip install -r requirements_gpu.txt
+            cd ivadomed
 
+            pip install -e .
```
mariehbourget, line 86: I got this error when installing in `venv`:
`ERROR: packaging 21.2 has requirement pyparsing<3,>=2.0.2, but you'll have pyparsing 3.0.5 which is incompatible.`

This prevent `ivadomed` from running. When running CLI I get:
```
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (pyparsing 3.0.5 (/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages), Requirement.parse('pyparsing<3,>=2.0.2'), {'packaging'})

During handling of the above exception, another exception occurred:

  File "/home/mhbourget/ivadomed_env_969_2/bin/ivadomed", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3252, in <module>
    def _initialize_master_working_set():
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3235, in _call_aside
    f(*args, **kwargs)
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3264, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 585, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 598, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (pyparsing 3.0.5 (/home/mhbourget/ivadomed_env_969_2/lib/python3.8/site-packages), Requirement.parse('pyparsing<3,>=2.0.2'), {'packaging'})
```

mariehbourget (changes requested):
The doc looks really great! :)

I still have some issue with installation in a `venv` in CPU (Ubuntu 20.04 in WSL1). I left details about the errors. Unfortunately, `ivadomed` in CLI does not work with this setup.

`unit_tests` are passing though.

docs/source/installation.rst
joshuacwnewton: Ah! Very good point! `python` runs the risk of mapping to `python2`, but if we tell the user to be explicit about their python version, then we avoid this issue.

We could say something like:

```rest
First, make sure that a compatible version of Python 3 is installed on your system by running:

.. code::

   python3 --version

If your system's Python is not 3.7, 3.8, or 3.9 (or if you don't have Python 3 installed at all), please [install Python](https://realpython.com/installing-python/) before continuing.

Once you have a supported version of Python installed, run the following command (where `python3.X` corresponds to the installed Python version)

.. code::

   python3.X -m venv ivadomed_env
```

docs/source/installation.rst
joshuacwnewton: Interesting! It looks like `sphinx-tabs` only updated its requirements (`0.16 -> 0.17`) last month (https://github.com/executablebooks/sphinx-tabs/pull/144), so I'm guessing this hasn't made its way to a release yet.

That said, `pip` should be smart enough to resolve dependencies between all of the requirements in `requirements-dev.txt`. I'm surprised it's throwing an error here. :thinking:

docs/source/installation.rst
joshuacwnewton: Ah, this has a better solution in https://github.com/ivadomed/ivadomed/pull/969#discussion_r744307483.

docs/source/installation.rst
dyt811: This issue is weirdly related to venv/pip common dependency version which several fixes are available (simplest is to update pip to 21). Could you check if you have Pip v20? In depth details as well as other many useful solutions are reported by @kousu here: https://github.com/neuropoly/computers/issues/218.

We will probably update the dependencies to be more specific to globally address this odd problem.

mariehbourget: You are right, I had `pip 20.0.2`.
I did `pip install --upgrade pip` and I now have `pip 21.3.1` and the installation and CLI run without error.

docs/source/installation.rst
dyt811: Very good point. `conda` was useful for GPU related installation but so far, guide has not really leveraged that much. Switching to `venv`.

docs/source/installation.rst
dyt811: Very good point. I agree the title could be more clear. Changed. :D

docs/source/installation.rst
dyt811: Very good point. I didn't consider this. Added. :)

docs/source/installation.rst
dyt811: Thanks to Nick, we have also added that pyparsing as a requirement and merged to master to include additional details. See #987

docs/source/installation.rst
dyt811: Thanks for the in-depth explanation.

dyt811:
All changes incorporated. Thanks for all the detailed review y'all!

docs/source/installation.rst
```diff
@@ -2,152 +2,133 @@ Installation
...
 
-Save this change and restart your terminal to apply the change.
+                .. code::
 
-There are fundamentally two different approaches to install ``IvadoMed``:
+                   python3 --version
 
-1) Install via Conda
-    This is the easiest way for personal computers.
+                If your system's Python is not 3.7, 3.8, or 3.9 (or if you don't have Python 3 installed at all), please `install Python <https://realpython.com/installing-python/>`_ before continuing.
```
joshuacwnewton, line 28: Ack, sorry. RealPython is now prompting people to create accounts to read their pages.

Screenshot from 2021-11-13 10-56-55

Normally I like linking to RealPython because I find their pages very well-written, but this is enough of a nuisance that I think we should maybe link somewhere else.

```suggestion
                If your system's Python is not 3.7, 3.8, or 3.9 (or if you don't have Python 3 installed at all), please `install Python <https://wiki.python.org/moin/BeginnersGuide/Download/>`_ before continuing.
```

This new link is tied to the official Python wiki, so it should be less susceptible to that kind of thing.

docs/source/installation.rst
```diff
@@ -2,152 +2,133 @@ Installation
...
 
-Step 3 : Activate environment and use
-++++++++++++++++++++++++++++++++++++++
+                ::
 
-::
+                    conda env create --name ivadomed_env
 
-    conda activate IvadoMedEnv
+            2. Activate the new conda environment (default named ``ivadomed_env``)
```
joshuacwnewton, line 50: Tiny nitpick: The text "(default named ``ivadomed_env``)" might be a bit unnecessary, since the name ``ivadomed_env`` was chosen in the step immediately before.

docs/source/installation.rst
```diff
@@ -2,152 +2,133 @@ Installation
...
-    pip install --upgrade pip
-    pip install ivadomed
+                pip install -r requirements.txt
 
-(Optional) Alternative Step 4 for Developers: Install from source
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-Bleeding-edge developments are available on the project's master branch
-on Github. Installation procedure is the following at repository root:
+        .. tab:: Nvidia GPU Support
```
joshuacwnewton, line 112: > Alternatively, if you have already downloaded/cloned the repo with access to the ``requirement.txt`` file, run the following command:

Hmm... so, if the user is installing `ivadomed` using `pip install ivadomed` (which is the most prominent method in these new docs), then they won't have access to the requirements files, right? 

Because of that, I feel like the user might get stuck, especially for the GPU tab ("Where is `requirements_gpu.txt`? What do I install?"). And, it feels like a pretty big ask to expect the user clone the repo just to get access to `requirements_gpu.txt`. It kind of defeats the purpose of PyPI, aha. 

But, at the same time, I don't have a great solution to this, because the root cause is deeper than just documentation (https://github.com/ivadomed/ivadomed/issues/861).

docs/source/installation.rst
mariehbourget, line 28:
```suggestion
                If your system's Python is not 3.6, 3.7, 3.8, or 3.9 (or if you don't have Python 3 installed at all), please `install Python <https://realpython.com/installing-python/>`_ before continuing.
```

docs/source/installation.rst
joshuacwnewton: As a short-term fix for the docs, maybe we could explicitly list the `torch`/`torchvision` requirements from the requirements files:

CPU Only, Mac:

```rst
.. code::

   pip install torch==1.8.0 torchvision==0.9.0 --find-links https://download.pytorch.org/whl/torch_stable.html
```

CPU-only, Windows/Linux:

```rst
.. code::

   pip install torch==1.8.0+cpu torchvision==0.9.0+cpu --find-links https://download.pytorch.org/whl/torch_stable.html
```

GPU:

```rst
.. code::

   pip install torch==1.8.1+cu111 torchvision==0.9.1+cu111 --find-links https://download.pytorch.org/whl/torch_stable.html
```

This doesn't feel like a great solution, because we're duplicating information from the requirements files. But, addressing #861 probably deserves much more thought/discussion, and I understand that your plan was to finish this PR by today. :thinking:

docs/source/installation.rst
```diff
@@ -2,152 +2,133 @@ Installation
...
 
+    .. tabs::
 
-Step 4: Install from release (recommended)
-++++++++++++++++++++++++++++++++++++++++++
+        .. tab:: CPU Support
 
-Install ``ivadomed`` and its requirements from
-`Pypi <https://pypi.org/project/ivadomed/>`__:
+            If you plan to run ``ivadomed`` on CPU only, install pytorch per instructions `provided here for 1.8 <https://pytorch.org/get-started/previous-versions/#wheel>`_.
```
mariehbourget, line 104: I find this line a bit confusing.

If I understand correctly:
- if installed with PyPI, torch must be installed from the external link
- if the repo is cloned, it can be installed using requirements

If that is the case, it would need to be more explicitly explained.
Also, linking to external site for instructions can be confusing for user, so I would suggest listing the command to run directly in our instructions.

docs/source/installation.rst
```diff
@@ -2,152 +2,133 @@ Installation
...
 
-We use ``pre-commit`` to enforce a limit on file size.
-After you've installed ``ivadomed``, install the hooks:
+Developer-only Installation Steps
++++++++++++++++++++++++++++++++++
 
-::
+    The additional steps below are only necessary for contributors to the ``ivadomed`` project.
+
+    For contributors to the ``ivadomed`` project, the ``pre-commit`` package is used to enforce a size limit on committed files. ``requirements_dev.txt`` also contain additional dependneices related to documentation building etc.
```
mariehbourget, line 127:
```suggestion
    For contributors to the ``ivadomed`` project, the ``pre-commit`` package is used to enforce a size limit on committed files. ``requirements_dev.txt`` also contain additional dependencies related to documentation building and testing.
```

docs/source/installation.rst
```diff
@@ -2,152 +2,133 @@ Installation
...
-    This is compatible with ComputeCanada cluster environment.
+                Once you have a supported version of Python installed, run the following command (replacing ``3.X`` with the Python version number that you installed):
 
-Approach 1: Conda
-------------------
+                .. code::
 
-Step 1: Clone the `ivadomed <https://github.com/ivadomed/ivadomed>`_ repository.
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+                   python3.X -m venv ivadomed_env
```
mariehbourget, line 34: This works great now.
I wanted to mention that the first time I tried this, I had to install python3.8-venv separately (I already had python 3.8 without venv apparently), see this https://github.com/ivadomed/ivadomed/pull/969#discussion_r744296259.

So I don't know if that is normal or not vut a user may have to do the same (I don't know how to test this now that my python3.7-venv is already installed)

docs/source/installation.rst
joshuacwnewton: Same thought! (https://github.com/ivadomed/ivadomed/pull/969#discussion_r748739159) :smile:

docs/source/installation.rst
joshuacwnewton: > (I already had python 3.8 without venv apparently)

Huh. It was my impression that `venv` came [bundled with Python](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#installing-virtualenv):

> Note: If you are using Python 3.3 or newer, the venv module is the preferred way to create and manage virtual environments. venv is included in the Python standard library and requires no additional installation.

But, as you've pointed out, `python-venv` needs to be installed separately on Debian/Ubuntu systems, or else the following error is given:

```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.8-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.
```

Yikes. I didn't know that, so thank you for pointing it out. (I missed it the first time you pointed it out, oops.) :sweat:

joshuacwnewton: Apparently this is a bug in Debian/Ubuntu specifically: https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847

We might benefit from adding an additional message in this section that says:

> If you use Debian or Ubuntu, make sure to install the `python3-venv` module alongside whichever version of Python you choose to install. For other operating systems, `venv` will be installed by default.

But, at the same time, the error message given by Python is quite clear, and provides a command for the user to run, so perhaps we can just leave it be?

mariehbourget: I like the additional message to help inexperienced users, but I would also be OK to leave it be because the python message is indeed very clear 🙂
In any case, this is minor and I don't think it should block the merge of the PR.

docs/source/installation.rst
dyt811: Funny I think maybe you visited the site wayyy too often? Just tested it in incognito and it was working? Did they cookie trap you lol. But I agree probably more legit linking the original repo. Going to make the change to official wiki.

docs/source/installation.rst
dyt811: Good point. I mean, I feel like it almost feel better just to point user to the relevant section in the PyTorch installation guide than trying to duplicate information from them, even though technically the Torch website present these information more cleanly. I have adopted the above into a slightly more elaborate form including original solution.

docs/source/installation.rst
dyt811: Oh yeah, good point. Not sure why we dropped 3.6 support, even though we technically still test them.

docs/source/installation.rst
dyt811: Good point! Adopted in conjunction with Joshua's comment. Explicitly stated now for all scenarios.

docs/source/installation.rst
dyt811: Let's be a bit more verbose. Tab already hide tons of parallel path already for those not interested so our main path is actually already much shorter than others. Better safe than sorry. Thanks for testing this out!

dyt811:
All changes incorporated and expanded. Just realized that WIndows further complicated the issue. Thank @joshuacwnewton and @mariehbourget for tireless detailed feedback. Greatly appreciated the insightful discussions and dozens of corner cases I didn't think of before. Please keep them coming!

docs/source/installation.rst
```diff
@@ -2,152 +2,199 @@ Installation
...
 
-Step 4 : Install from source
-++++++++++++++++++++++++++++++++++++++
+                        Replacing ``3.X`` with the Python version number that you installed):
 
-::
+                        .. code::
 
-    pip install -e .
+                            python3.X -m venv ivadomed_env
```
joshuacwnewton, line 55: Nitpick: I'd maybe move some of the text around so that the command comes first:

```rst
                        .. code::

                           # Note: Replace ``3.X`` with the specific Python version number that you installed
                           python3.X -m venv ivadomed_env

                        .. warning::

                           If you use ``Debian`` or ``Ubuntu``, you may be prompted to install the ``python3-venv`` module when creating the virtual environment. This is expected, so please follow the instructions provided by Python. For other operating systems, ``venv`` will be installed by default.
```

(I suggest this because the preceding text is "run the following command:". I found it a bit jarring to see a warning, rather than the actual command. So, I think this way might flow a bit better?)

docs/source/installation.rst
```diff
@@ -2,152 +2,199 @@ Installation
...
+
+
+Step 3: Install ``torch`` and ``torchvision`` with CPU or GPU Support
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+    .. tabs::
+
+        .. tab:: CPU Support
+
+            If you plan to run ``ivadomed`` on CPU only, install pytorch per instructions provided below for your specific operating system:
```
joshuacwnewton, line 142: Nitpick: `pytorch` -> `PyTorch`

```suggestion
            If you plan to run ``ivadomed`` on CPU only, install PyTorch per instructions provided below for your specific operating system:
```

joshuacwnewton (approved):
I only have 2 very minor nitpicks, so I'm going to give my approval in advance. 🙂

docs/source/installation.rst
mariehbourget: I agree with Joshua here. I would also put the command first, followed by a note about Debian/Ubuntu.

mariehbourget (approved):
Thank you @dyt811!
The instructions look good, all the different cases are well explained, and easy to follow.

docs/source/installation.rst
dyt811: Great point.

dyt811:
Changes adopted. We are almost there!

PR #980, opened 2021-10-29
Apply filter parameters on 2D patches to remove empty patches

### LINKED ISSUE #979: Apply `slice_filter_params` on 2D patches
In the `loader parameters`, we have a [slice_filter_params](slice_filter_params
) to filter out empty input, empty mask or absent class.
The filters are applied [here](https://github.com/ivadomed/ivadomed/blob/2e5f70faa67e2a21f9ae11ada69145256aa6629b/ivadomed/loader/mri2d_segmentation_dataset.py#L109) on the entire image.
For microscopy, we would like to be able to filter on each patch of an image.

To filter on the "patch" level, we would need to apply the filters in [prepare_indices](https://github.com/ivadomed/ivadomed/blob/2e5f70faa67e2a21f9ae11ada69145256aa6629b/ivadomed/loader/mri2d_segmentation_dataset.py#L135), where the coordinates for the patches of the dataset are indexed.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR aims to apply filters on patches in addition to entire image to remove empty patches from dataset, as described in #979.

I added a `PatchFilter` class in `loader/patch_filter.py` (similar to `SliceFilter`) with 3 filters on a patch basis:
`filter_empty_mask`
`filter_absent_class`
`filter_empty_input`
contrary to "slice" filters, the "patch" filters only apply at training time (training and validation) as we don't want to remove patches for testing or segment.

I updated the documentation accordingly [here](https://ivado-medical-imaging--980.org.readthedocs.build/configuration_file.html#patch-filter-params), added a test, and updated doctrings for `SliceFilter`.

The `SliceFilter` behavior remains unchanged except for the Default value `filter_empty_mask=False`, now harmonized with the default value from the config file.

## Linked issues
Fixes #979


mariehbourget:
This PR is ready for review. See comment above for description.

Questions:
The documentation and docstrings do not indicate the default behavior for the slice filters.
In addition, the default behavior is different between `config_default.json` and the `SliceFilter init` for `filter_empty_mask`.
- config_default.json:
```
        "slice_filter_params": {
            "filter_empty_mask": false,
            "filter_empty_input": true
        },
```
- SliceFilter init:
```
    def __init__(self, filter_empty_mask=True,
                 filter_empty_input=True,
                 ...
```
Do we want to document the default behavior of the filters?
If yes, what should be the default behavior of `filter_empty_mask`?
@charleygros, @andreanne-lemay, maybe you have some insights on this?
Thanks!

ivadomed/loader/mri2d_segmentation_dataset.py
```diff
@@ -147,17 +147,28 @@ def prepare_indices(self):
...
 
-            for x in range(0, (shape[0] - self.length[0] + self.stride[0]), self.stride[0]):
-                if x + self.length[0] > shape[0]:
-                    x = (shape[0] - self.length[0])
-                for y in range(0, (shape[1] - self.length[1] + self.stride[1]), self.stride[1]):
-                    if y + self.length[1] > shape[1]:
-                        y = (shape[1] - self.length[1])
+            for x_min in range(0, (shape[0] - self.length[0] + self.stride[0]), self.stride[0]):
+                if x_min + self.length[0] > shape[0]:
+                    x_min = (shape[0] - self.length[0])
```
dyt811, line 152: I see this right align the mini patch to the entire shape, which might cause partial overlap with earlier patch. I assume this is an expect behaviour that would not affect training? similar for bottom aligning the mini patch for Y?

ivadomed/loader/mri2d_segmentation_dataset.py
```diff
@@ -147,17 +147,28 @@ def prepare_indices(self):
...
+
                     self.indexes.append({
-                        'x_min': x,
-                        'x_max': x + self.length[0],
-                        'y_min': y,
-                        'y_max': y + self.length[1],
+                        'x_min': x_min,
+                        'x_max': x_max,
+                        'y_min': y_min,
+                        'y_max': y_max,
```
dyt811, line 171: 👍 thanks for making it more readable here.

ivadomed/loader/mri2d_segmentation_dataset.py
```diff
@@ -147,17 +147,28 @@ def prepare_indices(self):
...
+                    if y_min + self.length[1] > shape[1]:
+                        y_min = (shape[1] - self.length[1])
+                    y_max = y_min + self.length[1]
+
+                    patch = {'input': list(np.asarray(self.handlers[i][0]['input'])[:, x_min:x_max, y_min:y_max]),
+                             'gt': list(np.asarray(self.handlers[i][0]['gt'])[:, x_min:x_max, y_min:y_max]) \
+                                   if self.handlers[i][0]['gt'] else [],
+                             'input_metadata': self.handlers[i][0]['input_metadata'],
+                             'gt_metadata': self.handlers[i][0]['gt_metadata']}
+                    if self.slice_filter_fn and not self.slice_filter_fn(patch, is_2d_patch=self.is_2d_patch):
```
dyt811, line 164: I am not sure sure what `self.handlers[i][0]` is? Is it a dict, dataframe or something else? I checked the earlier comment on handler and it seems that it is just a list of indexes? So... where did this stem from? My guess so far is that the respective input/gt/keys are the actual image/slice/(source) shape? 

I feel we can afford to be a lot more verbose here to expand it out for ease of understanding as we are packing a lot of actions into a few lines.

dyt811, line 164: I think slice_filter_fn is a `SliceFilter` instance, and no longer a `dict` as the docstring indicated, is that right?

ivadomed/loader/slice_filter.py
```diff
@@ -4,65 +4,101 @@
...
+                    return False
+
+            if self.filter_classification:
+                if not np.all([int(
+                        self.classifier(
+                            imed_utils.cuda(torch.from_numpy(img.copy()).unsqueeze(0).unsqueeze(0),
+                                            self.cuda_available))) for img in input_data]):
+                    return False
 
         return True
```
dyt811, line 104: Overall, I think these parts needs to be simplified. 

Multilayered `if`s are typically error prone. Same logic can be expressed as shown below: 

```
# Filter 2D patches at training time that 
if is_2d_patch and self.is_train:
    # ... do not have ANY ground truth (i.e. all masks are empty)
    if self.filter_empty_mask_patch and not np.any(gt_data):
        return False

    # ... have absent classes (i.e. one or more masks are empty)
    if self.filter_absent_class_patch and not np.all([np.any(mask) for mask in gt_data]):
        return False
else:

  # Filter slices that do not have ANY ground truth (i.e. all masks are empty)
  if self.filter_empty_mask and not np.any(gt_data):
      return False

  # Filter slices that have absent classes (i.e. one or more masks are empty)      
  if self.filter_absent_class and not np.all([np.any(mask) for mask in gt_data]):
      return False

  # Filter set of images if one of them is empty or filled with constant value (i.e. std == 0)      
  if self.filter_empty_input and np.any([img.std() == 0 for img in input_data]):
      return False

  # Filter unmatched class? 
  matching_class = np.all([int(self.classifier(
                  imed_utils.cuda(torch.from_numpy(img.copy()).unsqueeze(0).unsqueeze(0),
                                  self.cuda_available))) for img in input_data])
  if self.filter_classification and not matching_class :    
      return False

return True
```

I also recommend avoid one liners like the `matching_class` boolean assignment I extracted out. IF a one liner like that is needed, there probably should be at least a paragraph explaining what it is doing. Typically it is far more efficient to write it out into smaller (albeit less efficient) statements that make it crystal clear what each steps are doing. It also benefit from type hinting and IDE validation along each small step.

testing/unit_tests/test_slice_filter.py
```diff
@@ -88,5 +89,58 @@ def test_slice_filter(download_data_testing_test_files, transforms_dict, train_l
...
+        "model_params": {"name": "Unet", "length_2D": [32, 32], "stride_2D": [32, 32]},
+        "slice_filter_params": slice_filter_params,
+        "slice_axis": "axial",
+        "multichannel": False
+    }
+    # Get Training dataset
+    bids_df = BidsDataframe(loader_params, __tmp_dir__, derivatives=True)
+    ds = imed_loader.load_dataset(bids_df, **loader_params)
+
+    print('\tNumber of loaded patches: {}'.format(len(ds)))
```
dyt811, line 124: I recommend use loguru.logger as most of the other underlying modules are already using so the output will be more consistent and auto time stamped. 

Also, I recommend to use [`f`string format ](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python)for easier readability. 

image

testing/unit_tests/test_slice_filter.py
```diff
@@ -88,5 +89,58 @@ def test_slice_filter(download_data_testing_test_files, transforms_dict, train_l
...
+
+    loader = DataLoader(ds, batch_size=BATCH_SIZE,
+                        shuffle=True, pin_memory=True,
+                        collate_fn=imed_loader_utils.imed_collate,
+                        num_workers=0)
+    print('\tNumber of Neg/Pos patches in GT.')
```
dyt811, line 130: ditto above.

ivadomed/loader/slice_filter.py
```diff
@@ -4,65 +4,101 @@
...
 
     Attributes:
-        filter_empty_mask (bool): If True, samples where all voxel labels are zeros are discarded.
-        filter_empty_input (bool): If True, samples where all voxel intensities are zeros are discarded.
-        filter_absent_class (bool): If True, samples where all voxel labels are zero for one or more classes are discarded.
-        filter_classification (bool): If True, samples where all images fail a custom classifier filter are discarded.
+        filter_empty_mask (bool): If True, slices where all voxel labels are zeros are discarded.
+        filter_absent_class (bool): If True, slices where all voxel labels are zero for one or more classes are discarded.
+        filter_empty_input (bool): If True, slices where all voxel intensities are zeros are discarded.
+        filter_classification (bool): If True, slices where all images fail a custom classifier filter are discarded.
```
dyt811, line 27: Just wondering if it is "all" images or "any" images? The code seems to suggest that as soon as any image failed the classifier, it will return false?

ivadomed/loader/slice_filter.py
```diff
@@ -4,65 +4,101 @@
 
 
 class SliceFilter(object):
```
dyt811, line 6: Just curious, do you think we should rename this to `SlicePatchFilter` now? Or something like `GenericFilter` or `ConfigurableFilter`? 

Also, now that I read through the class a bit, it seems more like a kind of parameterized filter function than a class. 

I have not used many callable class before so pretty cool TIL to see class structured in this way. 

Also I sort of agree with Julien's other commends on the issue, perhaps, SliceSelector or SlicePatchSelector is more 
distinct because in image/3D data domain, filter has many very unique purpose (e.g. gaussian filter) so maybe `selector` would make the purpose of this class more clear.

ivadomed/loader/mri2d_segmentation_dataset.py
mariehbourget: Yes, this is normal and expected, we might optimize this behavior at a later time, but it does not affect the training significantly.

testing/unit_tests/test_slice_filter.py
mariehbourget: Again, I'm building on previous code here. So I duplicated the previous test. I can make the changes for both tests, or we could open a separate PR to clean up "print" in all our tests at once.

ivadomed/loader/slice_filter.py
mariehbourget: Ok, as discussed in meeting. Let's plan to make the "refactoring" changes separately as it deals with retrocompatiblity with old config files.
Depending on the complexity of the current PR, we may want to do the refactoring before.

mariehbourget (commented):
Thank you @dyt811 for the detailed review.
Some of the points you raise are very important, and I think I may have to review my implementation in more depth. I left comments but did not change anything for now.

There are about 5 different subjects that we may want to treat separately
1) Renaming/refactoring of the SliceFilter Class
2) Maybe split SliceFilter and PatchFilter in two classes
3) Refactoring of the classifier filter
4) Documentation on handlers/indexes
5) Usage of print in testing

Let me know your thoughts. Thanks!

ivadomed/loader/mri2d_segmentation_dataset.py
dyt811: Yup. Good point. I strongly agree this is not something that has to be changed per this PR since you are just carrying out the earlier design patterns. I was just curious. Thanks for the explanation.

ivadomed/loader/slice_filter.py
dyt811: Yup. Good point. I think the complexity warrant a split for clarity and you won't have to worry about the complex if conditions here and should be greatly simplified and more readable. Please disregard my overly complicated example lol. 

I will probably add a classifier filter refactor down the road to avoid complexity like that. Again, out of scope of this particular PR, thanks for the clarification.

testing/unit_tests/test_slice_filter.py
dyt811: Good idea. I thought that was done but it seems we definitely missed some. I am going to write up a ticket to review this across the code base.

ivadomed/loader/mri2d_segmentation_dataset.py
mariehbourget: Ok thanks, I'll keep that in mind while re-designing the PR

charleygros:
> This PR is ready for review. See comment above for description.

Sorry for the 8790217390 delay.

> The documentation and docstrings do not indicate the default behavior for the slice filters.

I agree, it should be documented: IMO, the default behaviour should be: `"filter_empty_input": true` and `"filter_empty_mask": false`.

> In addition, the default behavior is different between `config_default.json` and the `SliceFilter init` for `filter_empty_mask`.

Whooops.. yea so IMO, we should correct this and follow the values currently used in `config_default.json`.

ivadomed/loader/mri2d_segmentation_dataset.py
mariehbourget: As discussed, I added this line for now:
https://github.com/ivadomed/ivadomed/blob/29f24072c6d6b374f7c23292881cf1a44c10ae00/ivadomed/loader/mri2d_segmentation_dataset.py#L163
Let's move the indexes vs. handlers documentation to another issue/PR.

mariehbourget: Done

mariehbourget:
@dyt811
I revisited this PR and chose to split the SliceFilter and PatchFilter in two separate classes. It is much easier to understand the code logic (no nested if loops) and will be easier to maintain as well.
I also corrected all the docstrings and documentation issues raised in the comments.

I went with `PatchFilter` for now because I'm still unsure about the "right" wording. I will be able to change it afterwards once we are settled.

My issue with `Selector` instead of `Filter` is that intuitively, I would think it selects or "keeps" some patches based on a condition, whereas in reality it's the opposite, it "removes" patches (same goes for SliceFilter).
And the params in the config file are for example:
```
"filter_empty_mask": false,
"filter_empty_input": true
```
If we change them for `select_empty_something`, it's even more confusing because true means they are not selected and vice-versa.

We would need to replace `Filter` by a synonym like `Remove` or `Discard`.
If we want `Select`, then we need to change the implementation and make it compatible with the previous `Filter` params, which may not be trivial.

Thoughts?

docs/source/configuration_file.rst
```diff
@@ -447,6 +446,47 @@ See details in both ``train_validation`` and ``test`` for the contrasts that are
...
+
+
+.. code-block:: JSON
+
+    {
+        "loader_parameters": {
+            "patch_filter_params": {
+                "filter_empty_mask": false,
+                "filter_empty_input": false
+            }
```
dyt811, line 485: missing `filter_absent_class` here?

ivadomed/loader/patch_filter.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, filter_empty_mask=False,
+                 filter_absent_class=False,
+                 filter_empty_input=False,
+                 is_train=False):
+        self.filter_empty_mask = filter_empty_mask
+        self.filter_absent_class = filter_absent_class
+        self.filter_empty_input = filter_empty_input
+        self.is_train = is_train
+
+    def __call__(self, sample):
```
dyt811, line 37: can we add a type hint for this `sample` here? Is it a dataframe or something? I wasn't super sure.

ivadomed/loader/patch_filter.py
```diff
@@ -0,0 +1,55 @@
...
+        input_data, gt_data = sample['input'], sample['gt']
```
dyt811, line 38: Same for `input_data` and `gt_data` here. Are they `panda` `dataframe` or `series` or `list`?

ivadomed/loader/slice_filter.py
dyt811: Yeah, I think it is complicated enough to worry about the refactoring in the future.

testing/unit_tests/test_patch_filter.py
```diff
@@ -0,0 +1,91 @@
...
+def _cmpt_slice(ds_loader):
+    cmpt_label = {0: 0, 1: 0}
+    for i, batch in enumerate(ds_loader):
+        for gt in batch['gt']:
+            # TODO: multi label
+            if np.any(gt.numpy()):
+                cmpt_label[1] += 1
+            else:
+                cmpt_label[0] += 1
+    print(cmpt_label)
```
dyt811, line 32: I would recommend to use loguru's logger.debug/info or the like as most of the code base has migrated to those and it has auto time stamps etc (and allow centralized sink down the road if you want to keep logs etc).

testing/unit_tests/test_patch_filter.py
```diff
@@ -0,0 +1,91 @@
...
+    return cmpt_label[0], cmpt_label[1]
+
+
+@pytest.mark.parametrize('transforms_dict', [{"CenterCrop": {"size": [128, 128], "applied_to": ["im", "gt"]}}])
+@pytest.mark.parametrize('train_lst', [['sub-rat3_ses-01_sample-data9_SEM.png']])
+@pytest.mark.parametrize('target_lst', [["_seg-axon-manual", "_seg-myelin-manual"]])
+@pytest.mark.parametrize('patch_filter_params', [
+    {"filter_empty_mask": False, "filter_empty_input":  True},
+    {"filter_empty_mask": True, "filter_empty_input": True}])
+@pytest.mark.parametrize('dataset_type', ["training", "testing"])
```
dyt811, line 42: FYI, in the near future, we will most likely start the pytest-cases migration to avoid these gigantic parametrize block which is very hard to keep track.

testing/unit_tests/test_patch_filter.py
```diff
@@ -0,0 +1,91 @@
...
+        "path_data": [os.path.join(__data_testing_dir__, "microscopy_png")],
+        "bids_config": f"{path_repo_root}/ivadomed/config/config_bids.json",
+        "target_suffix": target_lst,
+        "extensions": [".png"],
+        "roi_params": {"suffix": None, "slice_filter_roi": None},
+        "model_params": {"name": "Unet", "length_2D": [32, 32], "stride_2D": [32, 32]},
+        "slice_filter_params": {"filter_empty_mask": False, "filter_empty_input": False},
+        "patch_filter_params": patch_filter_params,
+        "slice_axis": "axial",
+        "multichannel": False
```
dyt811, line 63: We had began the migration of the all keywords (KW) classes to centrally store these keyword strings for the various params to allow easier refactoring and consistency change down the road. You might want to take a look at those: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/keywords.py

ivadomed/loader/mri2d_segmentation_dataset.py
dyt811: Agreed. Thanks for the clarification. I will open an internal refactoring ticket for this.

testing/unit_tests/test_slice_filter.py
dyt811: NPLY-80/NPLY-84

ivadomed/loader/slice_filter.py
dyt811: Thanks for the explanation! Still not sure when that is really used or if we have a test case for it but seems like a very edge feature. Your explanation makes sense.

ivadomed/loader/slice_filter.py
dyt811: I agree. Let's keep it simple.

dyt811:
> We would need to replace `Filter` by a synonym like `Remove` or `Discard`. If we want `Select`, then we need to change the implementation and make it compatible with the previous `Filter` params, which may not be trivial.
> 
> Thoughts?

Yeah, I think we totally don't need to worry about it in this PR yet. Default values certainly clarified the expected behavior a lot more so great change there. 

I reviewed all the latest change as of 2021-12-31T175842EST. I don't have any issues with them but since I am not exactly the subject expertise, I am not sure if I should approve it or not. since it is still in draft. Let me know if I should pull it down and test it further, From all the code changes so far, I think it is more readable than before so great job there! 

A lot of the other concerns raised, I think can be noted in other future refactoring PRs.

docs/source/configuration_file.rst
mariehbourget: `config_default.json` is used to "fill" missing config key with default values [here](https://github.com/ivadomed/ivadomed/blob/d6fe24db4d6a4319c57802b7a6d68e17f6d8b9e0/ivadomed/main.py#L582). In this case (and for slice filter as well), as long as there is at least one param in the config, it will work. I did the same as slice filter with the 2 most likely used filters.

ivadomed/loader/patch_filter.py
mariehbourget: Good point. It's a dict from [here](https://github.com/ivadomed/ivadomed/blob/29f24072c6d6b374f7c23292881cf1a44c10ae00/ivadomed/loader/mri2d_segmentation_dataset.py#L164).
I'll update it here and in slice filter as well.

mariehbourget: I added docstrings and type hint in d19127b for patch filter and slice filter.

ivadomed/loader/patch_filter.py
mariehbourget: Resolved in d19127b

testing/unit_tests/test_patch_filter.py
mariehbourget: I'm not sure if you mean add new keywords to the list, or to change the parameters in the tests for keywords?
If it's the latter, it may be preferable to keep the patch filter tests "consistent" with the slice filter tests and not change for keywords in a single test file.

mariehbourget:
Thank you @dyt811 for your help on this PR.
I addressed all your comments and the PR is now ready for a final review.

ivadomed/loader/slice_filter.py
```diff
@@ -40,24 +47,24 @@ def __init__(self, filter_empty_mask=True,
             else:
                 self.classifier = torch.load(classifier_path, map_location='cpu')
 
-    def __call__(self, sample):
+    def __call__(self, sample: dict):
         input_data, gt_data = sample['input'], sample['gt']
```
dyt811, line 51: Minor line shift change. I think this line went above comment.

ivadomed/loader/patch_filter.py
```diff
@@ -0,0 +1,59 @@
...
+        filter_empty_input (bool): If True, 2D patches where all voxel intensities are zeros are discarded
+            at training time. Default: False.
+        is_train (bool): Indicates if at training time.
+
+    """
+
+    def __init__(self, filter_empty_mask=False,
+                 filter_absent_class=False,
+                 filter_empty_input=False,
+                 is_train=False):
```
dyt811, line 31: I would recommend to type hint them as bool just to be extra clear.

ivadomed/loader/patch_filter.py
```diff
@@ -0,0 +1,59 @@
...
+        self.filter_empty_mask = filter_empty_mask
+        self.filter_absent_class = filter_absent_class
+        self.filter_empty_input = filter_empty_input
+        self.is_train = is_train
```
dyt811, line 35: Also these ones as well.  `is_` prefix is pretty good at indicating binaries. but `filter_` flag might be less intuitive new devs

ivadomed/loader/slice_filter.py
```diff
@@ -40,24 +47,24 @@ def __init__(self, filter_empty_mask=True,
...
+        """Extract input_data and gt_data lists from sample dict and discard them if they don't match certain
+        conditions.
 
+        """
         if self.filter_empty_mask:
-            # Filter slices that do not have ANY ground truth (i.e. all masks are empty)
+            # Discard slices that do not have ANY ground truth (i.e. all masks are empty)
```
dyt811, line 57: 👍

ivadomed/loader/slice_filter.py
```diff
@@ -40,24 +47,24 @@ def __init__(self, filter_empty_mask=True,
...
             if not np.any(gt_data):
                 return False
-
         if self.filter_absent_class:
```
dyt811, line 60: Actually you raise an excellent point about using "discard". Do you think `self.discard_empty_mask` would be a better flag? It certainly convey the message more clear than `filter` as filter can be both filter in or filter out.

Up to your preference. Just thought about it per your earlier change.

testing/unit_tests/test_patch_filter.py
dyt811: Hey, no worry. Test refactoring is coming down the road for JSON consolidation reasons. So no worry about it :D.

dyt811 (approved):
Thank you for the great work and sorry for the delay. I do not have any concerns or issues with the contributions. 

I mentioned a few points for discussion more as food for thoughts. I do not think they are mandatorily enforced for the PR acceptance. 

That being said, I can't say I am an expert with the filter application in patch scenarios but the test converge seems comprehensive enough to warrant its inclusion. 

We will likely have to make a eventual 3.0 release based on this since I believe these two parameters are mandatory now going forward, if I am reading correctly?

ivadomed/loader/patch_filter.py
mariehbourget: Agreed, done in 0761115.

ivadomed/loader/patch_filter.py
mariehbourget: Agreed, done in 0761115.

ivadomed/loader/slice_filter.py
mariehbourget: Good catch! Resolved in b111bfb.

ivadomed/loader/slice_filter.py
mariehbourget: - I agree, we could go with `discard_empty_something` instead of `filter_empty_something` for both `PatchFilter` and `SliceFilter`.
- `PatchFilter` and `SliceFilter` themselves would need to be renamed... `PatchDiscard` and `SliceDiscard`?
- Then we need to account for retrocompatibility by changing the keys from old config files to the new keys in the KEY_CHANGE_DICT in the config manager [here](https://github.com/ivadomed/ivadomed/blob/76b36a0a0f7141feb2d5b00b33e4c3a06865fc2c/ivadomed/config_manager.py#L71).

We could either:
1. Do another PR to deal with renaming SliceFilter and retrocompatiblity that would go in before this PR, then update this PR with PatchDiscard from issue Also in this issue: https://github.com/ivadomed/ivadomed/issues/997.
2. Merge this PR first and rename afterwards.

I prefer option 1 because it's cleaner to make the change in SliceFilter before implementing PatchFilter, but it may take more time to finally get this new feature in.

At this point, I think we just need to take a decision about the name :)
More discussion about the naming in a comment above https://github.com/ivadomed/ivadomed/pull/980#issuecomment-1000535086

I could try to have a PR ready for review for the renaming by the next meeting so we can move on.
Let me know what you think.

mariehbourget:
Thank you @dyt811 for the review and for your patience with this PR.

> I mentioned a few points for discussion more as food for thoughts. I do not think they are mandatorily enforced for the PR acceptance.

As mentioned in https://github.com/ivadomed/ivadomed/pull/980#discussion_r792076016 , I would be ready to merge this PR but we may want to take care of renaming before. Let me know your thoughts.

> We will likely have to make a eventual 3.0 release based on this since I believe these two parameters are mandatory now going forward, if I am reading correctly?

I don't think a new major release is necessary. I updated the `config_default.json` so if the parameters are missing in the config file, they are added by the config manager. I don't see this new feature as more important than any other and I would only release a 2.9.3 version eventually when we need it for SCT or ADS.
(I have another draft PR (https://github.com/ivadomed/ivadomed/pull/1050) that will be necessary for ADS, so I will need a new release soon anyway).

ivadomed/loader/loader.py
```diff
@@ -26,6 +27,7 @@ def load_dataset(bids_df, data_list, transforms_params, model_params, target_suf
         roi_params (dict): Contains ROI related parameters.
         contrast_params (dict): Contains image contrasts related parameters.
         slice_filter_params (dict): Contains slice_filter parameters, see :doc:`configuration_file` for more details.
+        patch_filter_params (dict): Contains patch_filter parameters, see :doc:`configuration_file` for more details.
```
jcohenadad, line 30:
```suggestion
        patch_filter_params (dict): Contains patch_filter_params values, see :doc:`configuration_file` for more details.
```

docs/source/configuration_file.rst
```diff
@@ -447,6 +446,47 @@ See details in both ``train_validation`` and ``test`` for the contrasts that are
...
+
+    {
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "title": "patch_filter_params",
+        "description": "Discard a 2D patch from the dataset if it meets a condition at training time, see below.",
+        "type": "dict",
+        "options": {
+            "filter_empty_input": {
+                "type": "boolean",
+                "description": "Discard 2D patches where all voxel intensities are zeros.Default: ``False``."
```
jcohenadad, line 460:
```suggestion
                "description": "Discard 2D patches where all voxel intensities are zeros. Default: ``False``."
```

docs/source/configuration_file.rst
```diff
@@ -447,6 +446,47 @@ See details in both ``train_validation`` and ``test`` for the contrasts that are
...
+            },
+            "filter_empty_mask": {
+                "type": "boolean",
+                "description": "Discard 2D patches where all voxel labels are zeros.Default: ``False``."
```
jcohenadad, line 464:
```suggestion
                "description": "Discard 2D patches where all voxel labels are zeros. Default: ``False``."
```

docs/source/configuration_file.rst
```diff
@@ -447,6 +446,47 @@ See details in both ``train_validation`` and ``test`` for the contrasts that are
...
+            },
+            "filter_absent_class": {
+                "type": "boolean",
+                "$$description": [
+                    "Discard 2D patches where all voxel labels are zero for one or more classes\n",
+                    "(this is most relevant for multi-class models that need GT for all classes).\n",
```
jcohenadad, line 470:
```suggestion
                    "(this is most relevant for multi-class models that need GT for all classes). \n",
```

docs/source/configuration_file.rst
```diff
@@ -447,6 +446,47 @@ See details in both ``train_validation`` and ``test`` for the contrasts that are
         }
     }
 
+
+.. jsonschema::
+
+    {
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "title": "patch_filter_params",
+        "description": "Discard a 2D patch from the dataset if it meets a condition at training time, see below.",
```
jcohenadad, line 455:
```suggestion
        "description": "Discard a 2D patch from the dataset if it meets a condition defined below. Contrary to the field 'slice_filter_params' which applies at training and testing time, this parameter only applies during training time.",
```

ivadomed/loader/slice_filter.py
mariehbourget: Hi @dyt811,
Thank you again for the review on this complicated PR.
I discussed twith @jcohenadad  and we agreed to merge this PR and take care of the renaming/refactoring afterwards with issue #997.

mariehbourget:
Hi @dyt811,
Just for your information, I was able to have all the CI tests passing on the 2nd attempt.

As I mentioned in a previous comment, I discussed with Julien and we agreed to merge this PR as is and take care of the renaming/refactoring afterwards with issue #997.

Following this merge, we will need a patch release (2.9.3) for integration with ADS.

Issue #977, opened 2021-10-26
Long training time with microscopy images


mariehbourget:
**Disclaimer**
This issue is based on preliminary observations and does not contain a lot of details/examples at the moment.

## Issue description

I started recently to train microscopy models in `ivadomed`.
I found that the training time is superior compared to ADS.

For example, training on the SEM model with 9 images takes 4 hours in `ADS` on `bireli` for 400 epochs with data augmentation i.e. about 36sec/epoch.

With the same 512x512 patch_size in ivadomed (37 for training and 45 for validation) I get:
- about 62 sec/epoch (with `RandomAffine` and `ElasticTransform` data augmentation)
- about 21 sec/epoch (without data augmentation)

Note: Because microscopy images are all different sizes, a different "split" between training and validation sets could affect those timings (longer with more patch in training set) and I don't know what the ADS split was.

While 62sec vs. 36sec. is not that big of a deal on this small dataset, it is more problematic on my 2 other larger datasets. So I wanted to open this issue to see if some optimization of the training process could be done somewhere.

With smaller patch_size of 256x256 and data augmentation (which I'm actually using in ivadomed), I got:
- 6 hours training on `rosenberg` for 110 epochs i.e. about 198 sec/epoch

To investigate:
- Data augmentation seems to be an important bottleneck here, there may be some optimization to be done.
- This may be linked in part to the issue described in #972 (with the `MRI2DSegmentation` `get_item`).
- This may be linked in part to the issue described in #975 (with the `MetricManager`).
- This may be due in part to the architecture of the model itself, as the default `ivadomed` Unet has a lot more features per layer than `ADS` (starting at 64 vs. 16).

I would be happy to provide images and config files if someone is interested in looking into this issue.
Thanks!

mariehbourget:
### Further investigations:

I ran some profiles on my problematic optical microscopy dataset.
This dataset is very large with:
- 649 validation patches and 1933 training patches at size 256x256
- 184 validation patches and 577 training patches at size 512x512

Here are some training times per epoch  (excluding first loading time):
- **Test 1)** patches of 512x512 without data augmentation, training time ~ 500 sec/epoch
- **Test 2)** patches of 512x512 with "medium" data augmentation, training time ~ 6000 sec/epoch
- **Test 3)** patches of 512x512 with "aggressive" data augmentation, training time ~ 13000 sec/epoch
- **Test 4)** patches of 256x256 without data augmentation, training time ~ 1400 sec/epoch

[...]

Issue #972, opened 2021-10-19
Long segmentation time on large microscopy images


mariehbourget:
**Disclaimer**
This issue is based on preliminary observations and does not contain a lot of details/examples at the moment.

## Issue description

I noticed that the segmentation time on large microscopy images is quite long in ivadomed compared to what I was used to in ADS.
To this effect, I fixed a bug in image reconstruction in PR #956.
In addition, we found that using the `.pt` version of the model (on GPU) and the `.onnx` version (on CPU) help with segmentation time as described in issue #961.

Nevertheless, for a microscopy image of ~8000x6000 pixels, the segmentation time with my ivadomed model (`.pt` on GPU on rosenberg) is around 2.5 minutes.
In comparison, an ADS model (on GPU on bireli) would take less than 30 seconds to segment the same image.

I investigated a little bit and the bottleneck seems to come from this [for loop](https://github.com/ivadomed/ivadomed/blob/5207e7a5f0c882ae66248322ffbc518c0d2d6ce7/ivadomed/inference.py#L448). More specifically, there is a delay between the return of `reconstruct_3d_object` and the beginning of the next iteration of the loop (is it the time to fetch the batch in DataLoader?)

I could use some help here, thanks!

kousu:
Oh you could also try this!

```
python3 -m cProfile -s cumulative your-training-script.py
```

That will tell you what's eating your time. If it lists `_MultiProcessingDataLoaderIter._next_data()` near the top then that will reinforce this diagnosis.

mariehbourget:
Update:
I fixed an issue in PR #982 where the transforms were applied on the entire image for each patch instead of on each individual patch. This reduce the segmentation time from 171 sec to 118 sec.

<details>
<summary> cProfile output </summary>

```
         7467214 function calls (7365196 primitive calls) in 117.724 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   2483/1    0.050    0.000  117.821  117.821 {built-in method builtins.exec}
        1    0.014    0.014  117.821  117.821 segment_rosenberg.py:1(<module>)
        1    0.030    0.030  101.423  101.423 inference.py:314(segment_volume)
      108    0.001    0.000   64.071    0.593 dataloader.py:344(__next__)
      108    0.003    0.000   64.070    0.593 dataloader.py:383(_next_data)
      107    0.001    0.000   62.155    0.581 fetch.py:42(fetch)
[...]

PR #982, opened 2021-11-03
Apply transforms on 2D patches

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR aims to apply transforms on each individual 2D patches instead of the entire image in `mri2d_segmentation_dataset.py/get_item` when loading a batch. This will partially fix performance issues described in #977 when training on large microscopy images. It also help with performance when running segmentation as described in #972.

The "slice" behavior (without patches) is unchanged.
No new test is required, all current tests are passing.
I also tested the the commands `--train`, `--test` and `--segment` in single-channel, multi-contrasts and multi-channels.

To test the PR:
The PR can be tested by training and testing a model with the addition of `length_2D` and `stride_2D` options in the config file of any `default_model` (2D unet).

## Linked issues
Related to #972 and #977.


ivadomed/loader/mri2d_segmentation_dataset.py
```diff
@@ -196,81 +196,92 @@ def __getitem__(self, index):
...
+                # stack_gt is a tensor of size 1x1, values: 0 or 1
+                # "expand(1)" is necessary to be compatible with segmentation convention: n_labelxhxwxd
+                stack_gt = torch.from_numpy(seg_pair_slice["gt"][0]).expand(1)
 
-            if stack_gt is not None:
-                for _ in range(len(stack_gt)):
-                    data_dict['gt'] = stack_gt[:,
-                                       coord['x_min']:coord['x_max'],
-                                       coord['y_min']:coord['y_max']]
+        else:
```
andreanne-lemay, line 244: There seems to be a lot of code duplication here between `if self.is_2d_patch` and the `else`. Could we simply keep the code under the first part of the if and simply set 
```python
if not self.is_2d_patch:
    coord = {}
    coord['x_min'], coord['x_max'], coord['y_min'], coord['y_max'] = 0, seg_pair_input["input"][0].shape[0], 0, seg_pair_input["input"][0].shape[1]
``` 
(something like that) I might have missed some technicalities why this is not possible, so please ignore if it's too complicated.

andreanne-lemay (approved):
Looks good overall! Makes a lot of sense.
To consider in next PRs or this one depending on how urgently we want to merge this PR:
* Do the same thing for the 3D loader.
Thanks for taking care of this!

ivadomed/loader/mri2d_segmentation_dataset.py
mariehbourget: I had tried to minimize the duplicated code but could not find a good way. I think your suggestion could work well, thank you!
I'll try it, the only issue I might have is to test the ROI correctly. Currently ROI only applies to entire slice and not patches and I'm not very familiar with the feature. Is the [Cascaded architecture](https://ivadomed.org/tutorials/cascaded_architecture.html#cascaded-architecture) tutorial using this feature? (by cropping around the segmented spinal cord?)

mariehbourget: Thank you @andreanne-lemay!
I was able to remove the code duplication following your advice, it looks a lot better!

I answered my own question: the "cascaded" tutorial does not use the ROI feature. I ended up testing it with `ROICrop` and `suffix` in `roi_params` which I think is the right thing to do.

I tested several scenarios (Microscopy 2D with patches, MRI 2D with and without patches, and MRI with ROICrop.
Everything looks good to me and the tests are passing.

Would you be available to take a look again to make sure I did not missed something before I merge?

As for the 3D loader, I will open a separate issue as I don't have time or resources to take of it at the moment.
Thank you again!

andreanne-lemay: Thanks for looking into it and sounds good for the separate 3D issue! I'm looking through it again now, sorry for the delay!

andreanne-lemay (approved):
LGTM! Thanks again :D

mariehbourget:
For reference, I opened the issue #993 for the similar 3D subvolume case.

ivadomed/loader/mri2d_segmentation_dataset.py
```diff
@@ -196,81 +196,66 @@ def __getitem__(self, index):
...
+            # ROI goes first because params of ROICrop are needed for the followings
+            stack_roi, metadata_roi = self.transform(sample=roi_pair_slice["gt"],
+                                                     metadata=metadata_roi,
+                                                     data_type="roi")
+            # Update metadata_input with metadata_roi
+            metadata_input = imed_loader_utils.update_metadata(metadata_roi, metadata_input)
+
+        # Add coordinates of slices or patches to input metadata
+        for metadata in metadata_input:
+            metadata['coord'] = [coord["x_min"], coord["x_max"],
```
cakester, line 217: coord is not defined here if you didn't go into the else statement

PR #998, opened 2021-12-06
Experimental incorporation of tutorial jupyter notebooks open in Colab path

dyt811:
#### GitHub

#### PR contents

## Description

@naga-karthik has added additional useful PRs incorporating tutorials that are highly user friendly and streamline optimized for Google Colab usage. This is an attempt to not only slim downed version of the colab jupyter notebook (nuking the hefty 11mb meta data) but also enable directly open a copy in Colab option bypassing the hassle of requring user to save a copy etc while allowing ivadomed developer too update the jupyter notebook over time. 

@naga-karthik could you see if this is what you have in mind in terms of Google Colab integration? I preserved ALL the raw output from jupyter notebook already so almost nothing should change. It should also open DIRECTLY in CoLab via the readme.md. 

This is a proof of concept PR and we have yet to decide whether to expand it to include all other tutorials. 

Once merged, we will need to update the badge link as well!


testing/tutorials/readme.md
```diff
@@ -0,0 +1,5 @@
+[Tutorial 1 2D Segmentation UNet ![Open n Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/ivadomed/ivadomed/blob/yd/incorporating_colab_tutorials/testing/tutorials/tutorial_1_2d_segmentation_unet.ipynb)
```
naga-karthik, line 1: Just a minor type here! (I think it doesn't matter as the badge shows it correctly)

testing/tutorials/tutorial_3_uncertainty_estimation.ipynb
```diff
@@ -0,0 +1,463 @@
...
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "TOFLk8iPrzkf"
+      },
+      "source": [
+        "### Ivadomed Tutorial 2: **Uncertainty Estimation**\n",
+        "\n",
```
naga-karthik, line 27: so the link says "Tutorial 3 Uncertainty Estimation" but the title of the tutorial says "Tutorial 2". I have fixed this issue and here is the updated link: https://colab.research.google.com/drive/1L8MSyzqZbJbV336KhT3_B9RQEPnLNI2J#scrollTo=uzOt6ugooys4
(I thought it would get updated automatically, but it didn't, hence posting the link here)

naga-karthik (approved):
Thanks for taking the time to create this PR, @dyt811! I tested the links and it seems like they run okay. I had a concern that the users shouldn't be able to edit the master version (created from my Gmail) which is also sorted. I tried running with my other account and noticed that I cannot make changes to the actual tutorial without saving a copy first. So, it works out. 

Also, I was also wondering how the tutorials could be updated in the future. Since it is on Colab, I was thinking that any change I make to the master Colab file should update all the links automatically. Can you please confirm if that's the case? Or, is it something like any updates should go through the PR process every time?

jcohenadad:
I would suggest to remove the outputs on the colab books

testing/tutorials/readme.md
dyt811: Sharp eyes! Thanks!

dyt811:
> Also, I was also wondering how the tutorials could be updated in the future. Since it is on Colab, I was thinking that any change I make to the master Colab file should update all the links automatically. Can you please confirm if that's the case? Or, is it something like any updates should go through the PR process every time?

So to summarize, the process would be like this:

1) The "master copy" is now on GitHub repo. It is no longer from your Colab files. 

2) This is a regular Jupyter notebook file CONVERTED from the Colab ipynb download (Colab stored some nasty large metadata that had to be stripped out manually to bring down the size from 18mb to kb sizes. 

3) If you want to make changes to these tutorial you should be able to do one of the following depending on how big the changes you are looking at;

   a) Minor changes such as typo/strings etc, you should be able to edit the jupyter notebook file in the master via PR then merge it. 

   b) Major changes that require validation on Colab platforms (e.g. new commands, new display format etc or anything require actually running check on Colab). You will have to open the latest master jupyter notebook in colab (as shown in the example readme.md buttons to open OR via direct upload of Jupyter notebook file to Colab) and make the changes, then save as jupyter notebook files and make sure to: 
   
       i) wipe the output, 
   
       ii) delete the meta data key shown below which I believe is related to tensorboard? 

      image
     A simple check of the ipynb file to see. Keep them small.  
   
This way we can keep track the changes made to the tutorial jupyter notebook while allowing end users to create copies from it. The downside is developer needs a bit more work than usual to update the master notebook.

naga-karthik:
Thank you for the detailed explanation, @dyt811! The procedure is clear and makes sense. You can go ahead and merge the PR!

Issue #730, opened 2021-03-22
Parameter controlling the Input-level dropout


charleygros:
The "input-level dropout" has been implemented by the #705 PR. Future work could include the addition of a parameter controlling this dropout.

### Motivation for the feature

- We may want to have a different behavior across modalities: (e.g., always train with T2w or T1w, but always keep T2star).
- We may want to control the frequency of the dropping during the training.

### Description of the feature

Add a parameter that controls the frequency of this dropping (in the same way that we have a parameter to control the neurons dropping). This dropping could be modality dependent. For instance, having in the config file: [0.5, 0.7, 1] for "T2w or T1w can drop but always keep T2star".

dyt811:
Hey, Amine Elameri 

Recommended course of action and general tips to maybe point you in the general directions

* Go through the ivadomed tutorials just to have a high level overview of the entire repositories organization and architecture. If you run into problems with tutorial, ping both me and @copperwiring so we can address the issues. 
* This should also get your python environment in a reasonable place to run tests and validate functions and see how they work and step through them etc.  
* Pay particular attention and understand how the `config.json` related configuration systems works and especially in relation to the `loader parameters` related to `is_input_dropout`. 
* Also pay attention of how data modalities (MRI vs CT vs Microscopy vs others) are specified in the configuration files.  It is obviously related to above since you will be specifying modality specific dropout. 
* Carefully review the content of PR#705 and the change it introduced in addition to linked resources @andreanne-lemay can address any questions related to that. 
* Propose the new contrast specific parameters as described within this issue as @charleygros described at a Dev meeting before implementing them is probably best before you go further. 
* I would recommend leave HEMIS related comparison stuff until after this stage. If you want, you can read a bit more about HEMIS model which can handling missing data but it might be very hard to digest that until you get familiar enough with the codebase.

copperwiring:
I would recommend to install [FSL](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FslInstallation) (recommended) or [mango](http://ric.uthscsa.edu/mango/) (if you issues installing FSL and need quick glance at any image). It makes a world of difference when you see the final output in the image to understand what you did in the code. This will be particularly helpful when following tutorials.

PR #999, opened 2021-12-06
Aea/input dropout params

Elameri:
## Description
This PR implements the input-level dropout controlling parameters. When input-level dropout is enabled and the parameters [p1, p2, p3] (with p in [0, 1]) are passed, T1w, T2w and T2star have respectively p1, p2, and p3 chance to be dropped. If a channel is already empty due to an actual missing modality, it will be considered by the function as a having the parameter p=1.

To apply the input-level dropout and use the parameters [0.5, 0.5, 0.5], one can edit the config file as follow:
```
"loader_parameters": {
    "is_input_dropout": true,
    "input_dropout_params": [0.5, 0.5, 0.5]
}
```

## Linked issues
Following PR #705 Input-level dropout feature
and issue #730 about adding a parameter that controls the frequency of this dropping.

## PR contents

- [x] Modify dropout_input function to add dropping parameters
- [x] Adapt all relevant files (mri2d_segmentation_dataset, mri3d_subvolume_segmentation_dataset, bids_dataset, bids3d_dataset, keywords, loader, config_default)

### DIFF
--- a/ivadomed/loader/bids_dataset.py
+++ b/ivadomed/loader/bids_dataset.py
@@ -32,6 +32,7 @@ class BidsDataset(MRI2DSegmentationDataset):
         soft_gt (bool): If True, ground truths are not binarized before being fed to the network. Otherwise, ground
         truths are thresholded (0.5) after the data augmentation operations.
         is_input_dropout (bool): Return input with missing modalities.
+        input_dropout_params (list): Dropout parameters
 
     Attributes:
         filename_pairs (list): A list of tuples in the format (input filename list containing all modalities,ground \
@@ -46,7 +47,7 @@ class BidsDataset(MRI2DSegmentationDataset):
     def __init__(self, bids_df, subject_file_lst, target_suffix, contrast_params, model_params, slice_axis=2,
                  cache=True, transform=None, metadata_choice=False, slice_filter_fn=None, roi_params=None,
                  multichannel=False, object_detection_params=None, task="segmentation", soft_gt=False,
-                 is_input_dropout=False):
+                 is_input_dropout=False, input_dropout_params=[0.5, 0.5, 0.5]):
 
         self.roi_params = roi_params if roi_params is not None else \
             {ROIParamsKW.SUFFIX: None, ROIParamsKW.SLICE_FILTER_ROI: None}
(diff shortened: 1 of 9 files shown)


ivadomed/loader/bids_dataset.py
```diff
@@ -32,6 +32,7 @@ class BidsDataset(MRI2DSegmentationDataset):
         soft_gt (bool): If True, ground truths are not binarized before being fed to the network. Otherwise, ground
         truths are thresholded (0.5) after the data augmentation operations.
         is_input_dropout (bool): Return input with missing modalities.
+        input_dropout_params (list): Dropout parameters
```
jcohenadad, line 35: https://github.com/ivadomed/ivadomed/pull/999/files#r763065892

ivadomed/loader/mri3d_subvolume_segmentation_dataset.py
```diff
@@ -31,10 +31,11 @@ class MRI3DSubVolumeSegmentationDataset(Dataset):
         soft_gt (bool): If True, ground truths are not binarized before being fed to the network. Otherwise, ground
         truths are thresholded (0.5) after the data augmentation operations.
         is_input_dropout (bool): Return input with missing modalities.
+        input_dropout_params (list): Dropout parameters.
```
jcohenadad, line 34: https://github.com/ivadomed/ivadomed/pull/999/files#r763065892

ivadomed/loader/utils.py
```diff
@@ -426,23 +426,45 @@ def update_filename_to_nifti(filename):
...
 
 
-def dropout_input(seg_pair):
+def dropout_input(seg_pair, drop_param):
     """Applies input-level dropout: zero to all channels minus one will be randomly set to zeros. This function verifies
     if some channels are already empty. Always at least one input channel will be kept.
 
     Args:
         seg_pair (dict): Batch containing torch tensors (input and gt) and metadata.
+        drop_param (list): [p1, p2, p3], T1w, T2w and T2star have respectively p1, p2 and p3 chances to be dropped.
```
jcohenadad, line 435: I advise against using fixed T1w, T2w and T2star contrasts. The code needs to be as modular as possible. Eg: what if a user is only interested in microscopy and knows nothing about MRI: these acronyms won't be understood by that user. I suggest going at a higher abstract level when introducing this parameter. For example: if this is to be applied to multi channel, then name it a 'channel dropout' instead of 'T1w', etc.

jcohenadad (changes requested):
Added a few comments-- also, the documentation might need to be updated to reflect this case scenario

PR #1015, opened 2021-12-17
Implementation of Random Gamma Contrast Augmentation

### LINKED ISSUE #1014: Random Gamma Contrast Data Augmentation

Random gamma contrast is a popular and simple data augmentation method, implemented in other libraries such as [`torchio`](https://torchio.readthedocs.io/transforms/augmentation.html#randomgamma). We found it to be helpful in [our work](https://github.com/ivadomed/ms-challenge-2021) on MS Challenge 2021.

Random gamma contrast randomly changes the contrast of an image by exponentiating the values of the image with a parameter gamma: `new = old ** gamma`. More information about this operation can be found [here](https://en.wikipedia.org/wiki/Gamma_correction).

### Additional context

More context about this feature request can be found [here](https://github.com/ivadomed/model_seg_ms_mp2rage/issues/16).

uzaymacar:
#### GitHub

#### PR contents

## Description

This PR (i) adds the `gamma` keyword in `ivadomed/keywords.py` for the random gamma contrast augmentation and (ii) implements random gamma contrast augmentation for images through a `RandomGamma()` class  in `ivadomed/transforms.py`. The formula used for this augmentation is `new = old ** gamma`. 

We ask `log_gamma_range` as an input argument from the user, and the log gamma values are later converted to gamma values with `np.exp()`. The input argument is in the log space for more convenient user inputs. Typical log gamma values lie between -3.0 and 3.0, and therefore typical gamma values lie between 0.0498 and 20.0855.

## To-do
- [x] Give some visual example augmentation results in this PR.
- [x] Update the main ivadomed documentation to mention `RandomGamma()`.

## Linked issues

Resolves #1014.

### DIFF
--- a/docs/source/configuration_file.rst
+++ b/docs/source/configuration_file.rst
@@ -1625,6 +1625,38 @@ Available Transformations:
         }
     }
 
+.. jsonschema::
+
+    {
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "title": "RandomGamma",
+        "type": "dict",
+        "$$description": [
+            "Randomly changes the contrast of an image by gamma exponential."
+        ],
+        "options": {
+            "log_gamma_range": {
+                "type": "(float, float)",
+                "description": "Log gamma range for changing contrast."
+            },
+            "p": {
+                "type": "float"
+            }
+        }
+    }
+
+.. code-block:: JSON
+
+    {
+        "transformation": {
+            "RandomGamma": {
+                "log_gamma_range": [-3.0, 3.0],
+                "p": 0.5,
+                "applied_to": ["im"],
+                "dataset_type": ["training"]
+            }
+        }
+    }
 
 .. _Uncertainty:
 
--- a/ivadomed/keywords.py
+++ b/ivadomed/keywords.py
@@ -180,6 +180,7 @@ class MetadataKW:
     OFFSET = "offset"
     ELASTIC = "elastic"
     GAUSSIAN_NOISE = "gaussian_noise"
+    GAMMA = "gamma"
     DATA_SHAPE = "data_shape"
     SLICE_INDEX = "slice_index"
     MISSING_MOD = "missing_mod"
--- a/ivadomed/transforms.py
+++ b/ivadomed/transforms.py
@@ -1043,13 +1043,11 @@ class RandomGamma(ImedTransform):
                 # Apply gamma contrast
                 data_out = np.sign(sample) * (np.abs(sample) ** gamma)
 
-                # Return original sample and empty metadata if overflow occurs
-                if np.any(np.isinf(data_out)):
-                    metadata[MetadataKW.GAMMA] = [None]
-                    return sample, metadata
+                # Keep data type
+                data_out = data_out.astype(sample.dtype)
 
-            # Keep data type
-            data_out = data_out.astype(sample.dtype)
+                # Clip +/- inf values to the max/min quantization of the native dtype
+                data_out = np.nan_to_num(data_out)
 
             return data_out, metadata
 
--- a/ivadomed/transforms.py
+++ b/ivadomed/transforms.py
@@ -1038,8 +1038,15 @@ class RandomGamma(ImedTransform):
             metadata[MetadataKW.GAMMA] = [None]
 
         if any(metadata[MetadataKW.GAMMA]):
-            # Apply gamma contrast
-            data_out = np.sign(sample) * (np.abs(sample) ** gamma)
+            # Suppress the overflow case (due to exponentiation)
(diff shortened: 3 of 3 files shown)


uzaymacar:
Given below are visualizations of the result of random gamma contrast with gamma values of 1.0, 0.5, -0.5, 1.0.

RandomGamma_slice172
RandomGamma_slice141
RandomGamma_slice146
RandomGamma_slice147

uzaymacar:
The latest commit [52ba219](https://github.com/ivadomed/ivadomed/pull/1015/commits/52ba2194e481d61b8bae1e9f471111c2316ec45a) detects overflow cases due to exponentiation (with large values) and returns the original sample (i.e. without applying gamma contrast) when this is the case.

jcohenadad:
> The latest commit 52ba219 detects overflow cases due to exponentiation (with large values) and returns the original sample (i.e. without applying gamma contrast) when this is the case.

wouldn't it be OK to accept clipped voxels?  This could happen in 'real world' MRI data

docs/source/configuration_file.rst
```diff
@@ -1625,6 +1625,38 @@ Available Transformations:
         }
     }
 
+.. jsonschema::
+
+    {
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "title": "RandomGamma",
```
jcohenadad, line 1632: How about simply calling it `Gamma`?
Usually these transformations are random, and it makes it clearer from a user standpoint (when looking at the doc)

uzaymacar: I think the explicit usage of "Random" here is helpful because:
* first input param is a range of log gammas, as opposed to a single value of gamma, from where we *randomly* sample a gamma value
* second input param is probability of applying this augmentation, which makes the transformation *random*

It is important to make the distinction between such random augmentations and deterministic augmentations (e.g. `CenterCrop`, `NormalizeInstance`, etc.). Moreover, I see a similar logic applied in other libraries such as [`torchio`](https://torchio.readthedocs.io/transforms/augmentation.html) and [`torchvision`](https://pytorch.org/vision/stable/transforms.html). `ivadomed` also features `RandomAffine`, `RandomShiftIntensity`, and `RandomReverse`.

I am happy to be convinced otherwise, but whatever we decide here, I think we should unify `ivadomed` transformations in terms of this logic. For example, `ElasticTransform` should instead be `RandomElasticTransform` by my logic as it includes ranges and probability as input parameters.

uzaymacar:
> wouldn't it be OK to accept clipped voxels? This could happen in 'real world' MRI data

This is a good point! I am assuming I can't leave `inf` and `-inf` values in the images. I am thinking perhaps we can let overflow happen but convert the infinity values to finite values via [`numpy.nan_to_num()`](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html). Is this a plausible solution?

docs/source/configuration_file.rst
jcohenadad: you convinced me 😊 

and yes, i suggest the following changes:
- making 'Random' prefix systematic when that applies
- listing all transformatinons alphabetically in the doc for better clarity

jcohenadad:
> I am assuming I can't leave inf and -inf values in the images. I am thinking perhaps we can let overflow happen but convert the infinity values to finite values via numpy.nan_to_num(). Is this a plausible solution?

indeed, we definitely don't want `inf` 😅 . I would suggest clipping to the max/min quantization of the native dtype. Eg: if the input image is encoded in unsigned 8bit, then the min/max would be 0/256

docs/source/configuration_file.rst
uzaymacar: Opened issue #1030 regarding this.

uzaymacar:
> I would suggest clipping to the max/min quantization of the native dtype

Yep! This is exactly what  [`numpy.nan_to_num()`](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html) does.

jcohenadad:
> > I would suggest clipping to the max/min quantization of the native dtype
> 
> Yep! This is exactly what [`numpy.nan_to_num()`](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html) does.

right, but we just need to make sure that the dtype is the same between the native image and the image this numpy function will be applied to

uzaymacar:
The last commit [05d5dd3](https://github.com/ivadomed/ivadomed/pull/1015/commits/05d5dd3f8af14055c66185850b390516fb95b4f0) implements clipping to max/min quantization of sample dtype.

jcohenadad (approved):
tested, and looks good 👌   Thank you @uzaymacar !

PR #1024, opened 2021-12-21
Fix pixel size keyword in run_segment_command

mariehbourget:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
This PR fixes a bug in the `run_segment_command` of `main.py` with the pixel size keywords.
In earlier PR #946 , `PixelSize` was replaced by the `OptionKW.PIXEL_SIZE` (`pixel_size`) instead of the `MetadataKW.PIXEL_SIZE` (`PixelSize`), so the `--segment` command will not run for microscopy:
```
Traceback (most recent call last):
  File "/home/mhbourget/venv-ivadomed-38c/bin/ivadomed", line 11, in <module>
    load_entry_point('ivadomed', 'console_scripts', 'ivadomed')()
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 589, in run_main
    run_command(context=context,
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 362, in run_command
    run_segment_command(context, model_params)
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 286, in run_segment_command
    options[OptionKW.PIXEL_SIZE] = bids_df.df.loc[bids_df.df['filename'] == subject][OptionKW.PIXEL_SIZE].values[0]
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/pandas/core/frame.py", line 2906, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/mhbourget/venv-ivadomed-38c/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 2897, in get_loc
    raise KeyError(key) from err
KeyError: 'pixel_size'
```

## Linked issues
No related issue.

PR #1025, opened 2021-12-21
Update microscopy following BEP release

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR updates the code following Microscopy BEP release in BIDS release 1.7.0.

1. Include the `PixelSizeUnits` metadata key when loading microscopy files.
    - `PixelSizeUnits` can be either `mm`, `um` or `nm` (previously only `um`) and the pixel size of the image is converted to `mm` in the `segmentation_pair.py/get_microscopy_pixelsize` function. This  works for the `--train` and `--test` command.
    - A new test (`test_microscopy_pixelsize`) is added in `test_loader.py` to confirm the correct conversion of units.
    - For segmentation, the pixel size units metadata is also added to the options of the `inference.py/segment_volume` API (used by ADS) and in `main.py/run_segment_command` for the command `--segment`.

2. Update the force indexing of microscopy file to remove file extensions not BIDS compliant (.tiff, .ome.tiff, .ome.tf2, .ome.tf8) and `microscopy` data type directory (replaced by `micr` only).

TO DO before the merge:

- [x] PR #1024 needs to go first.
- [x] Update data_axondeepseg_tem dataset
- [x] Update wakehealth (Osvaldo's) dataset
- [x] Update data-testing --> release r20220221 from [PR](https://github.com/ivadomed/data-testing/pull/17).
- [x] Add link to new release of data-testing [here](https://github.com/ivadomed/ivadomed/blob/d1b9e118f491d9e5e88523f8d0f0cae65b26ecca/ivadomed/scripts/download_data.py#L26) (the tests will not pass without this update).

## Linked issues
No related issues.


mariehbourget:
This PR is ready for review. It updates the ivadomed implementation for microscopy following the release of BIDS 1.7.0.
Details in the first comment.

ivadomed/keywords.py
```diff
@@ -194,6 +195,7 @@ class MetadataKW:
     INPUT_FILENAMES = "input_filenames"
     ROI_METADATA = "roi_metadata"
     PIXEL_SIZE = "PixelSize"
+    PIXEL_SIZE_UNITS = "PixelSizeUnits"
```
dyt811, line 198: Just a question, somewhat related to this PR, I see this keyword is added both at option/metadata KeyWords class. I need to review a bit more to tell but off the top of your head, do you think it is necessary to have these two separate classes? We did the keywords refactoring using best estimation from code perspective so not sure if we could have been more efficient or consolidated here?

ivadomed/loader/bids_dataframe.py
```diff
@@ -81,9 +81,9 @@ def create_bids_dataframe(self):
             # validate=True by default for both indexer and layout, BIDS-validator is not skipped
             # Force index for samples tsv and json files, and for subject subfolders containing microscopy files based on extensions.
             # Force index of subject subfolders containing CT-scan files under "anat" or "ct" folder based on extensions and modality suffix.
-            # TODO: remove force indexing of microscopy files after BEP microscopy is merged in BIDS
+            # TODO: remove force indexing of microscopy files after Microscopy-BIDS is integrated in pybids
             # TODO: remove force indexing of CT-scan files after BEP CT-scan is merged in BIDS
-            ext_microscopy = ('.png', '.tif', '.tiff', '.ome.tif', '.ome.tiff', '.ome.tf2', '.ome.tf8', '.ome.btf')
```
dyt811, line 86: Just curious, wasn't `tiff` the same as `tif` as equally valid extension? Or is it because we are doing partial matching and `.tif` can pickup both?

ivadomed/loader/bids_dataframe.py
```diff
@@ -94,8 +94,8 @@ def create_bids_dataframe(self):
                     subject_path = path_object.parts[subject_path_index]
                     if path_object.name == "samples.tsv" or path_object.name == "samples.json":
                         force_index.append(path_object.name)
-                    if (path_object.name.endswith(ext_microscopy) and (path_object.parent.name == "microscopy" or
-                            path_object.parent.name == "micr") and subject_path.startswith('sub')):
+                    if (path_object.name.endswith(ext_microscopy) and path_object.parent.name == "micr" and
```
dyt811, line 97: Just wondering. Given now that "micr" is standarized in BIDS, should we actually create a keyword for this? I am not sure if we have a relevant class already.

ivadomed/loader/segmentation_pair.py
```diff
@@ -337,58 +337,66 @@ def convert_file_to_nifti(self, filename, extension, is_gt=False):
...
+            pixel_size_units = self.metadata[0][MetadataKW.PIXEL_SIZE_UNITS]
+            if pixel_size_units == "mm":
+                conversion_factor = 1          # Conversion factor from mm to mm
+            elif pixel_size_units == "um":
+                conversion_factor = 0.001      # Conversion factor from um to mm
+            elif pixel_size_units == "nm":
+                conversion_factor = 0.000001   # Conversion factor from nm to mm
+            else:
+                raise RuntimeError(f"The PixelSizeUnits '{pixel_size_units}' of '{Path(filename).stem}' is not "
+                                   f"supported. ivadomed supports the following PixelSizeUnits: 'mm', 'um' and 'nm'.")
```
dyt811, line 370: Great friendly useful error!

dyt811 (commented):
Looking solid on my side from GitHub reviews. Going to change to approval once locally tested shortly.

ivadomed/keywords.py
mariehbourget: That is a good question!

So here is the catch:
- In the `MetadataKW` class, the name of the keywords come from the name of the JSON field as standardized in BIDS and indexed by pybids. So `PixelSize` and `PixelSizeUnits` in Pascal case, non-negotiable as per BIDS ;) They are used in the loader [here](https://github.com/ivadomed/ivadomed/blob/8090c86d63218a82b0c1f3c1709d36b6e1dcd47f/ivadomed/loader/segmentation_pair.py#L340).
- In the `OptionKW` class, the name of the keywords are based on the ivadomed variables name convention, so `pixel_size` and `pixel_size_units` in snake_case (I had them in PascalCase at the begining but was asked to changed them to snake case). In this case, the variable are unrelated to BIDS. They are used by the segment_volume API [here](https://github.com/ivadomed/ivadomed/blob/8090c86d63218a82b0c1f3c1709d36b6e1dcd47f/ivadomed/inference.py#L321).

A great example of their different usage is in the run_segment_command (when running `--segment`) [here](https://github.com/ivadomed/ivadomed/blob/8090c86d63218a82b0c1f3c1709d36b6e1dcd47f/ivadomed/main.py#L284), where the PixelSize (from BIDS metadata) is converted to pixel_sixe (for segment_volume options).

ivadomed/loader/bids_dataframe.py
mariehbourget: You are right, `tiff` and `tif` are exactly the same file type and are valid "TIFF" files. However, they are not considered valid for BIDS.
In BIDS, they asked us to minimize the number of extensions considered valid to be a "BIDS-compliant" dataset. So the 4 extensions I left there (`png`, `tif`, `ome.tif` and `ome.btf`) are the only ones that are microscopy BIDS compliant.

As of today, we could continue to force-index the non-valid extensions, but eventually we will integrate Microscopy to pybids and they won't work anymore. So, I think it's best to remove them from the possibilities as soon as possible.

ivadomed/loader/bids_dataframe.py
mariehbourget: I don't think we need to create keywords for this. It is a BIDS "data type" like any other. For example in MRI, there are several data types (ex: `anat`, `dwi`, `func`). We normally don't see them in our ivadomed code because it's pybids and not ivadomed that deals with them to index the files.
Again, in the microscopy case, we see "micr" in the code because we need to force-index the files as microscopy in not integrated in pybids yet.

ivadomed/keywords.py
dyt811: Great! Thanks a lot for the detailed explanation.

ivadomed/loader/bids_dataframe.py
dyt811: Thanks for the info. Greatly appreciate it!

dyt811 (approved):
Past local test. Looking good.  :D. Sorry took a while for me to get around to test it. Greatly appreciate the in-depth background explanations.

Issue #1030, opened 2021-12-22
Making "Random" prefix systematic for random transformations 


uzaymacar:
### Motivation for the feature

The "Random" prefix is usually added to transformations that have a random component, and this can be seen in libraries like [`torchio`](https://torchio.readthedocs.io/transforms/augmentation.html) and [`torchvision`](https://pytorch.org/vision/stable/transforms.html). `ivadomed` also features `RandomAffine`, `RandomShiftIntensity`, and `RandomReverse` by this logic.

It is important to  make a distinction between such random transformations and deterministic transformations (e.g. `CenterCrop`, `NormalizeInstance`, etc.) for sake of clarity from a usage perspective.

### Description of the feature

Add a "Random" prefix for `ivadomed` transformations that have a random component within such as:
* having an input param that is a range of values - as opposed to a single value - from where we *randomly* sample a single value
* having an input param `p` for the *probability* of applying this transformation per iteration

By this logic, for example, `ElasticTransform` would instead be `RandomElasticTransform`. While doing this, we can also alphabetize the transformations in the docs.

### Additional context

Discussion arised from [here](https://github.com/ivadomed/ivadomed/pull/1015#discussion_r773988550).

jcohenadad:
it would be great to have the feedback from @mariehbourget @andreanne-lemay @charleygros

andreanne-lemay:
I think it's fine, however, we should make sure the old names still work (e.g., the old class could call the new random class). Otherwise, older config files won't work anymore.

Issue #1029, opened 2021-12-22
Consider merging "Usage" and "Scripts" into a single documentation entry


jcohenadad:
### Motivation for the feature

Currently, the CLI script "ivadomed" is under "Usage", while the other CLI scripts are under "Scripts". I find this distinction is confusing and suggest to simply merge all CLI scripts under a section called "Command-Line Tools" or simply "Usage".

uzaymacar:
Further suggestions on this are [here](https://github.com/ivadomed/ivadomed/pull/1061#pullrequestreview-868294862) and [here](https://github.com/ivadomed/ivadomed/pull/1061#issuecomment-1033111713).

PR #826, opened 2021-06-18
Remove all NumpyToTensor configs keys

cakester:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

https://github.com/ivadomed/ivadomed/pull/818 this has been merged yestarday, I found there is more NumpyToTensor in config files in the repo, so this pr removes them all

## Linked issues

resolves #815

PR #1061, opened 2022-01-29
Diffused transformations in docs to preprocessing, DA, and utilities

### LINKED ISSUE #1059: Transformations documentation: separate between preprocessing and data augmentation
As seen in the below lines

https://github.com/ivadomed/ivadomed/blob/76b36a0a0f7141feb2d5b00b33e4c3a06865fc2c/ivadomed/transforms.py#L1181

we specify (i) resampling, (ii) center-cropping, and (iii) roi-cropping as preprocessing steps amongst all the transformations.

As per the discussion today, we should probably make this separation clear in the documentation. We can have a "preprocessing" and "data augmentations" subsection under [all the transformations listed](https://ivadomed.org/configuration_file.html#transformations).

uzaymacar:
#### GitHub

#### PR contents

## Description

This PR diffuses the "Transformations" heading into (i) "Preprocessing", (ii) "Data Augmentations", and (iii) "Utility Transformations" headings at the same level. 

In doing so, I also alphabetized the transformations under the new "Data Augmentations" as previously mentioned [here](https://github.com/ivadomed/ivadomed/issues/1030#issue-1086998894).

In the reviewing of this PR, it would be great to get some feedback for
* perhaps a better / shorter name for "Utility Transformations"
* currently customizable parameters (e.g. `"applied_to"`, `"dataset_type"`, etc.) are repeated in all three headings which is not ideal 

## Linked issues

Resolves #1059

### DIFF
--- a/docs/source/configuration_file.rst
+++ b/docs/source/configuration_file.rst
@@ -1223,30 +1223,47 @@ Cascaded Architecture Features
     }
 
 
-Transformations
----------------
+Preprocessing
+-------------
 
-Transformations applied during data augmentation. Transformations are sorted in the order they are applied to the image samples. For each transformation, the following parameters are customizable: 
+Transformations applied during preprocessing. Transformations are sorted in the order they are applied to the image samples. For each transformation, the following parameters are customizable:
 
 - ``applied_to``: list between ``"im", "gt", "roi"``. If not specified, then the transformation is applied to all loaded samples. Otherwise, only applied to the specified types: Example: ``["gt"]`` implies that this transformation is only applied to the ground-truth data.
 - ``dataset_type``: list between ``"training", "validation", "testing"``. If not specified, then the transformation is applied to the three sub-datasets. Otherwise, only applied to the specified subdatasets. Example: ``["testing"]`` implies that this transformation is only applied to the testing sub-dataset.
-
 
 .. jsonschema::
 
     {
         "$schema": "http://json-schema.org/draft-04/schema#",
-        "title": "NumpyToTensor",
+        "title": "Resample",
         "type": "dict",
-        "description": "Converts nd array to tensor object."
+        "options": {
+            "hspace": {
+                "type": "float",
+                "range": "[0, 1]",
+                "description": "Resolution along the first axis, in mm."
+            },
+            "wspace": {
+                "type": "float",
+                "range": "[0, 1]",
+                "description": "Resolution along the second axis, in mm."
+            },
+            "dspace": {
+                "type": "float",
+                "range": "[0, 1]",
+                "description": "Resolution along the third axis, in mm."
+            }
+        }
     }
 
 .. code-block:: JSON
 
     {
         "transformation": {
-            "NumpyToTensor": {
-                "applied_to": ["im", "gt"]
+            "Resample": {
+                "hspace": 0.75,
+                "wspace": 0.75,
+                "dspace": 1
             }
         }
     }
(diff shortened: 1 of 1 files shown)


kanishk16:
Would it make sense to have a subsection as transformations with subsubsections as (i) "Preprocessing", (ii) "Data Augmentations", and (iii) "Utilities" as shown below:

image

(Why?) ivadomed at its core uses PyTorch and this aligns with `torchvision.transforms.Compose` where we eventually chain several transformations together ranging from preprocessing to data augmentations. This even resolves repetition of customizable parameters that could be stated just once under Transformations subsection.

ref: [Sphinx subsubsections](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections)

uzaymacar:
@kanishk16 It would definitely make sense, thanks for your suggestion! Another supporting argument for your idea is that the key used in the config file is `"transformations"` and it would be good to retain a "Transformations" heading (similar to how we have it with `"training_parameters"` key and "Training Parameters" heading and so on).

A counter-argument for this idea is non-consistency with "Preprocessing" heading as @jcohenadad noted [here](https://github.com/ivadomed/ivadomed/issues/1059#issuecomment-1024373651), but I would be in favor of your idea because it solves multiple problems.

kanishk16:
> A counter-argument for this idea is non-consistency with "Preprocessing" heading as @jcohenadad noted [here](https://github.com/ivadomed/ivadomed/issues/1059#issuecomment-1024373651), but I would be in favor of your idea because it solves more problems.

@uzaymacar Thanks for highlighting this, I completely missed this. Apart from this trade-off, another plausible solution could be to have a different key (`preprocessing`) and a corresponding subsection for it to be more consistent. And, data augmentations along with utility transformations under a single subsection of transformations. I guess this solution is more intuitive (feedback appreciated on this thought). The solution we choose depends upon our target audience for ivadomed. 

Since we have more pressing issues as of now, we could go with the first one for the time being and perhaps open up another issue to resolve this later on.

jcohenadad:
> @uzaymacar Thanks for highlighting this, I completely missed this. Apart from this trade-off, another plausible solution could be to have a different key (preprocessing) and a corresponding subsection for it to be more consistent. And, data augmentations along with utility transformations under a single subsection of transformations. I guess this solution is more intuitive (feedback appreciated on this thought). The solution we choose depends upon our target audience for ivadomed.

@kanishk16 could you please show us in a 'tree'-like display what your TOC suggestion would look like?

naga-karthik:
Thanks, @uzaymacar for restructuring the docs! It is much required! I am in favor of keeping the name "Transformations" as it is currently in the TOC and it also corresponds well with the keys that we have in the config file. However, as "Preprocessing" shouldn't technically be under "Transformations", I agree with @kanishk16's comment on creating new key-value pairs for preprocessing in the config file and having another subsection (I think preprocessing keys are important enough to justify a new subsection!). 

At the risk of extending the discussion, here are my 2 more cents on this: 

1. Personally, I would do away with the `NumpyToTensor` key in the config file in the first place, because this conversion is something I expect to be done by default without having the user know about this (I mean, the basic building block in PyTorch is a Tensor). Moreover, in all the examples I have seen, the value for that key is always empty ie `NumpyToTensor: {}`, which suggests that removing it at some point should be considered.
2. If (1) above is implemented, then we can push "Normalize" under "transformations". This is because, from [here](https://pytorch.org/vision/main/generated/torchvision.transforms.Normalize.html), I see that normalize is indeed under `torchvision.transforms` and we also do not have to worry about "utilities" or "utility transformations" anymore.

jcohenadad (commented):
I would personally rename "Configuration File" into "Usage"

currently, "Usage" is pretty light and only has the CLI info, whereas Configuration file _should_ be part of Usage (moreover, some of the params could also be called via the CLI, right?).

But maybe that should be part of another PR?

uzaymacar:
Thanks all for the excellent feedback! I would like to pick a few points to **handle in separate issues + PRs** if possible:
* > I would personally rename "Configuration File" into "Usage" > 

     We should think of this together with #1029.

* > I would do away with the NumpyToTensor key

     Agreed. Why was this implemented in the first place -- if no answer let's proceed with your suggestion!
     
* > have a different key (preprocessing)

  > creating new key-value pairs for preprocessing in the config file

     Agreed. 

Moving onto this PR, IIUC currently we have two proposals on top of the current state of the branch:

* h1: Transformations, h2: Preprocessing, h2: Data Augmentations, h2: Utility by @kanishk16's first suggestion
* h1: Preprocessing, h1: Transformations by @kanishk16's second suggestion + @naga-karthik's suggestion  

I was in favor of first option because all of preprocessing, data augmentations, and utility are implemented as and are transformations: (a) being implemented in [`ivadomed/transforms.py`](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/transforms.py), (b) based on the [`ImedTransform`](https://github.com/ivadomed/ivadomed/blob/97b5772374660b12895f7a458941f468090e8bf8/ivadomed/transforms.py#L83) class, and (c) have common parameters as mentioned [here](https://github.com/ivadomed/ivadomed/pull/1061#issue-1117979313). I like the second option because it (a) doesn't hide everything under "Transformations", (b) establish consistency with "Postprocessing", and (c) remove Utility as a heading.

How can we fuse the two? LMK if I have missed anything! Perhaps first taking care of the tangential but excellent points mentioned at the beginning might shine some light on this discussion.

mariehbourget:
Very interesting discussion!
Here is my two-cents:

**Regarding Usage, Configuration File and Scripts:**
> * > I would personally rename "Configuration File" into "Usage" >
>   
>   We should think of this together with [Consider merging "Usage" and "Scripts" into a single documentation entry #1029](https://github.com/ivadomed/ivadomed/issues/1029).

I would be in favor of keeping the term and section "configuration file" but reorganizing the "Usage" section as:
h1: Usage, h2: Command line tools, h2: Configuration file, h2: Scripts

**About NumpyToTensor:**
> * > I would do away with the NumpyToTensor key
>   
>   Agreed. Why was this implemented in the first place -- if no answer let's proceed with your suggestion!

As per https://github.com/ivadomed/ivadomed/pull/818, https://github.com/ivadomed/ivadomed/pull/826 and https://github.com/ivadomed/ivadomed/pull/830, the NumpyToTensor key is not necessary in the config file and is applied directly in the code [here](https://github.com/ivadomed/ivadomed/blob/cc28e360ca2f09de51e2d4be165d8db20f9d79a2/ivadomed/transforms.py#L157) (for transforms that are not preprocessing). The code is retro-compatible for config files that contains the key [here](https://github.com/ivadomed/ivadomed/blob/cc28e360ca2f09de51e2d4be165d8db20f9d79a2/ivadomed/transforms.py#L123) (the key is ignored).
So I think it's safe to remove it from the transforms since it is not used anymore on a config file basis.

**About preprocessing:**
> * > have a different key (preprocessing)
>   > creating new key-value pairs for preprocessing in the config file
>   
>   Agreed.

I'm less in favor of this one because preprocessing transforms ARE transforms (same arguments as Uzay for option 1).
They are in the `transforms.py` file and share classes with data augmentation (ex: `Compose`). I don't see them as the opposite of postprocessing since preprocessing applies to the image + GT whereas postprocessing applies to the predictions.
Moreover, we have a [preprocessing.py](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/preprocessing.py) file that is something else completely.
I'm not against it completely but would be careful moving forward with this refactoring.

**About the doc hierarchy:**
> Moving onto this PR, IIUC currently we have two proposals on top of the current state of the branch:
> 
> * h1: Transformations, h2: Preprocessing, h2: Data Augmentations, h2: Utility by @kanishk16's first suggestion
> * h1: Preprocessing, h1: Transformations by @kanishk16's second suggestion + @naga-karthik's suggestion

Because of my previous point, I will suggest a third option.
- h1: Transformations, h2: Preprocessing, h2: Data augmentation (including NormalizeInstance), NumpyToTensor is removed.

uzaymacar:
Latest commit [3035fa5](https://github.com/ivadomed/ivadomed/pull/1061/commits/3035fa57fac95492f330f601915c01e67a3fdcde) implements `h1: Transformations, h2: Preprocessing, h2: Data Augmentations` and removes NumPyToTensor as suggested by @mariehbourget -- I think it is a good conclusion of all of the ideas presented in this discussion, thanks Marie-Hélène! I am marking this PR ready-to-review as a result.

mariehbourget:
> where would 'postprocessing' be located? I would suggest using this more intuitive display to get the discussion going:
> 
> ```
> ├── GETTING STARTED
> │   ├── Installation
> │   ├── Data
> │   ├── Usage
> │   │   ├── Command line tools
> │   │   ├── Configuration file
> │   │   └── Scripts
> │   ├── Transformations
> │       ├── Preprocessing
> │       ├── Data Augmentations
> │       └── Postprocessing
> ├── Architectures
> ├── Pre-trained models
> ├── Help
> ```

Just to clarify: "Transformations" is nested under "Configuration file" with all the other config parameters (including Postprocessing), see updated structure below.

As for "Postprocessing", I don't think it should be inside "Transformations" because:
- They are not tranformations i.e. not based on ImedTransform class or torchvision_transforms
- Transformations apply to images + GT whereas Postprocessing apply to predictions

```
├── GETTING STARTED
│   ├── Installation
│   ├── Data
│   ├── Usage
│   │   ├── Command line tools
│   │   ├── Configuration file
│   │   │   ├── <other previous configuration file categories>
│   │   │   ├── Transformations
│   │   │   │   ├── Preprocessing
│   │   │   │   └── Data Augmentations
│   │   │   ├── Uncertainty
│   │   │   ├── Postprocessing
│   │   │   └── Evaluation Parameters
│   │   └── Scripts
│   ├── Architectures
│   ├── Pre-trained models
│   ├── Help
```

jcohenadad:
from https://github.com/ivadomed/ivadomed/pull/1061#issuecomment-1034116998, i feel that an h3 nesting for 'Transformations' (which is a pretty important section) will likely be missed.

Maybe we should reconsider some of the proposed changes about 'transformations'? And if we wish to merge 'scripts' with usage, if we consider that `ivadomed` _is_ a script (which it is), then we could maybe do something like this:

~~~
├── GETTING STARTED
│   ├── Installation
│   ├── Data
│   ├── Usage  --> list all scripts here at the same level: 'ivadomed', 'ivadomed_visualize_transforms', etc.
│   ├── Configuration file
│   │   ├── <other previous configuration file categories>
│   │   ├── Transformations
│   │   │   ├── Preprocessing
│   │   │   └── Data Augmentations
│   │   ├── Uncertainty
│   │   ├── Postprocessing
│   │   └── Evaluation Parameters
│   ├── Architectures
│   ├── Pre-trained models
│   ├── Help
~~~

to sum up the proposed changes are:
- Move 'Configuration file' _after_ 'Usage', because I fear people will get overwhelmed by the 'Configuration file' and should first get the 'big picture' of how ivadomed works, and _then_ get into the details of how exactly they should use it
- Consider `ivadomed` a script
- Move all scripts to 'Usage'

PR #1069, opened 2022-02-02
Support for WandB Experimental Tracking 

### LINKED ISSUE #1010: ML Experiment Tracking and Management
### Discussed in https://github.com/ivadomed/ivadomed/discussions/1004

<div type='discussions-op-text'>

<sup>Originally posted by **naga-karthik** December 13, 2021</sup>
The aim of this post is to open a stage for discussing various alternatives for ML experiment tracking. 

### Motivation:

As we train more and more DL models, keeping track of the models (i.e. how many filters were used, what learning rate, epochs, etc.) becomes increasingly difficult. The brute-force way is to manually add entries onto a spreadsheet but there are several much better alternatives that make this task a lot easier.

### Trackers Explored:

- [Weights and Biases (wandb.ai)](https://wandb.ai/site): This was tested (by @naga-karthik) while working on a course project. My overall experience has been good. There are several useful features, for e.g. grouping experiments into "Runs", visualization of all loss curves, logging of hyperparameters and gradients. Most importantly, everything can be found online (on the user's wandb profile), which avoids running it on localhost (as it is done on Tensorboard). 
   - While it is easy to use on a personal basis (an account can be created free of charge), I am unaware of the potential issues it might create when we want to integrate it into the ivadomed system. Particularly, will it be considered as an "enterprise" account, or, are the users who will use ivadomed expected to create wandb accounts as a prerequisite? Need some discussion here.
   - [Here](https://wandb.ai/site/pricing) are the pricing details for wandb 
  - Pros:
    - @uzaymacar and @naga-karthik are familiar with it (maybe others in the lab?) and can fill up the pros/cons

- [Comet.ml](https://www.comet.ml/site/): has been tried @dyt811 
   -  Comet.ml has apparently good integration with a wide variety of ML run time environments/framewokrs: troch/tensorflow/SKLearn/MLFlow, etc. 
   - Key commands for logging [Comet.ml cheat sheet: supercharge your machine learning experiment management](https://medium.com/comet-ml/comet-ml-cheat-sheet-supercharge-your-machine-learning-experiment-management-7786e08f8e9e) 
   - [Hyperparameters search](https://www.comet.ml/docs/python-sdk/introduction-optimizer/)
      - Also check out Optuna also for optimization
   - Pro: 
      - Off the shelf, centralized web portal to get started with API keys within literally minutes to log metrics/data
      - Good for metrics/parameters/experiments tracking with minimum amount of code. 
      - compatible with multiple experimental tracking solutions
      - has built in optimizer
      - Probably more commercial support if you are on paid tier? 
      - Free for academia affiliated researchers
   - Con: 
      - Expensive if you have to buy it (a few hundred per month per seat), 
      - not open source
      - not locally deployable AFAIK. 

- [MLFlow](https://mlflow.org): has been around for a while too
    - Pro: 
       - Free. Open source
       - Can be used to reproduce/replicate simple experiments (have not tried run on more complicated deep learning)
       - Probably more community support? 
    - Con:
       - Lack an off the shelf centralized SAAS solution. You will have to host your own (e.g. kind of like a tensorboard portal)
       - Slightly less feature than the Comet.ml solution in the track space. Not 100% sure. ? 

### Future:

In my opinion, having an experiment tracker (that can be accessed online without localhost) is definitely a plus point. It turns out [this issue](https://github.com/ivadomed/ivadomed/issues/801) already brought up this topic and some more information on this can be found here. I hope that this post will motivate more discussions and possibly others can chip in with other (newer) experiment trackers that they might have used.

</div>

naga-karthik:
#### GitHub

#### PR contents

## Description

This PR aims to integrate [WandB](https://wandb.ai) in ivadomed for experimental tracking and management. Currently, it is still a WIP and the necessary wandb calls have been added just below the Tensorboard calls in the corresponding files. Once wandb is fully integrated and all the features we had with Tensorboard are merged, the Tensorboard calls would be removed. Since we want wandb to be optional (like the `debugging` key in the config file), a few variables reflecting this have been added in `keywords.py`.

Since this is a draft PR, it would be great if there are some suggestions for the files I should be looking at (in addition to `training.py` and `visualize.py`), a better way to aggregate all hyperparameters to be logged (see variable `cfg` in `training.py`), possibly better key names for the config file etc. 

## Linked issues

Resolves #1010


arbalest2:
Shouldn't wandb be added to [requirements](https://github.com/ivadomed/ivadomed/blob/master/requirements_common.txt)?

arbalest2:
You might want to add wandb package to the [requirements file](https://github.com/ivadomed/ivadomed/blob/master/requirements_common.txt)

ivadomed/training.py
```diff
@@ -54,6 +55,25 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
     # Write the metrics, images, etc to TensorBoard format
     writer = SummaryWriter(log_dir=path_output)
 
+    # Collect all hyperparameters into a dictionary
+    cfg = {
```
mariehbourget, line 59: I noticed some possible parameters are not included in the `cfg`, is that expected? (ex: `dropout_rate`, `final_activation`, `length_2D`)
I feel like adding params individually may be very high maintenance in the long term and prone to error (if the goal is to have everything).

Would it be possible to merge both the `training_params` dict and `model_params` dict into this `cfg`?

mariehbourget:
> Since this is a draft PR, it would be great if there are some suggestions for the files I should be looking at (in addition to `training.py` and `visualize.py`),

The only one I see is the [training_curve](https://ivadomed.org/scripts.html#ivadomed-training-curve) script to plot graph from tensorboard. This may be obsolete depending on the capabilities of wandb?

ivadomed/keywords.py
```diff
@@ -8,6 +8,10 @@ class ConfigKW:
     MODEL_NAME = "model_name"
     MODIFIED_3D_UNET = "Modified3DUNet"
     DEBUGGING = "debugging"
+    EXPERIMENT_TRACKING = "experiment_tracking"
+    PROJECT_NAME = "project_name"
+    GROUP_NAME = "group_name"
+    RUN_NAME = "run_name"
```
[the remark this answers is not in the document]
naga-karthik: Thanks for the suggestion! I'm afraid it would be difficult to do so. There's a hierarchy that wandb follows (see figure attached) - experiments are categorized according to project name (sci-zurich) --> group name (SagT2w_training) --> runs. If we condense this, it would be diffcult for tracking the experiments efficiently (which is indeed the purpose of wandb). 

wandb_crop

That said, I'll read up more on this and see if they can be shortened in any way possible!

ivadomed/training.py
naga-karthik: > I feel like adding params individually may be very high maintenance in the long term and prone to error (if the goal is to have everything).

Thank you! Yes, the goal is to log as many hyperparameters as possible but while I was adding them 1-by-1 I felt that it is indeed error-prone (like the parameters you pointed out which I missed)! I'll look into the linked config manager file but it seems that merging both` training_params` and `model_params` is an efficient option. Thanks again for the pointers! :)

jcohenadad: > Would it be possible to merge both the training_params dict and model_params dict into this cfg?

that would be brilliant 🙌

ivadomed/keywords.py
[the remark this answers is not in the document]
dyt811: @naga-karthik I think Uzay mean the keyword in ConfigKW. 

If you create a new keyword class e.g. `WandBConfigKW` (which essentially indicates a new context) this can potentially allow people to know that `WandBConfigKW.GROUP` can only mean group in the WandB context and allow to you to keep it shorter while keeping the code more readable as it will appear int eh code as "WandBConfigKW.GROUP` which hints the context pretty explicitly. 

Also I think EXP_TRACKING is probably clearly enough. This is also the advantage of using keywords class: easy refactoring/hinting to developers without changing the underlying key string.

ivadomed/training.py
dyt811: @naga-karthik so I wasn't sure if WandB have this, but Comet.ML actually picked up some parameters/metrics automatically when I was playing with its integration with tensorflow. In your experience, does WandB do some auto tracking once invoked in the context of Torch? The easiest way to usually tell is to log into WandB and see what is tracked vs what you told it to explicitly track. 

Also, it will probably be good to directly log entire fully overwritten JSON file once the myriad of configuration integration is complete.

naga-karthik: Thank you for the suggestion @mariehbourget! I could merge both `training_params` and `model_params` and the result now is that everything is being logged. It can be seen [here](https://wandb.ai/naga-karthik/ivado-wandb-testing/groups/temp_v2/table?workspace=user-naga-karthik) in a table form. Right now, I feel that the column names are too big (since they have been directly taken from the above dicts, I am looking into how to make it more informative.

@dyt811 So as far as the parameters/metrics related to your model are concerned, wandb logs them only when you tell it to. There are other external aspects as the environment, system, num_gpus, usage, etc. which it logs automatically (see "System" drop-down in [this](https://wandb.ai/naga-karthik/ivado-wandb-testing/groups/temp_v2/workspace?workspace=user-naga-karthik) link). In addition, it also logs some metadata and media which can be seen in "Files" in [this](https://wandb.ai/naga-karthik/ivado-wandb-testing/groups/temp_v2/files?workspace=user-naga-karthik) link.   

Also, could you please elaborate a bit more on this:

> directly log the entire fully overwritten config JSON

Right now, the input config file is already saved in the output folder after training, right? It will still be saved with the wandb keys added to the config file. Or, did you mean something else?

naga-karthik:
Yes, this script should be obsolete with wandb logging. Currently, with the script, I see that the train and val losses are plotted together and likewise for multiple output paths, however, wandb compares train losses and val losses independently across several runs (which, imo, is better).

ivadomed/keywords.py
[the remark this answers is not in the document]
naga-karthik: Thank you for the clarification @dyt811! I have created a new keyword class as suggested (see 265f8d28c75eef5234dfcb989f49c5acf1ec7b2c for details). 

Also, I have removed the variable EXP_TRACKING because I have observed that currently ivadomed logs the loss curves on tensorboard by default and the images are logged only when `debugging=true`. To follow this, I did the same with wandb as well (log losses/metrics by default and images when debugging=true).

mariehbourget:
> Yes, this script should be obsolete with wandb logging.

Thinking about it, the training_curve script was also outputting a csv file with all the values, which was useful to plot "custom" graph of the training. Is that something we can output from wandb as well?

naga-karthik:
Yes, the WandB UI has an option just for this! You can find a small video about how to export a runs table into a csv file [here](https://docs.wandb.ai/ref/app/features/runs-table#export-runs-table-to-csv) (thank you @Andjelaaaa for finding the link!).

ivadomed/training.py
```diff
@@ -54,6 +55,12 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
     # Write the metrics, images, etc to TensorBoard format
     writer = SummaryWriter(log_dir=path_output)
 
+    # Collect all hyperparameters into a dictionary
+    cfg = { **training_params, **model_params}
+    # Initialize WandB with metrics and hyperparameters
+    wandb.init(project=wandb_params["project_name"], group=wandb_params["group_name"], 
+                                name=wandb_params["run_name"], config=cfg)
```
dyt811, line 62: Again, we should not always assume the `wand_params` parental key even exists here. Here using the` .get()`  may still cause a crash as the `wand_params` could be `None` and cannot call `.get()` on `None`.  For more about [why use get](https://towardsdatascience.com/18-common-python-anti-patterns-i-wish-i-had-known-before-44d983805f0f#4d61) and [SO simple explanations](https://stackoverflow.com/a/11041421).

I recommend something like this: 
```
# Establish "wandb_params" exists and also its has a API key in it. 
if wandb_params and wandb_params[WandbKW.API_KEY]:

  # Setup default names for wandb.init, not assuming they exists by providing default alternatives.
   project_name = wand_params.get("WandbKW.PROJECT_NAME", "Default_Project")
   group_name= wand_params.get("WandbKW.GROUP_NAME", "Default_Group")
   run_name = wand_params.get("WandbKW.RUN_NAME", "Default_Run")
   
   # Login OR you can also use os.environ["WANDB_API_KEY"] to set the key into the environment. 
   wandb.login(key=wandb_params[WandbKW.API_KEY])

   # Init Project
   wandb.init(project=project_name, group=group_name , name=run_name, config=cfg)
```

ivadomed/training.py
```diff
@@ -118,6 +125,9 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
                                                     num_epochs)
     logger.info("Scheduler parameters: {}".format(training_params["scheduler"]["lr_scheduler"]))
 
+    # Logs gradients (at every log_freq steps) to the dashboard.
+    wandb.watch(model, log="gradients", log_freq=wandb_params["log_grads_every"])
```
dyt811, line 129: You may wish to add a check like above: 
```
# Establish "wandb_params" exists and also its has a API key in it. 
if wandb_params and wandb_params[WandbKW.API_KEY]:
```

ivadomed/training.py
```diff
@@ -154,7 +164,7 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
         start_time = time.time()
 
         lr = scheduler.get_last_lr()[0]
-        writer.add_scalar('learning_rate', lr, epoch)
+        wandb.log({"learning_rate": lr})
```
dyt811, line 167: Same as above.

ivadomed/training.py
```diff
@@ -261,20 +272,21 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
...
 
             # METRICS COMPUTATION FOR CURRENT EPOCH
             val_loss_total_avg_old = val_loss_total_avg if epoch > 1 else None
             metrics_dict = metric_mgr.get_results()
             metric_mgr.reset()
-            writer.add_scalars('Validation/Metrics', metrics_dict, epoch)
+            wandb.log({"validation-metrics": metrics_dict})
             val_loss_total_avg = val_loss_total / num_steps
-            writer.add_scalars('losses', {
+            wandb.log({"losses": {
```
dyt811, line 286: Same as above.

ivadomed/visualize.py
```diff
@@ -128,7 +129,7 @@ def convert_labels_to_RGB(grid_img):
     return rgb_img
 
 
-def save_tensorboard_img(writer, epoch, dataset_type, input_samples, gt_samples, preds, is_three_dim=False):
+def save_wandb_img(writer, epoch, dataset_type, input_samples, gt_samples, preds, is_three_dim=False):
```
dyt811, line 132: FYI, we MAY eventually abstract this... because I suspect we MAY hear push backs when tensor board feedback is gutted, so if that happens, we may need to generalize this portion of the code to handle multiple streams (e.g. Wandb+tensorboard)

But for this PR, probably not need to be too concerned as it is.

requirements_common.txt
```diff
@@ -21,3 +21,4 @@ tensorboard>=1.15.0
 tqdm>=4.30
 scipy
 torchio>=0.18.68
+wandb
```
dyt811, line 24: You may wish to fix a specific version on this.

dyt811 (changes requested):
Hey, @naga-karthik. Great stuff. This will make future experiments a lot easier to incorporate. 

I believe there might be a few issues to review/address to fully wrap things up:


1) When an API key is not detected. It is VITAL that none of the WandB related code are executed because without a properly setup API key, most execution on new system (outside  of the already setup wandb.init() system) would fail/stall. This applies to all users once your change is merged to master and released subsequently so it will affect 100% of the user, and forcing them to create WandB account or stuck at an interactive CLI is not a good idea. Addressing this will ALSO fix your CI issue currently experiencing. 

2) You SHOULD include a CI compatible  unit test that actually validates the WandB functions. One possible way is to load WANDB_API_KEY as an environment variable, MVP shown here: 
image based on the official documentation [here](https://docs.wandb.ai/guides/technical-faq#how-do-i-switch-between-accounts-on-the-same-machine).    

   Or you can use the wandb.login function

   image

   The key here is for the unit test to actually have a valid API key such that you can submit the data even on GitHub CI and this ensures all these WandB (soon to be conditional) code will work with AND without wandb setup. 

Aside from these two points, I don't see any other  issues that prevents the merge and look forward to testing it out more.

ivadomed/config/config.json
```diff
@@ -8,8 +8,14 @@
         "object_detection_path": null,
         "safety_factor": [1.0, 1.0, 1.0]
     },
+    "wandb": {
+        "project_name": "ivado-wandb-testing",
+        "group_name": "temp",
+        "run_name": "run-1",
+        "log_grads_every": 100
```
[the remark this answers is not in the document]
naga-karthik: That sounds reasonable! I have added the key name "wandb_api_key" as per your suggestion.

ivadomed/training.py
naga-karthik: This suggestion and the reference links were super useful! Thank you very much for that! I do think having `wandb.login(...)` right here would be useful in terms of code-understandability (as important wandb methods would be at one place), instead of setting the environment variable. lmk if we should proceed with `os.environ` in case it has any other advantages!

requirements_common.txt
naga-karthik: Just out of curiosity, why specify a version? Is it because updates to wandb may deprecate some of the functions and we'd have to keep updating those in ivadomed as well?

naga-karthik:
Thank you very much @dyt811 for taking the time to review! The detailed suggestions were really helpful and useful to learn as well! There is one issue I wanted to clarify a bit more: 

> The key here is for the unit test to actually have a valid API key such that you can submit the data even on GitHub CI and this ensures all these WandB (soon to be conditional) code will work with AND without wandb setup.

how do you suppose we have a (general) valid API here? Because, an API exists as a consequence of having an account, and that key for a particular account should be kept private. Should we create a dummy NeuroPoly account and use its API key for all required testing? This seems like a good option as people who don't want a personal account (for now) can use the dummy key and test the PR as well! What are your thoughts?

kanishk16:
> 1. When an API key is not detected. It is VITAL that none of the WandB related code are executed because without a properly setup API key, most execution on new system (outside  of the already setup wandb.init() system) would fail/stall. This applies to all users once your change is merged to master and released subsequently so it will affect 100% of the user, and forcing them to create WandB account or stuck at an interactive CLI is not a good idea. Addressing this will ALSO fix your CI issue currently experiencing.

After pondering a bit on this nice catch by @dyt811 and the suggestion from @jcohenadad to keep tensorboard (during the initial stages of wandb's integration when suggested deprecation of tensorboard), I was wondering how about having tensorboard as a fallback rather than deprecating it completely.

Perhaps we could add `initialize_wandb` as a boolean key that gives the user a choice to use wandb. Furthermore, `wandb_tracking` as a boolean variable, that would signify the successful authentication of wandb. This would then be used anywhere in the script to handle the cases when the authentication wasn't successful (HOW? refer to the code...) 

```
# signifies the intention of the user to use wandb for experiment tracking
if wandb_params[WandbKW.initialize_wandb]:
  try:
    if wandb_params and wandb_params[WandbKW.API_KEY]:
              
      ...  
      
      # setting variable for checks after successful authentication
      wandb_tracking = True

  except Exception as e:

    # log error mssg for unsuccessful wandb authentication
    # switching to tensorboard
    wandb_tracking = False

else:
  # fallback in case of preference to tensorboard
  wandb_tracking = False

....  

if wandb_tracking:
  # wandb logging
else:
 # tensorboard logging
``` 
There is a trade-off for flexibility over a bit of redundancy and extra dependency of tensorboard but would like to know if this is a reasonable tradeoff.

naga-karthik:
Thank you @kanishk16 for the suggestion and for providing a code example! However, as discussed, there are two things: (i) we would not like to add something extra to something that's already working well (i.e. tensorboard logging), and (ii) while I agree that there is indeed a redundancy double-logging, I would argue that it has a marginal effect on the total training time (unless we log gradients as well, which tensorboard does not do anyway), so it will not be a huge issue. However, as for maintaining tensorboard, we will have to wait and see what happens in the future (how future TF versions go about tensorboard), but until then we should be fine with both wandb and tensorboard.

kanishk16 (commented):
@naga-karthik Thanks for the detailed response... IIUC the following cases are addressed:

- [x] Valid API Key found -> wandb code executed as well as rest of the pipeline
- [x] No API Key found -> wandb code skipped but rest of the pipeline executed 

but the following case still needs to be addressed:

- [ ] Invalid API Key found (random string or expired API key entered by the user) -> wandb code skipped but rest of the pipeline executed 

- How do we handle this? 
Since there is no way to suppose a general valid API Key and it also needs to be kept private. Therefore, we capitalize on the fact that any invalid API key would raise an exception which we could definitely catch. Now, I would recommend to also use a flag that denotes if the authentication was successful or not rather than establishing if "wandb_params" exists and also its has a API key in it (may or may not be valid) which makes us miss the above case. Moreover, the flag could be sent as an argument to `save_img()` to further check and log (only in case of successful authentication) to W&B. 

```
def initialize_wandb(wandb_params):
  try:
    # login would raise an exception for the cases invalid API key and not found  
    wandb.login(key=wandb_params[WandbKW.API_KEY])

    # Setup default names for wandb.init, not assuming they exists by providing default alternatives.
    project_name = wand_params.get("WandbKW.PROJECT_NAME", "Default_Project")
    group_name= wand_params.get("WandbKW.GROUP_NAME", "Default_Group")
    run_name = wand_params.get("WandbKW.RUN_NAME", "Default_Run")
    
    # Init Project
    wandb.init(project=project_name, group=group_name , name=run_name, config=cfg)
    
  except Exception as e:

    # log error mssg for unsuccessful wandb authentication

    # set flag
    wandb_tracking = False

  else:
    # setting flag after successful authentication
    wandb_tracking =True

  return wandb_tracking

....  

# calling this function in train & setting the flag
wandb_tracking = initialize_wandb(wandb_params) 

if wandb_tracking:
  # wandb logging
```

`initialize_wandb(wandb_params)` could be added to the `utils.py` script. I guess this way it would not only be modular but should also ease in testing as we would just need to check whether the function returns `True` or `False`.

ivadomed/training.py
dyt811: 👍 for `wandb.login`. 

Since you asked, I think `os.environ` approach's only other pro I can potentially think of is when you set it, you can potentially not need to change it again (as API keys tend to stick around). So in "theory" if you say update `.bashrc` to set this or something similar semi-permanent, it will always prevent you ever worry about it. Easy from a seasoned ivadomed dev/user/experimenter perspective but not sure how easy that it would be for not experienced Linux users etc. Alternative to that would be load config and set the os environment but it usually doesn't stick around permanently so probably not worth the hassle. 

All in short, I agree `wandb.login` call is probably far more scalable and readable.

ivadomed/visualize.py
dyt811: Strongly agree. Nice one. 👍

dyt811:
> ```
> def initialize_wandb(wandb_params):
>   try:
>     # login would raise an exception for the cases invalid API key and not found  
>     wandb.login(key=wandb_params[WandbKW.API_KEY])
> 
>     # Setup default names for wandb.init, not assuming they exists by providing default alternatives.
>     project_name = wand_params.get("WandbKW.PROJECT_NAME", "Default_Project")
>     group_name= wand_params.get("WandbKW.GROUP_NAME", "Default_Group")
>     run_name = wand_params.get("WandbKW.RUN_NAME", "Default_Run")
>     
>     # Init Project
>     wandb.init(project=project_name, group=group_name , name=run_name, config=cfg)
>     
>   except Exception as e:
> 
>     # log error mssg for unsuccessful wandb authentication
> 
>     # set flag
>     wandb_tracking = False
> 
>   else:
>     # setting flag after successful authentication
>     wandb_tracking =True
> 
>   return wandb_tracking
> 
> ....  
> 
> # calling this function in train & setting the flag
> wandb_tracking = initialize_wandb(wandb_params) 
> 
> if wandb_tracking:
>   # wandb logging
> ```
> 
> `initialize_wandb(wandb_params)` could be added to the `utils.py` script. I guess this way it would not only be modular but should also ease in testing as we would just need to check whether the function returns `True` or `False`.

I agree. Good catch. I didn't really consider the corner case of invalid API key because I assumed whoever is knowledge enough to put/update the API key in the configuration JSON (since we don't include default API keys), can probably read the exception and understand that their key is wrong etc. However, your solution is more explicit and would also solve the problem of what if they forgot default project name etc in JSON. 

@naga-karthik I think the proposed boolean flag COULD work better than my initial thoughts. We might even be able to write that flag back into the `context` WandB dictionary and as it passes around, allowing downstream process to just read that instead and now have to repetitive checks.

ivadomed/training.py
```diff
@@ -54,6 +55,22 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
...
 
+    # Only initialize wandb if required params are found in the config file
+    if wandb_params is not None and wandb_params[WandbKW.WANDB_API_KEY]:
+        # Collect all hyperparameters into a dictionary
+        cfg = { **training_params, **model_params}
+
+        # Get the actual project, group, and run names if they exist, else choose the temporary names as default
+        project_name = wandb_params.get(WandbKW.PROJECT_NAME, "temp_project")
+        group_name = wandb_params.get(WandbKW.GROUP_NAME, "temp_group")
+        run_name = wandb_params.get(WandbKW.RUN_NAME, "temp_run")
```
dyt811, line 66: I would also recommend maybe give an output to user to mention that. Because otherwise, they won't look for `temp_project`/`temp_group`/`temp_run` etc.

ivadomed/training.py
```diff
@@ -54,6 +55,22 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
...
+
+        # Log on to WandB (since the API_KEY is already provided)
+        wandb.login(key=wandb_params[WandbKW.WANDB_API_KEY])
```
dyt811, line 69: I haven't tied it yet, but curious how this would work if the `config.JSON` has this key set to "" like what you have in the PR. Would it default back to system wandb login? or would it throew exception, requiring what @kanishk16 has recommended try/catch loop?

ivadomed/training.py
```diff
@@ -118,6 +135,11 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
                                                     num_epochs)
     logger.info("Scheduler parameters: {}".format(training_params["scheduler"]["lr_scheduler"]))
 
+    # Only call wandb methods if required params are found in the config file
+    if wandb_params is not None and wandb_params[WandbKW.WANDB_API_KEY]:
```
dyt811, line 139: Again, per @kanishk16's good suggestion, if we centralize/modularize the check into its function and encode that into the context dictionary for wandb, it might be easier than invoking this line constantly? Also... counteract the scenario of wrong/empty API key?

ivadomed/training.py
```diff
@@ -155,6 +177,8 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
 
         lr = scheduler.get_last_lr()[0]
         writer.add_scalar('learning_rate', lr, epoch)
+        if wandb_params is not None and wandb_params[WandbKW.WANDB_API_KEY]:
```
dyt811, line 180: Same as above.

dyt811 (approved):
Thanks for the great update. I think this is a great start for incorporating WandB tracking and making all future experiments easier. 

I strongly recommend incorporating @kanishk16 's suggestion as it makes future changes more modular and also reduce the check clutter and fail points as it is hard to fail with booleans. One benefit of this is that it will let you pass CI tests easier. Alternative is to either disable the API key in config or provide a "generic ones" to just get around CI to merge for now. 

We may consider future PRs abstracting out metrics overflow further like what you have done with `save_img` but it is not absolutely required at this point and can certainly be more patched out in the near future. 

Ping me if you have any other questions. :D

ivadomed/config/config.json
```diff
@@ -40,12 +47,12 @@
         "test_fraction": 0.2
     },
     "training_parameters": {
-        "batch_size": 18,
+        "batch_size": 32,
```
jcohenadad, line 50: any reason for changing that? (unrelated to this PR)

ivadomed/config/config.json
```diff
@@ -40,12 +47,12 @@
...
         "loss": {
             "name": "DiceLoss"
         },
         "training_time": {
-            "num_epochs": 100,
+            "num_epochs": 15,
```
jcohenadad, line 55: same comment as https://github.com/ivadomed/ivadomed/pull/1069/files#r826242733

ivadomed/training.py
naga-karthik: Yes, good point! I will put out a print statement here with those names

ivadomed/config/config.json
naga-karthik: Again, no specific reason here as well. Waiting for 15 epochs is a lot lesser time than waiting for 100 epochs to know whether everything is running fine and not crashing anywhere during/after training. Well, 15 and not 5 because, segmentation outputs are not good for the latter (because too less training).

ivadomed/config/config.json
naga-karthik: Again, no specific reason here as well. Waiting for 15 epochs is a lot lesser time than waiting for 100 epochs to know whether everything is running fine and not crashing anywhere during/after training. Well, 15 and not 5 because, segmentation outputs are not good for the latter (because too less training).

ivadomed/training.py
naga-karthik: Okay, something interesting happened here - I set-up the try/catch loop as suggested by @kanishk16 and left the wandb_api_key value in the config file empty. Even then, the code ran without breaking i.e. wandb logging was enabled. This is because WandB does not always need the API Key to work. When the user is running it for the first time ever, only the API Key is required. This is because wandb stores the key in the `~/.netrc` file locally. 

This is something useful to potentially mention in the documentation that the it would suffice if the user only uses the API Key for the first time. Then, it can be left empty (but, the "wandb_api_key" key in the dictionary cannot be deleted)

kanishk16:
> 1. About this comment "wand_params exist + no API Key found -> wandb code skipped but rest of the pipeline executed" - this does not seem to be the case. Refer to my comment [here](https://github.com/ivadomed/ivadomed/pull/1069#discussion_r826412257) for details.

My bad, I was actually thinking about the first time users, but this is the beauty of try/catch. We don't need to worry about each and every case, as it covers all of them. In your case, it essentially found the API key irrespective of its location and handled it gracefully as expected.

> 2. I have more-or-less followed your code snippet, except I define the names and do the init inside `training.py` only. If you look at [this file](https://github.com/ivadomed/ivadomed/pull/1069/commits/8dae0332b01ef74882ccbc2dc679ff63882fcfa2), you'll see that only .login is being done in `utils.py`. For me, it made more sense to include wandb initialization in `training.py` itself.

Absolutely makes sense! I was just focusing on the edge cases at that time.

ivadomed/config/config.json
```diff
@@ -8,6 +8,13 @@
         "object_detection_path": null,
         "safety_factor": [1.0, 1.0, 1.0]
     },
+    "wandb": {
+        "wandb_api_key": "",
+        "project_name": "ivado-wandb-testing",
+        "group_name": "temp",
+        "run_name": "run-1",
+        "log_grads_every": 100
```
naga-karthik, line 16: Here's how the "wandb" dict works based on the latest tests:

1. **When "wandb_api_key" is included in the dict and the API key authentication is successful:** The training proceeds without issues, metrics are logged on **both** WandB and Tensorboard. 
2. **When "wandb_api_key" is included in the dict and the API key authentication is unsuccessful:** WandB tracking is disabled, the training proceeds (does not stop) by just logging metrics on Tensorboard. A print statement notifies that WandB is disabled.
3. **When "wandb_api_key" is NOT included in the dict:** Error thrown and the training stops. This key is necessary.
4. **When "wandb_api_key" is included but is an empty string:** For e.g. as `"wandb_api_key": ""`, the training proceeds with **both** WandB and Tensorboard enabled **only if** there was a previous successful authentication. This is because once wandb has successfully authenticated, it stores the api key locally so that every time the user would not need to authenticate it. So, this also means that ivadomed's users also need to authenticate it only once and leave the field empty thereafter.

I think this covers all the possible cases, so if there's no other issue then this PR can be merged!

dyt811: Cool. I think as long as the default template has been updated, this is not an issue. The `wandb` key probably should be updated in all the config JSONs?

naga-karthik: Yes, I thought of creating a separate PR just for updating the docs and config files!

kanishk16:
> Hey @dyt811 and @kanishk16, I found another approach for dealing with multiple loggers [here](https://github.com/chandar-lab/RLHive/blob/main/hive/utils/loggers.py). Do you think that this approach is more modular and potentially easy to maintain in the future? Let me know your thoughts!

@naga-karthik I reviewed a few more packages including [Avalanche: an End-to-End Library for Continual Learning](https://github.com/ContinualAI/avalanche). It took me a little more than expected. They follow a similar approach as you found above. It would be easier to maintain because it completely separates the loggers and in future any addition or removal of loggers would be more functionally independent. As a software design principle, functional independence is optimized. 

I pondered about it. Since we have a decent approach and introducing the feature is higher on priority, it would be great if we open an issue for refactoring. Perhaps it could be reviewed almost the same time we review tensorboard.

naga-karthik:
Thanks for going through more libraries, @kanishk16! There was an issue regarding this that was recently closed. Could you please re-open it and add your thoughts in [this issue](https://github.com/ivadomed/ivadomed/issues/1090)?

kanishk16:
Makes sense to add this review there rather than opening another issue.

ivadomed/config/config.json
naga-karthik: @uzaymacar I just checked for the situation when the parent "wandb" key in the config file is not included. The training runs as usual. See my edited [comment above](https://github.com/ivadomed/ivadomed/pull/1069#discussion_r846307049)! Thanks again for bringing up this edge-case which I had missed! 🙌

PR #1075, opened 2022-02-09
Refactor test_loader.py to be using pytest cases

konantian:
#### GitHub

#### PR contents

## Description

Refactor test_loader.py to be using pytest cases


testing/unit_tests/cases_loader.py
```diff
@@ -0,0 +1,242 @@
...
+    default_loader_parameters = {
+        LoaderParamsKW.PATH_DATA: [str(Path(__data_testing_dir__, "microscopy_png"))],
+        LoaderParamsKW.BIDS_CONFIG: f"{path_repo_root}/ivadomed/config/config_bids.json",
+        LoaderParamsKW.TARGET_SUFFIX: [["_seg-myelin-manual", "_seg-axon-manual"]],
+        LoaderParamsKW.EXTENSIONS: [".png"],
+        LoaderParamsKW.ROI_PARAMS: {"suffix": None, "slice_filter_roi": None},
+        LoaderParamsKW.CONTRAST_PARAMS: {"contrast_lst": []}
+    }
+
+    return default_loader_parameters
```
dyt811, line 19: I would recommend to also get the default JSON from the `config.json` root file within the current repo., and OVERWRITE its `loader_params` key with it. 

You will see below comments where when this was not done, you kind of have to redefine the model and transformation keys multiple times.

testing/unit_tests/cases_loader.py
```diff
@@ -0,0 +1,242 @@
...
+
+def case_bids_df_ctscan():
+    loader_parameters = {
+        LoaderParamsKW.PATH_DATA: [str(Path(__data_testing_dir__, "ct_scan"))],
+        LoaderParamsKW.BIDS_CONFIG: f"{path_repo_root}/ivadomed/config/config_bids.json",
+        LoaderParamsKW.TARGET_SUFFIX: ["_seg-manual"],
+        LoaderParamsKW.EXTENSIONS: [".nii.gz"],
+        LoaderParamsKW.ROI_PARAMS: {"suffix": None, "slice_filter_roi": None},
+        LoaderParamsKW.CONTRAST_PARAMS: {"contrast_lst": ["ct"]}
+    }
```
dyt811, line 60: I would still recommend to load from the `get_microscopy_loader_parameters()` for future compatibility reason. 

In the event some new future keys are missed here from being included during this loader_parameter's dictionary definition. Better safe than sorry.

testing/unit_tests/cases_loader.py
```diff
@@ -0,0 +1,242 @@
...
+        ModelParamsKW.NAME: "Unet",
+        ModelParamsKW.DROPOUT_RATE: 0.3,
+        ModelParamsKW.BN_MOMENTUM: 0.1,
+        ModelParamsKW.FINAL_ACTIVATION: "sigmoid",
+        ModelParamsKW.DEPTH: 3
+    }
+
+    transform_parameters = {
+        TransformationKW.NUMPY_TO_TENSOR: {},
+    }
```
dyt811, line 111: Instead of redefining models here, we should leverage the existing `config.json` which has both the `default_model` key which we can overwrite (then pass) to the test function as well as the transformation function. 

Note that you MAY have to empty out existing default configurations from config.json in order for some tests to pass as well. However, this is a good chance to pick up some learning as well.

testing/unit_tests/cases_loader.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def case_2d_patches_and_resampling():
+    loader_parameters = get_microscopy_loader_parameters()
+    loader_parameters.update({
+        LoaderParamsKW.TARGET_SUFFIX: ["_seg-myelin-manual"],
+        LoaderParamsKW.CONTRAST_PARAMS: {"contrast_lst": [], "balance": {}},
+        LoaderParamsKW.SLICE_AXIS: "axial",
+        LoaderParamsKW.SLICE_FILTER_PARAMS: {"filter_empty_mask": False, "filter_empty_input": True},
+        LoaderParamsKW.PATCH_FILTER_PARAMS: {"filter_empty_mask": False, "filter_empty_input": False},
```
dyt811, line 123: Just beware that some of these sub contrast/params dicts have further keywords that can be redefined using `keywords.py` values.

testing/unit_tests/cases_loader.py
```diff
@@ -0,0 +1,242 @@
...
+        ModelParamsKW.STRIDE_2D: [244, 116]
+    }
+
+    transform_parameters = {
+        TransformationKW.RESAMPLE: {
+            TransformationKW.W_SPACE: 0.0002,
+            TransformationKW.H_SPACE: 0.0001
+        },
+        TransformationKW.NUMPY_TO_TENSOR: {},
+    }
```
dyt811, line 143: Notice this pattern where you have to redefine these two dicts over and over again? It is probably safer to overwrite a default one solely on the modifications. 

Again, best to test it out as the very default value can introduce more aberrations. 

Set/Empty accordingly. 

Another benefit of a single default and overwrite/empty in unit is the advantage that such alterations are IMMEDIATELY visible and apparent for new coders joining, otherwise, a DIFF is required to really tell what changed. All part of ability to improve readability/ease of interpretation.

dyt811 (commented):
Thanks for the good work! I think with a few minor adjustments we should be able to improve the readability of this class substantially. 

We may also add some comments to the case files to better explain it in the context of the cases_loader.py file for people. Mostly copying the explanation is probably sufficient.

PR #1076, opened 2022-02-13
Prepatory PR for PR1044 about Multi-Contrast/Sessions Support

dyt811:
#### GitHub

#### PR contents

## Description

During the making of 1044, a lot of anchilary changes were made as issues were discovered along the way that were not core to the goal of 1044 and complicate the 1044 review process. This PR aims to target those same changes such that rebased 1044 will be a lot easier to review.

### DIFF
--- a/ivadomed/loader/film.py
+++ b/ivadomed/loader/film.py
@@ -163,12 +163,12 @@ def clustering_fit(dataset, key_lst):
     return model_dct
 
 
-def check_isMRIparam(mri_param_type, mri_param, subject, metadata):
+def check_isMRIparam(mri_param_type: str, mri_param: dict, subject: str, metadata: dict) -> bool:
     """Check if a given metadata belongs to the MRI parameters.
 
     Args:
         mri_param_type (str): Metadata type name.
-        mri_param (list): List of MRI params names.
+        mri_param (dict): List of MRI params names.
         subject (str): Current subject name.
         metadata (dict): Metadata.
 
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -8,6 +8,7 @@ from loguru import logger
 from pathlib import Path
 from ivadomed.keywords import ConfigKW, LoaderParamsKW
 from typing import List
+from difflib import SequenceMatcher
 
 AXIS_DCT = {'sagittal': 0, 'coronal': 1, 'axial': 2}
 
@@ -401,6 +402,17 @@ def format_path_data(path_data):
     if isinstance(path_data, str):
         path_data = [path_data]
     return path_data
+
+
+def similarity_score(a: str, b: str) -> float:
+    """
+    use DiffLIb SequenceMatcher to resolve the similarity between text. Help make better choice in terms of derivatives.
+    Args:
+        a: a string
+        b: another string
+    Returns: a score indicative of the similarity between the sequence.
+    """
+    return SequenceMatcher(None, a, b).ratio()
 
 
 def init_ivadomed():
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -6,6 +6,7 @@
 pytest~=6.2
 pytest-cov
 pytest-ordering~=0.6
+pytest-cases~=3.6.8
 sphinx
 flake8
 coverage
--- a/testing/common_testing_util.py
+++ b/testing/common_testing_util.py
@@ -1,7 +1,16 @@
+import os
 from pathlib import Path
-from loguru import logger
+
+import json
+import csv_diff
+import pandas as pd
 import pytest
+from loguru import logger
+
+from ivadomed.keywords import BidsDataFrameKW, LoaderParamsKW, ContrastParamsKW, ConfigKW
+from ivadomed.loader.bids_dataframe import BidsDataframe
 from ivadomed.scripts import download_data as ivadomed_download_data
+from ivadomed.main import set_loader_params
 import shutil
 import sys
 
@@ -22,6 +31,10 @@ path_data_testing_tmp: str = str(path_repo_root / "tmp" / Path(path_data_testing
 path_functional_tests: str = str(path_repo_root / "testing" / "functional_tests")
 path_data_functional_source: str = str(path_repo_root / "data_functional_testing")
 path_data_functional_tmp: str = str(path_repo_root / "tmp" / Path(path_data_functional_source).name)
+
+path_data_multi_sessions_contrasts_source: Path = path_repo_root / "data_multi_testing"
+path_data_multi_sessions_contrasts_tmp: Path = path_repo_root / "tmp" / Path(
+    path_data_multi_sessions_contrasts_source).name
 
 
 def download_dataset(dataset: str = 'data_testing'):
@@ -64,4 +77,3 @@ def remove_dataset(dataset: str = 'data_testing'):
 def remove_tmp_dir():
     """Recursively remove the ``tmp`` directory if it exists."""
     shutil.rmtree(path_temp, ignore_errors=True)
-
--- a/testing/functional_tests/test_automate_training.py
+++ b/testing/functional_tests/test_automate_training.py
@@ -1,5 +1,6 @@
 import logging
 import pytest
+import os
 from pytest_console_scripts import script_runner
 from pathlib import Path
 from testing.functional_tests.t_utils import __tmp_dir__, create_tmp_dir, __data_testing_dir__, \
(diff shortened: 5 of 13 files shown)


testing/unit_tests/t_utils.py
```diff
@@ -20,6 +22,15 @@ def download_data_testing_test_files():
     download_dataset("data_testing")
 
 
+@pytest.fixture(scope='session')
+def download_multi_data():
+    """
+    This Pytest fixture DOWNLOAD all the test data set REQUIRED for the multi-session, multi-contrast related unite
```
mariehbourget, line 28:
```suggestion
    This Pytest fixture DOWNLOAD all the test data set REQUIRED for the multi-session, multi-contrast related unit
```

testing/unit_tests/t_utils.py
```diff
@@ -20,6 +22,15 @@ def download_data_testing_test_files():
...
+    testing.
+    """
+    download_dataset("data_multi_testing")
```
mariehbourget, line 31: I think I'm missing something, I don't understand where is this new multi testing data?

testing/unit_tests/t_utils.py
```diff
@@ -38,3 +49,19 @@ def create_tmp_dir(copy_data_testing_dir=True):
...
+    """Create a temporary directory for data related to multi-session unit tests and copy test data files.
+    1. Remove the ``tmp`` directory if it exists.
+    2. Copy the ``data_testing_multi`` directory to the ``tmp`` directory. Ignoring the `.git` folder within
+    Any data files created during testing will go into ``tmp`` directory.
+    This is created/removed for each test.
+    """
+    ignore_git_pattern = shutil.ignore_patterns(str(path_data_multi_sessions_contrasts_source / '.git'))
+    remove_tmp_dir()
+    os.mkdir(path_temp)
+    if os.path.exists(path_data_multi_sessions_contrasts_source):
```
mariehbourget, line 64: Here we use "os" whereas in `create_tmp_dir` above, we use "Path". Is there a preferred method? I would suggest using the same if possible.

testing/unit_tests/t_utils.py
```diff
@@ -1,9 +1,11 @@
+import os
```
mariehbourget, line 1: Maybe use "Path" instead, see other comment at lines 63-64.

testing/common_testing_util.py
```diff
@@ -1,7 +1,16 @@
+import os
```
mariehbourget, line 1: I don't think "os" is used in the changes in this file.

testing/unit_tests/t_utils.py
```diff
@@ -38,3 +49,19 @@ def create_tmp_dir(copy_data_testing_dir=True):
     if Path(path_data_testing_source).exists() and copy_data_testing_dir:
         shutil.copytree(path_data_testing_source,
                         path_data_testing_tmp)
+
+
+def create_tmp_dir_multi_session():
+    """Create a temporary directory for data related to multi-session unit tests and copy test data files.
+    1. Remove the ``tmp`` directory if it exists.
```
mariehbourget, line 56: Just to be sure, this "tmp" folder is the same as the one for `create_tmp_dir`?
If that is the case, I assume the tests are run sequentially and the 2 "tmp" folders are not in competition.

testing/functional_tests/test_automate_training.py
```diff
@@ -34,6 +35,38 @@ def test_automate_training(download_functional_test_files, script_runner):
...
+    subprocess mode which cannot be stepped. Other than that, nothing else really changed and is exactly the same.
+    Very useful for debugging this high level function to spot problems
+
+    Fixture Required:
+        download_functional_test_files:
+    """
+    file_config = os.path.join(__data_testing_dir__, 'automate_training_config.json')
+    file_config_hyper = os.path.join(__data_testing_dir__,
+                                     'automate_training_hyperparameter_opt.json')
+    __output_dir__ = os.path.join(__tmp_dir__, 'results')
```
mariehbourget, line 49: Here we use "os" whereas in `test_automate_training_run_test` below, we use "Path". Is there a preferred method? I would suggest using the same if possible.

testing/functional_tests/test_automate_training.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import pytest
+import os
```
mariehbourget, line 3: Maybe use "Path" instead, see other comment at lines 46-49.

mariehbourget (commented):
The PR looks good, I left a few questions here and there, nothing major.
I'll let Naga review and approve as well.

ivadomed/loader/film.py
```diff
@@ -163,12 +163,12 @@ def clustering_fit(dataset, key_lst):
...
 
 
-def check_isMRIparam(mri_param_type, mri_param, subject, metadata):
+def check_isMRIparam(mri_param_type: str, mri_param: dict, subject: str, metadata: dict) -> bool:
     """Check if a given metadata belongs to the MRI parameters.
 
     Args:
         mri_param_type (str): Metadata type name.
-        mri_param (list): List of MRI params names.
+        mri_param (dict): List of MRI params names.
```
naga-karthik, line 171: Change to "Dictionary of MRI params names" to be precise?

ivadomed/utils.py
```diff
@@ -403,6 +404,17 @@ def format_path_data(path_data):
     return path_data
 
 
+def similarity_score(a: str, b: str) -> float:
```
naga-karthik, line 407: I would recommend we be explicit in naming this function as `string_similarity_score` instead. This is because "similarity score" has different meanings in ML (eg cosine similarity, distance-based similarity like Euclidean, L1, etc.), just to avoid unnecessary confusion.

testing/common_testing_util.py
```diff
@@ -1,7 +1,16 @@
...
+import pandas as pd
 import pytest
+from loguru import logger
+
+from ivadomed.keywords import BidsDataFrameKW, LoaderParamsKW, ContrastParamsKW, ConfigKW
+from ivadomed.loader.bids_dataframe import BidsDataframe
 from ivadomed.scripts import download_data as ivadomed_download_data
+from ivadomed.main import set_loader_params
 import shutil
 import sys
```
naga-karthik, line 15: Unnecessary import - even `sys` is unused in this file.

testing/functional_tests/test_automate_training.py
naga-karthik: Agreed, I suggest we go with `Path` for maintaining consistency. See also [this](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/) link in favor of using `pathlib`, in general.

testing/unit_tests/t_utils.py
naga-karthik, line 64:
```suggestion
    if Path(path_data_multi_sessions_contrasts_source).exists():
```

naga-karthik (approved):
Everything lgtm! I mostly agree with @mariehbourget's comments about the discrepancies in the usage of `os` and `Path`. Additionally, I have two suggestions regarding a docstring and a function name. Thanks for looking into this!

testing/functional_tests/test_automate_training.py
```diff
@@ -34,6 +35,38 @@ def test_automate_training(download_functional_test_files, script_runner):
     check_sha256(str(file_config))
 
 
+def test_automate_training_run_test_debug(download_functional_test_files):
+    """A unit test similar to test_automate_training_run_test but allow step through (instead of using script caller/
```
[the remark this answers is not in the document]
dyt811: Happy to elaborate on this. So what happened is for "subprocess" version of the unit test above this current one, it spawn a system subprocess to carry out the CLI based action (invoke the automated training in this case, the test was designed to test a CLI command after all, so make sense to do that). The problem with that is, while it can conduct [black box testing](https://en.wikipedia.org/wiki/Black-box_testing), it does NOT allow debug/stepping through the code (also known as [white-box testing](https://en.wikipedia.org/wiki/White-box_testing)). `Automated Training` test being a very high level test, it frequently fails during development just because it covers such a huge area of the code. If it fails, it is actually very hard to figure out what goes wrong when you cannot step through the code. 

Hence, why adding this test to really help development process actually.

requirements_dev.txt
```diff
@@ -6,6 +6,7 @@
 pytest~=6.2
 pytest-cov
 pytest-ordering~=0.6
+pytest-cases~=3.6.8
```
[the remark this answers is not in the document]
dyt811: VERY correct. In fact, only "new" tests pertaining to multi-session, multi-contrasts are using the pytest-cases. There are PLANS to convert everything to pytest eventually (to clean up and avoid those massive pages of `@pytest.parametrize` calls and make it way more human readable. BUT that is coming down the road in a few months likely and will be gradually phased in with minimal intrusion via much specific PRs on those.

testing/unit_tests/t_utils.py
dyt811: Good point. I didn't update the DITC_URL for the data. Good catch!

Should be something similar to this in `download_data.py`'s DICT_URL'

image

testing/unit_tests/t_utils.py
dyt811: Great point. I must woke on the wrong  side of the bed for this. Fixed to be consistent with Pathlib.

testing/unit_tests/t_utils.py
dyt811: Yup. Fixed together with the other.

testing/common_testing_util.py
dyt811: Removed. 👍

testing/unit_tests/t_utils.py
dyt811: That is actually a very good point. 

So the current way is like how we are handling functions data testing vs unit data testing: 
image

A new sub folder is created under `tmp` (for any of the functional or unit) tests. They are all under the same "tmp" folder but in different sub folders. However, if we start parallelerize the tests, they WILL fail because one test finish, nuke the tmp folder, will cause the others to fail. 

This may not be a desired behaviour and we may need to address it in the future if we plan to parallelize the tests

testing/functional_tests/test_automate_training.py
dyt811: Good point! Updated.

testing/functional_tests/test_automate_training.py
dyt811: Yup. Corrected :)

testing/unit_tests/t_utils.py
mariehbourget: Hey @dyt811, did you push your changes? I see you wrote "fixed" for most of my comments but still see the code version before the fix, am I missing something?

PR #1086, opened 2022-02-22
Segment with ONNX or PT model based on CPU/GPU availability

### LINKED ISSUE #961: Option to choose model (`pt` or `onnx`) based on device (GPU or CPU) for segmentation
In the `path_output/model_name` folder, `ivadomed` saves by default the `onnx` model after training (unless conversion to `onnx` is not possible) here:
https://github.com/ivadomed/ivadomed/blob/ac29c7cbe299e6478854823ed6bd818c0e5f9642/ivadomed/training.py#L323-L334
The pytorch model (`best_model.pt`) is also saved to `path_output` only (not in `model_name` folder).

Then, when running the `--segment` command, the segmentation is done with the `pt` model if available `path_output/model_name`, otherwise with the `onnx` model here:
https://github.com/ivadomed/ivadomed/blob/ac29c7cbe299e6478854823ed6bd818c0e5f9642/ivadomed/inference.py#L68-L89

More often that not, the model saved is in `onnx` format and the segmentation is performed on CPU.
It seems that is was implemented to optimized segmentation on CPU for SCT in #234.

For microscopy, we often have very large images and running the segmentation on CPU when GPU is available is not optimal.
Examples on an image of 8000x6000 pixels:
- CPU inference with `onnx` model: 6 minutes
- GPU inference with `pt` model: 2.5 minutes

It would be great to save both models in the `path_output/model_name` folder after training and choose the model to use based on the device available at inference time (CPU or GPU).

Let me know what you think!

Note: This is not a bug at the moment since all of this can be achieved manually by moving and renaming `best_model.pt` in the `path_output/model_name` folder.

mariehbourget:
#### GitHub

#### PR contents

## Description

This PR fixes #961 and adds 2 features:
1) After training, both the ONNX and PT versions of the model are saved under the model folder. Up until now, when conversion to ONNX was possible, only the ONNX version was saved, otherwise, only the PT version.
2) When running `segment_volume` (through the `--segment` command or API), the version of the model used (ONNX or PT) is based on its availability and on GPU availability. Hence, when GPU is available, PT is used (if available) otherwise ONNX is used. And vice-versa for CPU using ONNX in priority.

Details of the motivation are explained in #961.
In addition for ADS, we chose to package our model repo with both the ONNX and PT version, ex: [SEM model](https://github.com/axondeepseg/default-SEM-model/tree/main/model_seg_rat_axon-myelin_sem).
This PR will allow the segmentation to be done more efficiently on large microscopy images by choosing the ONNX version (optimize for CPU) or the PT version (optimized for GPU) depending on the device.

Tagging @joshuacwnewton because the `segment_volume` API is also used by SCT.

## Linked issues
Fixes #961

### DIFF
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -50,7 +50,7 @@ def get_preds(context: dict, fname_model: str, model_params: dict, gpu_id: int, 
 
     Args:
         context (dict): configuration dict.
-        fname_model (str): name of file containing model (without .pt or .onnx extension).
+        fname_model (str): name of file containing model.
         model_params (dict): dictionary containing model parameters.
         gpu_id (int): Number representing gpu number if available. Currently does NOT support multiple GPU segmentation.
         batch (dict): dictionary containing input, gt and metadata
@@ -60,16 +60,6 @@ def get_preds(context: dict, fname_model: str, model_params: dict, gpu_id: int, 
     """
     # Define device
     cuda_available, device = imed_utils.define_device(gpu_id)
-
-    # Assign '.onnx' or '.pt' model based on availability and device
-    # On GPU, '.pt' model is prioritized whereas '.onnx' model is prioritized on CPU.
-    if Path(fname_model + '.onnx').is_file():
-        if cuda_available and Path(fname_model + '.pt').is_file():
-            fname_model = fname_model + '.pt'
-        else:
-            fname_model = fname_model + '.onnx'
-    else:
-        fname_model = fname_model + '.pt'
 
     with torch.no_grad():
 
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -50,7 +50,7 @@ def get_preds(context: dict, fname_model: str, model_params: dict, gpu_id: int, 
 
     Args:
         context (dict): configuration dict.
-        fname_model (str): name of file containing model.
+        fname_model (str): name of file containing model (without .pt or .onnx extension).
         model_params (dict): dictionary containing model parameters.
         gpu_id (int): Number representing gpu number if available. Currently does NOT support multiple GPU segmentation.
         batch (dict): dictionary containing input, gt and metadata
@@ -60,6 +60,16 @@ def get_preds(context: dict, fname_model: str, model_params: dict, gpu_id: int, 
     """
     # Define device
     cuda_available, device = imed_utils.define_device(gpu_id)
+
+    # Assign '.onnx' or '.pt' model based on availability and device
+    # On GPU, '.pt' model is prioritized whereas '.onnx' model is prioritized on CPU.
+    if Path(fname_model + '.onnx').is_file():
+        if cuda_available and Path(fname_model + '.pt').is_file():
+            fname_model = fname_model + '.pt'
+        else:
+            fname_model = fname_model + '.onnx'
+    else:
+        fname_model = fname_model + '.pt'
 
     with torch.no_grad():
 
--- a/ivadomed/models.py
+++ b/ivadomed/models.py
@@ -1499,19 +1499,20 @@ def get_model_filenames(folder_model):
         prefix_model = Path(folder_model).name
         fname_model_onnx = Path(folder_model, prefix_model + '.onnx')
         fname_model_pt = Path(folder_model, prefix_model + '.pt')
+        cuda_available = torch.cuda.is_available()
 
-        # Assign '.pt' or '.onnx' model based on availability and device.
-        # On GPU, '.pt' model is prioritized whereas the '.onnx' model is prioritized on CPU.
-        if fname_model_onnx.is_file():
-            if fname_model_pt.is_file() and torch.cuda.is_available():
-                fname_model = fname_model_pt
-            else:
-                fname_model = fname_model_onnx
-        elif fname_model_pt.is_file():
-            fname_model = fname_model_pt
-        else:
+        # Assign '.pt' or '.onnx' model based on file existence and GPU/CPU device availability
+        if not fname_model_pt.is_file() and not fname_model_onnx.is_file():
             raise FileNotFoundError(f"Model files not found in model folder: "
                                     f"'{str(fname_model_onnx)}' or '{str(fname_model_pt)}'")
+        # '.pt' is preferred on GPU, or on CPU if '.onnx' doesn't exist
+        elif ((    cuda_available and     fname_model_pt.is_file()) or
+              (not cuda_available and not fname_model_onnx.is_file())):
+            fname_model = fname_model_pt
+        # '.onnx' is preferred on CPU, or on GPU if '.pt' doesn't exist
+        elif ((not cuda_available and     fname_model_onnx.is_file()) or
+              (    cuda_available and not fname_model_pt.is_file())):
+            fname_model = fname_model_onnx
 
         fname_model_metadata = Path(folder_model, prefix_model + '.json')
         if not fname_model_metadata.is_file():
--- a/ivadomed/models.py
+++ b/ivadomed/models.py
@@ -1486,27 +1486,29 @@ def set_model_for_retrain(model_path, retrain_fraction, map_location, reset=True
 def get_model_filenames(folder_model):
     """Get trained model filenames from its folder path.
 
-    This function checks if the folder_model exists and get trained model (.pt or .onnx) and its configuration file
-    (.json) from it.
-    Note: if the model exists as .onnx, then this function returns its onnx path instead of the .pt version.
+    This function checks if the folder_model exists and get trained model path (without .pt or .onnx extension)
+    and its configuration file (.json) from it.
 
     Args:
         folder_name (str): Path of the model folder.
 
     Returns:
-        str, str: Paths of the model (.onnx) and its configuration file (.json).
+        str, str: Paths of the model (without .pt or .onnx extension) and its configuration file (.json).
     """
     if Path(folder_model).is_dir():
         prefix_model = Path(folder_model).name
-        # Check if model and model metadata exist. Verify if ONNX model exists, if not try to find .pt model
-        fname_model = Path(folder_model, prefix_model + '.onnx')
-        if not fname_model.is_file():
-            fname_model = Path(folder_model, prefix_model + '.pt')
-            if not fname_model.exists():
-                raise FileNotFoundError(fname_model)
+        fname_model = Path(folder_model, prefix_model)
+
+        # Check if at least one of '.pt' or '.onnx' model exists
+        if not Path(str(fname_model) + '.onnx').is_file() and not Path(str(fname_model) + '.pt').is_file():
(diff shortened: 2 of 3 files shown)


ivadomed/models.py
```diff
@@ -1486,27 +1486,29 @@ def set_model_for_retrain(model_path, retrain_fraction, map_location, reset=True
 def get_model_filenames(folder_model):
     """Get trained model filenames from its folder path.
 
-    This function checks if the folder_model exists and get trained model (.pt or .onnx) and its configuration file
-    (.json) from it.
-    Note: if the model exists as .onnx, then this function returns its onnx path instead of the .pt version.
+    This function checks if the folder_model exists and get trained model path (without .pt or .onnx extension)
+    and its configuration file (.json) from it.
```
joshuacwnewton, line 1490: > Tagging @joshuacwnewton because the segment_volume API is also used by SCT.

Thank you for the heads-up! 

As far as I understand, these changes will pretty much be invisible to SCT? Just because all of the models we currently list as part of `sct_deepseg` contain either an `.onnx` file or a `.pt` model but not both:

image

But, I appreciate being made aware of these changes nonetheless.

ivadomed/inference.py
```diff
@@ -61,6 +61,16 @@ def get_preds(context: dict, fname_model: str, model_params: dict, gpu_id: int,
...
 
+    # Assign '.onnx' or '.pt' model based on availability and device
+    # On GPU, '.pt' model is prioritized whereas '.onnx' model is prioritized on CPU.
+    if Path(fname_model + '.onnx').is_file():
+        if cuda_available and Path(fname_model + '.pt').is_file():
+            fname_model = fname_model + '.pt'
+        else:
+            fname_model = fname_model + '.onnx'
+    else:
+        fname_model = fname_model + '.pt'
```
joshuacwnewton, line 72: When I read this snippet of code, the first observation that came to mind was: 

> "If neither the `.pt` nor the `.onnx` model exists, then `fname_model` will still be assigned `{fname_model}.pt` because of the last `else:` block. That doesn't seem right!"

Then I kept reading the PR, and realized this couldn't happen because of:

https://github.com/ivadomed/ivadomed/blob/1a913e079e3472b5784a2140804a51233fa8e0a0/ivadomed/models.py#L1503-L1505

My takeaway here is that the logic is a little difficult to follow right now, because the file-parsing code is split into 2 separate places (`get_preds` and `get_model_filenames`). So, what if we combined the `.pt`/`.onnx`-selection logic with the existing logic inside the `get_model_filenames` function? 

(This also feels a little more intuitive to me -- my expectation is that `get_model_filenames` would return the model filename, as opposed to `get_preds` taking care of that logic.)

ivadomed/models.py
mariehbourget: Indeed, we the current model list, those changes won't affect SCT but could we future models, depending on how they are packaged.

ivadomed/inference.py
mariehbourget: That is a great point. I agree with you, it would be a lot clearer to choose onnx or pt directly in `get_model_filenames`, thanks!
The only downside I see is that we will need to import `imed_utils` in `models.py` to call the `define_device` function, but I think it's worth it.

joshuacwnewton: > The only downside I see is that we will need to import imed_utils in models.py to call the define_device function, but I think it's worth it.

Hmm... is that necessary, actually? 🤔 

It seems like it we only require the `cuda_available` property for the the `if`/`else` block. And, looking at [`define_device`](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/utils.py#L117-L137), we this get directly from `torch`:

https://github.com/ivadomed/ivadomed/blob/860f44ff98eb58a8cce64a8bcc69b73df5a57b7f/ivadomed/utils.py#L128

mariehbourget: You are right, thank you! And torch is already imported in models. Perfect, I'm doing the changes now.

ivadomed/models.py
```diff
@@ -1486,27 +1485,39 @@ def set_model_for_retrain(model_path, retrain_fraction, map_location, reset=True
...
+        if fname_model_onnx.is_file():
+            if fname_model_pt.is_file() and torch.cuda.is_available():
+                fname_model = fname_model_pt
+            else:
+                fname_model = fname_model_onnx
+        elif fname_model_pt.is_file():
+            fname_model = fname_model_pt
+        else:
+            raise FileNotFoundError(f"Model files not found in model folder: "
+                                    f"'{str(fname_model_onnx)}' or '{str(fname_model_pt)}'")
```
joshuacwnewton, line 1514: Hmm... when reading this code, I found that it required a bit of thinking to figure out the different cases. I found myself having to manually trace each section to work out all the possibilities:

1. Neither file exists (error is thrown)
* At least one file exists:
  2. GPU is available, `.pt` exists (_`.pt` used_)
  3. GPU is available, `.pt` does not exist (_`.onnx` used_)
  4. Only CPU is available, `.onnx` exists (_`.onnx` used_)
  5. Only CPU is available, `.onnx` does not exist (_`.pt` used_)

So, one other suggestion I have is to try making these 5 different cases explicit in the code itself:

```python
        # Assign '.pt' or '.onnx' model based on file existence and GPU/CPU device availability
        if not fname_model_pt.is_file() and not fname_model_onnx.is_file():
            raise FileNotFoundError(f"Model files not found in model folder: "
                                    f"'{str(fname_model_onnx)}' or '{str(fname_model_pt)}'")
        # .pt is preferred on GPU, or on CPU if .onnx doesn't exist  
        elif ((    torch.cuda.is_available() and     fname_model_pt.is_file()) or 
              (not torch.cuda.is_available() and not fname_model_onnx.is_file())):
            fname_model = fname_model_pt
        # .onnx is preferred on CPU, or on GPU if .pt doesn't exist 
        elif ((not torch.cuda.is_available() and     fname_model_onnx.is_file()) or 
              (    torch.cuda.is_available() and not fname_model_pt.is_file())):
            fname_model = fname_model_onnx
```

Even though this way contains a lot of redundant checks, my thinking here is that... if the code is self-explanatory, then that will allow a dev to just read the code (without having to spend time inferring the different cases from the code).

ivadomed/training.py
```diff
@@ -330,12 +330,15 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
...
                                            model_params[ModelParamsKW.FOLDER_NAME] + ".onnx")
             imed_utils.save_onnx_model(model, input_samples, str(best_model_path))
+            logger.info(f"Model saved as '.onnx': {best_model_path}")
         except:
-            # Save best model in model directory
-            best_model_path = Path(path_output, model_params[ModelParamsKW.FOLDER_NAME],
-                                           model_params[ModelParamsKW.FOLDER_NAME] + ".pt")
-            torch.save(model, best_model_path)
-            logger.warning(f"Failed to save the model as '.onnx', saved it as '.pt': {best_model_path}")
+            logger.warning(f"Failed to save the model as '.onnx'")
```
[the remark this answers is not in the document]
mariehbourget: Of course, than you for the info, fixed in [a77bdfb](https://github.com/ivadomed/ivadomed/pull/1086/commits/a77bdfbdb7fdbd4c6296a4d26bec4c8ed73aa71d)

ivadomed/models.py
mariehbourget: Thank you again for your excellent suggestion, I agree that it's better to have it self-explanatory. I integrated your suggestion with a small change to call `torch.cuda.is_available()` only once.

ivadomed/training.py
```diff
@@ -330,12 +330,15 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
...
-                                           model_params[ModelParamsKW.FOLDER_NAME] + ".pt")
-            torch.save(model, best_model_path)
-            logger.warning(f"Failed to save the model as '.onnx', saved it as '.pt': {best_model_path}")
+            logger.info(f"Model saved as '.onnx': {best_model_path}")
+        except Exception as e:
+            logger.warning(f"Failed to save the model as '.onnx': {e}")
+
+        # Save best model as PT in the model directory
+        best_model_path = Path(path_output, model_params[ModelParamsKW.FOLDER_NAME],
+                                       model_params[ModelParamsKW.FOLDER_NAME] + ".pt")
```
joshuacwnewton, line 339: Very very minor nitpick, but the spacing it a little off here, hehe.

```suggestion
        best_model_path = Path(path_output, model_params[ModelParamsKW.FOLDER_NAME],
                               model_params[ModelParamsKW.FOLDER_NAME] + ".pt")
```

joshuacwnewton (approved):
Personally, this looks good to me! I'm not sure if someone else from the imed team should also review, but I'm happy to give my approval for now. :)

ivadomed/models.py
```diff
@@ -1486,27 +1485,40 @@ def set_model_for_retrain(model_path, retrain_fraction, map_location, reset=True
...
 
     Args:
         folder_name (str): Path of the model folder.
 
     Returns:
-        str, str: Paths of the model (.onnx) and its configuration file (.json).
+        str, str: Paths of the model (.pt or .onnx) and its configuration file (.json).
     """
     if Path(folder_model).is_dir():
+
```
uzaymacar, line 1498: We can remove the extra space here.

ivadomed/models.py
```diff
@@ -1486,27 +1485,40 @@ def set_model_for_retrain(model_path, retrain_fraction, map_location, reset=True
...
+        fname_model_onnx = Path(folder_model, prefix_model + '.onnx')
+        fname_model_pt = Path(folder_model, prefix_model + '.pt')
+        cuda_available = torch.cuda.is_available()
+
+        # Assign '.pt' or '.onnx' model based on file existence and GPU/CPU device availability
+        if not fname_model_pt.is_file() and not fname_model_onnx.is_file():
+            raise FileNotFoundError(f"Model files not found in model folder: "
+                                    f"'{str(fname_model_onnx)}' or '{str(fname_model_pt)}'")
+        # '.pt' is preferred on GPU, or on CPU if '.onnx' doesn't exist
+        elif ((    cuda_available and     fname_model_pt.is_file()) or
```
uzaymacar, line 1509: Not sure if the spacing after `((` and `and` is intended but if not we can remove them. Same thing for lines 1513 and 1514 as well.

uzaymacar (approved):
Added two minor styling comments but otherwise LGTM!

ivadomed/models.py
joshuacwnewton: It's definitely nonstandard formatting, aha, so if you feel like removing it, go for it.

The reason I added it in the first place was because I thought it made it a little easier to parse (since everything lined up nicely, and it was clear whether or not `not` was present.)

uzaymacar: Ah I see, no worries if it has a purpose :)

PR #1093, opened 2022-03-07
Seed ivadomed training for reproducibility

etetteh:
The current `ivadomed` implementation only uses a `random seed` during a dataset split into train, validationa, and test sets. 
This ignores other sources of randomness from libraries such as NumPy, and Torch. Also, when a user decides to manually split datasets using a [`.joblib` file](https://ivadomed.org/configuration_file.html#fname-split) , the entire training is not seeded, making it difficult to reproduce results.

This PR is to seed, not only the dataset `splitting phase`, but the `training phase` as well as other phases.


ivadomed/keywords.py
```diff
@@ -50,7 +51,7 @@ class SplitDatasetKW:
     SPLIT_METHOD = "split_method"
     FNAME_SPLIT = "fname_split"
     DATA_TESTING = "data_testing"
-    RANDOM_SEED = "random_seed"
+    #RANDOM_SEED = "random_seed"
```
jcohenadad, line 54: avoid commenting code-- git is here to recover legacy code if necessary

ivadomed/loader/utils.py
```diff
@@ -8,9 +8,10 @@
 from pathlib import Path
 from loguru import logger
 from sklearn.model_selection import train_test_split
-from torch._six import string_classes, int_classes
+from torch._six import string_classes #, int_classes
```
jcohenadad, line 11: same comment here about code commenting

ivadomed/loader/utils.py
```diff
@@ -254,7 +255,7 @@ def imed_collate(batch):
         if elem.shape == ():  # scalars
             py_type = float if elem.dtype.name.startswith('float') else int
             return __numpy_type_map[elem.dtype.name](list(map(py_type, batch)))
-    elif isinstance(batch[0], int_classes):
+    elif isinstance(batch[0], int): #int_classes):
```
jcohenadad, line 258: and here

etetteh:
I will resolve that now.

ivadomed/random.py
```diff
@@ -0,0 +1,8 @@
+import torch
+import random
+import numpy as np
+
+def set_seed(seed=6, rank=0):
```
jcohenadad, line 5: please add docstrings.

ivadomed/loader/utils.py
jcohenadad, line 258: does this change have to do with the purpose of this PR? if so, how? if not, can you please fix it in another PR?

ivadomed/random.py
jcohenadad: pls indicate what the function does

ivadomed/loader/utils.py
etetteh: Some versions of `torch` do not have the `int_classes` in the `torch._six` attribute, and according to  this [discussion](https://discuss.pytorch.org/t/cannot-import-name-int-classes-from-torch-six/126611) users can fix that by simply using `int` instead of importing `int_classes`

jcohenadad: looks a bit out-of-scope-- i will let @dyt811 take over

ivadomed/loader/utils.py
```diff
@@ -219,7 +220,7 @@ def get_subdatasets_subject_files_list(split_params, df, path_output, subject_se
         train_lst, valid_lst, test_lst = get_new_subject_file_split(df=df,
                                                                     split_method=split_params[SplitDatasetKW.SPLIT_METHOD],
                                                                     data_testing=split_params[SplitDatasetKW.DATA_TESTING],
-                                                                    random_seed=split_params[SplitDatasetKW.RANDOM_SEED],
+                                                                    random_seed=random_seed,
```
andreanne-lemay, line 223: I am worried here about retro-compatibility. Old config files where the seed is specified in the split dataset will probably crash. I think it's find to move it, but we should make sure at least for now that the current config files don't crash. There is this file that helps to deal with these scenarios: https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config_manager.py#L70

ivadomed/loader/utils.py
etetteh: Okay. We may add it in a future PR since it works for all torch versions, unlike the `int_classes` .

jcohenadad: in this case can you pls open an issue so it doesn't fall through the cracks-- thx

ivadomed/loader/utils.py
andreanne-lemay: People using ivadomed use these config files as templates, but often have copies. For some studies, we have dozens of config file, so it's not viable to change the config file every time the code changes. When possible, we like to make everything retrocompatible which is why we created the class ConfigManager. This way it will support the new and previous versions of the config file.

ivadomed/random.py
```diff
@@ -0,0 +1,16 @@
...
+    This function controls sources of randomness to aid reproducibility.
+    Args:
+        seed: int to initialize random number generator (RNG)
+        rank: int specifying the GPU id for training
+    Returns:
+        seeded RNG in the specified libraries/packages.
+    """
+    random.seed(seed + rank)
+    np.random.seed(seed + rank)
+    torch.manual_seed(seed + rank)
```
andreanne-lemay, line 16: Why do we need the `rank` parameter? I'm not sure we want the seed to be different if we change GPUs or train on multiple GPUs.

ivadomed/loader/utils.py
etetteh: Well, noted.

ivadomed/random.py
etetteh: We don't necessarily need it for the current implementation of ivadomed. However, it is a convention used for multi GPU training and I left it for future works. Also, it doesn't hurt anything as of now. We can remove it, if it violates ivadomed conventions.

ivadomed/loader/utils.py
etetteh: Yes, please.

ivadomed/__init__.py
```diff
@@ -1 +1,2 @@
 from .utils import __version__, __ivadomed_dir__
+from .random import set_seed
```
dyt811, line 2: I usually do not recommend relative imports, unless absolutely necessary. A good explanation is provided here: https://realpython.com/absolute-vs-relative-python-imports/#pros-and-cons-of-relative-imports

ivadomed/config/config_vertebral_labeling.json
```diff
@@ -27,7 +27,7 @@
     },
     "split_dataset": {
         "fname_split": null,
-        "random_seed": 8,
+        "random_seed": 6,
```
dyt811, line 30: Curious about your thoughts on centralizing the seeds all to a single place. So if that to happen, would we still need to indicate `random_seed` here? Also, just curious why this particular JSON was changed instead of the main `config.json`?

I see below, you used this as the anchor to set the random seed for all dependencies. I agree from a practical perspective, this is the "quickest" way to get to proper seeding with minimal change to the code base. 

From a design/user configuration perspective, do you think it would easier for you (as a user role) to see this setting at the "top" configuration json level instead? You have probably seen how seeding is being done in other libraries and code bases, any interesting examples we can reference here as a good practice example?

ivadomed/loader/utils.py
```diff
@@ -8,9 +8,10 @@
 from pathlib import Path
 from loguru import logger
 from sklearn.model_selection import train_test_split
-from torch._six import string_classes, int_classes
+from torch._six import string_classes
 from ivadomed import utils as imed_utils
-from ivadomed.keywords import SplitDatasetKW, LoaderParamsKW, ROIParamsKW, ContrastParamsKW
+from ivadomed.keywords import SplitDatasetKW, LoaderParamsKW, ROIParamsKW, ContrastParamsKW, ConfigKW
+from ivadomed.random import set_seed
```
dyt811, line 14: I like how you used absolute import here. 👍

ivadomed/loader/utils.py
```diff
@@ -141,7 +142,7 @@ def get_new_subject_file_split(df, split_method, data_testing, random_seed,
             raise ValueError("All lists in subject_selection parameter should have the same length.")
 
         sampled_dfs = []
-        random.seed(random_seed)
+        set_seed(random_seed)
```
dyt811, line 145: Just to recap, this is setting the seed for all dependencies base on the  `"split_dataset": {"random_seed": 6}` in the config JSON right?

ivadomed/loader/utils.py
```diff
@@ -254,7 +255,7 @@ def imed_collate(batch):
         if elem.shape == ():  # scalars
             py_type = float if elem.dtype.name.startswith('float') else int
             return __numpy_type_map[elem.dtype.name](list(map(py_type, batch)))
-    elif isinstance(batch[0], int_classes):
+    elif isinstance(batch[0], int):
```
dyt811, line 258: This is for the Torch 1.9+ fix right? I.e. if we leave the original import `int_classes`, it will break on Torch 1.9+?

ivadomed/main.py
```diff
@@ -9,6 +9,7 @@
 import multiprocessing
 import re
 
+from ivadomed.random import set_seed
```
dyt811, line 12: I also like how you did absolute import here. 👍

ivadomed/main.py
```diff
@@ -347,6 +348,8 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False):
     logger.add(str(path_log))
     logger.add(sys.stdout)
 
+    set_seed(context[ConfigKW.SPLIT_DATASET][SplitDatasetKW.RANDOM_SEED])
```
dyt811, line 351: Same comments as above with regard to how to best set seeds. Maybe we can think about a higher level key at context level to set the Random Seed for all instead of having to delve down into the SplitDataSet Random Seed level.

ivadomed/random.py
```diff
@@ -0,0 +1,16 @@
...
+import numpy as np
+
+def set_seed(seed=6, rank=0):
+    """
+    This function controls sources of randomness to aid reproducibility.
+    Args:
+        seed: int to initialize random number generator (RNG)
+        rank: int specifying the GPU id for training
+    Returns:
+        seeded RNG in the specified libraries/packages.
```
dyt811, line 12: Can probably remove this line, as there is no "return" _per se_.  Sorry for being pedantic.

ivadomed/random.py
dyt811: I also recommend adding type hinting, i.e. `(seed: int = 6, rank: int = 0)`

ivadomed/random.py
dyt811: Great idea and initiative 👍. I think that is a great future supporting perspective on controlling the variability at the hardware level. 

The potential downside (and this is a hunch for now) is that if  we use something like Compute Canada, the rank information... may be very unspecified? Sort of an edge case for now. Beside that point...

For localized training with controlled hardware environment, I think this works pretty flawlessly and FORCE the ORIGINAL experimenter to really replicate the experiment controlling the specific GPU used. I think that is a great idea. 

However, this situation may prove very... challenging for non ORIGINAL experimenter. E.g. Student A in another lab trying to reproduce our work will probably not be able to use the exact same hardware and GPU rank as we have (in fact a known issue with our current unit tests are failing with GPU unit tests because config requested a non-existent Rank), and or forcibly using a GPU in a particular slot just because we used that slot. Or worse, if  we have a Rank 7 GPU in a 8 GPU cluster setup, the students will never be able to replicate the results with same config UNLESS they tweak that `set_seed` function to manually set the values (which is beyond what we normally  expect from novice users/students outside of our contributor circle). 

Therefore I think it is probably safer at this point to not included it in line with the original goal of this PR, which is the allow seed fixing. By including Rank, we may in scenarios inadvertently  forcing unreproducible behavior via different seed, which goes against what you hope to achieve. 

I do think this is a valuable idea that when multi GPU support comes alive, we should make this an optional setting potential for people to set it. But that is well beyond current scope in terms of implementation/complexity so I do not recommend implemented at the moment.

ivadomed/training.py
```diff
@@ -22,11 +22,12 @@
 from ivadomed.loader import utils as imed_loader_utils
 from ivadomed.loader.balanced_sampler import BalancedSampler
 from ivadomed.keywords import ModelParamsKW, ConfigKW, BalanceSamplesKW, TrainingParamsKW, MetadataKW
+from ivadomed.random import set_seed
```
dyt811, line 25: In case you don't realize it, I really really appreciate absolute path import. 👍

ivadomed/training.py
```diff
@@ -54,6 +55,8 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
     # Write the metrics, images, etc to TensorBoard format
     writer = SummaryWriter(log_dir=path_output)
 
+    set_seed(random_seed)
```
dyt811, line 58: Coming back to a few points I mentioned earlier, **when** to set the seed is actually a great and not so simple question. If we do centralize the seed somehow, we probably want to set it sometimes as early as possible. However, current seed is in the `config` file so maybe a better place to set seed is in the loading function for config? What are your thoughts on this? Ideally, we probably only need to set the seed once as well.

dyt811 (commented):
Lots of great ideas. I think we should discuss a bit at the meeting about the pros vs cons of centralized seed setting (as you have started doing in this PR) and how to do it more comprehensively. 

A strong evidence to support any approach is to look at other mature project code bases which implement seed setting from config file and how they do it and when they do it to draw analogy from and should help lend support to your proposed approach.

ivadomed/random.py
etetteh: I agree with the fact that `rank` is not needed at this point, and it is mainly a feature of multi GPU use cases. The reason I included it is for us not to come back to this when we start supporting multi GPU.
However, I can take it out at this point.

ivadomed/__init__.py
etetteh: Okay. I will remedy that. I was following the existing convention here.

ivadomed/config/config_vertebral_labeling.json
etetteh: > Also, just curious why this particular JSON was changed instead of the main config.json?

I don't really understand what you mean here. This is the main `config.json` file, unless I might have changed something that has skipped my mind.

> From a design/user configuration perspective, do you think it would easier for you (as a user role) to see this setting at the "top" configuration json level instead? You have probably seen how seeding is being done in other libraries and code bases, any interesting examples we can reference here as a good practice example?

Most of the libraries/codebases I have come across use the seed at the top level. Initially, that was what I did, but @andreanne-lemay  and @mariehbourget mentioned that would mean changes to the existing config files (Comment is at the top), so I had to undo everything and use the one inside `split_dataset` as the anchor. I don't think it is a good design principle to do that, but it is a fix for now until I can work around the `config_manager` file to make it "retrospective".

ivadomed/loader/utils.py
etetteh: Right

ivadomed/loader/utils.py
etetteh: Yes, that is right.

ivadomed/training.py
etetteh: Okay. I do too, will ensure I continue doing that.

ivadomed/training.py
etetteh: The way the ivadomed codebase is, I couldn't come up with a better fix (I have to analyse more, and we need to discuss), than to have one in the `training` and another in the `loader`. 
Usually, the seed is to be set once. Anyways, I don't think the current implementation hurts anything, except the issue with where the key needs to be in the config file.

etetteh:
I have made the changes concerning the use of the `rank` argument and the data type hinting (I thought I had already done that).

ivadomed/random.py
```diff
@@ -0,0 +1,13 @@
...
+
+def set_seed(seed: int = 6):
+    """
+    This function controls sources of randomness to aid reproducibility.
+    Args:
+        seed: int to initialize random number generator (RNG)
+    """
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
```
naga-karthik, line 13: Just adding a reference here - https://pytorch-lightning.readthedocs.io/en/stable/_modules/pytorch_lightning/utilities/seed.html#seed_everything to set the seed across the GPUs also.
I think that also should be added here.

naga-karthik: This is also a good reference about setting global seed - https://github.com/chandar-lab/RLHive/blob/c5b1b77a3daf87aeb877e10c4f65bcd7e2bc1ff8/hive/utils/utils.py#L34

etetteh: Thank you @naga-karthik. 
With the current implementation of `random.py`, I am able to reproducible the results of an experiment on CPU. I ran this [tutorial](https://ivadomed.org/tutorials/one_class_segmentation_2d_unet.html) with this [config.json](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config.json) for 8 runs using the same seed, 5 epochs each and I obtained the same results.

Also, the experiment was done with the seed set inside the [main.py](https://github.com/etetteh/ivadomed/blob/a7125d5e5f3c946eefc2f9aec8b92e505d6c6244/ivadomed/main.py#L351)  only, without those in [loader/utils.py](https://github.com/etetteh/ivadomed/blob/a7125d5e5f3c946eefc2f9aec8b92e505d6c6244/ivadomed/loader/utils.py#L145), and [training.py](https://github.com/etetteh/ivadomed/blob/a7125d5e5f3c946eefc2f9aec8b92e505d6c6244/ivadomed/training.py#L58).

However, I have not been able to obtain reproducible results on a GPU (V100), even upon adding the GPU specific features referenced above by @naga-karthik

ivadomed/__init__.py
dyt811: Totally understandable. I was in the same shoes. There were some dated practices that I see around that could be improved and I found myself frequently introduce them in PRs and bloated PR to large size than usual. I think you have many great ideas and feel free to experiment them. I found that what helps me the most is do some trailblazer branches and then look back on how that can be brokeninto smaller/modular PRs to make the incorporate easier to adopt. I can't say I mastered that myself but I find that smaller PR proportionally increase approval time magically somehow. 👍 . Keep up the good recommendations.

ivadomed/config/config_vertebral_labeling.json
dyt811: Good point. Thanks for the discussion. Regarding the first point. I think [config_vertebral_labeling.json] is only a specific config.json instead of the [main config.json](https://github.com/ivadomed/ivadomed/blob/master/ivadomed/config/config.json)? So depending on what your goal is, would it be more applicable to modify the main config JSON file?

ivadomed/__init__.py
etetteh: Thank you, and advice well taken note of.

ivadomed/random.py
kanishk16: I reviewed a few more widely used libraries which could serve as references:

- MONAI - https://github.com/Project-MONAI/MONAI/blob/a676e3876903e5799eb3d24245872ea4f3a5b152/monai/utils/misc.py#L224
-  MMSegmentation - https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/apis/train.py#L19

etetteh:
I strongly think being able to reproduce results is crucial and should be prioritized, inasmuch as it is a challenging task. 
One of the things I asked myself was that if the results are changing for each experiment run even with the same seed, how sure am I that a hyperparameter tweak/change is having an effect.

ivadomed/random.py
etetteh: Thanks @kanishk16. I will run experiments with those suggestions, to see if they fix the issue.

Previously, I tested the [torch.backends.cudnn.deterministic = True](https://github.com/Project-MONAI/MONAI/blob/a676e3876903e5799eb3d24245872ea4f3a5b152/monai/utils/misc.py#L275), but it raises an error for some objects in ivadomed, because not all objects have a deterministic capability. Will dive deeper.

PR #1100, opened 2022-03-11
Fix loading when names of multiple target_suffix overlap

### LINKED ISSUE #1055: Target suffix loading: use of `contains` may pick multiple targets
### Current behavior

The current implementation of target suffix loading involves the use of the `contains()` string method to capture target files as shown in the below lines:

https://github.com/ivadomed/ivadomed/blob/76b36a0a0f7141feb2d5b00b33e4c3a06865fc2c/ivadomed/loader/bids_dataframe.py#L125-L128

This means that

* if your derivatives contains multiple filenames with overlap (e.g. `_lesion-manual.nii.gz` and `_lesion-manual2.nii.gz`) and
* if  you use a `"target_suffix"` containing the overlap (e.g. `_lesion-manual`),

then the current loading process picks all of the filenames with the given overlap as potential targets. 

This is problematic because the user might (reasonably) think that only `_lesion-manual.nii.gz` will be used as the ground-truth when the `"target_suffix"` is specified as `_lesion-manual`, whereas in reality both `_lesion-manual.nii.gz` and `_lesion-manual2.nii.gz` are used. 

I didn't go deeper in the codebase yet to see what happens when multiple targets are picked during the loading process, but I can confirm that the GT that ends up being used changes from run to run based on my experiments with `ivadomed --test`.

### Expected behavior
 

I would expect the `"target_suffix"` to **exactly match** the target filename. Two options we have are:
1. Changing line 128 as shown above with something like
   ```
   & df_next['filename'].str.split(os.extsep).apply(lambda x: x[0]).str.endswith('|'.join(self.target_suffix)))]
   ```
   which enforces an exact match of target suffix.

2. In case the use of `contains()` is justified in some scenarios and its implemented that way for a specific reason, the user can instead give a full `"target_suffix"` including the file extension such as `_lesion-manual.nii.gz`. I have tested this and can confirm it solves the problem.

The latter option is notably a lot easier to implement as it doesn't require a change in the codebase. However, I think this is an issue many people might face without knowing it in the future.

### Steps to reproduce

1. Download the `basel-mp2rage` dataset as shown [here](https://github.com/ivadomed/model_seg_ms_mp2rage#get-the-data).
2. Run preprocessing on the data as shown [here](https://github.com/ivadomed/model_seg_ms_mp2rage#prepare-the-data).
3. Insert `print(bids_df.get_deriv_fnames())` [after line 370 in `ivadomed/main.py`](https://github.com/ivadomed/ivadomed/blob/76b36a0a0f7141feb2d5b00b33e4c3a06865fc2c/ivadomed/main.py#L370) which initializes the BIDS data frame.
4. Run the [following config file](https://github.com/ivadomed/model_seg_ms_mp2rage/blob/main/config/seg_lesion.json) with `ivadomed --train`.
5. Take a look at the output of the print statement from step 3.

However, this is specific to one dataset and it is understandable that not everybody will be willing to run preprocessing etc. on this particular dataset. In this case, I would like to point out that this issue is reproducible with **any dataset which has multiple annotations per subject where the string target suffix for these annotations have an overlap**.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR attempts to fix issue #1055 where the derivatives were index with "contains" in the filename that could pick-up multiple targets in certain case.

I replaced "contains" by "endswith" which will prevent the 2 problems we have encountered so far when the target_suffix names overlap which are:
- `_lesion-manual` and `_lesion-manual2` files were used with target_suffix `_lesion-manual` ([ref comment](https://github.com/ivadomed/ivadomed/issues/1055#issue-1114505527))
- `_seg-axon` and `_seg-axonmyelin` files were used with target_suffix `_seg-axon` ([ref comment](https://github.com/ivadomed/ivadomed/issues/1055#issuecomment-1064226989))

## Discussion points:

**EDIT: I'll leave the discussion points here for future reference. However, as commented below, the "endswith" approach is sufficient for now and further improvements of pairing visibility of image and grounds-truths for the users will be done separately (see #1102).**
1. Some cases may not be covered because "endswith" is still not an exact match.
ex: `manual` could be used to match `lesion-manual` ([ref comment](https://github.com/ivadomed/ivadomed/issues/1055#issuecomment-1023904525))
2. I was unable to enforce an exact match because I don't know where to split the filename to get the beginning of the `target_suffix`. My original idea was to split after the `modality_suffix` (columns `suffix` in the dataframe), but unfortunately the `modality_suffix` are not indexed by pybids for derivatives files.
3. We could split at the last underscore but that implies that the `target_suffix` in the filenames must not have underscore in them except for the leading underscore.
4. In addition, an exact match implies that the underscore must always be included in the `target_suffix` in the config file.

## Linked issues
Fixes #1055

### DIFF
--- a/ivadomed/loader/bids_dataframe.py
+++ b/ivadomed/loader/bids_dataframe.py
@@ -115,8 +115,11 @@ class BidsDataframe:
             # Drop rows with json, tsv and LICENSE files in case no extensions are provided in config file for filtering
             df_next = df_next[~df_next['filename'].str.endswith(tuple(['.json', '.tsv', 'LICENSE']))]
 
-            # Update dataframe with subject files of chosen contrasts
-            # and with derivative files of chosen target_suffix
+            # The following command updates the dataframe by doing 2 things:
+            # 1. Keep only subject files of chosen contrasts (for files that are not in the 'derivatives' folder)
+            #    (ex: '<dataset_path>/sub-XX/anat/sub-XX_T1w.nii.gz' with contrast_lst:["T1w"])
+            # 2. Keep only derivatives files of chosen target_suffix (for files that are in the 'derivatives' folder)
+            #    (ex: '<dataset_path>/derivatives/labels/sub-XX/anat/sub-XX_T1w_seg-manual.nii.gz' with target_suffix:["_seg-manual"])
             df_next = df_next[(~df_next['path'].str.contains('derivatives')
                                & df_next['suffix'].str.contains('|'.join(self.contrast_lst)))
                                | (df_next['path'].str.contains('derivatives')
--- a/ivadomed/loader/bids_dataframe.py
+++ b/ivadomed/loader/bids_dataframe.py
@@ -119,8 +119,8 @@ class BidsDataframe:
             # and with derivative files of chosen target_suffix
             df_next = df_next[(~df_next['path'].str.contains('derivatives')
                                & df_next['suffix'].str.contains('|'.join(self.contrast_lst)))
-                              | (df_next['path'].str.contains('derivatives')
-                                 & df_next['filename'].str.contains('|'.join(self.target_suffix)))]
+                               | (df_next['path'].str.contains('derivatives')
+                               & (df_next['filename'].str.split('.').apply(lambda x: x[0])).str.endswith(tuple(self.target_suffix)))]
 
             # Update dataframe with files of chosen extensions
             df_next = df_next[df_next['filename'].str.endswith(tuple(self.extensions))]


jcohenadad:
Regarding the discussion points, I would favor the approach where the user has more visibility about the pairing that was done by ivadomed (ie: image & target ground truth), and that way, if they realize the pairing was wrong (because of the '-manual' example cited), then they would understand and change the target 'endswith' key in the config file for the value '-lesion-manual').

ivadomed/loader/bids_dataframe.py
```diff
@@ -119,8 +119,8 @@ def create_bids_dataframe(self):
             # and with derivative files of chosen target_suffix
             df_next = df_next[(~df_next['path'].str.contains('derivatives')
                                & df_next['suffix'].str.contains('|'.join(self.contrast_lst)))
-                              | (df_next['path'].str.contains('derivatives')
-                                 & df_next['filename'].str.contains('|'.join(self.target_suffix)))]
+                               | (df_next['path'].str.contains('derivatives')
+                               & (df_next['filename'].str.split('.').apply(lambda x: x[0])).str.endswith(tuple(self.target_suffix)))]
```
dyt811, line 123: I have to say, this is some of the most complex one liners I have seen in Python ever among all my projects. I sort of understand we have to write it mainly due to the df[ ] filtering syntax but I keep wondering if we can make it more readable by either breaking it down or simplifying the overall expression. 

My understanding is for line 123, you are taking file name, convert to string, split it by `.` into list then for the FIRST part, (pre extension?), filter for elements in that above list which ends with one of the target_suffix and applying this boolean selection to filter the df_next along with previous expression before |...

... and I only explained line 123, not 120, 121 and 122.

I think it is fine for now if that is your goal to do what I described above?  

I will ask @konantian to see if we can find a way to reword this to be more readable. I strongly recommend adding more comment explaining the mechanism and maybe best give a few example file names. 

I do agree this will be better than the earlier contain approach is MORE specific than before though. 

I will update to approval once the unit test passes.

dyt811 (commented):
Thanks for the great work. I certainly think this will be better handling of  the naming than before. 

I will update to approval once local test passes.

dyt811 (approved):
Looking good! Thanks a lot!

ivadomed/loader/bids_dataframe.py
mariehbourget: > I have to say, this is some of the most complex one liners I have seen in Python ever among all my projects. I sort of understand we have to write it mainly due to the df[ ] filtering syntax but I keep wondering if we can make it more readable by either breaking it down or simplifying the overall expression.
> 
> My understanding is for line 123, you are taking file name, convert to string, split it by `.` into list then for the FIRST part, (pre extension?), filter for elements in that above list which ends with one of the target_suffix and applying this boolean selection to filter the df_next along with previous expression before |...
> I think it is fine for now if that is your goal to do what I described above?

Yes, you explanation is right!

> I will ask @konantian to see if we can find a way to reword this to be more readable. I strongly recommend adding more comment explaining the mechanism and maybe best give a few example file names.

I agree, this one liner is hard to understand, any help to simplify it would by greatly appreciated!
In the meantime, I updated the comment above the command with more details on the behavior and a filename example for each, let me know if that is better (or worse 😅)

Issue #1049, opened 2022-01-18
Inference on images with shapes different from training


andreanne-lemay:
## Issue description
This issue has for goal to document inference on images with dimensions different from the ones used during training.

## 2D UNet
The 2D model trained for spinal cord segmentation in the module `sct_deepseg` named `seg_sc_t2star` was used for testing.

1. I ran the original model with the following command line. The resulting segmentation is presented in red (see image). 
```bash
sct_deepseg -i data_example_spinegeneric/sub-cardiff03/anat/sub-cardiff03_T2star.nii.gz -o cropping.nii.gz -task seg_sc_t2star
```
2. I locally modified the configuration file in `spinalcordtoolbox/data/deepseg_models/t2star_sc/t2star_sc.json` and removed the cropping step that center cropped the image to a 160x160 pixel image. Each image passed through the model was now 307x307 pixels. I re-run the command listed in step 1. The resulting segmentation is presented in blue.

![image](https://user-images.githubusercontent.com/49137243/150023705-a35395d9-4c83-4807-a760-57904f87e51f.png)

Observation: It was possible to run the model on images of the cropped image of size 160x160 and on the whole image of 307x307 pixels. Both models have similar outputs (but not exactly the same).

## 3D UNet
The task trained for spinal cord tumor segmentation composed of 2 cascaded model (1-SC detection and cropping and 2-Tumor segmentation) in the module `sct_deepseg` named `seg_tumor-edema-cavity_t1-t2` was used for testing.

1. I ran the original model with the following command line.
```bash
sct_deepseg -i 'sub-Astr144/anat/sub-Astr144_T2w.nii.gz' 'sub-Astr144/anat/sub-Astr144_T1w.nii.gz' -c t2 t1 -o cropping.nii.gz -task seg_tumor-edema-cavity_t1-t2
```
2. I locally modified the configuration file in `spinalcordtoolbox/data/deepseg_models/model_seg_sctumor-edema-cavity_t2-t1_unet3d-multichannel/model_seg_sctumor-edema-cavity_t2-t1_unet3d-multichannel.json` and modified the length and stride of patches to match the size of the resampled image (240x240x19). This way, only one big patch would be processed at once. I run the command following the command line to run only the tumor segmentation model (without the initial cropping).
```bash
[...]

PR #1101, opened 2022-03-11
Segment 2D images without patches

mariehbourget:
#### GitHub

#### PR contents

## Description
This draft PR:
- adds an option to segment 2D images with and without patches (default: without) as discussed in #1049 
- adds an argument to modify the patch overlap

Both only apply for `--segment` command.
This is only a prototype and further testing is needed.

## Linked issues
Related to #1049

### DIFF
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -349,6 +349,8 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
                             Length equals 2 [PixelSizeX, PixelSizeY] for 2D or 3 [PixelSizeX, PixelSizeY, PixelSizeZ] for 3D, \
                             where X is the width, Y the height and Z the depth of the image.
             * 'pixel_size_units': (str) Units of pixel size (Must be either "mm", "um" or "nm")
+            * 'patch': (bool) Indicate if 2d patching is used while segmenting (command '--segment'), otherwise inference is \
+                       performed on the entire image at once.
             * 'overlap_2D': (list of int) List of overlaps in pixels for 2D patching. Length equals 2 [OverlapX, OverlapY], \
                             where X is the width and Y the height of the image.
             * 'metadata': (str) Film metadata.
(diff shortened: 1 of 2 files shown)


hermancollin:
I can review this PR when it's ready. It would be nice to also offer this option in ADS. At a quick glance, I see you added an argument to `segment_volume` so I'm guessing it would be pretty easy to implement on the ADS side.

mariehbourget:
> I see you added an argument to `segment_volume` so I'm guessing it would be pretty easy to implement on the ADS side.

Yes exactly, we would only have to pass the argument to segment_volume without changes to the config file.

hermancollin:
@mariehbourget I will review this today but I don't understand how it works. My understanding was that u-nets have a fixed input size. Is the image resized to the input patch size the model was trained on? That doesn't make sense because your large segmented images in #1049 would never be as good. Or is the image subdivided in 512x512 patches (for example) without overlap and with padding when the size is smaller than the patch size?

mariehbourget:
Hi @hermancollin, great questions!
My understanding may not be complete either on what is happening, but here is what I gathered from my tests and from the tests in #1049.

> Is the image resized to the input patch size the model was trained on?

No, the only resampling is the resampling to the common resolution (`Resample` param), wihtout any ties to the original patch size used.

> Or is the image subdivided in 512x512 patches (for example) without overlap and with padding when the size is smaller than the patch size?

No, there is no patches at all.
A model with patches and without overlap would have a `length_2D == stride_2D`.
In 2D in ivadomed, there is never any padding (I think), only for sub-volumes in 3D which I'm not sure why.

> My understanding was that u-nets have a fixed input size.

> I get that these networks should theoretically be able to segment any image size because the learned weights are in the convolutional kernels, but since I'm not as familiar with ivadomed codebase, I don't see the underlying logic.

From what I understand, the model doesn't care about the size of the input. As you said, the weights are in the convolutional kernels and they are shared by all the pixels of the same layer, regardless of how many pixels there are.

So if I'm not mistaken, the "fixed" input size is only really to be able to use a batch size > 1 i.e. combine several images in the same tensor of size (batchsize, channels, X, Y).
If you use a batch size of 1, I think you should be able to train a model with images of different sizes.
However, the "fixed" smaller patch sizes are also probably necessary to avoid memory issues with large images.

So we are able to segment different image sizes because:
- each image is loaded independently of the others i.e. batch_size = 1
- in the case of a 3D volume segmented by 2D slices, all slices have the same dimensions so the batch size is not important.

Let me know if that makes sense.

hermancollin:
Yes that makes a lot of sense. Even though you reported the metrics were slightly worse when using `--no-patch`, I feel like they are more visually pleasing. The "patching artifacts" or patch-overlap problems we often see have a definite impact on the seg quality, especially for smaller axons. Without patches, I notice a bunch of small axons suddently appear out of thin air.

Everything LGTM. The option makes ivadomed load the image as a single slice. Also, it's faster than I expected. Using `segment_volume` directly (ONNX model):

with patches:
```
python script_patching.py  29,33s user 1,67s system 176% cpu 17,592 total
```
without patches:
```
python script_no_patching.py  17,69s user 0,93s system 169% cpu 10,984 total
```
Keep in mind the script with patching had some slight patch overlap.

hermancollin (approved):

hermancollin:
@mariehbourget yes, even better!

Issue #1035, opened 2021-12-24
Documentation not showing up in Scripts section in RTD


uzaymacar:
In the RTD, documentation for a few scripts currently doesn't show up at https://ivadomed.org/scripts.html. The missing scripts are:
* `ivadomed_visualize_transforms`
* `ivadomed_automate_training`
* `ivadomed_prepare_dataset_vertebral_labeling`

As can be seen in `docs/source/scripts.rst`, the documentation for this part is generated with `autofunction`. 

What is interesting is, when I do

```
cd docs
make html
```

as instructed [here](https://github.com/ivadomed/ivadomed/wiki/documentation) and locally look at the docs, **there is no problem**. Somehow this only happens when deployed.

I tried to find a `latest` build that doesn't have this issue from [here](https://readthedocs.org/projects/ivado-medical-imaging/builds/) but I haven't been successful yet.

jcohenadad:
Interesting. [This previous build](https://ivado-medical-imaging--980.org.readthedocs.build/scripts.html) does show all the scripts. I'm looking into this now.

jcohenadad:
There is a bunch of warnings involving `visualize_transforms` (and other scripts that fail to be displayed) in the latest build, which do not appear in build 980 linked above:

<details><summary>Terminal output</summary>

```console
WARNING: autodoc: failed to import module 'loader' from module 'ivadomed.loader'; the following exception was raised:
cannot import name 'int_classes' from 'torch._six' (/home/docs/checkouts/readthedocs.org/user_builds/ivado-medical-imaging/envs/1034/lib/python3.7/site-packages/torch/_six.py)
WARNING: autodoc: failed to import module 'utils' from module 'ivadomed.loader'; the following exception was raised:
cannot import name 'int_classes' from 'torch._six' (/home/docs/checkouts/readthedocs.org/user_builds/ivado-medical-imaging/envs/1034/lib/python3.7/site-packages/torch/_six.py)
WARNING: autodoc: failed to import module 'utils' from module 'ivadomed.object_detection'; the following exception was raised:
cannot import name 'int_classes' from 'torch._six' (/home/docs/checkouts/readthedocs.org/user_builds/ivado-medical-imaging/envs/1034/lib/python3.7/site-packages/torch/_six.py)
WARNING: autodoc: failed to import module 'evaluation' from module 'ivadomed'; the following exception was raised:
[...]

PR #1104, opened 2022-03-15
Fix import of int_classes from torch._six in torch version 1.9+

etetteh:
Although the current ivadomed implementation is built on torch version 1.8, newer versions of torch no longer support importing int_classes from torch._six, and we need to remedy this in a future update of ivadomed when upgrading the underlying version of torch.

This PR proposes the fix for the said issue.


konantian:
Maybe we could upgrade `torch` to 1.10 in `requirements.txt` as well.

etetteh:
Yes, that would be my best suggestion

konantian:
This PR also resolved issue #1035

dyt811:
> Maybe we could upgrade `torch` to 1.10 in `requirements.txt` as well.

I agree we should do more tastings as well and eventually move to 1.10 in the near future.

dyt811 (approved):
So far, this is one of the fixes that have been widely documented across the web to support forward compatibility. 

Please ensure in future PRs, follow the recommended format as indicated in the template to facilitate review and documentation purposes (especially when eventually release notes will link back to this PR for explanation and context).

PR #1000, opened 2021-12-06
fix: Unify current installation

kanishk16:
## Description
**tl;dr**: Unifies the installation process through `extras_require` argument in `setup.py`

As discussed in #996 and in light of discussion in #795 especially @kousu's [comment](https://github.com/ivadomed/ivadomed/issues/795#issuecomment-879952457) making torch non-mandatory resulted in a few blockers as well as overgrown installation instructions. Thus, increasing the priority to streamline the installation process along with mandating torch installation keeping in mind the user's freedom to choose a backend for torch either through CPU or GPU. 

This PR:
1. Reverts #757 by eliminating the need for cascading `requirements.txt` files and making torch mandatory through `extras_require`
2. Resolves #861 as users can now install ivadomed through pip via `pip install ivadomed[cpu]` or `pip install ivadomed[gpu]` which not only installs minimal requirements but also torch depending upon the backend (still pinned to CUDA-11). Moreover, `pip install ivadomed[dev_gpu]` installs minimal requirements + torch on GPU (CUDA-11) + dev requirements 
3. Consequently, a limitation of this plausible solution is `pip install ivadomed` wouldn't install anything requiring users to explicitly specify the backend to install ivadomed. However, it makes easier for downstream packages to specify ivadomed as a dependency `ivadomed[cpu]` or `ivadomed[gpu]`. 
4. ~#1057 as proposed in the [comment ](https://github.com/ivadomed/ivadomed/issues/1057#issuecomment-1031707720). As the proposed solution is much more general and handles concerns to Compute Canada, the idea of `pip install ivadomed[inference]` is dropped. `pip install ivadomed[inference]` installs minimal requirements + torch and torchvision on cpu this can be further tuned to install packages that would be required for inference only, related to #795.~
Update: Moved to Singularity as a fix since ERROR(s) still persist
5. Resolves #996

HOW?
This PR introduces two functions to `setup.py`:
1. `get_whl_info()` - returns tags required to handpick torch +torchvision wheels
2. `get_torch_whls(backend)` - handpicks torch + torchvision wheels based upon the backend

#### TODO:
- [x] Update Docs: Open a separate PR to incorporate the updated procedure for installation and all the minor readability issues

To test the PR, just try any of the following commands to install ivadomed (preferably in a new virtual environment):
- `pip install .[cpu]` - installs minimal requirements + torch on CPU
- `pip install .[gpu]` - installs minimal requirements + torch on GPU (CUDA-11)
[...]

Issue #728, opened 2021-03-16
Installation Instructions incomplete


ahill187:
## Issue description

The installation instructions for developers on the docs are not quite correct (at least for my version). If I do 
```
pip install -e .
```

in a clean virtual environment, I get an error first that `Cython` is not installed, and then that `numpy` is not installed.

### Suggestion

Update installation docs to the following:

```
pip install Cython
pip install numpy
pip install -e .
```

## Environment
venv

### System description
MacOS 10.16 (Big Sur)
Python 3.8

### Installed packages
None

ssinad:
Adding Cython to [requirements.txt](https://github.com/ivadomed/ivadomed/blob/master/requirements.txt) might solve the problem.

dyt811:
This problem is also somewhat happening on Windows. With pytorch 1.5.0 missing and require conda to resolve. I am attempting to add a conda YAML for possible alternative installation path that is more... cross platform compatible BUT require conda. Still an idea for now as I get it working at least on Windows. 

Do you all know the rationale/history as to why we initially went with pip instead of conda? Was it related to Compute Canada?

jcohenadad:
> Do you all know the rationale/history as to why we initially went with pip instead of conda? Was it related to Compute Canada?

a mix between CC and @kousu suggestions to not use conda

kousu:
Thanks for catching this @ahill187. Build dependencies -- as opposed to runtime dependencies -- are always a bit tricky because they skim a chicken-or-egg problem.

First, is this really our bug? I'm trying to reproduce it but there are no cython files that I can see in this project, so we shouldn't need it. It might be that one of the dependencies is mispackaged, and that `pip install cython numpy; pip install -e .` is a workaround for their bug. If so, we should find it and submit a PR to them.

If it is our bug then I'll share a tip: the most correct way to handle build dependencies is with `setup_requires`: https://stackoverflow.com/a/38057196/2898673 though there's [`pyproject.toml`](https://www.python.org/dev/peps/pep-0518/) which is [trying](https://martin-thoma.com/pyproject-toml/) to [replace this](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#package-requirement). IMO (and seemingly in the opinion of a ton of python devs) it's too new and unstable to trust yet so I'm not going to cover it. I think you just need to add:

```diff
diff --git a/setup.py b/setup.py
index ad4d638e..8a869751 100644
--- a/setup.py
+++ b/setup.py
[...]

PR #1125, opened 2022-05-02
docs: unify installation

### LINKED ISSUE #1019: Improve doc readability

Most users will want to use a GPU so i suggest to make this tab visible by default:

image

Also, I find the 'all OS' vs. 'Repo installation' confusing, because these two categories are not independent (is: choosing repo installation is included in the 'all OS' case). Why not having 'Package installation (recommended)' and 'git installation'?

Other related issues:
- https://github.com/ivadomed/ivadomed/issues/996
- https://github.com/ivadomed/ivadomed/issues/728

kanishk16:
#### GitHub

#### PR contents

## Description
This PR updates docs in reference to #1000 and addresses issues related to installation readability. 

A preview to the docs - https://ivado-medical-imaging--1125.org.readthedocs.build/installation.html

## Linked issues
Addresses #728 and Resolves #1019


kanishk16:
Tagging @joshuacwnewton in reference to #861 as this would remove the dependency to install torch explicitly.

joshuacwnewton:
> Tagging @joshuacwnewton in reference to #861 as this would remove the dependency for downstream packages to install torch explicitly

Thank you for the tag! 

Looking this over quickly, this would be _absolutely amazing_ for SCT, because it would unblock a longstanding issue we've been struggling with for years: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1526

Because we would be able to specify a `requirements.txt` file that contains:

```
ivadomed[cpu]
```

We would be able to install a CPU version of PyTorch by default without having to use a `--find-links` link. 

(We had been trying to brainstorm ways to avoid `--find-links` for months/years, and we had similar ideas to you: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1526#issuecomment-812740662. But, it's _much_ nicer to do `ivadomed[cpu]` than ask _our_ users to do `spinalcordtoolbox[cpu]`, so thank you.)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "esbonio.sphinx.confDir": "G:\\ivadomed_workspace\\ivadomed\\docs\\source\\conf.py"
+}
```
dyt811, line 3: This probably shouldn't be committed.

kanishk16: Lol, Nice Catch!

hermancollin:
I can see some great work has been done here. For axondeepseg, this would potentially solve a problem we are encountering more and more: when we want to update the ivadomed version, since its installed with `pip`, using `conda env update`, `pip install -e .` or similar commands do not update the ivadomed package (or any other pip packages for that matter) and the user needs to reinstall the whole environment because `conda` can't deal with pip dependencies properly.

In https://github.com/axondeepseg/axondeepseg/pull/631, we tried parsing the *environment.yml* file to update ivadomed when there is a mismatch with the version currently installed in the environment, but the problem is that torch/torchvision also need to be updated if bumped, and we have to make sure internally that ivadomed and torch packages match. 

This would make the whole process much simpler as we would only need to worry about ivadomed's version. Also, if the user wants to use GPUs, he would simply need to `pip install ivadomed[gpu]` in his ADS venv, bypassing the need for `--find-links`, as mentioned by @joshuacwnewton.

shuaisong9 (approved):

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
-                pip install -e .
+                        pip install --upgrade pip
 
+                        pip install ivadomed[gpu]
 
-Step 3: Install ``torch`` and ``torchvision`` with CPU or GPU Support
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-    .. tabs::
+                .. tab:: Source Installation
```
dyt811, line 127: Could try ` .. group-tab:: Source Installation` and allow sync between the GPU and CPU page when switching.

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
-                    ``ivadomed`` requires CUDA11 to execute properly. If you have a nvidia GPU, try to look up its Cuda Compute Score `here <https://developer.nvidia.com/cuda-gpus>`__, which needs to be > 3.5 to support CUDA11. Then, make sure to upgrade to nvidia driver to be at least v450+ or newer.
+                        cd ivadomed
 
-                    You can use ``nvidia-smi`` in both Linux and Windows to check for driver CUDA Version listed at the top right of the output console. On Linux, simply type in ``nvidia-smi`` in any console to see the output. On windows, you will need to locate the `nvidia-smi.exe` tool by following the instructions on `this page <https://stackoverflow.com/a/57100016>`__.
+                        pip install -e .[gpu]
 
-                    If you have a compatible NVIDIA GPU that supports CUDA11, and you have a recent enough driver installed, then run the following command:
 
-                    .. code::
+                .. tab:: Contributor or Developer Installation
```
dyt811, line 141: Ditto.

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
+                        git clone https://github.com/ivadomed/ivadomed.git
 
-                    If you plan to run ``ivadomed`` on CPU only, install PyTorch per instructions provided below for your specific operating system:
+                        cd ivadomed
 
-                    .. tabs::
+                        pip install -e .[dev_gpu]
 
-                        .. tab:: Windows/Linux
+        .. tab:: CPU Support
```
dyt811, line 153: Ditto.

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
+            To verify the CUDA version, simply run ``nvcc -V`` in both Linux and Windows. The release refers to your current CUDA version.
 
-            Bleeding-edge developments are available on the project's master branch
-            on Github. Install ``ivadomed`` from source:
-
-            ::
+            .. tabs::
 
-                git clone https://github.com/ivadomed/ivadomed.git
+                .. tab:: Package Installation (Recommended)
```
dyt811, line 116: See below.

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
-                    .. tabs::
+                        pip install -e .[dev_gpu]
 
-                        .. tab:: Windows/Linux
+        .. tab:: CPU Support
 
-                            .. code::
+            .. tabs:: 
+                
+                .. tab:: Package Installation (Recommended)
```
dyt811, line 157: Ditto.

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
 
-                        .. tab:: Mac
+                        pip install --upgrade pip
 
-                            .. code::
+                        pip install ivadomed[cpu]
 
-                               pip install torch==1.8.0 torchvision==0.9.0 --find-links https://download.pytorch.org/whl/torch_stable.html
 
+                .. tab:: Source Installation
```
dyt811, line 168: Ditto.

docs/source/installation.rst
```diff
@@ -100,98 +100,96 @@ Step 1: Setup dedicated python environment
...
-            .. code::
+                        git clone https://github.com/ivadomed/ivadomed.git
 
-               pip install -r requirements_gpu.txt
+                        cd ivadomed
 
+                        pip install -e .[cpu]
 
 
+                .. tab:: Contributor or Developer Installation
```
dyt811, line 182: Ditto.

dyt811 (commented):
Looking good. Recommend  `.. group-tab::` usage. 
Please merge once 1000 is merged, which MIGHT change if we go the route of using neuropoly packaged specific torch CPU.

kousu:
From https://github.com/ivadomed/ivadomed/pull/1129#issuecomment-1139730419, I'd like to suggest drastically culling the docs. We should get rid of most of the tabs (including the ComputeCanada tab) and the mentions of Windows vs macOS vs Linux, and just say

> ```
> pip install ivadomed
> ```

(which downloads everything from PyPI) because that works out of the box for nearly everyone.

If needed, somewhere we can add two alternate instructions, which are hardware- but not OS-specific:

> If you don't have a GPU, you can save some time and space with:
>
> ```
> pip install --extra-index-url https://download.pytorch.org/whl/cpu ivadomed
> ```

(which, for Linux users, downloads a smaller build)

and

> If you have a GPU and find it's not working with ivadomed, try
>
> ```
> pip uninstall -y torch torchvision
> pip install ivadomed \
>    --extra-index-url https://download.pytorch.org/whl/cu110 \
>    --extra-index-url https://download.pytorch.org/whl/cu111 \
>    --extra-index-url https://download.pytorch.org/whl/cu112 \
>    --extra-index-url https://download.pytorch.org/whl/cu113 \
>    --extra-index-url https://download.pytorch.org/whl/cu114 \
>    --extra-index-url https://download.pytorch.org/whl/cu115
> ```

(which, for Windows users, enables the GPU; and for Linux users; but does nothing for macOS users)

Trying to predict people's GPU problems should be out of scope for the install docs. If people are having problems we can [try to help](https://github.com/ivadomed/ivadomed/discussions), but if we just hew close to (i.e. to what Facebook decided to publish as the default builds) I think there won't be that many problems in the first place.

----

As for ComputeCanada -- they have their own builds of torch [pre-populated](https://docs.alliancecan.ca/wiki/Available_Python_wheels) -- `torch==1.10.0` for python 3.10, `torch==1.10.0`, `torch==1.9.0` for python 3.9, and many more including `torch==1.8.1` for python 3.8. So there's three strategies we can follow (in parallel!) that don't involve adding irrelevant info to our public docs:

* Document internally that we need to remember to run ivadomed on python3.8 when using computecanada
* Ask computecanada to make sure `torch~=1.8.0` builds exist for all python versions -- it is the torch LTS release afterall, that's a reasonable request
* Work to make ivadomed depend on `torch>=1.8.1,<1.11` instead of `torch==1.8.1`

kousu:
And idk maybe drop the `pip install -e` instruction from this? That can go into https://github.com/ivadomed/ivadomed/blob/master/CONTRIBUTING.rst instead.

kousu:
jcohenadad [says](https://github.com/neuropoly/computers/issues/202#issuecomment-1139710695)

> testing GPU on windows is low priority

which to me means supporting Windows GPUs is also a low priority. If we can cover Linux CUDA11 people and Windows people in one instruction like I think we can I want to do that, but if it turns out we have to give separate Linux and Windows instructions then we should just not say anything about Windows. If someone really wants to use Windows and GPUs they can ask us for tips and we can send them to https://pytorch.org/get-started/locally/ or https://discuss.pytorch.org/.

kousu (changes requested):
A reminder to ourselves that this PR needs to get updated to match what we've learned from #1000/#1129

.gitignore
```diff
@@ -21,6 +21,9 @@ log.txt
 # Pycharm stuff
 .idea
 
+# VSCode stuff
+.vscode
+
```
kousu, line 26: This isn't really part of this PR, is it?
 
```suggestion
```

kanishk16: Just updated in this PR itself as I was using VSCode

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
+            Ampere-based GPUs (with a Compute Capability of 8.x) natively
+            support CUDA>=11.1. Although CUDA 11.1 is backward compatible
+            but CUDA 10.2 is preferred for older GPUs.
+
+            CUDA 10.2 and CUDA 11.1 require an NVIDIA driver version >=440.33 
+            and >=450 respectively as indicated `here <https://docs.nvidia.com/deploy/cuda-compatibility/index.html#minor-version-compatibility>`__.
+            Please make sure to upgrade to the minimum NVIDIA driver version 
+            requirements for the respective CUDA builds. To verify the NVIDIA
+            driver version, just run the command `nvidia-smi` and you'll find 
+            your current driver version on the top.
```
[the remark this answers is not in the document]
kousu: There's also the `pip.conf` solution: instead of "please run this command" it could be "please create this file before installation"


```
# /etc/pip.conf or ~/.config/pip/pip.conf or C:\ProgramData\pip\pip.ini or %APPDATA%\pip\pip.ini
[install]
extra-index-url =
   https://download.pytorch.org/whl/cu111
   https://download.pytorch.org/whl/cu113
   https://download.pytorch.org/whl/cu115
```

That's essentially instructing people to set a global variable on their OS, but I think this is one of the rare cases where that's the right thing to do: a system has fixed hardware (usually), so every pretty much every package they use with their GPUs will need the same patch.

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
+        
+        .. tab:: NVIDIA GPU Support
+
+            ``ivadomed`` requires a minimum PyTorch version of either
+            1.8.1 or 1.8.2 which supports CUDA 10.2 and CUDA 11.1 builds
+            by default. 
+            
+            Ampere-based GPUs (with a Compute Capability of 8.x) natively
+            support CUDA>=11.1. Although CUDA 11.1 is backward compatible
+            but CUDA 10.2 is preferred for older GPUs.
```
kanishk16, line 116: Interesting updates #1141 & the upstream torch thread, at least for torch ~=1.8.1

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
+                            ::
 
-                    ``ivadomed`` requires CUDA11 to execute properly. If you have a nvidia GPU, try to look up its Cuda Compute Score `here <https://developer.nvidia.com/cuda-gpus>`__, which needs to be > 3.5 to support CUDA11. Then, make sure to upgrade to nvidia driver to be at least v450+ or newer.
+                                git clone https://github.com/ivadomed/ivadomed.git
 
-                    You can use ``nvidia-smi`` in both Linux and Windows to check for driver CUDA Version listed at the top right of the output console. On Linux, simply type in ``nvidia-smi`` in any console to see the output. On windows, you will need to locate the `nvidia-smi.exe` tool by following the instructions on `this page <https://stackoverflow.com/a/57100016>`__.
+                                cd ivadomed
 
-                    If you have a compatible NVIDIA GPU that supports CUDA11, and you have a recent enough driver installed, then run the following command:
+                                pip install -e . --extra-index-url https://download.pytorch.org/whl/cu102
```
kousu, line 201: My suggestion above drops Windows support entirely, because @jcohenadad [said we should](https://github.com/neuropoly/computers/issues/202#issuecomment-1139710695), but it would be good if the trick to getting GPUs working on Windows was saved _somewhere_.

I'm not sure where though. Maybe it could go in CONTRIBUTING.md along with the rest of this section, where we aren't as constrained by having to write for a wider audience and can dump reams of technical details with less care?

Or maybe we could tuck it into the GPU section in a `tabs:::` afterall and just ignore @jcohenadad's ask -- I wrote in my suggestion "Not officially supported nor tested", but at the moment GPU support on Linux isn't tested either, except by our own research, so maybe it's not so much of a stretch to just tuck in the Windows case too and get around to testing it [mañana](https://www.wordnik.com/words/ma%C3%B1ana).

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
 
-Developer-only Installation Steps
-+++++++++++++++++++++++++++++++++
+                                git clone https://github.com/ivadomed/ivadomed.git
 
-    The additional steps below are only necessary for contributors to the ``ivadomed`` project.
+                                cd ivadomed
 
-    The ``pre-commit`` package is used to enforce a size limit on committed files. The ``requirements_dev.txt`` also contain additional dependencies related to documentation building and testing.
+                                pip install -e .[dev]
```
kousu, line 365: The diff makes it hard to comment on the right part here, but again, I'd put all this in CONTRIBUTING.md.

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
+                
+                .. tab:: Package Installation (Recommended)
+
+                    .. tabs::
+
+                        .. tab:: Linux
+
+                            ::
+
+                                pip install ivadomed --extra-index-url https://download.pytorch.org/whl/cpu
```
kousu, line 280: I'm personally invested in this one line, because we [use it in SCT](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3790), but maybe it's better to drop it. The torch people think linux people should waste their space, so let's just roll with that?

Alternately, this could fit in a `tabs:::` above, next to the GPU section, maybe something like:

> Installation | CPU-based Installation | GPU-based Installation (Windows) | Ampere GPU-based Installation

So the 80% case is the first tab, which wastes some Linux users' space who haven't read closely but that's not the end of the world, the 15% case is the people who read the CPU instructions, another say ~5% of people are bold explorers using GPUs on Windows, and the <1% case is the people rich enough to have Ampere GPUs.

And the only difference between each tab is

a. Instructing people how to decide if that case applies to them
b. Giving different `extra-index-url` lines for their `pip.conf`

kousu (changes requested):
I want to see this doc get down to under 1 page. It should fit on a screen. It mostly comes down to telling people to run a single command, `pip install ivadomed` so surely we can streamline it.

The rest of the text is just special cases:

* `pip install --extra-index-url https://download.pytorch.org/whl/cpu ivadomed` ((linux) CPU)
* `pip install --extra-index-url https://download.pytorch.org/whl/cu111  --extra-index-url https://download.pytorch.org/whl/cu113  --extra-index-url https://download.pytorch.org/whl/cu115 ivadomed` (Ampere GPUs)
* `pip install --extra-index-url https://download.pytorch.org/whl/cu102 ivadomed` (Windows GPUs)

If we can include these three commands somewhere on the page, I think that's great, but I don't think each deserves a whole `# heading`. They should be after thoughts, something we can link someone in one of those special cases to in the rare case they ask for help.

And, I know you're in the Windows case @kanishk16, so this probably sucks to hear, but from [deciding not to support Windows](https://github.com/neuropoly/computers/issues/202#issuecomment-1139710695) and also the fact that the Windows case can be rolled into the Ampere case, maybe we should just skip that last special case entirely.

kousu:
> While writing an exhaustive draft, I found this [blog](https://arnon.dk/matching-sm-architectures-arch-and-gencode-for-various-nvidia-cards/) an interesting read.

This post is excellent! It really really clarifies the lower levels.

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
+
+                            and, with CUDA 11.1:
+
+                            ::
+
+                                git clone https://github.com/ivadomed/ivadomed.git
+
+                                cd ivadomed
+
+                                pip install -e . --extra-index-url https://download.pytorch.org/whl/cu111
```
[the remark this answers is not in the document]
dyt811: Good point. I like this approach of keeping things simple and ask others to do extra work. Why we keep things complicated and hard to update when we can ask the non-default path users to be doing more work right?

docs/source/installation.rst
dyt811: GPU testing will resume outside of the existing setup for sanity check reasons. It will be a completely independent and private process though as our existing GitHub workflow is not flexible enough to accommodate it and cannot securely make it integrate with GitHub in its current implementation.  I will open bugs on issue found manually for now.

dyt811:
I agree with @kousu's recommendation to strip out the `dev` installation section entirely into contributing.md or the like. Much easier to keep the doc simpler and primary user case facing. 

Can't we merge these two Package Installation tabs, if the installation are now pretty much the same across os? 
image

image


Similarly for the Source Installation tabs, 

image

image

docs/source/installation.rst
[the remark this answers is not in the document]
kanishk16: > My understanding is that it's not a question of the GPUs supporting CUDA, it's if CUDA supports the GPUs,

Agreed... looks like it didn't come through as thought, but your suggestion looks apt.

> the reason for preferring 10.2 is simply that it's the stable branch at the moment (and is for example [more optimized](https://github.com/ivadomed/ivadomed/issues/1141))

I'd guess other CUDA versions (11.1, etc) are stable as well, but it's just that this jump didn't propagate as expected until torch1.11 for older hardware. https://github.com/pytorch/pytorch/issues/47908#issuecomment-1148387263 So, skipping the reason should be good... What about?

```suggestion
            Ampere-based GPUs (with a Compute Capability of 8.x) only work with
            CUDA>=11.1. Although CUDA 11.1 is backward compatible with older hardware,
            CUDA 10.2 is preferred if available.
```

docs/source/installation.rst
```diff
@@ -100,98 +100,270 @@ Step 1: Setup dedicated python environment
...
+                            with CUDA 11.1:
 
-                               pip install torch==1.8.0+cpu torchvision==0.9.0+cpu --find-links https://download.pytorch.org/whl/torch_stable.html
+                            ::
+
+                                git clone https://github.com/ivadomed/ivadomed.git
+
+                                cd ivadomed
+
+                                pip install -e .[dev] --extra-index-url https://download.pytorch.org/whl/cu111
```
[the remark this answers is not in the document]
kanishk16: Absolutely... At that point, I was trying to consolidate everything.

docs/source/installation.rst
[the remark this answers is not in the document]
kanishk16: > I also wonder if even mentioning "10.2" will mislead people. 

Maybe or maybe not

> It's not really CUDA 10.2 that's important, but rather just using the default `torch` that's on PyPI, because that's the best-tested (and evidently most optimized) version of `torch` out there. Similarly the "minimum version" of torch is encoded in the .whl, so that's also something users shouldn't need to think about. So maybe:

Agreed

> (I replaced `nvidia-smi` with `cat` because `nvidia-smi` isn't installed with the drivers; on Ubuntu it comes [with `nvidia-cuda-toolkit`](https://github.com/neuropoly/computers/blob/99c8bf5c24365f7eb1f71d252011f605ead425c7/ansible/roles/neuropoly.gpu/tasks/main.yml#L19) but not with `nvidia-driver-XXX`; 

Thank you for adding this... I was sceptical about this as well

> I suppose for someone who was trying to get things working on Windows, this is a step back, since there maybe `nvidia-smi` does come? But on Windows someone can probably look in Control Panel to find the driver version?)
Yuss yuss

> `ivadomed-test-gpus` is a new suggestion, suggesting we make sure people can actually. We both know it could be `python -c 'import torch; print(torch.cuda.is_available())`, but maybe we should wrap it in a small helper script to prove to people that ivadomed is actually talking to the GPUs? Notice that except for "download.pytorch.org" we don't need to mention torch at all -- it's just a dependency, it's not what people are looking to learn about from ivadomed so the more we avoid opening the rabbit hole the better I think. But if we did ship such a script we'd have to maitnain it; so is there an obvious command that already exists in ivadomed that reports whether it's using GPUs or not that people could run on some small test data, just to verify everything?
I agree... I'm thinking something along the lines what ADS has ([Quick Test](https://axondeepseg.readthedocs.io/en/latest/documentation.html#quick-test)). It should come after the installation step... let's skip this for now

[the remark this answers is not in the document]
kanishk16: > There's also the `pip.conf` solution: instead of "please run this command" it could be "please create this file before installation"
> 
> ```
> # /etc/pip.conf or ~/.config/pip/pip.conf or C:\ProgramData\pip\pip.ini or %APPDATA%\pip\pip.ini
> [install]
> extra-index-url =
>    https://download.pytorch.org/whl/cu111
>    https://download.pytorch.org/whl/cu113
>    https://download.pytorch.org/whl/cu115
> ```
> 
> That's essentially instructing people to set a global variable on their OS, but I think this is one of the rare cases where that's the right thing to do: a system has fixed hardware (usually), so every pretty much every package they use with their GPUs will need the same patch.

I'm a little hesitant towards this solution bcz let's say someone follows these instructions and later on forgets this setting, and perhaps the torch installation evolves, then it'd be harder for them to debug.

docs/source/installation.rst
kanishk16: Flattening the tab hierarchy would result in repetition of some GPU instructions.

kanishk16:
> Can't we merge these two Package Installation tabs, if the installation are now pretty much the same across os? image
> 
> image

@dyt811 Notice the subtle difference for CUDA 10.2 in Linux & Windows tab. Anyway, we've moved the windows bit to developer section as suggested by Nick

.gitignore
kousu: I'm going to take this out of this PR and put it in as a separate commit to `master`.

environment.yml
```diff
@@ -1,9 +1,6 @@
 name: ivadomed_env
 channels:
   - conda-forge
-  - pytorch
 dependencies:
   - python>=3.7, < 3.9
   - pip
-  - pip:
-      - --requirement requirements.txt
```
kousu, line 9: This PR is supposed to be about docs, but this changes the actual code's behaviour subtly.

Do we even want this? It's probably better to use the pytorch conda channel if we have it available isn't it? Or, at worst, it's irrelevant, because `pip --extra-index-url ...` will override/ignore conda's repos?

kousu (changes requested):
I'm sorry @kanishk16, but I still think the docs could be more streamlined. I really think the page should start with

```
pip install ivadomed
```

and leave it at that for most people.

People will get lost with jargon ("what's an Ampere?" "what's a CUDA?" "why do I care?"). We need some amount of that documented, but we shouldn't front-load it.

If we're going to mention CUDA, we need to explain enough that someone can figure out if they need to care about what that is and why its version matters to them.

My suggestion with regard to ditching Windows from months ago was handwaving at a shortcut: I was thinking maybe we could get away with saying "and if your GPUs don't work, take these CUDA 11 builds and call me in the morning (because I'm a doctor)". But I think if we're going to be explicit about Ampere, it makes more sense to be explicit about Windows too.

Plus, in the meantime, I've discovered that torch 1.12 has ended up on PyPI, which has started taking precedence over the files in https://download.pytorch.org/whl/cu111/torch, which only goes up to `torch==1.10.2`. Now, in #1129 we pinned `torch~=1.8.1`, ~~so that should be irrelevant, but it looks like we haven't released that yet because the current version of `ivadomed` on PyPI is still `torch>=1.5.0`, so `pip`'s started preferring PyPI's `torch==1.12.0`, skipping the CUDA 11 version entirely~~ and this worked, but I was trying to install on Python 3.10, which was causing `pip` to fallback to an older ivadomed, one that didn't specify `python_requires`. In any case, to be safe, I think we need to use the trick I [applied to `romane`](https://github.com/neuropoly/computers/pull/367): use every `--extra-index-url` they offer, so that no matter which combination of pinning and python version and OS you're on there should be a CUDA 11 build that covers your case. I [mentioned this in passing before](https://github.com/ivadomed/ivadomed/pull/1125#issuecomment-1187320762) but I guess I didn't make clear why it was important; and anyway this version includes `cu116` (which has come out since we started working on this) and `cu117` and `cu118` for future-proofing. 

Rather than nitpick every little thing, I have written a PR-into-your-PR at https://github.com/ivadomed/ivadomed/pull/1184. You can see it in action at

* https://ivado-medical-imaging--1184.org.readthedocs.build/installation.html
* https://ivado-medical-imaging--1184.org.readthedocs.build/contributing.html

<details><summary>Installation</summary>

Screenshot 2022-07-19 at 20-50-18 Installation — ivadomed documentation

</details>

<details><summary>Contributing</summary>

Screenshot 2022-07-19 at 20-50-33 Contributing to ivadomed — ivadomed documentation
 
</details>

I:

1. Moved most of the venv stuff over to Contributing -- real software should be safe to install in people's existing environments
  1. Kept a remnant of it in a `.. note::`, which gestures at it. 
3. Dropped the irrelevant-to-99%-of-the-planet Compute Canada complaint
4. Expanded the section on verifying GPUs to cover explaining why you would care ("acceleration") and how to (`torch.cuda.is_available()`)
5. Dropped the source installation; if we want people to run bleeding edge code we should should either be releasing `rc`s aka Release Candidates aka `pip install --pre`, or we should just be pointing people at the developer install -- we probably will mostly point people at the developer install.
6. Replaced the marketing jargon "Ampere GPU" with the phrase "too new GPU"
7. Dropped `conda env create --name` because environment.yml already provides a name
8. Deduplicated the GPU instructions in Contributing by replacing them with a `.. note::` and a link back to the GPU instructions on Installation
9. Split the `.. group-tab::`s into "Linux"/"Windows"/"macOS"

    This made some for a bit of copy-pasting, but it has the nice benefit that the selected tab syncs everywhere, _even between the two pages_. And "macOS" is the official way to spell OS X nowadays, so I thought it was important to standardize on that spelling.
10. Replaced some `.. tab::`s with section headers; I'm not sure if we're on the same page about this or not, but I think the benefit of section headings are

    1. They're less deeply nested, which makes them easier to read
    2. They're linkable, so we can send people to https://ivado-medical-imaging--1184.org.readthedocs.build/installation.html#cpu-only-support or https://ivado-medical-imaging--1184.org.readthedocs.build/installation.html#cuda-11 if they ask. We can't do that with tabs.
3. Probably some other things.

So, that's the direction I think we should be going with this. But like I said in the PR, I submitted it for your consideration and to make it clearer what I think should be done on this branch, but it's not set in stone.

kanishk16:
Thanks @kousu for reviewing it extensively. I might have a slightly different take on starting the installation directly with `pip install ivadomed`. This is more or less my perspective in general - 

- My primary concern is missing the wood for the trees or perhaps creating a bit of confusion due to difference in behaviour on different OSes. More so, once M1 GPU acceleration performance is at par with NVIDIA GPUs (irrelevant as of now but soon to be).

You and me already know the default behaviour of pip & torch on Linux, but abstracting this info might be dangerous in case they run into some upstream issues (e.g.: performance), assuming related to CUDA & torch, then it might take them longer to figure it out.

> People will get lost with jargon ("what's an Ampere?" "what's a CUDA?" "why do I care?"). We need some amount of that documented, but we shouldn't front-load it.
>
> If we're going to mention CUDA, we need to explain enough that someone can figure out if they need to care about what that is and why its version matters to them.

I re-read the docs. CUDA runtime might be unclear, but it's sufficient to get an idea what CUDA runtime to choose. In case they're curious & want to dig in, they could always google it up.

> My suggestion with regard to ditching Windows from months ago was handwaving at a shortcut: I was thinking maybe we could get away with saying "and if your GPUs don't work, take these CUDA 11 builds and call me in the morning (because I'm a doctor)". But I think if we're going to be explicit about Ampere, it makes more sense to be explicit about Windows too.

According to the past discussions, I'm in favour of ditching Windows completely. Windows scenario is a highly unlikely one to be encountered. 

I'm yet to address other comments, and I believe it would be better if more people review this to get a perspective how are docs actually coming out to be.

kanishk16:
> There's also the `pip.conf` solution: instead of "please run this command" it could be "please create this file before installation"
> 
> ```
> # /etc/pip.conf or ~/.config/pip/pip.conf or C:\ProgramData\pip\pip.ini or %APPDATA%\pip\pip.ini
> [install]
> extra-index-url =
>    https://download.pytorch.org/whl/cu111
>    https://download.pytorch.org/whl/cu113
>    https://download.pytorch.org/whl/cu115
> ```
> 
> That's essentially instructing people to set a global variable on their OS, but I think this is one of the rare cases where that's the right thing to do: a system has fixed hardware (usually), so every pretty much every package they use with their GPUs will need the same patch.

image

@kousu and to the reviewers, I'm uncomfortable with the above propositions to the user. 

(Why?) According to the [docs](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#minor-version-compatibility), "From CUDA 11 onwards, applications compiled with a CUDA Toolkit release from within a CUDA major release family can run, **with limited feature-set**, on systems having at least the minimum required driver version as indicated below. **This minimum required driver can be different from the driver packaged with the CUDA Toolkit but should belong to the same major release**."  

Now, simplifying if I understood this correctly, the Linux x86_64 Minimum Required Driver Version for CUDA 11.x is >= 450.80.02 whereas the Linux x86_64 Minimum Required Driver Version for CUDA 11.3 is >=465.19.01 (reference to [docs](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html)). This essentially boils down to - even if I have an NVIDIA Driver Version of 450.80.02 it is possible to install and run pytorch compiled with CUDA 11.3 runtime but with limited feature-set. And, it is possible to not figure out some optimizations made in CUDA 11.3 for certain algos but everything appearing almost fine.

kousu:
I'm also concerned about missing the forest for the trees. But I think the forest in this case is `pip install ivadomed` and troubleshooting performance/hardware issues are the trees.

Another way to think of it is the 80%-20% rule: 80% of people are safe to just `pip install ivadomed`. The other 20% can use `--extra-index-url https://download.pytorch.org/whl/cu102` (or whatever) -- even putting that in their pip.conf -- but they are not the people we should be prioritizing.

> @kousu and to the reviewers, I'm uncomfortable with the above propositions to the user. [..]
> even if I have an NVIDIA Driver Version of 450.80.02 it is possible to install and run pytorch compiled with CUDA 11.3 runtime but with limited feature-set.  [..]

I'm sorry, I don't really understand. I think maybe we actually agree? But I'm not sure.

Are you saying that recommending CUDA 11 is bad? My read of what you linked isn't saying that CUDA 11 is less efficient, just that it is less efficient if used with nvidia-driver 440 or nvidia-driver 450; which is different than with CUDA 10, which (I'm guessing at the exact numbers here) didn't work at all with nvidia-driver 330. But can't we just assume people will be running the latest driver? On our local systems [I have us running nvidia-driver 510](https://github.com/neuropoly/computers/pull/349) and everything is stable and fine.

Taking a step back to look at the forest again: I know that debugging tricky CUDA issues is hard. I have [done it a lot at this point](https://github.com/neuropoly/computers/issues/76). That's exactly why I think we should bury it in `# Special Cases`; it should be rare (i.e. <20%) that anyone actually uses `--extra-index-url`.

As of July 2022, the CUDA 11 builds should only be used for people with Ampere ([Compute Capability](https://developer.nvidia.com/cuda-gpus#compute) >= 8.0) cards, which is what I meant by "too new" -- and those people, frankly, are rich and can afford to pay for their own tech support or losing a bit of performance. The default path, doing `pip install ivadomed`, will mean people on Linux will end up with the CUDA 10 builds which are

----

Anyway, I'll defer to the opinion of someone actually on the ivadomed team. My main blocker is that in this version, the basic `pip install ivadomed` command is buried and I think it needs to be way more prominent. @dyt811, @lifetheater57, can you give a read over our two versions of the docs and give us a tie-breaker vote? Or help us decide what to merge from each?

* https://ivado-medical-imaging--1125.org.readthedocs.build/installation.html
* https://ivado-medical-imaging--1125.org.readthedocs.build/contributing.html

* https://ivado-medical-imaging--1184.org.readthedocs.build/installation.html
* https://ivado-medical-imaging--1184.org.readthedocs.build/contributing.html

lifetheater57:
Maybe I'll understand why it would be required to have CUDA >= 11.0 when I compare your docs. However, in the meantime, I think requiring to have compute capability >= 8.0 is too early.

kousu:
> Maybe I'll understand why it would be required to have CUDA >= 11.0 when I compare your docs. However, in the meantime, I think requiring to have compute capability >= 8.0 is too early.

It's not. We used to be confused: `romane.neuro.polymtl.ca` has compute capability == 8.0 so ivadomed ended up with a requirement for CUDA >= 11. #1000, which became #1129 , loosened that requirement. It was artificial and unneccessary. This PR is the followup to #1129 to make the docs reflect the code. So, almost the opposite, even.

I was the one who initially put in the CUDA 11 requirement as a temporary workaround for `romane`, so now I'm trying to clean up the mess 😬

LuluSzutu:
I read the Ivadomed Installation Document in both PR1184 and PR1125. I feel it is better if we can combine these two versions.
In PR1125, I like the way of organize the content, and separate into Step1, Step2…..
Also the part of ‘ setup virtual environment for Ivadomed’ is very good to a new pyhton user (like me). So I think, for sure, we should keep this part. But this part is missing in PR1184.
In PR1184, I like the detail explanation for GPU support. Also includes the command for checking GPU and NVIDIA versions.
By merging these two version of installation documents, here is my proposal:
Dependencies:
OS Support: ivadomed officially supports GPU acceleration only on Linux and CPU on Linux, Windows and MacOS.
Python: ivadomed support python 3.5-3.10
              Checking python version:
              Linux, windows, macOS
GPU support (optional): To verify GPU hardware is detected, use:
               Linux, window, macOS
                         To verify NVIDIA driver, use:
               Linux, window, macOS
               Ivadomed support CUDA10, CUDA 11 …….
Installation
Step 1. Setup Virtual Environment
Step 2. Install Ivadomed
            For CPU-only Support
                   Linux, window, macOS
            For GPU support (Linux Only)
                   pip install ivadomed
             Troubleshoot Problems:
             Q1: I have CUDA 11.x, can I still use Ivadomed?
             Answer: *******
             Q1: I have GPU support, but after running ‘pip install ivadomed’, my ‘torch.cuda.is available’ is still ‘False’ why?
             Answer: *****

lifetheater57:
Here is my review for PRs #1125 and #1184. 

Note that I am commenting without checking most of the previous discussions to have a cold take on it. Also, it occurs to be quite aligned with @LuluSzutu [suggestion](https://github.com/ivadomed/ivadomed/pull/1125#issuecomment-1198462784) which I didn't see before writing my own review. 

I first begin with general comments that apply to both PRs and that my target issues present before these. Then, I give PR-specific comments for each PR like if we were to keep each of them as a whole. Finally, I propose how I would combine the PRs together.

In both cases: 
- `NeuroPoly Contributing Guidelines` link is broken.
- Why not use `ivadomed[cpu]` for the cpu version?
- The Copyrights got stuck in 2020.

PR #1125:
- Technically, the instruction should be to check whether a compatible Python version is installed on the system and not whether the _default_ Python has the right version (i.e. check if python3.7, python3.8 or python3.9 is available rather than checking if the version of python is in {3.7, 3.8, 3.9}).
- I guess it doesn't ship with _both_ CUDA 10.2 _and_ CUDA 11.1.

PR #1184:
- The main installation instructions are nice, but we need to be sure to test it very well since the steps to chose the right version regarding the GPU are abstracted.
- Also, do we aim to people that don't know how and why to setup a `venv`, if yes, then we should bring back the instructions about the `venv`. Also, if python needs to have a specific range of versions, then might as well give the instruction how to create the `venv` with one of these versions.
- Instead of saying "This should work with most combinations of OSes and pythons.", we should have a list of the tested/supported configurations.

Overall:

Each bullet is a section I would see on the installation page.
- Supported OS:
    - I feel like we should keep that section. However, I would suggest to have a list of what configuration we tested, what configuration we expect to work and what configuration we know don't work. 
    - If there is too much content, then I would go for a summary of what is supported or not and a complete section in a by-default collapsed section.
- Virtual environment:
    - We should still encourage the user to create a virtual environment since it is a great practice that prevents a lot of headaches.
    - I would write something like "You are strongly recommended to setup a virtual environment for `ivadomed` using Python>=3.7,<3.10", then have the specific instructions collapse by default.
- Installation:
    - I would by-default install the best version of PyTorch (to be defined) given the system configuration (if we can know it, i.e. like in PR #1184)
    - A (collapsed?) subsection "Alternative installations" explaining how to install a specific version if desired (cpu-only or a specific CUDA version) and why we could want to do so.
- Troubleshooting:
    - This section addresses the installation/runtime issues
    - I would put PR #1184 "GPU support" section here under something like "ivadomed doesn't train using GPU". 
        - This item would have to include some of the information under the "CUDA 11" subsection but may have to be changed depending on the way the "Installation" section is written (i.e. if there is some automation or not).
        - This item should also include a description of the limitations pertaining to compute capabilities 8.x (i.e. CUDA >= 11.1) found in PR #1125.
        - It would also be great to include the information related to the CUDA driver version required given the different CUDA version used by PyTorch.

PR #1126, opened 2022-05-04
Utilize System tmp Folder for Testing

### LINKED ISSUE #1099: Unit tests should use /temp instead of local source folder path directory

* During the development of Singularity images of ivadomed, we noticed that we are unable to carry out any unit tests. 
This was caused by the facts that singularity images set the source repository to "read only".

* When we trigger `pytest`, which typically happens on the ivadomed source code directory, it attempts to create a `/tmp` folder to start downloading the unit tests and execute them. 

* This is normally okay on dev environment (since we have write access to source code folder) or production environment (where user almost never run unit tests anyway). However, in the case of testing singularity contained ivadomed, it is a severe road block and those folders are read only. 

* To remedy this problem, the ivadomed needs to leverage a system independent location that is used to 1) download the training examples 2) write out intermediate results there. 3) finish executing the entire unit tests with this flexibility in mind. 

### Alternatives

* We can potentially make the source folder somehow writable during the singularity image building process. 
* We can also allow user to specify a potential path to carry out the download and caching of unit tests data. 

### Additional context

* This feature is crucial for validating successful containerization of the ivadomed in read only environment like Singularity container

cakester:
#### GitHub

#### PR contents

## Description

using system tmp/temp folder for testing instead of local tmp folder

## Linked issues

resolves #1099


testing/unit_tests/t_utils.py
```diff
@@ -45,8 +45,9 @@ def create_tmp_dir(copy_data_testing_dir=True):
             into the ``tmp`` folder.
     """
     remove_tmp_dir()
-    Path(path_temp).mkdir()
+    Path(path_temp).mkdir(parents=False, exist_ok=True)
     if Path(path_data_testing_source).exists() and copy_data_testing_dir:
+        shutil.rmtree(path_data_testing_tmp, ignore_errors=True)
```
cakester, line 50: note with python 3.8+, we can have exist_ok for copytree, so this line would not be needed

testing/unit_tests/test_testing.py
```diff
@@ -207,9 +207,10 @@ def test_inference_2d_microscopy(download_data_testing_test_files, transforms_di
                                                    ofolder=str(__output_dir__),
                                                    cuda_available=cuda_available)
 
-    assert len([x for x in __output_dir__.iterdir() if x.name.endswith(".nii.gz")]) == len(test_lst)
+    assert len([x for x in __output_dir__.iterdir() if x.name.endswith(".nii.gz")]) == len(test_lst) + 1
```
cakester, line 210: note there seem to be an extra file stored in the folders when I use tmp folder instead of local tmp folder, don't know why

testing/common_testing_util.py
```diff
@@ -76,4 +84,17 @@ def remove_dataset(dataset: str = 'data_testing'):
...
+
+    if count > 100:
+        num_to_delete = count - 100
+        lst = sorted(path_temp.iterdir(), key=os.path.getctime)
+        i = 0
+        while num_to_delete:
+            if lst[i].is_dir() and lst[i].stem.startswith("tmp"):
+                shutil.rmtree(lst[i], ignore_errors=True)
+                num_to_delete -= 1
+            i += 1
```
dyt811, line 100: Note that now you are nuking entire /tmp whenever ivadomed runs a unit test.... side effect is potentially tremendous for all other linux operations that leverage /tmp.  This is a potential undesirable side effect.

testing/common_testing_util.py
```diff
@@ -22,18 +23,25 @@
...
 
-path_temp: str = str(path_repo_root / "tmp")
+if not sys.platform.startswith('win'):
+    path_temp: Path = Path("/tmp")
+else:
+    path_temp: Path = Path(Path.home().parts[0], 'temp')
+
+timestr = time.strftime("%Y%m%d-%H%M%S")
+tmptimestr = "tmp" + timestr
+new_path_temp = path_temp / tmptimestr
```
dyt811, line 33: I think it is a good idea to add time stamp to this, as it prevents accidentally reusing older data.

testing/common_testing_util.py
```diff
@@ -22,18 +23,25 @@
 
 path_repo_root: Path = Path(__file__).parents[FOLDER_LEVELS_FROM_ROOT].absolute()
 
-path_temp: str = str(path_repo_root / "tmp")
+if not sys.platform.startswith('win'):
+    path_temp: Path = Path("/tmp")
+else:
+    path_temp: Path = Path(Path.home().parts[0], 'temp')
```
dyt811, line 29: What about using MKDTemp instead? https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

Those will persist in respective OS temp and stick around until deleted by user specifically.

testing/unit_tests/t_utils.py
dyt811: Yup. we are still supporting 3.7 sadly. Definitely removable in the future.

testing/unit_tests/test_testing.py
```diff
@@ -207,9 +207,10 @@ def test_inference_2d_microscopy(download_data_testing_test_files, transforms_di
...
     assert len([x for x in __output_dir__.iterdir() if x.name.endswith(".png")]) == 2*len(test_lst)
 
 
 def teardown_function():
-    remove_tmp_dir()
+    pass
```
dyt811, line 215: Makes sense now that we do not technically need to rigorously remove all `tmp` files.  However, we probably should stop teardown in ALL test functions?

testing/common_testing_util.py
kousu: Even better [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory):

```suggestion
with tempfile.TemporaryDirectory() as path_temp:
```

this form deletes itself when done.

dyt811: Yeah... one of the goal was to keep a limited number of these directories for comparisons and tracking unit tests outputs history.

PR #1131, opened 2022-05-17
Update publish_package.yml

kousu:
I've used this script in some of my own projects, and this reintegrates some of the improvements I've made as I've used it in practice:

- bump dependencies
- drop building on python 3.6 (the package is version agnostic, so this is just confusing)
- drop the workflow_dispatch because it's glitchy in some cases; use the Releases page instead to create a release
- add prerelease tagging (using a tag with an 'rc' suffix would mark it as pre-release on PyPI, this makes it also so on the releases page on GitHub)

### DIFF
--- a/.github/workflows/publish_package.yml
+++ b/.github/workflows/publish_package.yml
@@ -7,12 +7,6 @@ on:
   # publish from the Releases page:
   release:
     types: [published]
-  # publish from the Actions page:
-  workflow_dispatch:
-    inputs:
-      version:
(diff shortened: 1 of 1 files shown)


kousu:
This is a low-priority PR, but I just wanted to bring back the improvements I've foraged.

It might change your release workflow if you've been used to using the Actions page; but just get used to using the Releases page instead.

dyt811 (approved):
Going to be interesting test this out. Can't exactly test locally.

PR #1133, opened 2022-05-17
Use headless ("Agg") matplotlib backend in CI

### LINKED ISSUE #1132: CI failure: "RuntimeError: tk.h version (8.6) doesn't match libtk.a version (8.5)"
### Current behavior

CI is failing on {macOS} x {Python 3.7, Python 3.8}. Python3.6 and 3.9 are unaffected.

### Steps to reproduce

Run [CI](.github/workflows/run_tests.yml) on almost any branch currently and it will fail. Example failures:

* `master`: https://github.com/ivadomed/ivadomed/actions/runs/2340863632
* https://github.com/ivadomed/ivadomed/pull/1125: https://github.com/ivadomed/ivadomed/actions/runs/2295180486
  * Interestingly, this one only failed on 3.7, too
* https://github.com/ivadomed/ivadomed/pull/1126: https://github.com/ivadomed/ivadomed/actions/runs/2276043597
  * Interestingly, this one only failed on 3.7
* https://github.com/ivadomed/ivadomed/pull/1127: https://github.com/ivadomed/ivadomed/actions/runs/2321234595
* https://github.com/ivadomed/ivadomed/pull/1128: https://github.com/ivadomed/ivadomed/actions/runs/2327341106
* https://github.com/ivadomed/ivadomed/pull/1129: https://github.com/ivadomed/ivadomed/actions/runs/2340867212
* https://github.com/ivadomed/ivadomed/pull/1131: https://github.com/ivadomed/ivadomed/actions/runs/2341453544
 
Typical pattern of failures:

Screenshot 2022-05-17 at 18-15-15 Update publish_package yml · ivadomed_ivadomed@c4c425b

## Linked Issues

* Being tracked by GitHub staff at https://github.com/actions/setup-python/issues/402

kousu:
#### GitHub

#### PR contents

## Description

Force Agg matplotlib backend to avoid problems in CI. CI should be headless anyway, there's no need to try the other six backends.

## Linked issues

Fixes #1132, which is the most recent symptom, but it's not the only tricky thing that can come up when mixing and matching rendering backends and headless test machines.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -19,6 +19,11 @@ env:
   # (TravisCI quietly defined this on all their platforms, but we have to give it manually on GithubCI.)
   DEBIAN_FRONTEND: 'noninteractive'
   HDF5_USE_FILE_LOCKING: 'FALSE'
+  # Skip to the headless matplotlib renderer, which is less
+  # bug-prone in the constrained environment of CI
+  # Tip from a matplotlib dev: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3388#issuecomment-846091012
+  # Ref: https://matplotlib.org/stable/users/explain/backends.html
+  MPLBACKEND: 'Agg'
 
 jobs:
   ultra_matrix_test:
--- a/ivadomed/scripts/visualize_and_compare_testing_models.py
+++ b/ivadomed/scripts/visualize_and_compare_testing_models.py
@@ -9,6 +9,7 @@
 ###########################################################################################################
 
 import matplotlib
+from matplotlib import pyplot as plt
 import pandas as pd
 import numpy as np
 import itertools
--- a/ivadomed/scripts/visualize_and_compare_testing_models.py
+++ b/ivadomed/scripts/visualize_and_compare_testing_models.py
@@ -21,21 +21,7 @@ import argparse
 
 matplotlib.rcParams['toolbar'] = 'None'  # Remove buttons
 
-gui_env = ['TKAgg', 'GTKAgg', 'Qt4Agg', 'WXAgg']
-selected_gui_env = []
-for gui in gui_env:
-    try:
-        matplotlib.use(gui)
-        from matplotlib import pyplot as plt
-
-        selected_gui_env = gui
-        break
-    except:
-        continue
-# If none works
-if selected_gui_env == []:
-    from matplotlib import pyplot as plt
-
+if matplotlib.get_backend() == "agg":
     logger.warning("No backend can be used - Visualization will fail")
 else:
     logger.info(f"Using: {matplotlib.get_backend()}  gui")
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -222,7 +222,6 @@ def plot_transformed_sample(before, after, list_title=None, fname_out="", cmap="
     if fname_out:
         plt.savefig(fname_out)
     else:
-        matplotlib.use('TkAgg')
         plt.show()
 
 


kousu:
We use the same solution in SCT, but there we set it as a pytest setting:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/20d2f599c6bccffbcc0a852ad6acd3b71d509457/testing/conftest.py#L28-L33

@joshuacwnewton and I had a debate about which file was better to set this variable in: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3420#discussion_r652790703

kousu:
Drat, it didn't work:

Screenshot 2022-05-17 at 20-20-54 Use headless ( Agg ) matplotlib backend by kousu · Pull Request #1133 · ivadomed_ivadomed

I don't get it, that should have worked.

.github/workflows/run_tests.yml
```diff
@@ -19,6 +19,11 @@ env:
   # (TravisCI quietly defined this on all their platforms, but we have to give it manually on GithubCI.)
   DEBIAN_FRONTEND: 'noninteractive'
   HDF5_USE_FILE_LOCKING: 'FALSE'
+  # Skip to the headless matplotlib renderer, which is less
+  # bug-prone in the constrained environment of CI
+  # Tip from a matplotlib dev: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3388#issuecomment-846091012
+  # Ref: https://matplotlib.org/stable/users/explain/backends.html
+  MPLBACKEND: 'Agg'
```
joshuacwnewton, line 26: Could this be relevant, possibly? 

https://github.com/ivadomed/ivadomed/blob/bd904e5e139fa6a437abfc225216f6057824f1e3/ivadomed/utils.py#L225

kousu: Thank you for your eyes! When I got to that point I was just tired of the problem, so I'm really happy you picked up my baton to keep us going.

It could be either. `use()` _sets a **global**_, so its leaking into other tests, [e.g.](https://github.com/ivadomed/ivadomed/runs/6480501528?check_suite_focus=true):

```
ivadomed/metrics.py:305: in plot_roc_curve
    plt.figure()
[...]
../../../hostedtoolcache/Python/3.8.12/x64/lib/python3.8/tkinter/__init__.py:2272: in __init__
    self._loadtk()
```

```
testing/unit_tests/test_mixup.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ivadomed/mixup.py:36: in mixup
    save_mixup_sample(ofolder, data, targets, lambda_tensor)
ivadomed/mixup.py:63: in save_mixup_sample
    plt.figure(figsize=(20, 10))
[...]
../../../hostedtoolcache/Python/3.8.12/x64/lib/python3.8/tkinter/__init__.py:2272: in __init__
    self._loadtk()
```

```
 testing/functional_tests/test_visualize_transforms.py:21: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ivadomed/scripts/visualize_transforms.py:183: in main
    run_visualization(input=args.input, config=args.config, number=int(args.number),
ivadomed/scripts/visualize_transforms.py:171: in run_visualization
    imed_utils.plot_transformed_sample(before,
ivadomed/utils.py:210: in plot_transformed_sample
    plt.figure(figsize=(20, 10))
[...]
../../../hostedtoolcache/Python/3.8.12/x64/lib/python3.8/tkinter/__init__.py:2272: in __init__
    self._loadtk()
```

None of those three `figure()` calls are near the `use()` call but it's still up in their business.

ivadomed/scripts/visualize_and_compare_testing_models.py
```diff
@@ -21,7 +21,7 @@
 
 matplotlib.rcParams['toolbar'] = 'None'  # Remove buttons
 
-gui_env = ['TKAgg', 'GTKAgg', 'Qt4Agg', 'WXAgg']
+gui_env = ['Agg']
```
kousu, line 24: This is just experimental. I'll probably revert this.

But I don't think this loop should be here at all. matplotlib already does this loop: https://matplotlib.org/stable/users/explain/backends.html

> Without a backend explicitly set, Matplotlib automatically detects a usable backend based on what is available on your system and on whether a GUI event loop is already running. The first usable backend in the following list is selected: MacOSX, QtAgg, GTK4Agg, Gtk3Agg, TkAgg, WxAgg, Agg. The last, Agg, is a non-interactive backend that can only write to files.

kousu: So, I'm pretty sure this whole loop is redundant and can be replaced with just the last line of the block! Because matplotlib already does it. Check this:

```
[kousu@nigiri ~]$ python -c 'import matplotlib; print(matplotlib.get_backend())'
QtAgg
[kousu@nigiri ~]$ DISPLAY="" python -c 'import matplotlib; print(matplotlib.get_backend())'
agg
```

i.e. when I tell it I have no GUI, it automatically falls back to agg.

ivadomed/scripts/visualize_and_compare_testing_models.py
```diff
@@ -21,21 +21,7 @@
...
-
-        selected_gui_env = gui
-        break
-    except:
-        continue
-# If none works
-if selected_gui_env == []:
-    from matplotlib import pyplot as plt
-
+if matplotlib.get_backend() == "agg":
```
kousu, line 24: I'm pretty sure this loop is redundant: https://matplotlib.org/stable/users/explain/backends.html

> Without a backend explicitly set, Matplotlib automatically detects a usable backend based on what is available on your system and on whether a GUI event loop is already running. The first usable backend in the following list is selected: MacOSX, QtAgg, GTK4Agg, Gtk3Agg, TkAgg, WxAgg, Agg.

and calling `use()` in the middle of tests is messing the tests up by leaking state from one test to another, because `use()` is setting an internal global variable and hence causing some tests to use TkAgg or QtAgg or just agg whatever might be chosen depending on the order the tests happen to run in.

.github/workflows/run_tests.yml
kousu: Nice! Those were indeed the final pieces. CI is green: https://github.com/ivadomed/ivadomed/actions/runs/2344760419

Screenshot 2022-05-18 at 06-47-36 Use headless ( Agg ) matplotlib backend · ivadomed_ivadomed@23e018f

kousu:
@joshuacwnewton helped me make it work!

kanishk16 (approved):
Apart from just importing pyplot in `visualize_and_compare_testing_models.py`, LGTM!

kousu:
> Apart from just importing pyplot in `visualize_and_compare_testing_models.py`, LGTM!

Do you want

```
from matplotlib import pyplot as plt
```

to be moved to the top of the file instead of removed?

The tests passed without that, do you think it's really necessary?

joshuacwnewton:
> The tests passed without that, do you think it's really necessary?

The `visualize_and_compare_testing_models.py` is only [~16% covered](https://coveralls.io/builds/49246718/source?filename=ivadomed%2Fscripts%2Fvisualize_and_compare_testing_models.py) (though, for some reason Coveralls isn't giving line-by-line coverage). So, I'd be skeptical of trusting the test suite here?

Is it possible that the `onclick` and `visualize_and_compare_models` functions aren't tested at all?

kanishk16:
> > Apart from just importing pyplot in `visualize_and_compare_testing_models.py`, LGTM!
> 
> Do you want
> 
> ```
> from matplotlib import pyplot as plt
> ```
> 
> to be moved to the top of the file instead of removed?

Yup
> The tests passed without that, do you think it's really necessary?

I guess that'd be necessary if someone runs it as a script

https://github.com/ivadomed/ivadomed/blob/bd904e5e139fa6a437abfc225216f6057824f1e3/ivadomed/scripts/visualize_and_compare_testing_models.py#L111-L115

Moreover, I believe there are no tests yet for it.

kousu:
> Moreover, I believe there are no tests yet for it.


Oh true, I didn't notice that. I'll put it in then merge :)

PR #1143, opened 2022-06-02
Use setuptools_scm

kousu:
#### GitHub

#### PR contents

## Description

setuptools_scm is the PyPA-blessed way to integrate git with pip.

Replaces version.txt with the most recent git tag, and MANIFEST.in with
git's contents.  This makes it much less likely to accidentally publish
a broken build, and less work to do a release.

Drops _git_info() and its subroutines. This changes the version string's
format used in dev mode (pip install -e). It previously was

```
{install_type}-{ivadomed_branch}-{ivadomed_commit}
```

Now it's (roughly)

```
{tag+1}.dev{N}+g{commit}[.d{date}]
```

which is https://peps.python.org/pep-0440/#version-scheme approved.

e.g. right now the most recent tag is v2.9.5, 9 commits back before the
current ec136672i, so setuptools_scm generates "2.9.6.dev9+gec136672".
If the branch is edited with uncommited changes, it is marked "dirty"
with "2.9.6.dev9+gec136672.d20220602"; this is a feature the previous
format couldn't tell us. Losing the branch name is too bad but we can
probably live without it for the benefit of conforming to standards.

## Linked issues

The fragileness of MANIFEST.in showed up while trying to release 2.9.6: https://github.com/ivadomed/ivadomed/runs/6711207528?check_suite_focus=true#step:6:169. The immediate problem there was broken build introduced by https://github.com/ivadomed/ivadomed/pull/1131#issuecomment-1145207416, which was reverted in https://github.com/ivadomed/ivadomed/pull/1144, but setuptools_scm is a much stronger long term solution.

### DIFF
--- a/ivadomed/utils.py
+++ b/ivadomed/utils.py
@@ -2,6 +2,12 @@ import os
 import sys
 import subprocess
 import hashlib
+
+if sys.version_info < (3, 8):
+    import importlib_metadata
+else:
+    import importlib.metadata as importlib_metadata
+
 import numpy as np
 import wandb
 from enum import Enum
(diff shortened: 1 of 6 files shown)


ivadomed/utils.py
```diff
@@ -303,18 +280,15 @@ def get_arguments(parser, args):
     return args
 
 
-def __get_commit(path_to_git_folder=None):
-    """Get GIT ivadomed commit.
+def __get_commit(path_to_git_folder):
+    """Get a git commit.
```
kousu, line 284: This function is better implemented as

```
return subprocess.check_output(["git","describe","--always","--dirty"]))
```

that's the standard form most projects use. A dirty repo is reported as "-dirty", instead of with a "*", but otherwise it's the same, and can be done in one line. But that can be done in a follow-up.

PR #1169, opened 2022-06-16
Transformation on Subvolume for mri3d_subvolume_segmentation_dataset

### LINKED ISSUE #993: Transforms in 3D are applied on full volumes instead of sub-volumes
## Issue description

Transforms such as `NormalizeInstance` and data augmentation, such as `RandomAffine` and `ElasticTransform` are applied in [get_item](https://github.com/ivadomed/ivadomed/blob/6c56f91c614faf43567d558acbad349b6565c1fa/ivadomed/loader/mri3d_subvolume_segmentation_dataset.py#L111) of `mri3d_subvolume_segmentation_dataset.py` on the ENTIRE volume each time a subvolume is loaded. i.e. when loading a sub-volume for training, the entire volume is first [transformed](https://github.com/ivadomed/ivadomed/blob/6c56f91c614faf43567d558acbad349b6565c1fa/ivadomed/loader/mri3d_subvolume_segmentation_dataset.py#L138), then the [subvolume is extracted](https://github.com/ivadomed/ivadomed/blob/6c56f91c614faf43567d558acbad349b6565c1fa/ivadomed/loader/mri3d_subvolume_segmentation_dataset.py#L161) from its coordinates in metadata.

This can cause significant performance issues with large volumes and/or datasets.

This issue was first described for the 2D "patch" case in this https://github.com/ivadomed/ivadomed/issues/977#issuecomment-956408247 (Point 3 at the bottom)

The 2D case was fixed in PR #982 and a similar fix should be implemented for 3D subvolumes as well.

cakester:
#### GitHub

#### PR contents

## Description

transformation on subvolume for mri3d_subvolume_segmentation_dataset instead of full volume

## Linked issues

resolves #993


cakester:
following https://github.com/ivadomed/ivadomed/pull/982/files

ivadomed/loader/mri3d_subvolume_segmentation_dataset.py
```diff
@@ -180,15 +180,34 @@ def __getitem__(self, index):
         metadata_input = seg_pair['input_metadata'] if seg_pair['input_metadata'] is not None else []
         metadata_gt = seg_pair['gt_metadata'] if seg_pair['gt_metadata'] is not None else []
 
-        # Run transforms on images
-        stack_input, metadata_input = self.transform(sample=seg_pair['input'],
+        coord = {}
+        coord['x_min'], coord['x_max'] = 0, seg_pair["input"][0].shape[0]
+        coord['y_min'], coord['y_max'] = 0, seg_pair["input"][0].shape[1]
+        coord['z_min'], coord['z_max'] = 0, seg_pair["input"][0].shape[2]
```
dyt811, line 186: These are just adding place holder values, correct?

lifetheater57 (commented):
If I understand correctly, the current code changes don't change what is transformed are still the full size pairs (Related to @dyt811 comment.).

Also, if I understood correctly, previously we would have:
```
1. Take full volume -> 2. Apply transformation -> 3. Take subvolume -> 4. Do something with subvolume
```
Is the intent of this PR to get the same transformed subvolume to do `4`?

If we do:
```
1. Take subvolume -> 2. Apply transformation -> 3. Do something with subvolume
```
Depending on the transformation applied, we may get a different result because some transformations are sensitive to the change in coordinates. 

To, keep the same transformation given the change in coordinates, we might want to do something like:
```
1. Take subvolume -> 2. Transform the transformation -> 3. Apply transformed transformation -> 4. Do something with subvolume
```

dyt811:
Lucy's feedback a few days ago posted on Slack: 

> The order of image processing pipeline of doing subvolume transformation instead of full volume transformation do affect the transformation result.  For spacial transfer such as flip, rotate, etc, the result wouldn’t change. But for image smoothing/blur, elastic-transfromation, etc, because a convolution kernel is used, the transformation result is different. For ML training propers, I don’t think this difference matters. However, if the transformation is k-space transform, you definitely CANNOT apply k-space transformation on subvolume.

So overall, I agree that we may have to make this path mandatory and full volume augmentation optional?

ivadomed/loader/mri3d_subvolume_segmentation_dataset.py
dyt811: Nvm. I did not see the earlier context. =_=.

ivadomed/loader/mri3d_subvolume_segmentation_dataset.py
```diff
@@ -148,86 +148,112 @@ def __len__(self):
...
+        coord: dict = self.indexes[subvolume_index]
+        x_min = coord.get(SegmentationDatasetKW.X_MIN)
+        x_max = coord.get(SegmentationDatasetKW.X_MAX)
+        y_min = coord.get(SegmentationDatasetKW.Y_MIN)
+        y_max = coord.get(SegmentationDatasetKW.Y_MAX)
+        z_min = coord.get(SegmentationDatasetKW.Z_MIN)
+        z_max = coord.get(SegmentationDatasetKW.Z_MAX)
+
+        # Obtain tuple reference to the pairs of file references
+        tuple_seg_roi_pair: tuple = self.handlers[coord.get(SegmentationDatasetKW.HANDLER_INDEX)]
```
lifetheater57, line 172: The use of the dictionary is a nice addition. We should also use it in the [function](https://github.com/ivadomed/ivadomed/blob/ed964e7233d5f56d0597951669dedc71a68e27ba/ivadomed/loader/mri3d_subvolume_segmentation_dataset.py#L139-L145) that create the indices.

lifetheater57 (approved):
I approve this PR since there is no more functional work to do given my understanding. However, I suggest to do the changes related to the keywords usage for maintainability reasons and the changes related to the ranges to ease the code analysis.

ivadomed/loader/mri3d_subvolume_segmentation_dataset.py
dyt811: Yeah, refactoring and keywords extractions are going to be in separate PRs. Trying to limit change scopes within the same functions as main purpose.

PR #1129, opened 2022-05-17
unify installation (attempt #2)

kousu:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [x] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

~Use the [newish --extra-index](https://github.com/pytorch/pytorch/issues/26340#issuecomment-1121566590) that torch put up at https://download.pytorch.org/whl in preference to --find-links. This means a lot of the installation gymnastics can be streamlined.~

Streamline the installation by moving all dependencies back into setup.py. This essentially redoes #1000 but without the [disallowed](https://github.com/ivadomed/ivadomed/pull/1000#issuecomment-1119101714) direct .whl links or the `pip install ivadomed[cpu]` / `pip install ivadomed[gpu]` extras.

Effects:

* Each `requirements_x.txt` becomes an “extra”, giving a consistent installation process no matter whether people are installing from git, from pypi, or _as a dependency of another package_. There's no fumbling with `pip install -r requirements_dev.txt` or `pip install -r requirements_gpu.txt`
* Drops `requirements_gpu.txt` entirely,  declaring our dependency on `torch`/`torchvision` in `setup.py` directly. 

  Doing this fixes #1130 and #861, making sure downstream packages ([like SCT](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3790#issuecomment-1129234605)) can safely depend on ivadomed. This even resolves #996.
  
* `torch` is loosened to `torch~=1.8.1` (meaning 1.8.1, 1.8.2, 1.8.3, …), which enables `--extra-index-url https://download.pytorch.org/whl/....` for a wider set of platforms (see below)
* CUDA 11 is no longer mandatory -- it never really was.

[...]

PR #1180, opened 2022-07-11
Tests README: Add instructions to install testing-related packages

mathieuboudreau:
#### GitHub

#### PR contents

## Description

In the testing/ README, it was previously assumed that you already had `pytest` and `pytest_console_scripts` already installed, but these are not installed along with the IVADOMED requirements. I've added instructions to install them, otherwise users might get an error (I did).

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -12,7 +12,8 @@ Checkout `ivadomed/.github/workflows/run_tests.yml` to see how tests are run on 
 
 1. Install dependencies
 ```
-pip install pytest pytest_console_scripts
+cd ivadomed  # root of the repo
+pip install -e .[dev]
 ```
 
 2. Download the required dataset(s) using the `ivadomed` command line tools:
--- a/testing/README.md
+++ b/testing/README.md
@@ -10,13 +10,18 @@ Checkout `ivadomed/.github/workflows/run_tests.yml` to see how tests are run on 
 
 ## Running Locally
 
-1. Download the required dataset(s) using the `ivadomed` command line tools:
+1. Install dependencies
+```
+pip install pytest pytest_console_scripts
+```
+
+2. Download the required dataset(s) using the `ivadomed` command line tools:
 ```
 cd ivadomed  # root of the repo
 ivadomed_download_data -d data_testing -o data_testing  # for unit tests
 ivadomed_download_data -d data_functional_testing -o data_functional_testing  # for functional tests
 ```
-2. To run all tests:
+3. To run all tests:
 ```
 pytest
 ```
--- a/testing/README.md
+++ b/testing/README.md
@@ -16,21 +16,16 @@ cd ivadomed  # root of the repo
 pip install -e .[dev]
 ```
 
-2. Download the required dataset(s) using the `ivadomed` command line tools:
+2. To run all tests:
 ```
-cd ivadomed  # root of the repo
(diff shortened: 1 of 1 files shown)


dyt811:
Good call. I agree. This was actually because [we USED to have a requirement file for those for dev specifically](https://ivadomed.org/installation.html#developer-only-installation-steps) but since a recent PR streamlining the installation process, it has been borked.

testing/README.md
```diff
@@ -10,13 +10,18 @@ Checkout `ivadomed/.github/workflows/run_tests.yml` to see how tests are run on
 
 ## Running Locally
 
-1. Download the required dataset(s) using the `ivadomed` command line tools:
+1. Install dependencies
+```
+pip install pytest pytest_console_scripts
```
dyt811, line 15: I agree these are the minimal essential. There MAY be more dev related dependencies in #1129 here you want to take a look and see: https://github.com/ivadomed/ivadomed/pull/1129/files#diff-c9c796259f3852b51b531b79cbf07820145088d4f108fb006ab44b15f0b15934 . For now, pytest-cases are not used but I suspect it will be in the near future.

dyt811 (approved):
Thansk!

testing/README.md
kousu: #1129/#1000 put `pytest` into `[dev]`, and that's what CI uses:

https://github.com/ivadomed/ivadomed/blob/14f67f99812599840ef2195171cffc17feb3f643/.github/workflows/run_tests.yml#L66

So I think we should recommend that instead:

```suggestion
cd ivadomed  # root of the repo
pip install -e .[dev]
```

testing/README.md
```diff
@@ -10,13 +10,18 @@ Checkout `ivadomed/.github/workflows/run_tests.yml` to see how tests are run on
...
+```
+
+2. Download the required dataset(s) using the `ivadomed` command line tools:
```
kousu, line 18: I just testing the other day and I didn't have to do this. The tests are set up to auto-download these datasets (which is genius!)

kousu (changes requested):
Thanks for running through the install process and catching these problems. These are the sorts of things that fall through the cracks when people write software together. See my comments above for what I think should be done.

I'm sorry it got lost in my inbox @mathieuboudreau.

testing/README.md
```diff
@@ -10,13 +10,19 @@ Checkout `ivadomed/.github/workflows/run_tests.yml` to see how tests are run on
...
+cd ivadomed  # root of the repo
+pip install -e .[dev]
+```
+
+2. Download the required dataset(s) using the `ivadomed` command line tools:
 ```
 cd ivadomed  # root of the repo
 ivadomed_download_data -d data_testing -o data_testing  # for unit tests
 ivadomed_download_data -d data_functional_testing -o data_functional_testing  # for functional tests
 ```
```
kanishk16, line 24: As Nick suggested, this step is no longer required to run the test suite.

kanishk16 (approved):
Thanks @mathieuboudreau as we would have definitely missed this! LGTM apart from a minor suggestion which is continuation to Nick's comment.

testing/README.md
kousu: I took the liberty of deleting them :)

kousu (approved):

mathieuboudreau:
Thank you all for your reviews and suggestions! Merging!

PR #1181, opened 2022-07-11
Resolve imageio v2->v3 imread deprecation warnings

mathieuboudreau:
#### GitHub

#### PR contents

## Description

The IVADOMED test suite run is throwing warning about the deprecation of imageio's imread behaviour when it changes from version 2 to version 3:

Screen Shot 2022-07-11 at 1 37 08 PM

Because imageio is only implicitely installed in the current [requirements file](https://github.com/ivadomed/ivadomed/blob/e22b2e340261d48dc137a9a9c72ad627440073eb/requirements.txt) through other of the installed packages, but [explicitely used](https://github.com/mathieuboudreau/ivadomed/blob/87cf24db983f89d231b86d39cedd5f558b9b9819/ivadomed/loader/segmentation_pair.py#L2) in at least one file (the one throwing the warnings), this makes it such that whenever imageio releases version 3.0, the IVADOMED behaviour might break. Note that the dependency(ies) that install imagio don't appear have a cap on the version, since the most up-to-date version is installed during installation.

In this PR I suggest that the recommended deprecation solution is applied (`imageio.imread` -> `imageio.v2.imread`).

Doing so resolves 89 of the 90 warnings for me:

Screen Shot 2022-07-11 at 1 36 40 PM

The other alternative would be to add imageio as a dependency and fix it to version ~2.x, but I just chose this one as it was suggested by the imageio devs.

### DIFF
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -311,9 +311,9 @@ class SegmentationPair(object):
         # For '.png', '.tif', '.tiff', '.jpg' and 'jpeg' extentions
         # Read image as 8 bit grayscale in numpy array (behavior TBD in ivadomed for RGB, RBGA or higher bit depth)
         if "tif" in extension:
-            img = np.expand_dims(imageio.imread(filename, format='tiff-pil'), axis=-1).astype(np.uint8)
+            img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil'), axis=-1).astype(np.uint8)
             if len(img.shape) > 3:
-                img = np.expand_dims(imageio.imread(filename, format='tiff-pil', as_gray=True), axis=-1).astype(np.uint8)
+                img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil', as_gray=True), axis=-1).astype(np.uint8)
         else:
             img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
 
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -315,7 +315,7 @@ class SegmentationPair(object):
             if len(img.shape) > 3:
                 img = np.expand_dims(imageio.imread(filename, format='tiff-pil', as_gray=True), axis=-1).astype(np.uint8)
         else:
-            img = np.expand_dims(imageio.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
+            img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
 
         # Binarize ground-truth values (0-255) to 0 and 1 in uint8 with threshold 0.5
         if is_gt:
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@
 pyparsing<3,>=2.0.2
 csv-diff>=1.0
 loguru~=0.5
+imageio~=2.19
 joblib~=1.0
 matplotlib>=3.3.0
 nibabel~=3.2


kanishk16 (commented):
@mathieuboudreau Thanks for the fix! Since we're explicitly using `imageio`, I'm more inclined towards adding it as a dependency, otherwise it LGTM. In case you're not inclined in that direction, could you link to the discussion where the devs preferred the other approach, just to record why we chose this alternative, and I'll approve immediately.

mathieuboudreau:
>@mathieuboudreau Thanks for the fix! Since we're explicitly using imageio, I'm more inclined towards adding it as a dependency, otherwise it LGTM. 

That's fine with me - maybe we can both add it as a requirement **and** keep the fix that removes the deprecation warnings/fixes the behaviour to 2.x? I'll do that and push and you can approve if you're ok with that.

kanishk16 (approved):
Thanks @mathieuboudreau! I wasn't clear enough, but this is what I had in mind.

hermancollin:
@mathieuboudreau btw this will not solve the problem on the ADS side until ivadomed releases a new patch. Our environment.yml now uses this:
```
- pip:
   - ivadomed~=2.9.6
```
which is flexible enough to take a patch if we make a new release before they switch to 2.9.7

mathieuboudreau:
Good catch! Do you think we should update our req condition, or wait to manually update it when they make their new release?

hermancollin:
> Do you think we should update our req condition

@mathieuboudreau hmm I'm not sure we can do this. Ivadomed is pip installed. Do you think we can access the master branch directly?

mathieuboudreau:
Yeah, there's a way to install github master branches through "pip" installation, but I don't think there's a rush to do that here since these just resolved warnings. We can just wait for whatever new IVADOMED release includes it and then bump our ivadomed version in our requirements at that stage. Would that be ok with you?

hermancollin:
@mathieuboudreau yes I think it's better to wait.

mariehbourget:
Hi @mathieuboudreau and @kanishk16,
I just found out that `imageio 2.19+` that we added with this PR requires `pillow>8.3.2`.

Our requirements ask for `pillow>7.0.0` which is not compatible.
I was able to pip install the correct version but I was wondering if we should change our requirement for pillow or if we should just remove it and let imageio pick the right version?
What is the best practice here?

mathieuboudreau:
>Our requirements ask for pillow>7.0.0 which is not compatible.
I was able to pip install the correct version but I was wondering if we should change our requirement for pillow or if we should just remove it and let imageio pick the right version?
What is the best practice here?

Actually, since 8.3.2 > 7.0.0, I don't see a conflict here? If a requirement for IVADOMED was that it's `~=` to 7.0.0, that would be an issue, but because 

https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/requirements.txt#L16

It just means that you could now update that number version to be a bit higher

mariehbourget:
> Actually, since 8.3.2 > 7.0.0, I don't see a conflict here? If a requirement for IVADOMED was that it's `~=` to 7.0.0, that would be an issue, but because
> 
> https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/requirements.txt#L16
> 
> It just means that you could now update that number version to be a bit higher

Yes you are right, sorry I was not clear.
There is no conflict, what happened is:
- I had to redo a `pip install -e .` in an "old" environment in order to have the right version of imageio since this PR.
- In that environment, I had Pillow 8.0.1 which is higher that 7.0.0 so it was not re-installed.
- But because it's lower than 8.3.2, it would not work with imageio 2.20.
- So I pip installed 8.3.2 separately.

My question is: should we update the minimal requirement of pillow to 8.3.2 to avoid that kind of (very small) issue?

mathieuboudreau:
Ahh I understand now! Yup, that's probably the best thing we can do to avoid this issue moving forward I think! Good catch!

kousu:
> > Actually, since 8.3.2 > 7.0.0, I don't see a conflict here? If a requirement for IVADOMED was that it's `~=` to 7.0.0, that would be an issue, but because
> > https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/requirements.txt#L16
> > 
> > It just means that you could now update that number version to be a bit higher
> 
> Yes you are right, sorry I was not clear. There is no conflict, what happened is:
> 
>     * I had to redo a `pip install -e .` in an "old" environment in order to have the right version of imageio since this PR.
> 
>     * In that environment, I had Pillow 8.0.1 which is higher that 7.0.0 so it was not re-installed.
> 
>     * But because it's lower than 8.3.2, it would not work with imageio 2.20.
> 
>     * So I pip installed 8.3.2 separately.
> 
> 
> My question is: should we update the minimal requirement of pillow to 8.3.2 to avoid that kind of (very small) issue?

I'm a little confused about what the glitch was, but I have confidence that bumping that version number is not the best practice.

The best practice is to let `imageio` declare and manage its own dependencies. There's only one place that ivadomed actually calls pillow and it hardly even counts

https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/testing/unit_tests/test_tensorboard_save.py#L8

So I think ivadomed should remove `pillow` from `requirements.txt` entirely. From our perspective, it's a black-box component of `imageio`. If we did update the version, then in two years someone will be scratching their heads about why that particular version was chosen, but be too scared to consider changing or removing it. I'll submit a PR to do this.

Now, here's why I'm confused about the glitch: if you did `pip install -e .` on this branch you should have gotten `imageio~=2.19` because

https://github.com/mathieuboudreau/ivadomed/blob/b3e1c019337857e855b3b459ca22d12785c1d22b/requirements.txt#L10

and that should have pulled in `pillow >= 8.3.2` because of

https://github.com/imageio/imageio/blob/cd721d92fe096e6bb0ec82e5d0f2ed2fed44b06e/setup.py#L177

So something weird happened! But I've come to expect some weird things from `pip` so I wouldn't be surprised. If it happens again maybe

But generally whenever testing, especially when testing dependency changes, I think the usual method is to recreate the venv entirely. Or we should also be able to use `pip install --force-reinstall -e .`, which amounts to the same thing. Those are more conventional and -- I think in the long term -- less confusing than trying to second-guess the dependencies of our dependencies.

For example, in https://github.com/axondeepseg/axondeepseg/pull/642 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3790 we got rid of `torch` and instead let `ivadomed` manage it as a contained dependency, because neither ADS nor SCT uses `torch` directly, and this way we don't need to worry about version conflicts.

mariehbourget:
Thanks @kousu for the detailed answer and the PR, it is much appreciated!

mathieuboudreau:
Yes, thank you @kousu !

>.Or we should also be able to use pip install --force-reinstall -e ., which amounts to the same thing. 

Maybe this tip could be included in the documentation somewhere (if it is and I missed it, sorry!), I don't think all devs will know about the `--force-reinstall` flag (I didn't).

PR #1182, opened 2022-07-11
Control randomness for reproducibility

### LINKED ISSUE #1092: Controlling randomness for reproducibility
The current implementation of ivadomed only sets the random number generator seed for splitting a dataset into train, validation and test sets. However, there are other sources of randomness that we need to take into consideration, such as 1) the Numpy library 2) Pytorch library 3) Python itself 4) CUDA etc. 
We may use this [reference](https://pytorch.org/docs/stable/notes/randomness.html)

The main reason is seeding the dataset splitting alone, still leaves some sources of randomness unchecked. Also, if a user decides to use a `.joblib` file to split the data as documented [here ](https://ivadomed.org/configuration_file.html#fname-split), the `random_state` won't be used, but we still need to control other sources of randomness.

kanishk16:
#### GitHub

#### PR contents

## Description
This PR attempts to add a control over innumerable sources of randomness to enhance the reproducibility of the results of experiments. Thus, adds:
* an optional boolean `control_randomness` keyword as part of `split_dataset` with default as `False`
* a `set_global_seed(seed: int, control_randomness: bool)` function in `utils.py` to set a seed for different sources of randomness
* TODO: updates docs with this addition and its usage 
## Linked issues
Resolves #1092

### DIFF
--- a/ivadomed/keywords.py
+++ b/ivadomed/keywords.py
@@ -64,7 +64,7 @@ class SplitDatasetKW:
     TRAIN_FRACTION: str = "train_fraction"
     TEST_FRACTION: str = "test_fraction"
     BALANCE: str = "balance"
-
+    CONTROL_RANDOMNESS: str = "control_randomness"
 
 @dataclass
 class DataTestingKW:
--- a/ivadomed/loader/utils.py
+++ b/ivadomed/loader/utils.py
@@ -145,7 +145,6 @@ def get_new_subject_file_split(df, split_method, data_testing, random_seed,
             raise ValueError("All lists in subject_selection parameter should have the same length.")
 
         sampled_dfs = []
-        random.seed(random_seed)
         for m, n, v in zip(subject_selection["metadata"], subject_selection["n"], subject_selection["value"]):
             participants = random.sample(df[df[m] == v]['participant_id'].unique().tolist(), n)
             for participant in participants:
--- a/ivadomed/main.py
+++ b/ivadomed/main.py
@@ -25,8 +25,6 @@ from ivadomed.keywords import ConfigKW, ModelParamsKW, LoaderParamsKW, ContrastP
 from loguru import logger
 from pathlib import Path
 
-cudnn.benchmark = True
-
 # List of not-default available models i.e. different from Unet
 MODEL_LIST = ['Modified3DUNet', 'HeMISUnet', 'FiLMedUnet', 'resnet18', 'densenet121', 'Countception']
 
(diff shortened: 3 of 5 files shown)


docs/source/configuration_file.rst
```diff
@@ -794,6 +794,26 @@ Split Dataset
...
+            "Indicates if some sources of randomness in the experiments should be reduced by setting random_seed for numpy and torch. Default: ``False``.",
+        "type": "boolean"
+    }
+
+
+.. code-block:: JSON
+
+    {
+        "split_dataset": {
+            "control_randomness": true
```
kanishk16, line 812: In our internal meeting, dated 18-07-2022 @naga-karthik suggested to perhaps have something along the lines `use_deterministic`. I guess `use_deterministic` could be slightly misleading as we aren't setting `torch.use_deterministic_algorithms(True)` as it would throw a `RuntimeError` due to the following operation in our codebase `F.interpolate()` (ref: [PyTorch docs](https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html))

hermancollin:
@kanishk16 how can I test this PR? My idea was to setup a very minimalist training with 2 epochs and check the test image difference between 2 consecutive runs with the same config file. Would it cover everything that this feature adds in terms of randomness control?

I missed some discussions surrounding this but I'm guessing this only affects training.

kanishk16:
@hermancollin Thanks for this idea... I completely missed this direction as I was only thinking along the metrics that the logger reports. I guess this complements your idea. So, a sneak peek at the metrics reported by the logger along with your idea of testing image difference between 2 consecutive runs with the same config file should cover everything that this feature adds. 

A quick note that the results should be _exactly_ the same for 2 consecutive runs with same config file on CPU but _nearly_ same on GPU.

hermancollin (commented):
@kanishk16 I just finished testing this feature (sorry for the delay). The training losses are identical across runs using the same config file. Moreover, both models produce the exact same prediction given the same input image. 

(Tested with 2D microscopy data on CPU)

kanishk16:
Thanks @hermancollin! Since there were concerns about the results on the GPU and we'd like to know how similar the results are... So, could you also try the same experiment on the GPU and report the results here so we can discuss it with Naga?

hermancollin:
@kanishk16 of course, I'm curious too. I only need to get my GRAMES access back because I can't access the GPU clusters atm. Stay tuned.

hermancollin:
Hmm I got my access earlier than expected. So I did an identical test to what I did last week locally, but using a GPU. As you were expecting, the losses are slightly different, so this randomness control feature does indeed behave differently on CPU vs GPU.

I think this should be mentioned in the doc in the description of the `control_randomness` key.

PR #1184, opened 2022-07-20
docs: Unify Installation

kousu:
This is some edits to how I think https://github.com/ivadomed/ivadomed/pull/1125 should look (if merged this will merge **into** that PR). I thought showing what I had in mind would be easier than trying to back-seat drive as a reviewer.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -25,112 +25,97 @@ Thank you for your interest in contributing to ivadomed! This project uses the f
 
 .. _installation_contributor:
 
-Contributor or Developer ``ivadomed`` installation 
+Developer ``ivadomed`` installation
 ++++++++++++++++++++++++++++++++++++++++++++++++++
 
+1. Download the code:
+
+    ::
+
+        cd ivadomed
+        git clone https://github.com/ivadomed/ivadomed.git
+
+2. Isolate your work in a virtual environment:
+
+    We recommend developing inside of ``venv`` or ``conda``.
+    It makes debugging easier.
+
     .. tabs::
-        
-        .. tab:: NVIDIA GPU Support
 
-            PyTorch, an integral part of ``ivadomed``, ships 
-            CUDA 10.2 and CUDA 11.1 runtime by default with its
-            respective installation binaries.
+        .. group-tab:: ``venv``
 
-            In case if you're wondering about CUDA runtime, Ampere-based GPUs
-            (with a `Compute Capability <https://developer.nvidia.com/cuda-gpus>`_
-            of 8.x) only work with CUDA>=11.1. Although CUDA 11.1 is
-            backward compatible with older hardware, CUDA 10.2 is
-            preferred if available.
+            Create the environment:
 
-            Thus, to accelerate ``ivadomed`` with CUDA 10.2 on a Linux system, you'd
-            need to have GPUs installed with an `NVIDIA driver version >=440.33 
-            <https://docs.nvidia.com/deploy/cuda-compatibility/index.html#minor-version-compatibility>`_.
-            And, for CUDA 11.1 you'd rather need an upgraded NVIDIA driver version >=450.
-            
-            To verify the NVIDIA driver version, look in ``/sys`` by executing the
-            command:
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.rst
```diff
@@ -25,112 +25,97 @@ Thank you for your interest in contributing to ivadomed! This project uses the f
...
-            by following the instructions as specified in the :ref:`step 1 of the installation <installation_step1>`.
+            Create the environment:
 
-            Once you've set up a virtual environment and activated it, we recommend installing 
-            ``ivadomed`` from source along with some additional dependencies related to building
-            documentation, linting and testing:
+            .. code::
 
-            .. tabs::
+                python -m venv venv
```
kousu, line 51: Maybe

```suggestion
                python -m venv ivadomed_venv
```

but then that makes

```
ivadomed/ivadomed_venv
```

which is weird. And also `.gitignore` *already had an entry* for `venv/`. So if we change this we should change that too.

docs/source/installation.rst
```diff
@@ -1,210 +1,159 @@
...
+    If you are installing into your home folder (``pip install --user``), which is the default,
+    you need to make sure that ``~/.local/bin/`` (or ``%APPDATA%\Python\PythonXY\Scripts`` on Windows)
+    is on your ``$PATH`` for ivadomed's tools to be available. ``pip`` will warn you if it is not.
+    It should be pre-configured properly on Ubuntu and most common Linux distros.
 
-                ``ivadomed`` requires Python >= 3.7 and <3.10. First, make sure that a
-                compatible version of Python 3 is installed on your system by running:
+    If you are installing into a `venv <https://docs.python.org/3/library/venv.html>`_ or a
+    `conda env <https://docs.conda.io/projects/conda/en/latest/commands.html#conda-vs-pip-vs-virtualenv-commands>`_
+    then the tools will be available whenever that environment is activated.
```
kousu, line 22: Maybe this section would be better in a separate Troubleshooting section. 80% of people will be fine, it's only people on like, ArchLinux or Gentoo or Alpine that might have this problem, and it's a one-time fix for them too so they might already have fixed it before they try to install us.

docs/source/installation.rst
```diff
@@ -1,210 +1,159 @@
...
 
-                            source ivadomed_env/bin/activate
+        .. NOTE: we must periodically update the URL here to use the most recent CUDA,
+                 or else pip will start prefering the PyPI build to the pytorch.org build
+        .. NOTE: here we recommend CUDA 10.2, because that matches what Linux currently gets from PyPI.
+                 If/when torch starts pushing CUDA 11 to PyPI, update this.
+        ::
 
-                    .. group-tab:: Windows
+            pip install ivadomed --force-reinstall --extra-index-url https://download.pytorch.org/whl/cu102
```
kousu, line 85: In https://github.com/axondeepseg/axondeepseg/pull/642 I instead recommended people put this into pip.conf. I am leaning towards that being a better recommendation because the choice of which download.pytorch.org link to use really depends on people's hardware, and that doesn't change as rapidly as software, and if any other software they use uses pytorch...they're going to want to make the same call as for what build they want.

docs/source/installation.rst
```diff
@@ -1,210 +1,159 @@
...
+    pip install ivadomed --force-reinstall \
+        --extra-index-url https://download.pytorch.org/whl/cu110 \
+        --extra-index-url https://download.pytorch.org/whl/cu111 \
+        --extra-index-url https://download.pytorch.org/whl/cu112 \
+        --extra-index-url https://download.pytorch.org/whl/cu113 \
+        --extra-index-url https://download.pytorch.org/whl/cu114 \
+        --extra-index-url https://download.pytorch.org/whl/cu115 \
+        --extra-index-url https://download.pytorch.org/whl/cu116 \
+        --extra-index-url https://download.pytorch.org/whl/cu117 \
+        --extra-index-url https://download.pytorch.org/whl/cu118
```
kousu, line 125: ditto https://github.com/ivadomed/ivadomed/pull/1184/files#r927952094

PR #1081, opened 2022-02-17
Fix long evaluation time on microscopy images

mariehbourget:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [x] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
This PR aims to fix long evaluation time on large microscopy images with the axons and myelin segmentation task.

- It adds a `object_detection_metrics` param in the `evaluation_parameters` of the config file.
- The default is `True`.
- When `True`, it computes lesions true positive and false detection rates + Hausdorff score at evaluation time.
- When `False`, those metrics are skipped (they are not appropriate for axon+myelin segmentation and would create the bottleneck when running evaluation on large images).
- When `False`, the "painted" predictions in the `pred_masks` are also skipped since they are based on lesions true positive and false detection rates.
- Doc and microscopy tutorial are updated accordingly.

## Linked issues
Fixes #975

PR #1185, opened 2022-07-27
Convert pred data to uint8 prior to imwrite png

mathieuboudreau:
#### GitHub

#### PR contents

## Description

Writing the histology segmentation predictions throws a warning:

`Lossy conversion from float64 to uint8. Range [0, 1]. Convert image to uint8 prior to saving to suppress this warning.`

The root cause is that nibabel's `fdata()` call forces the returned data to always be float64, while we're saving to uint8.

This quick fix resolves that and gets rid of the warning.

## Linked issues

External: 

Axondeepseg
* https://github.com/axondeepseg/axondeepseg/pull/669
* https://github.com/axondeepseg/axondeepseg/issues/175
* https://github.com/axondeepseg/axondeepseg/issues/666


mariehbourget (commented):
Thanks @mathieuboudreau!
I tested the PR without binarizing the predictions (to have some gray values) and compared them with master and the predictions are identical.

However, I noticed that `pred_to_png` is also used with the comand `--test` to output "painted" versions of the predictions (with a different color for TP, FP, FN "objects") [here](https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/ivadomed/evaluation.py#L98).
The "colors" are set at values 1, 2 and 3 [here](https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/ivadomed/evaluation.py#L15).

With the imageio conversion, the values 0-1-2-3 were scaled between 0-255 at 0, 85, 170 and 255.
But with the changes in this PR, all the values end up at 255.

Honestly, this feature is useful for spinal cord of lesions segmentation on MRI but is of very little interest for axon and myelin segmentation, so I wonder if we should just remove it for microscopy.

Funny enough, I started working a while ago on another PR https://github.com/ivadomed/ivadomed/pull/1081 to skip the evaluation of TP, FP and FN objects and "painted" files for the axon and myelin task.

Let me revisit my old PR and think about this and I'll get back to you.

mariehbourget:
> I tested the PR without binarizing the predictions (to have some gray values) and compared them with master and the predictions are identical.
> 
> However, I noticed that `pred_to_png` is also used with the comand `--test` to output "painted" versions of the predictions (with a different color for TP, FP, FN "objects") [here](https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/ivadomed/evaluation.py#L98). The "colors" are set at values 1, 2 and 3 [here](https://github.com/ivadomed/ivadomed/blob/f0511a384ea1ff584c8ca136557655c2b0230ca8/ivadomed/evaluation.py#L15).
> 
> With the imageio conversion, the values 0-1-2-3 were scaled between 0-255 at 0, 85, 170 and 255.
> But with the changes in this PR, all the values end up at 255.

> Honestly, this feature is useful for spinal cord of lesions segmentation on MRI but is of very little interest for axon and myelin > segmentation, so I wonder if we should just remove it for microscopy.

@mathieuboudreau
Thinking out loud here.
I would rather not remove the "painted" option entirely in case we have tasks in the future with microscopy for which it is useful (task other that axon/myelin seg).

Maybe a good way to prevent issues with the conversion while remaining flexible to different scenarios could be to add a "range" argument to `pred_to_png`.

So everytime we call `pred_to_png`, we could specify the range (either `[0-1]` or `[0-3]`) and do the conversion to `[0-255]` accordingly. What do you think?

mathieuboudreau:
> What do you think?

I'm ok with that! For simplicity, may I suggest that the function argument be something like `num_classes` instead of range? This way you can pass on a single value (also, one thing as a user that always make me think twice is that some python functions exclude the last value in a range, i.e.:

```python

>>> import numpy as np
>>> a=np.arange(0,3)
>>> print(a)
[0 1 2]

```

so users risk passing on on the wrong "range" sometimes if they are also confused on if the last value is excluded or not).

mariehbourget:
> For simplicity, may I suggest that the function argument be something like `num_classes` instead of range?

I agree that range can be misleading.
But `num_classes` can be misleading as well, for example we have a 2-class training (axon and myelin) but each predictions has only a range of 0-1.
Maybe `max_value`?

mathieuboudreau:
@mariehbourget I implemented your suggestion, but am unsure if the tests cover it well because it's only called in integration tests I think and not a function-specific unit test. Would you mind taking a look, or provide me with information on how to test this using your example of a 3 class dataset?

ivadomed/testing.py
```diff
@@ -264,7 +264,8 @@ def run_inference(test_loader, model, model_params, testing_params, ofolder, cud
                         imed_inference.pred_to_png(output_list,
                                                    target_list,
                                                    fname_pred.split("_pred.nii.gz")[0],
-                                                   suffix="_pred.png")
+                                                   suffix="_pred.png",
+                                                   max_value=len(target_list))
```
mariehbourget, line 268: I think we have a misunderstanding in one of our previous discussion.
The `max_value` here is 1, no matter the number of classes.

`pred_to_png` is used at 3 different places in the ivadomed code:

- Here + in [main.py](https://github.com/ivadomed/ivadomed/blob/dee79e6bd90b30095267a798da99c4033cc5991b/ivadomed/main.py#L327) where the values of the `pred list` are predictions in float between [0-1], for each class, so the conversion is not related to the number of classes.
- In [evaluation.py](https://github.com/ivadomed/ivadomed/blob/dee79e6bd90b30095267a798da99c4033cc5991b/ivadomed/evaluation.py#L103) where the `pred_list` is a "painted" list (painted according to TP FP FN objects) in float with range [0-3].

So with a default `max_value` of 1, we don't have to specify the `max_value` here and we get the same behavior as before.
However, we need to specify the `max_value=3` in `evaluation.py` to have the correct output of the "painted" predictions.

mathieuboudreau:
Reminder: https://ivadomed.org/tutorials/two_class_microscopy_seg_2d_unet.html

ivadomed/evaluation.py
```diff
@@ -104,7 +104,8 @@ def evaluate(bids_df, path_output, target_suffix, eval_params):
                 imed_inference.pred_to_png(painted_list,
                                            target_list,
                                            str(path_preds.joinpath(subj_acq)),
-                                           suffix="_pred_painted.png")
+                                           suffix="_pred_painted.png",
+                                           max_value=len(target_list))
```
mariehbourget, line 108:
```suggestion
                                           max_value=3) # painted data contain 3 float values [0.0, 1.0, 2.0, 3.0] corresponding to background, TP, FP and FN objects)
```
This ensure the same behavior as the one currently on master for "painted data".
The max value is not related to the number of class in a model.

We always want to have a max_value of "3" here because they represent the 3 possible values that may be present in painted data for True Positive, False Positive and False Negative objects respectively (+ 0 for background).
By setting "3" here, we ensure that the float values 0.0, 1.1, 2.0 and 3.0 of painted data are converted to uint values 0, 85, 170 and 255 consistently in the png file.

ivadomed/inference.py
```diff
@@ -221,11 +221,12 @@ def pred_to_png(pred_list: list, target_list: list, subj_path: str, suffix: str
         subj_path (str): Path of the subject filename in output folder without extension
             (e.g. "path_output/pred_masks/sub-01_sample-01_SEM").
         suffix (str): additional suffix to append to the filename (e.g. "_pred.png")
+        max_value (int): Maximum mask value of the float mask to use during the converion to uint8.
```
mariehbourget, line 224:
```suggestion
        max_value (int): Maximum mask value of the float mask to use during the conversion to uint8.
```

mariehbourget (commented):
Hi Mathieu,
As discussed in meeting, I reviewed the PR and added 1 suggestion to deal properly with the "max_value" of "painted data".

For `pred_to_png` itself, the behavior with the new conversion is correct and as expected.
I will add a procedure for testing the PR shortly.

mariehbourget:
Here is a procedure to test the PR and to make sure that results are the same as with the master branch.
@mathieuboudreau let me know if you have any questions or concerns, thanks!

1. Create a log folder.
2. Copy the `model_seg_rat_axon-myelin_sem.pt` file from the [SEM model](https://github.com/axondeepseg/default-SEM-model/tree/main/model_seg_rat_axon-myelin_sem) in the log folder.
3. Rename the file `best_model.pt`.
4. Copy the config file `model_seg_rat_axon-myelin_sem.json` from the [SEM model](https://github.com/axondeepseg/default-SEM-model/tree/main/model_seg_rat_axon-myelin_sem) in the log folder.
5. In the config file, adjust the following:
    - `path_output` : the path of your log folder
    - `loader_parameters/path_data` : path to the [SEM dataset](https://github.com/axondeepseg/data_axondeepseg_sem)
    - Replace the path in `split_dataset/fname_split` with `null`
6. Run ivadomed with the --test command: `ivadomed -c path/to/your/config/file --test`
7. In `pred_masks` in the log folder you should have 2 PNG predictions files and 2 PNG "painted" files:
    - `sub-rat6_sample-data15_SEM_class-0_pred.png`
    - `sub-rat6_sample-data15_SEM_class-0_pred_painted.png`
    - `sub-rat6_sample-data15_SEM_class-1_pred.png`
    - `sub-rat6_sample-data15_SEM_class-1_pred_painted.png`
8. Predictions files should have 2 different values: 0 (background) and 1 (axon or myelin)
9. Painted files should have 4 values: 0 (background), 85 (True Positives), 170 (False Positives) and 255 (False Negatives)
10. Make sure that the results are the same on this branch vs. master branch.

As a reference, here are examples of TP, FP and FN on the painted files:

sub-rat6_sample-data15_SEM_class-0_pred_painted_example
sub-rat6_sample-data15_SEM_class-1_pred_painted_example

mathieuboudreau:
@mariehbourget I tested your (very well written!) testing procedure on 1) the master branch 2) my branch prior to your changes, and 3) the branch after your changes, and can confirm that your suggested changes have fixed the issue and now reproduced master. It looks good to me, so if you're ok with it I suggest you approve this PR and merge.

After that, I have it in my notes and the ADS action tasks that we should ask IVADOMED makes a new release with this update as soon as it's convenient for you, so that it would be included in our installed the version of IVADOMED installed with the ADS dependencies.

mariehbourget (approved):

mariehbourget:
> After that, I have it in my notes and the ADS action tasks that we should ask IVADOMED makes a new release with this update as soon as it's convenient for you, so that it would be included in our installed the version of IVADOMED installed with the ADS dependencies.

Yes, if I remember correctly, the goal of having a new ivadomed release was to also be able to implement the "no-patch" option from a previous PR (https://github.com/ivadomed/ivadomed/pull/1101) directly in ADS (https://github.com/axondeepseg/axondeepseg/issues/676).

I will merge this PR and setup for a release next week if it's ok with the ivadomed dev team.
Thanks @mathieuboudreau!

Issue #1188, opened 2022-07-30
"CUDA error: invalid device ordinal"


kousu:
### Current behavior

When run on a machine with GPUs, some tests fail with

```
Traceback (most recent call last):
  File "/home/GRAMES.POLYMTL.CA/p115628/.conda/envs/ivadomed/lib/python3.9/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/GRAMES.POLYMTL.CA/p115628/.conda/envs/ivadomed/lib/python3.9/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/home/GRAMES.POLYMTL.CA/p115628/src/ivadomed/ivadomed/scripts/automate_training.py", line 97, in train_worker
    ivado.run_command(config, thr_increment=thr_incr)
  File "/home/GRAMES.POLYMTL.CA/p115628/src/ivadomed/ivadomed/main.py", line 375, in run_command
    cuda_available, device = imed_utils.define_device(context[ConfigKW.GPU_IDS][0])
  File "/home/GRAMES.POLYMTL.CA/p115628/src/ivadomed/ivadomed/utils.py", line 167, in define_device
    torch.cuda.set_device(gpu_id)
  File "/home/GRAMES.POLYMTL.CA/p115628/.conda/envs/ivadomed/lib/python3.9/site-packages/torch/cuda/__init__.py", line 261, in set_device
    torch._C._cuda_setDevice(device)
RuntimeError: CUDA error: invalid device ordinal
```

### Expected behavior

The tests should pass.

### Steps to reproduce

```
pytest -v -s testing/functional_tests/test_automate_training.py::test_automate_training_run_test[subprocess]
pytest -v -s testing/functional_tests/test_automate_training.py::test_automate_training_run_test_debug
```

### Logs

<details>

```
(ivadomed) p115628@bireli:~/src/ivadomed$ pytest -v -s testing/functional_tests/test_automate_training.py::test_automate_training_run_test[subprocess]
========================================================================= test session starts =========================================================================
platform linux -- Python 3.9.13, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /home/GRAMES.POLYMTL.CA/p115628/.conda/envs/ivadomed/bin/python3.9
cachedir: .pytest_cache
rootdir: /home/GRAMES.POLYMTL.CA/p115628/src/ivadomed, configfile: pytest.ini
[...]

PR #1189, opened 2022-07-30
Bump data_functional_testing

kousu:
Fixes https://github.com/ivadomed/ivadomed/issues/1188. See that for details.


dyt811 (approved):
Thanks for catching this. I think with the de-emphasis of GPU based testing, this was shelfed when it really shouldn't have.

Issue #1102, opened 2022-03-14
Improve visibility of image and ground-truths pairing


mariehbourget:
### Motivation for the feature
First suggested by @jcohenadad in this comment: https://github.com/ivadomed/ivadomed/pull/1100#issuecomment-1065300055

Considering the new features we want to implement for pairing in multichannel in #1087, the multi-session/multi-contrast project in #1044, and problems with ground-truth pairing reported in #1055, it would be very useful for the user to have more visibility about the pairing that was done by ivadomed (ie: image & target ground truth), and that way, if they realize the pairing was wrong, they can change they config file accordingly.

Currently, wrong pairing can be very tricky to notice because we don't have any outputs of the true pairing done by ivadomed.
The only information about the files indexed is in the `bids_dataframe.csv` file. However, this file is over-written each time we launch ivadomed (ex: when using command `--segment` after training), and does not show the pairing itself.

### Description of the feature
One way to improve visibility would be to output in a CSV file the matched derivatives/label.
We may want to consider a mechanism to avoid the file being over-written when using different ivadomed command as well.

dyt811:
This is being somewhat addressed in 1009 discussion and will likely be rolled in as part of that major overhaul.

jcohenadad:
Note that this information is already accessible via a joblib file: https://ivadomed.org/configuration_file.html#fname-split

However, outputting a human-interpretable (eg CSV) file would be awesome.

Issue #1087, opened 2022-02-26
Expand the definition of `multichannel` to multiple images within the same contrast


naga-karthik:
### Motivation for the feature

Adds more functionality to ivadomed by making it possible to train multichannel models obtained by concatenating different images such as sagittal (`acq-sag_T2w)` and axial (`acq-ax_T2w`), having same dimensions within the same contrast.

### Description of the feature

Currently, the definition of `multichannel` as seen [here](https://ivadomed.org/configuration_file.html#contrast-params) is restrictive as it only allows the training of multichannel models if the image contrasts are different (i.e. if there are T1w and T2w images, then the number of input channels is 2). This leaves out the case where there exist multiple images _within_ the same contrast (i.e. sagittal and axial as mentioned above). 

This probably requires refactoring of the existing code so that the multichannel aspect does not depend on the existing contrast-only suffix string (i.e. T1w, T2w) but also is expanded to include the substring `acq-sag_<existing-contrast-suffix-here>` and `acq-ax_<existing-contrast-suffix-here>`. That is, for e.g. `acq-sag_T1w`, `acq-ax_T1w`, `acq-sag_T2w`, `acq-sag_T2-star', etc. 

### Additional context

This feature can also help accommodate (registered) inputs from multiple timepoints (by essentially concatenating them) as done in [ms-challenge-2021](https://github.com/ivadomed/ms-challenge-2021).

jcohenadad:
> but also is expanded to include the substring acq-sag_<existing-contrast-suffix-here> and acq-ax_<existing-contrast-suffix-here>.

I suggest to be more inclusive, otherwise in 6 months we will have other cases (eg: `*acq-flip1*`, `*run1*`, etc.) and we will have to refactor again.

mariehbourget:
> > but also is expanded to include the substring acq-sag_ and acq-ax_.
> 
> I suggest to be more inclusive, otherwise in 6 months we will have other cases (eg: `*acq-flip1*`, `*run1*`, etc.) and we will have to refactor again.

I agree with Julien here.
If we based the pairing for multichannel only on substrings, it will work for a particular case only (ex:  `acq-sag_T2w` and `acq-ax_T2w`). However, BIDS allows for other entities between `acq` and `suffix`, so we could end-up with filenames like `acq-sag_ce-XX_run-1_T2w` and  `acq-ax_ce-YY_run-2_T2w` in the future.

[...]

PR #1044, opened 2022-01-15
Add Multi-Session/Contrasts Support to Ivadomed

dyt811:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [ ] I've assigned a reviewer

#### PR contents

- [X] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [X] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR is the first attempt address multi-session and contrasts compatibility issue to the existing `ivadomed` process by significantly overhaul the 

### Preamble
   * This PR fundamentally change the nature of how indexing of the derivatives files work. Those changes are concentrated in the `ivadomed/loader/bids_dataframe.py` with its significantly revised multi stage filtering process for bids dataframes. Before examining the detailed changes, **I recommend to review the new unit tests at `test_loader_multi`** to see if those cases and the expected output (which you can get form the data_test_multi repo linked here: 
   * https://github.com/MotionCorrect/data_multi-sessions-contrasts/archive/refs/tags/v2022-01-06.zip
   * Beside that, this PR also add a few minor things here and there that was used to help develop it
      * Add [stepping debugging unit test to automate_training to enable step wise debugging](https://github.com/ivadomed/ivadomed/pull/1044/files#diff-8b2d531f36d5b0a30be8abc33bf837becdebede8930610c0d325bc9b40c39777R38) by calling its core function directly instead of via CLI/API to enable easier debugging during development
      * Adopt [pytest-cases](https://github.com/ivadomed/ivadomed/pull/1044/files#diff-c9c796259f3852b51b531b79cbf07820145088d4f108fb006ab44b15f0b15934R9) for easier test case scenario management process. Another PR is being prepared by @konantian to revamp it for ALL unit/functional tests to streamline the adopt the pytest-cases. https://github.com/ivadomed/ivadomed/pull/1071
[...]

Issue #1009, opened 2021-12-13
Multi-Contrast Multi-Sessions Feature Potential Discussions.


dyt811:
### Discussed in https://github.com/ivadomed/ivadomed/discussions/1003

<div type='discussions-op-text'>

<sup>Originally posted by **dyt811** December 13, 2021</sup>
Hi all. 

We are finally at a sufficiently refactored and base code improved stage to merge earlier radical experimental branch of changes. This would be a good time to revisit some key assumptions which are needed to resolve challenges with adopting multi-sessions/contrasts support: 

## Key Assumptions: 

* All modality/session input data are coregistered by user already (and of same dimension!). 

* For a subject to be included, it must have ALL [`Targeted Modality`] x [`Targeted Session`] data. Missing any session or modality will result in exclusion. 
   * Rationale: if we start accommodating either missing modalities or missing sessions, the input vector will not be stable (since they need to be FIXED dimension) and will lead to require some very non-scalable solution to guess which session/modality to use for the missing ones or interpolation across modality/session which require many experiment specific knowledge. 
   * For the initial release, we focus on the bare minimal cases: all modality all session data are present, fixed and of **same dimension** at least within each subject. 
  
## Clarification Questions: 

* Currently, it seems we don't require all `target_suffix` data to be present (I believe?). Keeping this? 
   * Rationale: if all are data are coregistered with each other, at least ONE of the `target_suffix` (aka ground truth) is sufficient?  

* What about `roi_suffix`? Right now it is incorporated into `target_suffix` and the code just check if it is one of the derivatives detected? 

* Should we consider ground truth outside of the targeted modality? See below: Currently STRICT [target modality] match.
![image](https://user-images.githubusercontent.com/6720068/145754667-52378b70-9c00-48b3-b804-27fe02890947.png)
</div>

dyt811:
Per answers gathered from the meeting: 

1) We want to keep this. As long as at least one ground truth is defined. We keep the subject. 
   * Example: if subject has t1 target_suffix when config parameters requires both t1 and t2 target suffix (a list of t1/t2). 

2) ROI suffix is used to define ROI region (as a kind of spatial filtering). Combining ROI suffix and Target suffix for indexing purpose is FINE BUT we cannot leave ROI suffix inside target suffix for ground truth definition. 
   * [The `suffix` property in `roi_params`](https://ivadomed.org/configuration_file.html?highlight=roi_params#roi-params) is not required for all studies 
[...]

PR #1201, opened 2022-10-17
Prototype Incorporation of Loader v2 with Alternative File based Loader Config JSON Support

dyt811:
#### GitHub

#### PR contents

## Description

For context, please see #1087 #1044 #1009 #1102

TLDR: 

* This introduce a optional LoaderV2 data loading path which COEXISTS with existing BIDS Based file indexing system, entirely triggered by three additional keywords in the config json files: 
  * `dataset_groups` which is a list of dictionaries. See example at `ivadomed/config/example_loader_v2_configs.py`
  * `expected_input` determines the channels of input to the model. This will overwrite any subdataset and filter out those without matching number of input. 
  * `expected_gt`  determines the channels of input to the model. This will overwrite any subdataset and filter out those without matching number of gt. 

* New Data Loader Classes at 
   * `ivadomed/loader/consolidation.py`   Consolidation class used to combine data from various dataset_groups etc. 
   * `ivadomed/loader/all_dataset_group.py`   Highest level data organization class used to represent ALL data used for training
   * `ivadomed/loader/dataset_group.py`  Mid level data organization class used to represent a set of train/val/test data used
   * `ivadomed/loader/files_dataset.py`  Low level data organization class used to represent a set of files. 
   * `ivadomed/loader/generalized_loader_configuration.py`  Aggregated configuration information to streamline IO. 

* `Train.py` and `Main.py` reorganization to support this optional analyses path. 
   
* Mocker: to help generate BIDS like mocking NII file structure utilized for testing. 
  * Located at `testing/mocker` module. 

* Schematics available at https://drive.google.com/file/d/1G69tgZzktcdbvSu48dCRL6QPVavZryxF/view?usp=sharing 
* Example JSON POC at https://drive.google.com/file/d/1-XrLOGc-4vW-H-px7rrDH1r2ca4aEdlZ/view?usp=sharing 

## Original Objectives of the POC: Aggregator and Centralized Config
  
    [X] Centralize Parameters to MRI2DSeg/3DSeg calls (~10 places)
    [X] Build JSON parser for File Path Type Dataset
    [X] Introduce Min/Max File Number validation mechanism
    [X] Introduce Incorrect File Number handling mechanism
    [X] Introduce Backward compatible Loader-Parameters key changes to existing JSON
    [ ] Update BIDS DataSet to incorporate the additional files
    [X] Build DataSetAggregator to concatenate the filename_pairs from File Path dataset types
  
## Other clean ups: 
* load_fileset.py can probably be removed. 

## Limitations: 

There are still MANY limitations that we aim to address in the future update:
* No metadata handling 
* Conflicting/unexpected files/dataset are dropped during aggregation process
* No train/val/test splitting automatically. Must manually specify the split. 

## How to Test:
Since this is a fully featured PR, it comes with its own unit test at `testing/functional_tests/test_training.py`. Beside that, it is fully transparently integrated with existing Ivadomed training process. The only update required are all on the JSON aspect so as long as the new keywords are properly prepared using Loader V2 format, everything else stays mostly the same for a basic training process in theory. See  `testing/functional_tests/test_training.py` (only ~30 lines) to get a sense of how the JSON is updated to enable the V2 loader training.  We expect test training to reveal more issues. 

## Linked issues

Partially address some of the problem of #1087 #1044 #1009 #1102


dyt811:
Running into some minor hard coded pathing issue for unit tastings. Will fix shortly.

ivadomed/main.py
```diff
@@ -362,185 +415,401 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False, no_
...
 
     """
-    command = copy.deepcopy(context[ConfigKW.COMMAND])
+    # Command can only be: segment/train/test
+    command: str = copy.deepcopy(context[ConfigKW.COMMAND])
+
+    #################################
+    #  Loader Version Determination:
+    #################################
+    # If the "contrast_params" key is present, we utilize that new loader pathway.
```
mariehbourget, line 424:
```suggestion
    # If the "contrast_params" key is not present, we utilize that new loader pathway.
```

IIUC the code below:
"contrast_param" present = TRADITIONAL_BIDS_LOADER
"contrast_param" not present = MULTI_PATH_LOADER (V2 loader)

ivadomed/loader/generalized_loader_configuration.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
```
mariehbourget, line 1: I have some questions about this file (which I don't think we covered during the meeting).

Is it only used for the V2 loader?
Does it supersede the default configuration keys from `config_default.json` that are set by `config_manager.py` when running ivdomed?

ivadomed/main.py
```diff
@@ -362,185 +415,401 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False, no_
...
+    cuda_available, device = imed_utils.define_device(context.get(ConfigKW.GPU_IDS)[0])
+
+    # Probably need to invoke the aggregator around here.
+    # Parse the DataSets keys.
+    ####
+    # BIDS Data Set Section
+    ####
+
+    #
+    from ivadomed.loader.all_dataset_group import AllDatasetGroups
```
mariehbourget, line 681: I'm confused here.
My understanding was that the new "execute_multi_path_training" that we are reviewing now is for the "FILES" path only (as in the config in `example_loader_v2_configs.py`).

Here the code is under " # BIDS Data Set Section" and the "# File Data Set Section" on line 706 is empty.
I also assume that the "# RegEx Data Set Section" section on line 710 is empty as well and that the code from line 713 apllies for all 3 sections (BIDS, FILES and REGEX).

So just to be sure, the goal here is to review for the new FILES path only (not the new BIDS one)?

ivadomed/keywords.py
```diff
@@ -274,3 +276,58 @@ class SegmentationPairKW:
     INPUT_METADATA = "input_metadata"
     GT = "gt"
     INPUT = "input"
+
+@dataclass
```
kanishk16, line 280: I'd guess you'd like to add `@dataclass(frozen=True)`

ivadomed/keywords.py
```diff
@@ -274,3 +276,58 @@ class SegmentationPairKW:
...
+class CommandKW:
+    TRAIN = "train"
+    SEGMENT = "segment"
+    TEST = "test"
+
+@dataclass(frozen=True)
+class DataloaderKW:
+    # AllDatasetGroups Level
```
kanishk16, line 288: Perhaps a one-liner doc string just to give a little context

ivadomed/config/config_dataset.json
```diff
@@ -0,0 +1,191 @@
...
+                    "expected_gt": 1,
+                    "missing_files_handle": "drop_subject",
+                    "excessive_files_handle": "use_first_and_warn"
+
+                }
+            ],
+            "validation":[ Same Above Dict struct],
+            "test":[ Same Above Dict struct],
+            "training_validation":[ Same Above Dict struct],
+            "training_validation_test":[ Same Above Dict struct]
```
kanishk16, line 89: A note to myself: maybe I'll learn something new, but check if this works or skipped while testing

ivadomed/config/example_loader_v2_configs.py
```diff
@@ -0,0 +1,398 @@
+from __future__ import annotations
```
kanishk16, line 1: Optional in Python 3.7 to 3.10 ref: https://docs.python.org/3.11/library/__future__.html

ivadomed/keywords.py
```diff
@@ -274,3 +276,58 @@ class SegmentationPairKW:
...
+    MISSING_FILES_HANDLE: str = "missing_files_handle"
+    EXCESSIVE_FILES_HANDLE: str = "excessive_files_handle"
+
+    SUBSET_LABEL: str = "subset_label"
+    PATH_DATA: str = "path_data"
+
+
+@dataclass(frozen=True)
+class DatasetTypeKW:
+    # Used to specify the datatype in the display_selected_transfoms
```
kanishk16, line 316: Just converting the comment into a doc string would be a little more pythonic...

ivadomed/keywords.py
```diff
@@ -274,3 +276,58 @@ class SegmentationPairKW:
...
+    TRAINING = "training"
+    VALIDATION = "validation"
+    TESTING = "testing"
+
+@dataclass(frozen=True)
+class FileMissingHandleKW:
+    # Keywords used to specify the handling of missing files
```
kanishk16, line 323: same as above

ivadomed/keywords.py
```diff
@@ -274,3 +276,58 @@ class SegmentationPairKW:
...
+    SKIP = "skip"
+    RAISE = "raise"
+    IGNORE = "ignore"
+
+@dataclass(frozen=True)
+class FileExcessiveHandleKW:
+    # Keywords used to specify the handling of excessive files
```
kanishk16, line 330: same as above

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+        self.name: str = dict_files_pairing.get(DataloaderKW.SUBSET_LABEL, "DefaultFileDatasetLabel")
+
+        # We assume user explicitly provide the subject lists so WE do not do any additional filtering
+
+        # Currently does not support contrast balance (See BIDS Data 2D for reference implementation)
+        # Create a dictionary with the number of subjects for each contrast of contrast_balance
+
+        # NOT going to support bounding boxes
+
+        # NOT going to care about multi-contrast/channels as we assume user explicitly provide that.
```
kanishk16, line 66: Just a nitpick: I was thinking of having something like 
```
# Assumptions:
1.
2.
...

TODOs: Add support for bounding box
```

ivadomed/loader/loader.py
```diff
@@ -69,21 +69,6 @@ def load_dataset(bids_df, data_list, transforms_params, model_params, target_suf
...
-    #                                           target_suffix=target_suffix,
-    #                                           slice_axis=imed_utils.AXIS_DCT[slice_axis],
-    #                                           transform=tranform_lst,
-    #                                           metadata_choice=metadata_type,
-    #                                           slice_filter_fn=SliceFilter(**slice_filter_params,
-    #                                                                                         device=device,
-    #                                                                                         cuda_available=cuda_available),
-    #                                           roi_params=roi_params,
-    #                                           object_detection_params=object_detection_params,
-    #                                           soft_gt=soft_gt)
```
kanishk16, line 86: I'm kinda inclined towards not changing this partly bcz I lack knowledge about HEMIS_UNET and the related functionality, but there might definitely be a reason for using HDF5Dataset that might help us in giving a lead to deal with large datasets.

ivadomed/main.py
kanishk16: This direction is absolutely amazing. I was afraid of breaking stuff in our current pipeline until everything with multi-contrasts & multi-session lands in place but just dividing till the multi-contrast & multi-session pipeline would give us time to fix if there are any hiccups that are realized later on.

kanishk16 (commented):
Just did the first pass to this PR. More or less everything looks seamless, and in line with the conversations we've been having... added a few comments on the files I had a look at. I'll have to read about mocker and then get started with testing it the loaders.

Note for first-time reviewers as well as to me: Although the PR looks really big, it's essentially config + loader + tests and training pipeline. It might be helpful to this PR to break into these components or perhaps something Yang might be thinking of.

ivadomed/config/example_loader_v2_configs.py
```diff
@@ -0,0 +1,398 @@
...
+            ],
+            DataloaderKW.EXPECTED_INPUT: 2,
+            DataloaderKW.EXPECTED_GT: 1,
+        },
+    ],
+    DataloaderKW.EXPECTED_INPUT: 2,
+    DataloaderKW.EXPECTED_GT: 1,
+}
+
+example_1i1o_all_dataset_groups_config_json: dict = {
```
mariehbourget, line 233: Just for my understanding:

2i1o above is the multichannel case.
1i10 here is a single channel case, and when 2 subject files (inputs) are provided instead of 1, the second one is discarded.

Is that right?

ivadomed/config/config_dataset.json
```diff
@@ -0,0 +1,191 @@
+{
```
mariehbourget, line 1: Would it be possible to update this config with a working "FILES" path?
I tried doing a "real-life" training from the CLI by following the structures in `example_loader_v2_configs.py` but I was unable to modify the config and make it work.
In particular, I'm not sure to understand what the different levels of `expected_input` and `expected_gt` stand for and I got this error:
```
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 898, in run_main
    run_command(context=context,
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 541, in run_command
    execute_multi_path_training(
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 693, in execute_multi_path_training
    all_data = AllDatasetGroups(
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/all_dataset_group.py", line 45, in __init__
    self.n_expected_input: int = int(dict_all_datasets_group_spec.get(DataloaderKW.EXPECTED_INPUT, 0))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

ivadomed/loader/dataset_group.py
```diff
@@ -0,0 +1,159 @@
...
+            if a_test_dataset[DataloaderKW.TYPE] == "FILES":
+                test_set = Files3DDataset(a_test_dataset, config)
+                self.test_dataset.append(test_set)
+                self.test_filename_pairs.extend(test_set.filename_pairs)
+
+    def validate_update_2d_train_val_test_dataset_and_filename_pairs(self, config, dict_file_group_spec):
+        # Instantiate all the 2D Train/Val/Test Datasets
+        for a_train_dataset in dict_file_group_spec[DataloaderKW.TRAINING]:
+
+            if a_train_dataset[DataloaderKW.TYPE] == "FILES":
```
mariehbourget, line 86: Just for my understanding, in the long-term, we would have an `elif` for the REGEX path and `elif` for the BIDS path here?
Is that correct?

ivadomed/loader/dataset_group.py
```diff
@@ -0,0 +1,159 @@
...
+                self.val_filename_pairs.extend(val_set.filename_pairs)
+        for a_test_dataset in dict_file_group_spec[DataloaderKW.TEST]:
+            if a_test_dataset[DataloaderKW.TYPE] == "FILES":
+                test_set = Files3DDataset(a_test_dataset, config)
+                self.test_dataset.append(test_set)
+                self.test_filename_pairs.extend(test_set.filename_pairs)
+
+    def validate_update_2d_train_val_test_dataset_and_filename_pairs(self, config, dict_file_group_spec):
+        # Instantiate all the 2D Train/Val/Test Datasets
+        for a_train_dataset in dict_file_group_spec[DataloaderKW.TRAINING]:
```
mariehbourget, line 84: I think I'm missing something here. I thought that at this level, we were dealing with only one TRAIN/VALID/TEST list par group, so I'm not sure to understand why there are several TRAIN dataset in `dict_file_group_spec[DataloaderKW.TRAINING]`.

I would have expected to have a for loop going through TRAIN, VALID and TEST and not 3 different for loops.

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+            stride,
+            config.slice_axis,
+            config.nibabel_cache,
+            config.transform,
+            config.slice_filter_fn,
+            config.patch_filter_fn,
+            config.task,
+            config.roi_params,
+            config.soft_gt,
+            config.is_input_dropout,
```
mariehbourget, line 104: Even if the behavior is always `None`, would it make sense to have also the `disk_cache` argument here (it's the only one missing for the `MRI2DSegmentationDataset` constructor.

ivadomed/loader/files3d_dataset.py
```diff
@@ -0,0 +1,38 @@
...
+
+        super().__init__(
+            dataset.filename_pairs,
+            length=config.model_params[ModelParamsKW.LENGTH_3D],
+            stride=config.model_params[ModelParamsKW.STRIDE_3D],
+            transform=config.transform,
+            slice_axis=config.slice_axis,
+            task=config.task,
+            soft_gt=config.soft_gt,
+            is_input_dropout=config.is_input_dropout,
```
mariehbourget, line 37: Even if the behavior is always `None`, would it make sense to have also the `disk_cache` argument here (it's the only one missing for the `MRI3DSubVolumeSegmentationDataset` constructor.

mariehbourget (commented):
Amazing work @dyt811!
I tested everything I could and the logic of the config, code, mocker and tests all look good to me. I left some comments/suggestions and mostly questions to make sure I understand some details correctly.

At the moment, I was not able to run it seamlessly from the CLI (see other comments) but with some minor fixes, I think it will work well for FILES and I don't foresee issues for the future BIDS path either.

I must say that while the config makes sense, it is a bit of a learning curve to manipulate so I would recommend having working examples in JSON formats for the FILES path.

As discussed before, this is not going to work for Microscopy because of missing metadata. For reference, the function [get_microscopy_pixelsize](https://github.com/ivadomed/ivadomed/blob/22a384c9e0a6d25570eada78171703d0d8d2ef85/ivadomed/loader/segmentation_pair.py#L342) is where it's going to fail because it looks for `MetadataKW.PIXEL_SIZE_UNITS` and `MetadataKW.PIXEL_SIZE` that normally come from the BIDS dataframe.

I would also have some  suggestions for the mocker regarding its compliance with BIDS but I didn't want to bloat this PR so I thought I should wait for the dedicated mocker PR instead. Let me know if you prefer I add them in this review.

ivadomed/main.py
dyt811: Yes, that is a great point. This entire section of the code is only used when doing multi_path training (see upstream screenshot) (as oppose to the original V1 path)

image

The comment section is a bit out dated now. I will update it. 

The multi-path code instantiate object types based on JSON data object type FileDataset or BIDSDataset (V2) or RegexDataset type. So far, we only implemented FIleDatset type but we plan to support long term merging of all other data type as well. 

I will correct the comment here.

ivadomed/keywords.py
dyt811: Yeah, really need a PR to fix all these actually. Good point.

ivadomed/config/config_dataset.json
dyt811: Yeah, my bad. I think that is just some example. I believe this JSON will be discarded in the PR update.

ivadomed/keywords.py
dyt811: Good point. Will make these changes.

ivadomed/loader/loader.py
dyt811: So, what happened was a while back, HEMIS_UNET was deprecated by us, so that was why it was commented out to be eventually removed.

ivadomed/config/example_loader_v2_configs.py
dyt811: Absolutely. Perfectly correct.

ivadomed/loader/loader_files.py
```diff
@@ -0,0 +1,73 @@
+import copy
```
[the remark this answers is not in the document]
dyt811: Correct. I will remove this file. I think the original plan was to have a dedicated loader per each Dataset type (files vs Bids vs Regex) but that has since been superseded by the JSON specific FIlesDataset and future BIDsDataset/RegexDataset design.

ivadomed/loader/generalized_loader_configuration.py
mariehbourget: Ok so ultimetately, will we need to maintain both config_default.json and this file up-to-date or only this file?

ivadomed/loader/balanced_sampler_generic.py
```diff
@@ -0,0 +1,74 @@
+import torch
```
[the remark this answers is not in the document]
dyt811: So this file depends on BIDSDataset to function and we may need its FIlesDataset version IF we enable balanced sampling. In this PR I have not enabled it as it was disabled by default in training.py get_sampler. 

I stubbed it out but turns out for now, we won't need it. 

Will remove as well.

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
dyt811: Yeah, partially. 

So `files_dataset` has two major part, one part ,is the `files_pairs`, the other part is the config: `GeneralizedLoaderConfiguration`. 

The 2nd part ties into your earlier questioon on why that class is used (mostly when we construct new FIlesDataset) and contain information to make it compatible with MRI2DSegmentationDatset.

ivadomed/config/config_dataset.json
dyt811: Good point. I think config_dataset.json was suppose to fill that role but I did not fully flush out that example so I will do that to make CLI testing more viable as well based on Mocker results instead of just from unit test.

ivadomed/loader/dataset_group.py
dyt811: Yes, precisely! That is the place where other path will be introduced.

ivadomed/loader/dataset_group.py
dyt811: Great question. We technically support a LIST of training dataset. This way, train dataset could compose of many FIlesDataset/RegexDataset/BIdsDataset etc. All aggregated at the Datset Group level. Limit to a single TRAIN is a bit arbitrary and not flexible enough and the code is not that much complex (as you saw, three loops now instead of one loop). But underlying the dataset object work similar.

ivadomed/loader/files_dataset.py
dyt811: Great catch I will add this.

ivadomed/loader/files_dataset.py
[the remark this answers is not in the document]
mariehbourget: Thanks for the clarification. It took me a while but I think I get it now.
It is called `FilesDataset` because it serves the same purpose as the current (and future) `BidsDataset` and we would eventually have a `RegexDataset` as well, is that correct?

ivadomed/loader/dataset_group.py
mariehbourget: Oh yeah, ok I get it, you could have a list of multiples FILES, or a FILES and a BIDS dataset, in the same Dataset Group.

I still wonder if that is necessary because we already have a list of DatasetGroup, so if we want to specify 2 different training lists, we could separate them in 2 DatasetGroup, no? With each training group being either FILES, BIDS or REGEX.

I'm saying this because right now we have 2 differents ways of dealing with this (IIUC). Let's say we have a FILES training set and a BIDS training set, we could:
1. Split the FILES and BIDS in different DatasetGroup
2. List them both in the training set of the same DatasetGroup.
From a user standpoint, if this flexibility is not absolutely required, it may be confusing to have 2 ways of doing this (and 2 ways of filling the config file).

Which brings me to yet another question:
It is possible to have only a training set in a DatasetGroup (no valid/test sets) or do we need to specify train/valid/test for all DatagsetGroup?

dyt811: > I still wonder if that is necessary because we already have a list of DatasetGroup, so if we want to specify 2 different training lists, we could separate them in 2 DatasetGroup, no? With each training group being either FILES, BIDS or REGEX.

Yes. Correct. Current design is flexible enough that however users want to do it, they can do it. The backend code does not really have strong preference either way. 

From what I have seen so far, I believe at DatasetGroup level, there is no requirement for validation/test set (given that we have mostly focused on the core training path). 

However, at the AllDatasetGroup level, I believe at least validation set is required for proper training.

ivadomed/config/config_dataset.json
dyt811: I have now updated the `config_dataset.json` with a fully flushed out example adapted from test_training.py. Note that you MAY have to search and replace/adjust the `"path_data": "C:\\git\\ivadomed\\tmp"` part as that is the default folder of mocker output for my running which will likely be different from yours.

ivadomed/loader/files_dataset.py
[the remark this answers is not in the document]
dyt811: yeah, `FilesDataset` is like... the current v1 `BidsDataset` but without `BIDS`. I am still debating if `RegexDataset` needs to be a separate class or based on existing `FilesDataset` implementation. BidsDataset V2 will likely be a rewrite but leverage some similar high level logic from `FilesDataset`... to be finalized.

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+        logger.info(f"\t\tFileDataset object from {self.path_data}, {len(self.filename_pairs)} pairs of data files")
+
+        if verbose:
+            for pair_index, a_pair in enumerate(self.filename_pairs):
+                logger.info(f"\t\t\tImage Pair {pair_index}, Subject Image(s):")
+                for a_image in a_pair[0]:
+                    logger.info(f"\t\t\t\t{a_image}")
+                logger.info(f"\t\t\tImage Pair {pair_index}, Ground Truth Image(s):")
+                for a_gt in a_pair[1]:
+                    logger.info(f"\t\t\t\t{a_gt}")
```
kanishk16, line 239: Since this was unaligned in terms of whitespace and a little unintuitive to me so I refactored it to:
```
    def preview(self, verbose: bool = False):
        """
        Preview the FINALIZED FileDataSet included that has undergone validation and is ready to be used for training.
        Args:
            verbose: whether to print out the actual data path
        """
        logger.info(f"FileDataset object from {self.path_data}, {len(self.filename_pairs)} pairs of data files")
        
        if verbose:
            for pair_index, (imgs, gts, _, _) in enumerate(self.filename_pairs):
                logger.info(f"Image Pair {pair_index}, Subject Image(s):")                
                for img in imgs:
                    logger.info(f"\t{img}")
                
                logger.info(f"Image Pair {pair_index}, Ground Truth Image(s):")
                for gt in gts:
                    logger.info(f"\t{gt}")
```
In a lot of places I feel we could use something like 
`for pair_index, (imgs, gts, _, _) in enumerate(self.filename_pairs):` which makes everything more readable 🤷‍♂️

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+        length = (
+            config.model_params[ModelParamsKW.LENGTH_2D]
+            if ModelParamsKW.LENGTH_2D in config.model_params
+            else []
+        )
+        stride = (
+            config.model_params[ModelParamsKW.STRIDE_2D]
+            if ModelParamsKW.STRIDE_2D in config.model_params
+            else []
+        )
```
kanishk16, line 89: Todo: I'm not sure if it's really needed and if required we could move it to `GeneralizedLoaderConfig`

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+            config.task,
+            config.roi_params,
+            config.soft_gt,
+            config.is_input_dropout,
+        )
+
+    def parse_spec_json_and_update_filename_pairs(self, loader_json: dict) -> Tuple[str, list]:
+        """Load the json file and return the dictionary"""
+        # Given a JSON file, try to load the file pairing from it.
+        assert loader_json.get(DataloaderKW.TYPE) == "FILES"
```
kanishk16, line 110: To avoid typos & waste time on debugging:

`assert loader_json.get(DataloaderKW.TYPE).upper() == "FILES", f'Invalid {loader_json.get(DataloaderKW.TYPE)}!'`

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+
+        list_image_ground_truth_pairs: list = loader_json.get(DataloaderKW.IMAGE_GROUND_TRUTH)
+        list_image_ground_truth_pairs_filtered: list = []
+        # Go through each subject
+        for a_subject_image_ground_truth_pair in list_image_ground_truth_pairs:
+
+            skip_subject_flag = False
+
+            # 2 lists: Subject List + Ground Truth list
+            assert len(a_subject_image_ground_truth_pair) == 2
```
kanishk16, line 154: Similarly something here as: 
`assert len(a_subject_image_ground_truth_pair) == 2, f'Either subject imgs or Ground Truth(s) missing!'
`

ivadomed/loader/files_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+
+            # Go check every file and if any of them don't exist, skip the subject
+            for subject_images_index, path_file in enumerate(
+                    list_subject_specific_images
+            ):
+                list_subject_specific_images[subject_images_index] = ensure_absolute_path(
+                    path_file, path_data
+                )
+                if not list_subject_specific_images[subject_images_index]:
+                    skip_subject_flag = True
```
kanishk16, line 174: I was thinking to push this above i.e before trimming the lists bcz it may happen, an edge case, there are 3 inputs given but while slicing first 2 are taken, and one of them doesn't exist and the one rejected while slicing exists then we might unnecessarily be rejecting the subject...

Also, I would like to maintain a list of subjects or filename pairs being rejected and then display all of them in the end as a warning rather than each one (Todo: check again how does it come across!)

kanishk16 (commented):
Phase 2 (Second pass): We could have a lot of tests added whenever an assertion error is thrown, and I believe we could also slightly improve the variable names realizing that this is not a polished version and more of an end-to-end sanity check for the complete pipeline.

There are still a small set of things that I'd be focusing on in my next review primarily related to the integration of loader v2 with existing json (intuitive keys, checking & raising errors for keys not found and perhaps test the end to end with the CLI!)

ivadomed/main.py
dyt811: Good point. This is the relevant CLI section you were talking about. I will play around with the trigger to use simpler conditions: 

image

dyt811: Now updated to check solely for the three new keys as per suggested. Worked like a charm.

ivadomed/config/config_dataset.json
```diff
@@ -0,0 +1,480 @@
...
+                    ],
+                    "expected_input": 2,
+                    "expected_gt": 1,
+                    "missing_files_handle": "skip",
+                    "excessive_files_handle": "use_first_and_warn",
+                    "path_data": "C:\\git\\ivadomed\\tmp"
+                }
+            ],
+            "expected_input": 2,
+            "expected_gt": 1
```
mariehbourget, line 163: Sorry Yang, I know you explained it a few times but I'm still missing something regarding the different levels of `expected_input` and `expected_gt` 😅 

So the tensor that we feed the network is always the same size i.e. all pairs must have the same dimensions. If we have the expected numbers and do the "clean-up" (missing, excessive files) at the dataset level, why do we need them at higher level? Won't they always be the same? What happens if a higher level has a different expected numbers than the dataset level?

I just don't get how they could be different from dataset to dataset and from levels to levels. And if they have to always be the same, could we specify those expected numbers at an unique place in the config?

ivadomed/main.py
```diff
@@ -362,185 +414,381 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False, no_
...
+                                     testing_params=testing_params,
+                                     path_output=path_output,
+                                     device=device,
+                                     cuda_available=cuda_available,
+                                     metric_fns=metric_fns,
+                                     postprocessing=context[ConfigKW.POSTPROCESSING])
+    # RUN EVALUATION
+    df_results = imed_evaluation.evaluate(bids_df, path_output=path_output,
+                                          target_suffix=loader_params[LoaderParamsKW.TARGET_SUFFIX],
                                               eval_params=eval_params)
```
mariehbourget, line 787: Just a heads-up, while the "test" function uses a "dataset" and will be compatible with the V2 loader, the evaluate function still relies heavily on the bids_dataframe and the fact that the dataset is organized as BIDS.

We will have to fix this before having the V2 FILES loader up and running (as phase 5 maybe?)

ivadomed/main.py
```diff
@@ -362,185 +414,381 @@ def run_command(context, n_gif=0, thr_increment=None, resume_training=False, no_
...
+    transform_valid_params, \
+    transform_test_params = imed_transforms.get_subdatasets_transforms(context[ConfigKW.TRANSFORMATION])
 
-    # MODEL PARAMETERS
+    # Update and populate model parameters AND loader parameters
     model_params, loader_params = set_model_params(context, loader_params)
 
-    if command == 'segment':
+    if command == CommandKW.SEGMENT and loader_version == LoaderParamsKW.TRADITIONAL_BIDS_LOADER:
         run_segment_command(context, model_params, no_patch, overlap_2d)
```
mariehbourget, line 465: Just a heads-up, the `--segment` command also relies heavily on BIDS (but is less critical than the `--test` command because we can more easily use the `segment_volume `API instead).

As a reference, we were already thinking about revamping this here: https://github.com/ivadomed/ivadomed/issues/1002

mariehbourget (commented):
Thanks @dyt811 for the new config file and fixing the CLI.
I was able to run a real training with MRI data and complete the review.

I can only run a microscopy training if I hardcode the needed metadata [here](https://github.com/ivadomed/ivadomed/blob/dc9df1fad04fb3845a836dcee4e3729b75b8c6d9/ivadomed/loader/segmentation_pair.py#L329) and bypass the `get_microscopy_pixelsize` function.
So we will need to think about a way to incorporate metadata as well in the FILES pathway.

Other than that, I would consider the PR ready to be split into smaller chunks.

kanishk16 (commented):
Thanks @mariehbourget, for raising this issue (https://github.com/ivadomed/ivadomed/pull/1201#discussion_r1022108012). I second this, and apart from this, I've some other design issues to add as well. IIUC looks like this PR deviates from the discussed specification https://github.com/ivadomed/ivadomed/issues/1087#issuecomment-1129253222

How the json (in its current form in this PR):
```
"dataset_groups": [
    {
        "dataset_group_label": "DataSetGroup1",
        "training": [
            {
                "type": "FILES",
                "subset_label": "TrainFileDataSet1",
                "image_ground_truth": [
                    [
                    ],
                    [
                    ],
                    [
                    ],
                    [
                    ]
                ],
                "expected_input": 2,
                "expected_gt": 1,
                "missing_files_handle": "skip",
                "excessive_files_handle": "use_first_and_warn",
                "path_data": "path\2\data"
            }
        ],
        "validation": [
        ],
        "test": [
        ],
        "expected_input": 2,
        "expected_gt": 1
    },
    {
        "dataset_group_label": "DataSetGroup2",
        "training": [
        ],
        "validation": [
        ],
        "test": [
        ],
        "expected_input": 2,
        "expected_gt": 1
    }
],
"expected_gt": 1,
"expected_input": 2,
```

How I'd like it to be:
```
"dataset_groups": [
    {
        "dataset_group_label": "DataSetGroup1",
        "type": "FILES",
        "training": {
                "image_ground_truth": [
                    [
                    ],
                    [
                    ],
                    [
                    ],
                    [
                    ]
                ],
                "missing_files_handle": "skip",
                "excessive_files_handle": "use_first_and_warn",
            },
        "validation": {
        },
        "test": {
        },
        "path_data": "path\2\data",
        "expected_gt": 1,
        "expected_input": 2,
    },
    {
        "dataset_group_label": "DataSetGroup2",
        "type": "FILES",
        "training": [
        ],
        "validation": [
        ],
        "test": [
        ],
        "path_data": "path\2\data",
        "expected_gt": 1,
        "expected_input": 2,
    }
]
```

Changes that I'd like to propose to align with the discussed specification: (assuming IIUC each dataset group represents data from a single source. Hence, it might be either in BIDS format or organized as per their lab (in that case, we'd be inclining either to BIDsify or use File format or Regex File format):

1. Move the `type` and `path_data` to the dataset group level
2. Removal of `subset_label` since I couldn't think of its use case (happy to change my mind in case there's one)
3. `training` (`validation` and `test`) to be a `dict` instead of a `list` for faster traversal in terms of time complexity as well as use case described in 4. (edit: read it later but similar opinions https://github.com/ivadomed/ivadomed/pull/1201#discussion_r1014155368, yet another reason; as well as counter-intuitive to the use case presented in 1.)
4. `training` (`validation` and `test`) to be optional for each dataset group, but while combining each training, validation, and test must contain data (Why? Currently, it isn't possible to have training data from one dataset group, validation from another, and testing from the last one. I feel having data from multiple sites but not being able to use it in this way would defeat the purpose of having data from multiple sites) 

Let me know if this makes sense 😅

ivadomed/config/config_dataset.json
kanishk16: Related https://github.com/ivadomed/ivadomed/issues/1087#issuecomment-1081039335 but still agree that the `expected_input` and `expected_gt` must be at the highest level to accommodate the above comment by Naga.

ivadomed/loader/consolidation.py
```diff
@@ -0,0 +1,141 @@
...
+
+if typing.TYPE_CHECKING:
+    from ivadomed.loader.all_dataset_group import AllDatasetGroups
+from ivadomed.loader.generalized_loader_configuration import GeneralizedLoaderConfiguration
+from ivadomed.loader.mri2d_segmentation_dataset import MRI2DSegmentationDataset
+
+from loguru import logger
+
+
+class ConsolidatedDataset(MRI2DSegmentationDataset):
```
kanishk16, line 16: IIUC the hierarchy: `FilesDataset` [2D/3D] < `DatasetGroup` < `AllDatasetGroup` < `ConsolidatedDataset`

On reviewing the code, I guess: 
1. The `ConsolidatedDataset` class should also inherit `MRI3DSubVolumeSegmentationDataset` to address the 3D case. So, while calling their constructors, we could do something along the lines of:
```
class ConsolidatedDataset(MRI2DSegmentationDataset, MRI3DSubVolumeSegmentationDataset):
    def __init__(
        self, filename_pairs: list, config: GeneralizedLoaderConfiguration
    ):
        ...

        if (ModelParamsKW.IS_2D not in config.model_params and not config.model_params.get(ModelParamsKW.IS_2D)):
            # call constructor of MRI3DSubVolumeSegmentationDataset 
            MRI3DSubVolumeSegmentationDataset.__init__(
                ...
            )
        else:
            # call constructor of MRI2DSegmentationDataset
            MRI2DSegmentationDataset.__init__(
                ...
            )

```
2. Moreover, I would rather suggest merging `ConsolidatedDataset` into `AllDatasetGroup` based on their functionality. This is just a suggestion as I believe creating a separate class for the functions we have in `ConsolidatedDataset` is an overkill.

kanishk16 (commented):
Phase 3 reviews: We could do better with the naming convention, and I've addressed a few design-related issues that bugged me. Once the design issues are fixed including https://github.com/ivadomed/ivadomed/pull/1201#pullrequestreview-1186663781, I might add another set of reviews to handle a few edge cases.

ivadomed/loader/files_dataset.py
kanishk16: Todo: add a test for the edge case I have in mind as mentioned above

ivadomed/loader/generalized_loader_configuration.py
dyt811: This file will hopefully replace all the above signatures parameters at the end of BIDSDataframe update. FileDataset no longer use above parameters anymore. It uses this class instead (same information, organized slightly differently)

BIDSDataframe will be udpated to use it during its upcoming rewrite. 

MRI2DSegmentation Dataset constructor (shown above) will also by then use the same configuration object. 

User perspective, they will be using V2 standard. User do not touch this configuration class file. This class is used to internally store and represents all the configurations to simplify interfaces and provide internal validations/consistency checks. Nothing much more than a a bit fancier tuple structure in a sense.

ivadomed/config/config_dataset.json
dyt811: Great question. It is a design choice to cover scenarios where there are conflicts... 

So... AllDataset Group Level: Highest. If not the same, hard pass any DatasetGroup within it. 
At Dataset Group Level: Higher but not the highest. If not the same at BIDSDataset/FIleDataset, hard pass any of these "dataset" regardless if they are in train/test/val etc. 
At DataSet level, it will skip any subjects that have less, usually trim (based on list order) of more than provided. So at the end, that dataset is guarnteed to have X for input and Y for GT, with validated file presence guarnteed.

ivadomed/main.py
dyt811: Yeah, correct. I think Metadata and Test function (or most things outside of "training) will need to be in another PR before BIDS Dataset update.

ivadomed/main.py
dyt811: Same as above. This is another significant part of the FileDataset changes I think we should undertake before BIDSDataframe changes.

testing/unit_tests/test_file_dataset.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+
+def setup_function():
+    create_tmp_dir()
+
+
+def test_FilesDataset():
+
+    create_mock_bids_file_structures(path_mock_data),  # pytest fixture, do not remove.
```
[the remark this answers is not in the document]
dyt811: Mocker has been reparametrized.

testing/mocker/mocker_fixture.py
```diff
@@ -0,0 +1,81 @@
...
+        },
+        {
+            "flip": [1, 2],  # Flip angle
+            "mt": ["on", "off"],  # MT
+            "MODALITY": ["MTS"],  # Modality
+        },
+        {
+            "flip": [2],  # Flip angle
+            "mt": ["off"],  # MT
+            "MODALITY": ["MTS"],  # Modality
```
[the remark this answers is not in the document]
dyt811: The previous two dicts in the example should cover those scenarios already. Also.. since we technically don't have "contrast" concept anymore at generation time.. all keywords are just combined multiplicatively consecutively. In here the third example, the goal was actually to test single contrast stability in terms of its generation.

testing/mocker/create_common.py
```diff
@@ -0,0 +1,150 @@
...
+        self.list_bids_details = list_anatomy_modalities
+        self.list_bids_keywords = ["acq", "desc", "run"]
+
+    @staticmethod
+    @abstractmethod
+    def default_constructor(path_to_mock_data: str):
+        raise NotImplementedError
+
+    @abstractmethod
+    def run(self):
```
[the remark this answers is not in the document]
dyt811: Yeah, it is based on the parent class that must have a run. This gets defined explicitly in the inheriting class.

ivadomed/config/config_dataset.json
```diff
@@ -0,0 +1,480 @@
...
+    "model_name": "unit_test",
+    "object_detection_params": {
+        "object_detection_path": null
+    },
+    "dataset_groups": [
+        {
+            "dataset_group_label": "DataSetGroup1",
+            "training": [
+                {
+                    "type": "FILES",
```
mariehbourget, line 17: Would it be possible to update this file to reflect the latest changes to help future reviewer?
example here:  "type" is at the dataset level and not the group level and does not match the config in [example_loader_v2_configs.py](https://github.com/ivadomed/ivadomed/pull/1201/files#diff-d1fba057563ae49e7163984ee5c4160f02af3d5bbcea113f31b9a86d06e257b3).
Thanks!

ivadomed/loader/dataset_group.py
```diff
@@ -0,0 +1,295 @@
...
+            dict_file_group_spec.get(DataloaderKW.TRAINING)
+        )
+
+        self.val_filename_pairs = self.parse_simple_spec_dict_for_filename_pairs(
+            dict_file_group_spec.get(DataloaderKW.VALIDATION)
+        )
+
+        self.test_filename_pairs = self.parse_simple_spec_dict_for_filename_pairs(
+            dict_file_group_spec.get(DataloaderKW.TEST)
+        )
```
mariehbourget, line 99: From my tests, it seems that with the current implementation, all 3 datasets (train/valid/test) must be present for each dataset_group or it throws an error:
```
  File "/home/mhbourget/venv-ivadomed-296/bin/ivadomed", line 11, in <module>
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 872, in run_main
    run_command(context=context,
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 535, in run_command
    execute_multi_path_training(
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 680, in execute_multi_path_training
    all_data = AllDatasetGroups(
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/all_dataset_group.py", line 56, in __init__
    dataset_group = FileDatasetGroup(dict_dataset_group_spec, config)
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/dataset_group.py", line 61, in __init__
    self.validate_update_2d_train_val_test_dataset_and_filename_pairs(dict_file_group_spec)
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/dataset_group.py", line 93, in validate_update_2d_train_val_test_dataset_and_filename_pairs
    self.val_filename_pairs = self.parse_simple_spec_dict_for_filename_pairs(
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/dataset_group.py", line 160, in parse_simple_spec_dict_for_filename_pairs
    if a_simple_loader_json.get(DataloaderKW.MISSING_FILES_HANDLE) == FileMissingHandleKW.SKIP:
AttributeError: 'NoneType' object has no attribute 'get'
```

From previous discussion (https://github.com/ivadomed/ivadomed/pull/1201#pullrequestreview-1186663781), I think we should be able to list only 1 or 2 of these datasets per dataset_group.
ex: train and validation comes from a the dataset group 1, and test comes from the dataset group 2.
Let me know if that makes sense or if I'm mistaken.

ivadomed/config/example_loader_v2_configs.py
```diff
@@ -0,0 +1,168 @@
...
+                         ["derivatives/labels/sub-04/ses-02/anat/sub-04_ses-02_mt-off_MTS_lesion-manual-rater1.nii"]],
+                        [["sub-04/ses-03/anat/sub-04_ses-03_flip-1_mt-off_MTS.nii",
+                          "sub-04/ses-03/anat/sub-04_ses-03_flip-1_mt-on_MTS.nii",
+                          "sub-04/ses-03/anat/sub-04_ses-03_flip-1_mt-on_MTS.nii"],
+                         ["derivatives/labels/sub-04/ses-03/anat/sub-04_ses-03_mt-off_MTS_lesion-manual-rater1.nii"]],
+                        [["sub-04/ses-04/anat/sub-04_ses-04_flip-1_mt-off_MTS.nii", ],
+                         ["derivatives/labels/sub-04/ses-04/anat/sub-04_ses-04_mt-off_MTS_lesion-manual-rater1.nii"]],
+                    ],
+                    DataloaderKW.MISSING_FILES_HANDLE: FileMissingHandleKW.SKIP,
+                    DataloaderKW.EXCESSIVE_FILES_HANDLE: FileExcessiveHandleKW.USE_FIRST_AND_WARN,
```
mariehbourget, line 102: This is a detail but I could not find where this warning is raised/did not see it in my test (it may not be implemented yet, which if fine as well, just wondering).

mariehbourget (commented):
Thanks @dyt811!
I reviewed the PR with the latest changes.
* I mainly tested the CLI with my own config file and was able to start a training with MRI data.
* The new "trigger" for the multi path loader is working well (except when the unused `path_data` in the traditional loader doesn't have a valid path, but from our discussion in meeting, this is a known issue for now).
* The `config_dataset.json` currently in the branch doesn't seem to reflect the latest changes. I was able to make my own config file much more easily than before with the new JSON architecture.
* I had another small issue when a train, valid and/or test dataset was not present for a given dataset_group (see other comment in `dataset_group.py`)

Overall, I find the new JSON hierarchy a lot easier to work with, and I think the PR is mature enough to be broken in smaller chunks.

ivadomed/config/config_dataset.json
dyt811: Yeah, thanks for catching that. the example config .py was updated but I missed this example JSON. Updated now.

ivadomed/loader/dataset_group.py
dyt811: Excellent point.  I now build a early return for the validation process when the Dict is none and also carry out an internal check for when al three are missing. Thanks for bringing up this excellent point.

ivadomed/config/example_loader_v2_configs.py
dyt811: Good point. just implemented it now. (had the warning for LESS but missed MORE for some reason)

ivadomed/loader/dataset_group.py
mariehbourget: Thanks! It's working well now. I think we would also need to add a check at a higher level (all dataset or consolitdation?) in case one or more of the datasets (train/valid or test) is missing, which does not seem to be implemented at the moment and end up with an out of range error:
```
  File "/home/mhbourget/venv-ivadomed-296/bin/ivadomed", line 11, in <module>
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 870, in run_main
    run_command(context=context,
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 533, in run_command
    execute_multi_path_training(
  File "/home/mhbourget/code/ivadomed/ivadomed/main.py", line 688, in execute_multi_path_training
    ds_valid = ConsolidatedDataset.consolidate_AllDatasetGroups_to_a_specific_filedataset_type(all_data, DataloaderKW.VALIDATION)
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/consolidation.py", line 93, in consolidate_AllDatasetGroups_to_a_specific_filedataset_type
    return ConsolidatedDataset(filename_pairs, all_dataset_groups.config)
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/consolidation.py", line 45, in __init__
    super().__init__(
  File "/home/mhbourget/code/ivadomed/ivadomed/loader/mri2d_segmentation_dataset.py", line 118, in __init__
    self.n_contrasts = len(self.filename_pairs[0][0])
IndexError: list index out of range
```

ivadomed/loader/dataset_group.py
```diff
@@ -0,0 +1,307 @@
...
+            # 2 lists: Subject List + Ground Truth list
+            assert len(a_subject_image_ground_truth_pair) == 2, \
+                f"Either subject images list or groundtruth files list are missing!"
+
+            # Validate and trim Subject files list
+            list_subject_specific_images: list = a_subject_image_ground_truth_pair[0]
+            if len(list_subject_specific_images) > self.n_expected_input:
+                logger.warning(f"HIGHER number of image files found than expected, for subject specification "
+                               f"{a_subject_image_ground_truth_pair}. Expected {self.n_expected_input} "
+                               f"but found {len(list_subject_specific_images)} from {list_subject_specific_images}")
```
mariehbourget, line 206:
```suggestion
                               f"but found {len(list_subject_specific_images)}. Using image files {list_subject_specific_images[:self.n_expected_input]}.")
```

Just a suggestion to warn the user of which files are indeed used explicitly.

ivadomed/loader/dataset_group.py
```diff
@@ -0,0 +1,307 @@
...
+                list_subject_specific_images = list_subject_specific_images[:self.n_expected_input]
+
+            # Validate and trim Ground Truth files list
+            list_subject_specific_gt: list = a_subject_image_ground_truth_pair[1]
+            if len(list_subject_specific_gt) > self.n_expected_gt:
+                logger.warning(f"HIGHER number of ground truth files found than expected, for subject specification "
+                               f"{a_subject_image_ground_truth_pair}. Expected {self.n_expected_gt} "
+                               f"but found {len(list_subject_specific_gt)} from {list_subject_specific_gt}")
```
mariehbourget, line 214:
```suggestion
                               f"but found {len(list_subject_specific_images)}. Using ground truth files {list_subject_specific_images[:self.n_expected_gt]}.")
```

Same as above

ivadomed/config/config_dataset.json
mariehbourget: Thanks for the update!
May I suggest that we use the same order everywhere for the JSON fields (in the config_dataset.json and in example_loader_v2_configs.py) , for clarity for the user.

Maybe the same order as in `example_loader_v2_configs.py` :
```
DATASET_GROUPS
    DATASET_GROUP_LABEL
    TYPE
    TRAINING
        IMAGE_GROUND_TRUTH
        MISSING_FILES_HANDLE
        EXCESSIVE_FILES_HANDLE
    VALIDATION
        IMAGE_GROUND_TRUTH
        MISSING_FILES_HANDLE
        EXCESSIVE_FILES_HANDLE
    TEST
        IMAGE_GROUND_TRUTH
        MISSING_FILES_HANDLE
        EXCESSIVE_FILES_HANDLE
    PATH_DATA
    EXPECTED_INPUT
    EXPECTED_GT
EXPECTED_INPUT
EXPECTED_GT
```

OR the following which I find kind of easier to follow:
```
DATASET_GROUPS
EXPECTED_INPUT
EXPECTED_GT
    DATASET_GROUP_LABEL
    TYPE
    PATH_DATA
    EXPECTED_INPUT
    EXPECTED_GT
    TRAINING
        IMAGE_GROUND_TRUTH
        MISSING_FILES_HANDLE
        EXCESSIVE_FILES_HANDLE
    VALIDATION
        IMAGE_GROUND_TRUTH
        MISSING_FILES_HANDLE
        EXCESSIVE_FILES_HANDLE
    TEST
        IMAGE_GROUND_TRUTH
        MISSING_FILES_HANDLE
        EXCESSIVE_FILES_HANDLE
```

ivadomed/loader/consolidation.py
mariehbourget: > 1. The `ConsolidatedDataset` class should also inherit `MRI3DSubVolumeSegmentationDataset` to address the 3D case.

Yes I agree, thanks @kanishk16 for pointing this out.

I just re-tested this PR in light of this [comment](https://github.com/ivadomed/ivadomed/pull/1224#discussion_r1055622479) in PR #1224 and the current implementation is only working in  2D, which I missed before. So yes , `ConsolidatedDataset` absolutely needs to take account of both cases.

In the same way, the call for consolidated dataset [here](https://github.com/ivadomed/ivadomed/blob/11c67dc5e5a69b3922c9e7e7d79a00da8413c7e9/ivadomed/main.py#L684) will need to take into account that `load_filenames` is inside the constructor in 3D and is outside the constructor in 2D. This difference can be seen in the current bids pathway in [loader.py](https://github.com/ivadomed/ivadomed/blob/11c67dc5e5a69b3922c9e7e7d79a00da8413c7e9/ivadomed/loader/loader.py#L78) where `load_filenames` is only explicitly called for 2D.
(I don't know where this difference comes from, it may be possible to harmonize 2D and 3D?)

PR #1209, opened 2022-12-02
Fix GPU behavior in segment_volume

### LINKED ISSUE #1208: Unexpected gpu-id behaviours
## Issue description

**Full disclosure: I discovered these while reviewing a PR for AxonDeepSeg ([PR# 701](https://github.com/axondeepseg/axondeepseg/pull/701)), so I've only tested it with the settings/API that @mariehbourget used to implement that functionality there**

### Current behavior

When selecting a gpu-id other than 0 (eg. 5), memory is allocated to gpu-id 0 regardless, and it does so both prior and during the use of the given gpu-id (eg. 5).

Also, if `CUDA_VISIBLE_DEVICES` is set for the session to something other than starting from 0 (eg `CUDA_VISIBLE_DEVICES=3,4,5`, then gpu-id does not set the device number that is set, but is in fact the ID relative to the visible devices (eg for `CUDA_VISIBLE_DEVICES=3,4,5`, a gpu-id of 1 (remember that in python, count starts at 0), the gpu device that would be set would be the gpu-id 4, and not 1.

### Expected behavior
 

Only the selected gpu-id should have memory used.

gpu-id documentation should also reflect how it behaves with CUDA_VISIBLE_DEVICES values set.

### Steps to reproduce

While testing our PR, I would set the gpu-id to 5 through the ADS/IVADOMED API, but noticed while monitoring the the `nvidia-smi` that the GPU ID would get some memory allocated in addition, and it would start before the memory started being used in the ID I set to 5.

Screenshot 2022-11-25 at 3 30 58 PM

Command: `axondeepseg -t SEM -i AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test/image.png --gpu-id 5`

Looking online, some commentators say that it's because when pytorch "turns on", it automatically initializes the CUDA device using device 0 (but more on that later): https://github.com/pytorch/pytorch/issues/2383#issuecomment-323112926

People in another suggested using CUDA_VISIBLE_DEVICES so that the device 0 isn't used.

Trying this, I discovered the second behaviour bug.

Command: `CUDA_VISIBLE_DEVICES=5,6 axondeepseg -t SEM -i AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test/image.png --gpu-id 5`

Screenshot 2022-11-25 at 3 23 21 PM

The code now errors.

This is because the gpu-id (as now implemented in IVADOMED) is in fact not the ID of the GPU, but instead the ID relative to the visible devices: https://discuss.pytorch.org/t/runtimeerror-cuda-error-invalid-device-ordinal/67488

So changing it to 

Command: `CUDA_VISIBLE_DEVICES=5,6 axondeepseg -t SEM -i AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test/image.png --gpu-id 0`

Now works:

Screenshot 2022-11-25 at 3 25 23 PM

**However**, note that if the list visible devices **don't** begin with the device I want to use, then the first visible device will still have some memory used from the point earlier on, it's not **always** device 0 that has memory taken to boot up pytorch (and to this point, I don't know where and if this can be stopped/changed), but it's the **first** in the list of available devices that this will occur for, for example:

Command: `CUDA_VISIBLE_DEVICES=1,3,5,7 axondeepseg -t SEM -i AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test/image.png --gpu-id 2`

Will result in:

Screenshot 2022-11-25 at 3 28 20 PM

I thought you should all be aware of these behaviours, as I've looked at your documentation and didn't see these details noted.

I've also shown all of this to @mariehbourget on Zoom, so if anything isn't well explained, hopefully she can clarify to you.

## Environment

See the AxonDeepSeg PR linked above.

### System description
`rosenberg` machine at Poly.

### Installed packages

See the AxonDeepSeg PR linked above.

mariehbourget:
#### GitHub

#### PR contents

## Description
This PR fixes #1208 where GPU 0 was used in addition to `gpu_id` when calling `segment_volume` directly (from ADS).
GPU `0` was used in `segment_volume` and `gpu_id` was used in the subsequent `get_preds`.
So this PR moves the `define_device` function from `get_preds` to `segment_volume`.

@joshuacwnewton: As far as I know, SCT doesn't use the `gpu_id`  but I prefer to notify you when we do changes to `segment_volume`. Let me know if you foresee any issues.

## Linked issues
Fixes #1208

### DIFF
--- a/ivadomed/inference.py
+++ b/ivadomed/inference.py
@@ -45,21 +45,20 @@ def onnx_inference(model_path: str, inputs: tensor) -> tensor:
     return torch.tensor(ort_outs[0])
 
 
-def get_preds(context: dict, fname_model: str, model_params: dict, gpu_id: int, batch: dict) -> tensor:
+def get_preds(context: dict, fname_model: str, model_params: dict, cuda_available: bool, device: torch.device, batch: dict) -> tensor:
     """Returns the predictions from the given model.
 
     Args:
         context (dict): configuration dict.
         fname_model (str): name of file containing model.
         model_params (dict): dictionary containing model parameters.
-        gpu_id (int): Number representing gpu number if available. Currently does NOT support multiple GPU segmentation.
+        cuda_available (bool): True if cuda is available.
+        device (torch.device): Device used for prediction.
         batch (dict): dictionary containing input, gt and metadata
 
     Returns:
         tensor: predictions from the model.
     """
-    # Define device
-    cuda_available, device = imed_utils.define_device(gpu_id)
 
     with torch.no_grad():
 
@@ -367,6 +366,9 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
 
     """
 
+    # Define device
+    cuda_available, device = imed_utils.define_device(gpu_id)
+
     # Check if model folder exists and get filenames to be stored as string
     fname_model: str
     fname_model_metadata: str
@@ -493,7 +495,7 @@ def segment_volume(folder_model: str, fname_images: list, gpu_id: int = 0, optio
     preds_list, slice_idx_list = [], []
     last_sample_bool, weight_matrix, volume, image = False, None, None, None
     for i_batch, batch in enumerate(data_loader):
-        preds = get_preds(context, fname_model, model_params, gpu_id, batch)
+        preds = get_preds(context, fname_model, model_params, cuda_available, device, batch)
 
         # Set datatype to gt since prediction should be processed the same way as gt
         for b in batch[MetadataKW.INPUT_METADATA]:


joshuacwnewton:
> @joshuacwnewton: As far as I know, SCT doesn't use the gpu_id

[You are correct!](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/74d4367fea98ab27bbabf7db36c25de1251b48fa/spinalcordtoolbox/scripts/sct_deepseg.py#L215) 

> Let me know if you foresee any issues.

Since the default value for `gpu_ids` is `0`, if `gpus_id` is never specified, then I think the GPU used will be `0` both before and after the changes in this PR. So, I  think that these changes should have no effect on SCT's use of ivadomed. :)

> but I prefer to notify you when we do changes to segment_volume. 

Thank you for the ping. :hearts:

kanishk16 (approved):
Sorry for delaying this but anyways, FTW 🚀

PR #1224, opened 2022-12-17
Introduced generalized_loader_configuration with a test, tests for loading files, and some base for future steps.

kaidalisohaib:
#### GitHub

#### PR contents

## Description

This PR, is the first part of the breakdown of PR #1201.
All of those modifications are from PR #1201.

- New GeneralizedLoaderConfiguration Class at `ivadomed/loader/generalized_loader_configuration.py`
  - Aggregated configuration information to streamline IO.
> A generalized class to store ALL key data requirement for the loading purpose.
    It is a ONE-STOP configuration data class to enable subsequent downstream analyses.
    Used to significantly simply the constructor parameters
- New keywords and added `frozen=True` for each dataclass at `ivadomed\keywords.py`.
- Tests
  - `testing/unit_tests/test_generalized_loader_configuration.py`, unit test for the new GeneralizedLoaderConfiguration class.
  - `testing/unit_tests/test_loader_files.py`, unit tests for multiple file verifications.
- Formatting / Type hinting / Comments
  - `ivadomed/loader/mri2d_segmentation_dataset.py`, more comments.
  - `ivadomed/loader/mri3d_subvolume_segmentation_dataset.py`, type hinting.
  - `ivadomed/training.py`, formatting and type hinting.
  - `ivadomed/transforms.py`, formatting and type hinting.
  - `ivadomed/utils.py`, added a simple line.
  - `testing/common_testing_util.py`, removed a non-used import.
- Files for future PR
  - `ivadomed\config\config_dataset.json`
  - `ivadomed\config\example_loader_v2_configs.py`
## Linked pull requests

#1201


ivadomed/loader/generalized_loader_configuration.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
```
mariehbourget, line 1: General comment about this file.

From my understanding, it will be used in future PR to simplified the dataset constructor parameters.
* I find the name "loader" to be misleading because we also gave a concept of "loader parameters" in the config files which is only a part of all of these parameters here. I think `generalized_loading_configuration` would alleviate this ambiguity.
* Where are the reference for the default value? Is it in `config_default.json` or in the current definitions of `MRI2D` and `MRI3D` datasets?
* How de we plan to keep this up-to-date during the transition? I would be cautious to introduce a file that supercedes other parameters without a control mecanism.

I think I don't understand correctly the purpose of this file in the "big picture" so I don't know exactly how to review it.

ivadomed/loader/generalized_loader_configuration.py
```diff
@@ -0,0 +1,84 @@
...
+            multichannel: bool = False,
+            soft_gt: bool = False,
+            is_input_dropout: bool = False,
+            slice_axis: int = 2,
+            slice_filter_fn: callable = None,
+            patch_filter_fn: callable = None,
+            roi_params: dict = None,
+            task: str = "segmentation",
+            length: Tuple[int, int, int] = (64, 64, 64),
+            stride: Tuple[int, int, int] = (0, 0, 0),
```
mariehbourget, line 41: In particular, I'm not sure how we should deal with `length` and `stride`.
Note that we have `length/stride_2D` that are _optional_ in `default_model` and of dimension 2, whereas we also have `length/stride_3D` that are _required_ in `Modified3DUNet` and of dimension 3.

Again, I don't understand how `length` and `stride` from here will be use down the line. I also wonder if other parameters have different _default_ values between 2D and 3D and if that will be an issue or not.

mariehbourget (commented):
Thanks @kaidalisohaib for the detailed description that helps reviewing the changes. It's nice to see the first smaller PR of this big project coming to life :)

I left many comments/suggestions that are mainly about formatting on things I did not catch before in PR #1201.

Feature-wise, I only have questions about the `generalized_loader_configuration` and its use down the line.

There are also conflicts with master that I would suggest to resolve sooner than later.

ivadomed/keywords.py
```diff
@@ -268,9 +270,73 @@ class SegmentationDatasetKW:
...
+    TRAINING_VALIDATION: str = "training_validation"
+    TRAINING_TEST: str = "training_test"
+    TRAINING_VALIDATION_TEST: str = "training_validation_test"
+
+    # Dataset Level
+    TYPE: str = "type"
+    DATASET_LABEL: str = "dataset_label"
+    IMAGE_GROUND_TRUTH: str = "image_ground_truth"
+    EXPECTED_INPUT: str = "expected_input"
+    EXPECTED_GT: str = "expected_gt"
```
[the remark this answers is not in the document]
dyt811: Great idea.

ivadomed/loader/mri2d_segmentation_dataset.py
```diff
@@ -57,8 +57,13 @@ class MRI2DSegmentationDataset(Dataset):
...
         stride (list): Size of the pixels' shift between patches, length equals 0 (no patching) or 2 (2d patching).
         is_2d_patch (bool): True if length in model params.
+
         prepro_transforms (Compose): Transformations to apply before training.
-        transform (Compose): Transformations to apply during training.
+
+        transform (List[Compose, UndoCompose]): Transformations to apply during training.
+        # Compose transforms
+        tranform_lst, _ = imed_transforms.prepare_transforms(copy.deepcopy(transforms_params), requires_undo)
+
```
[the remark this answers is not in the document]
dyt811: Odd. Removed. Good catch.

ivadomed/training.py
```diff
@@ -23,12 +23,18 @@
...
 
-def train(model_params, dataset_train, dataset_val, training_params, path_output, device, wandb_params=None,
-          cuda_available=True, metric_fns=None, n_gif=0, resume_training=False, debugging=False):
+def train(model_params: dict,
+          dataset_train: BidsDataset,  # to be replaced with consolidated dataset
+          dataset_val: BidsDataset,  # to be replaced with consolidated dataset
+          training_params: dict,
+          path_output: str,
+          device: str,
+          wandb_params=None, cuda_available=True, metric_fns=None, n_gif=0, resume_training=False, debugging=False):
```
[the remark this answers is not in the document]
dyt811: Done. Also added missing WandB params docstring

testing/unit_tests/test_loader_files.py
```diff
@@ -0,0 +1,452 @@
+import os
```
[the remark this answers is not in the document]
dyt811: Hmm... good catch. I think it was way ahead of its time. It was suppose to be utilized in some bids related testing. Removing this for now.

testing/unit_tests/test_generalized_loader_configuration.py
```diff
@@ -0,0 +1,19 @@
...
+        "bn_momentum": 0.1,
+        "final_activation": "sigmoid",
+        "depth": 3,
+    }
+
+    # Build a GeneralizedLoaderConfiguration:
+    config: GeneralizedLoaderConfiguration = GeneralizedLoaderConfiguration(
+        model_params=model_dict,
+    )
+    print(config)
```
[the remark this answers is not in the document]
dyt811: Good point. Altered to loguru output and added a simple assert for the dictionary comparisons.

ivadomed/loader/generalized_loader_configuration.py
dyt811: Valid concern. Renaming adopted. 

As for the default values, they are taken from all existing constructor function defaults which assigned these values already as they: 

You can see it here: 

image

It should have no impact on all functions because... the default we use is the same as the mri2d and mri3d datasets. If there is upstream overwrites, it gets past over and has precedent. if not those defaults have been used for a while. 

It is superseding in the sense that it overwrites the default for MRI2D/MRI3D, which we may eventually default to mandatorily require parameters in the future to avoid duplication of defaults but not needed to determined that at the moment. 

As for all existing changes, it does not affect anything outside of V2 loader pathway as this is not used anywhere else.

ivadomed/loader/generalized_loader_configuration.py
dyt811: Good point. I think it would be worth while to probably change these to NONE for now and expect upstream overwrite. 

The changes here are suppose to be a shim layer that is mirroring underlying values. Whatever existing config will overwrite them as before so none of it should in theory change but this length/stride dimensionality issue is a good example of 2D vs 3D difference.

ivadomed/loader/generalized_loader_configuration.py
mariehbourget: Thanks for the detailed explanation.
My concern is mainly for `length`, `stride` and `slice_axis` which have different default values in 2D and 3D in the screenshot above.
In 3D, length is always used because we always use subvolumes (or patches).
In 2D, we do both "patch" and "slice" (no-patch) training/testing. And the "patch" way is triggered by the presence of the length param.
When I tested the PR #1201, I had no issues with this param so I think it's OK but this is something to keep in mind.

ivadomed/loader/generalized_loader_configuration.py
mariehbourget: > Good point. I think it would be worth while to probably change these to NONE for now and expect upstream overwrite.

Yes, especially because the default is not the same for 2D and 3D as explained above.

ivadomed/loader/generalized_loader_configuration.py
mariehbourget: So I just tested again in PR #1201 and the default values of length here (in 3D) don't seem to impact the behavior in 2D with or without patches.

However, as described in this [comment](https://github.com/ivadomed/ivadomed/pull/1201/files#r1069617005) in PR #1201, the prototype does not work in 3D at the moment.

kanishk16: To move forward, a note in the class should be added for 3D scenarios to revisit this and is not lost in the complete integration process.

ivadomed/loader/generalized_loader_configuration.py
```diff
@@ -0,0 +1,84 @@
...
+
+from ivadomed.keywords import ROIParamsKW
+from ivadomed.transforms import Compose, prepare_transforms
+
+if typing.TYPE_CHECKING:
+    from ivadomed.loader.patch_filter import PatchFilter
+    from ivadomed.loader.slice_filter import SliceFilter
+
+
+class GeneralizedLoaderConfiguration:
```
kanishk16, line 15: I second @mariehbourget's comment about the naming convention... I'm still not convinced with the name but let's move ahead meanwhile, we could think of a better name. Renaming to be adopted for the class name as well...

ivadomed/training.py
```diff
@@ -73,21 +85,32 @@ def train(model_params, dataset_train, dataset_val, training_params, path_output
         # Initialize WandB with metrics and hyperparameters
         wandb.init(project=project_name, group=group_name, name=run_name, config=cfg, dir=path_output)
 
-    # BALANCE SAMPLES AND PYTORCH LOADER
-    conditions = all([training_params[TrainingParamsKW.BALANCE_SAMPLES][BalanceSamplesKW.APPLIED],
-                      model_params[ModelParamsKW.NAME] != "HeMIS"])
-    sampler_train, shuffle_train = get_sampler(dataset_train, conditions,
-                                               training_params[TrainingParamsKW.BALANCE_SAMPLES][BalanceSamplesKW.TYPE])
+    # If balance sampler is applied and not using HeMIS model,
+    # Then use a sampler.
```
kanishk16, line 89: Merge it into a single-line comment

testing/unit_tests/test_generalized_loader_configuration.py
```diff
@@ -0,0 +1,19 @@
...
+    model_dict = {
+        "name": "Unet",
+        "dropout_rate": 0.3,
+        "bn_momentum": 0.1,
+        "final_activation": "sigmoid",
+        "depth": 3,
+    }
+
+    # Build a GeneralizedLoaderConfiguration:
+    config: GeneralizedLoaderConfiguration = GeneralizedLoaderConfiguration(
```
kanishk16, line 15: To be updated in case renaming of the class is adopted as suggested above!

kanishk16 (changes requested):
Looks in good shape in terms of functionality... Minor nitpicky comments and once addressed it should be good to go in!

PR #1193, opened 2022-09-19
feat: update default args for `wandb.login()`

kanishk16:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
Although WandB includes a feature of anonymous experimental tracking wherein, the user can track experiments even without a WandB account. Just changing the default argument to `anonymous='allow'` in the `wandb.login()` enables this feature.

Additionally, if a user logs in for the first time on a machine or the previous API Key authentication trials were unsuccessful, and the user leaves the wandb_api_key key blank `("wandb_api_key": "")` then a prompt appears as below which blocks the execution. Adding a timeout of 60s as `timeout=60` in `wandb.login()` prevents the above scenario.

image

This PR is an attempt to handle the edge cases raised while testing singularity containers locally.

PR #1253, opened 2023-01-02
Do not use wandb if not specified in the config file

### LINKED ISSUE #1252: Do not use wandb if not specified in the config file

My config file does NOT include any wandb field, yet training with ivadomed used wandb:
~~~
2023-01-01 18:25:54.402 | INFO     | ivadomed.loader.loader:load_dataset:118 - Loaded 6958 volumes of shape [32, 32, 16] for the training set.
2023-01-01 18:25:54.402 | INFO     | ivadomed.main:create_path_model:89 - Creating model directory: /home/GRAMES.POLYMTL.CA/p101317/data_nvme_p101317/model_seg_lesion_mp2rage_20230101_182457-batch_size-8/model_seg_lesion_mp2rage
wandb: Currently logged in as: jcohenadad. Use `wandb login --relogin` to force relogin
2023-01-01 18:25:55.487 | INFO     | ivadomed.utils:initialize_wandb:56 - WandB API Authentication Successful!
wandb: wandb version 0.13.7 is available!  To upgrade, please run:
wandb:  $ pip install wandb --upgrade
wandb: Tracking run with wandb version 0.13.6
wandb: Run data is saved locally in /home/GRAMES.POLYMTL.CA/p101317/data_nvme_p101317/model_seg_lesion_mp2rage_20230101_182457-batch_size-8/wandb/run-20230101_182555-3eiodlnq
wandb: Run `wandb offline` to turn off syncing.
wandb: Syncing run run-1
wandb: ⭐️ View project at https://wandb.ai/jcohenadad/my_project
wandb: 🚀 View run at https://wandb.ai/jcohenadad/my_project/runs/3eiodlnq
~~~

This is probably because the code checks for an environment variable? Anyway, this is highly confusing and undesired. I suggest to not activate wandb if it is not mentioned in the config file.

jcohenadad:
#### GitHub

#### PR contents

## Description
#1252, even if the config file does NOT include any wandb field, training with ivadomed can use wandb in case wandb logging was previously done, indicates that deliberate omission of the value to `wandb_api_key` while logging in and the anonymous logging feature introduced in #1193 could sometimes be misleading and undesired. Hence, this PR undoes and simplifies it by bypassing `wandb.login` call if the key is not defined or is empty.

## Linked issues

Resolves #1252


jcohenadad:
@kanishk16 do you mind reviewing this? if you cannot please let me know

kanishk16:
I'm sorry to stall the progress... On it right away... weirdly all the notifications were sent to my spam

ivadomed/utils.py
```diff
@@ -35,6 +35,12 @@ def __str__(self):
 
 
 def initialize_wandb(wandb_params):
+
+    # If key is not defined or is empty, exit immediately.
+    # https://github.com/ivadomed/ivadomed/issues/1252
+    if wandb_params[WandbKW.WANDB_API_KEY] == '':
+        return False
```
kanishk16, line 42: Perhaps nitpicky, but this:
1. doesn't inform the user about wandb being disabled and the reason. 
2. Evaluating an empty string as `== ''` doesn't help in avoiding instances with whitespace (ref: https://stackoverflow.com/a/9573278) 
```
 >>> bool("")
 False
 >>> bool("   ")
 True
 >>> bool("   ".strip())
 False
```
Therefore, I'm inclined to something along the lines of which takes care of that:

```
try:
  if wandb_params[WandbKW.WANDB_API_KEY].strip():
    raise ValueError()

  wandb.login(...)
```

kanishk16 (changes requested):
#1252 indicates that deliberate omission of the value in `wandb_api_key` while logging in and the anonymous logging feature introduced in #1193 could sometimes be misleading and undesired. Hence, this undoes and simplifies it. 

We'd also like to remove this from the documentation [here](https://ivadomed.org/configuration_file.html#wandb-api-key). Specifically, the following should be removed from the docs:
> From then on, the value in this key-value pair in the config file could be omitted, like 'wandb_api_key': ''

jcohenadad:
> We'd also like to remove this from the documentation [here](https://ivadomed.org/configuration_file.html#wandb-api-key). Specifically, the following should be removed from the docs:

Yes! excellent catch

ivadomed/utils.py
jcohenadad: sounds good! can you please do all the changes you proposed in your review? thanks!

kanishk16 (approved):
LGTM 🚀 Will merge once tests pass!

PR #1292, opened 2023-04-26
Fix deprecated feature call due to newest imageio release

mathieuboudreau:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

As of imageio==2.28.0 (April 24, 2023), the feature `as_gray=True` is deprecated when loading images. I've replaced it with their recommended substitute, `mode='L'`.

## Linked issues

Resolves #1291

PR #1296, opened 2023-06-15
Fix 2D image conversion to 8 bit

mathieuboudreau:
#### GitHub

#### PR contents

## Description
IVADOMED's implentation of loading 2D images (eg histology) has a bug where if 16bit TIFs (maybe other formats too, only one tested) are loaded, 

Screenshot 2023-06-15 at 8 50 10 PM

the image isn't properly converted to 8 bits and instead gets "cut off", 

Screenshot 2023-06-15 at 8 40 46 PM

This proposed fix should do the conversion properly and resolve this issue, which was first raised in AxonDeepSeg,

Screenshot 2023-06-15 at 8 42 24 PM

## Linked issues
https://github.com/axondeepseg/axondeepseg/issues/737

### DIFF
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -316,11 +316,13 @@ class SegmentationPair(object):
         # For '.png', '.tif', '.tiff', '.jpg' and 'jpeg' extentions
         # Read image as 8 bit grayscale in numpy array (behavior TBD in ivadomed for RGB, RBGA or higher bit depth)
         if "tif" in extension:
-            img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil'), axis=-1).astype(np.uint8)
+            img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil'), axis=-1)
             if len(img.shape) > 3:
-                img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil', as_gray=True), axis=-1).astype(np.uint8)
+                img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil', as_gray=True), axis=-1)
         else:
-            img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
+            img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1)
+        
+        img = imageio.core.image_as_uint(img, bitdepth=8)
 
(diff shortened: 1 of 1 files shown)


mathieuboudreau:
To test, maybe I/we should apply these changes to the IVADOMED code in an ADS environment, and test that the segmentation runs correctly. I've so far only tested the change in a Python shell outside of the ADS workflow

hermancollin (approved):
On the master branch, loading a 16bit image into napari gives an empty segmentation

I then use this branch:
```
pip uninstall ivadomed
git clone https://github.com/neuropoly/ivadomed.git
cd ivadomed
pip install -e .
git checkout mb/8bitconversion
```
Loading the same 16bit image and apply the model, the image is now segmented correctly. I think this pretty much resolves this problem, thank you for figuring this out!

Note that we still have this warning when we apply the model, at which stage the image is loaded in ivadomed. 
```
14:56:29 WARNING Lossy conversion from uint16 to uint8. Losing 8 bits of resolution. Convert image to uint8 prior to saving to suppress this warning.
```
This warning is present both on this branch and on the master branch. It probably comes from this new line:
```
img = imageio.core.image_as_uint(img, bitdepth=8)
```

mathieuboudreau:
Thanks @hermancollin !

>Note that we still have this warning when we apply the model, at which stage the image is loaded in ivadomed.

I prefer a warning than an empty image =)

From the tests though, it seems like I need to use a different syntax as IVADOMED pulls a more recent version of Pillow,

```
TypeError: The keyword `as_gray` is no longer supported.Use `mode='F'` for a backward-compatible result, or  `mode='L'` for an integer-valued result.
```

I'll have to test to see if I resolve it in IVADOMED with this new syntax, that it doesn't break when we then later on import it from ADS's environment.

mathieuboudreau:
I just tested a 16bit image and it segmented well (i.e. ivadomed side ok) but napari didn't display it correctly (meaning changes needed on the ADS side). So this IVADOMED issue can be closed now, we'll deal with with the rest on the ADS repo!

kanishk16:
@mathieuboudreau As far as I remember, #1297 was supposed to fix `imageio` issues by migrating to `v3` essentially combining this PR & #1292 😅

mathieuboudreau:
I forgot to show the images in my last comment:

Here's the raw image:

Screenshot 2023-12-11 at 12 04 42 PM


And them loaded in napari and segmented masks:

Screenshot 2023-12-11 at 11 32 00 AM

kanishk16:
Lovely!!! 🚀

Issue #1191, opened 2022-08-07
Dependency Pinning


kousu:
Our current dependencies inconsistently use [version specifiers](https://peps.python.org/pep-0440/#version-specifiers), probably because we don't have a good team-wide understanding:

https://github.com/ivadomed/ivadomed/blob/c7b8b94dd564eac2e0c1685cb88602f703f58613/requirements.txt#L80-L26
 
In fact, maybe no one really understands python dependency management very well, judging from [these](https://bernat.tech/posts/version-numbers/) [articles](https://snarky.ca/why-i-dont-like-semver/) that @kanishk16 found. On top, we're vaguely using semver for ourselves, but poorly -- mostly we just guess if it's a Major, Minor or Patch release.

_Originally posted by @lifetheater57 in https://github.com/ivadomed/ivadomed/pull/1137#discussion_r938281300_:

> Removing the upper bound on the version exposes us to potential issues if some breaking changes versions are made in future versions of either packages, isn't it? I would only keep the versions that we did tested (i.e. restricting to these with `>=min, <=max` bounds).

Some specify minimums, like

```
wandb>=0.12.11
```

which gets written into the .whl on PyPI, which means that _if someone tries to `pip install ivadomed==2.9.6` in the _future_, after `wandb` has had breaking changes, then `ivadomed` will be broken. On the other hand, it means that in the more common case where `wandb` has improvements that aren't breaking, `ivadomed` will happily be able to use it **and so will any project that someone is trying to use in conjunction with ivadomed** because its declared dependencies are not overconstrained.

Other dependencies are constrained, like

```
nibabel~=3.2
```

which [is short for](https://peps.python.org/pep-0440/#compatible-release)

```
nibabel>=3.2,<4
```

That [second article points out](https://snarky.ca/why-i-dont-like-semver/), `~=` isn't actually a safe option because

> SemVer can't protect you from having to test your code to see if a micro version is compatible with your code.

i.e. we could still have the problem that, say nibabel 3.2.2 accidentally introduced a breaking change that means `pip install ivadomed==2.9.6` always creates a broken install.

But the downside of this is that the situation may be overconstrained, like the [first article says](https://bernat.tech/posts/version-numbers/)

> You’re requesting a version between four and five, while the service library is requesting a version between five and six. Both constraints cannot be satisfied.
>
[...]

PR #1107, opened 2022-03-17
Fix loading of TIF 16bits grayscale files

mariehbourget:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [x] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
This PR fixes issue #1105 where TIF 16bits grayscale gave an error while loading:
- It includes a new test to make sure every combination of PNG/TIF, 8/16 bits, Grayscale/RGB/RGBA files are loaded properly.
- It requires a new set of testing images being integrated in data-testing in PR https://github.com/ivadomed/data-testing/pull/18.
- PR https://github.com/ivadomed/data-testing/pull/18 needs to be approved, merged and a new release of data-testing done to update the link for download [here](https://github.com/ivadomed/ivadomed/blob/6e1e87b4251a1e8985901194a821fb638d9242cb/ivadomed/scripts/download_data.py#L26) before merging the current PR.
- I updated some keywords in the code of the test_loader.py file as well.

## Linked issues
Fixes #1105

PR #1297, opened 2023-06-26
chore: Dependency Maintenance (imageio 2->3, pyBIDS<0.15.6, readthedocs.yml v2, Python 3.7->3.8)

### LINKED ISSUE #1291: imageio "as_gray" option deprecated as of 2.28.0
## Issue description

These lines:

https://github.com/ivadomed/ivadomed/blob/e101ebea632683d67deab3c50dd6b372207de2a9/ivadomed/loader/segmentation_pair.py#L318-L323

Result in an error now when the most recent release of imageio is installed.

```

ads_gray) mathieuboudreau@Mathieus-MacBook-Pro ads_gray % axondeepseg_test
Computing the segmentation of axon and myelin on test image.
2023-04-25 11:40:32.582 | WARNING  | AxonDeepSeg.apply_model:axon_segmentation:49 - The 'no-patch' option was not selected for segmentation. Please note that this option could potentially produce better results but may not be suitable with large images depending on computer RAM capacity.
2023-04-25 11:40:32.582 | INFO     | ivadomed.utils:define_device:160 - Cuda is not available.
2023-04-25 11:40:32.582 | INFO     | ivadomed.utils:define_device:161 - Working on cpu.
2023-04-25 11:40:32.583 | INFO     | ivadomed.config_manager:_display_differing_keys:202 - Adding the following keys to the configuration file
2023-04-25 11:40:32.584 | INFO     | ivadomed.config_manager:deep_dict_compare:47 -     wandb: {'wandb_api_key': '', 'project_name': 'my_project', 'group_name': 'my_group', 'run_name': 'run-1', 'log_grads_every': 100}
2023-04-25 11:40:32.584 | INFO     | ivadomed.config_manager:deep_dict_compare:47 -     loader_parameters: patch_filter_params: {'filter_empty_mask': False, 'filter_empty_input': False}
2023-04-25 11:40:32.584 | INFO     | ivadomed.config_manager:deep_dict_compare:47 -     lr_scheduler: base_lr: 1e-05
2023-04-25 11:40:32.584 | INFO     | ivadomed.config_manager:deep_dict_compare:47 -     lr_scheduler: max_lr: 0.01
2023-04-25 11:40:32.584 | INFO     | ivadomed.config_manager:deep_dict_compare:47 -     evaluation_parameters: object_detection_metrics: True
2023-04-25 11:40:32.584 | INFO     | ivadomed.config_manager:_display_differing_keys:204 - 

2023-04-25 11:40:32.584 | WARNING  | ivadomed.inference:segment_volume:409 - fname_roi has not been specified, then the entire volume is processed.
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/bin/axondeepseg_test", line 33, in <module>
    sys.exit(load_entry_point('AxonDeepSeg', 'console_scripts', 'axondeepseg_test')())
  File "/Users/mathieuboudreau/neuropoly/github/ads_gray/AxonDeepSeg/integrity_test.py", line 35, in integrity_test
    axon_segmentation(path_testing, [str(path_testing / image)], path_model, acquired_resolution=0.13, gpu_id=0, overlap_value=[48,48])
  File "/Users/mathieuboudreau/neuropoly/github/ads_gray/AxonDeepSeg/apply_model.py", line 61, in axon_segmentation
    nii_lst, _ = imed_inference.segment_volume(str(path_model), input_filenames, gpu_id=gpu_id, options=options)
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/ivadomed/inference.py", line 475, in segment_volume
    ds.load_filenames()
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/ivadomed/loader/mri2d_segmentation_dataset.py", line 127, in load_filenames
    roi_pair = SegmentationPair(input_filenames,
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/ivadomed/loader/segmentation_pair.py", line 59, in __init__
    input_img = self.read_file(input_file)
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/ivadomed/loader/segmentation_pair.py", line 297, in read_file
    img = self.convert_file_to_nifti(filename, extension, is_gt)
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/ivadomed/loader/segmentation_pair.py", line 323, in convert_file_to_nifti
    img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/imageio/v2.py", line 345, in imread
    result = file.read(index=0, **kwargs)
  File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/imageio/plugins/pillow.py", line 193, in read
    raise TypeError(
TypeError: The keyword `as_gray` is no longer supported.Use `mode='L'` instead.
(ads_gray) mathieuboudreau@Mathieus-MacBook-Pro ads_gray % 

```

We found this in AxonDeepseg (https://github.com/axondeepseg/axondeepseg/issues/738), I fixed all the local `as_gray=True` to `mode=L` per their message, but the instances highlighted above in Ivadomed continues to give errors and would need to be changed and added to a new release here.

File "/Users/mathieuboudreau/opt/anaconda3/envs/ads_gray/lib/python3.8/site-packages/ivadomed/loader/segmentation_pair.py", line 323, in convert_file_to_nifti
    img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)

kanishk16:
#### GitHub

#### PR contents

## Description

This PR combines a number of maintenance changes, updating various dependencies that have evolved since the last release of ivadomed.

### 1. imageio 2->3

`imageio` had planned a pretty user-friendly (slow and steady) transition to v3, https://github.com/imageio/imageio/issues/725, for a year and a half. It is now approaching its stage to bump to v3. We initially made some hotfixes to move to v2, #1181, rather than migrating to v3, considering it to be in its early days - stabilizing & gathering user feedback. But, it appears to have stabilized after a significant number of minor and patch releases, especially in the last 2-3 months (https://github.com/imageio/imageio/issues/982, https://github.com/imageio/imageio/issues/931). 

Moreover, the recent `imageio` releases raised quite a few issues #1291 that were high-priority concerns originating from https://github.com/axondeepseg/axondeepseg/issues/738 and https://github.com/axondeepseg/axondeepseg/issues/737. There were attempts to hotfix with #1292 and #1296, but the tests were still broken. Since `imageio.v3` appears much more stable, this PR attempts to streamline and resolve the above issues by migrating to `imageio.v3`.

Further details about the migration itself can be found in this comment: https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1743338605

### 2. pyBIDS<0.15.6

In recent months, pyBIDS has introduced some backwards incompatible changes that break our tests in multiple ways: https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1622272746, https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1836724597.

We could update `ivadomed` to be compatible with the newest PyBIDS, however in the short term it's simplest just to pin <0.15.6. 0.15.5 was released in November 2022 and its wheel is not limited to specific versions of Python 3, so there should be no forwards-compatibility issues. 

### 3. Pillow <10.0.0

Pillow 10.0.0 is incompatible with the outdated version of `tensorboard` we use. This fix will be unnecessary once we update `torch` in #1304.

See: https://github.com/ivadomed/ivadomed/pull/1297#discussion_r1412547253

### 4. readthedocs.yml v2

See: https://blog.readthedocs.com/migrate-configuration-v2/

### 5. Python 3.7->3.8

Python 3.7 has reached EOL, so it needed to be updated in our tests workflow as well as the readthedocs workflow.

### 6. Ubuntu 18.04

Reached EOL.

## Linked issues

Resolves #1291


kanishk16:
@hermancollin Though most tests pass, I'd like you to test & verify as in #1296. Also, it would be great if you could check the inference too. 

@dyt811 Since windows tests aren't passing for some time (unrelated to the current PR) and are required before merging, could you remove the windows tests from the required statuses - Python 3.8 and Ubuntu-18.04, Python 3.8 (#1298) to unblock this PR?

mathieuboudreau:
>@dyt811 Since windows tests aren't passing for some time (unrelated to the current PR) and are required before merging, could you remove the windows tests from the required statuses - Python 3.8 and Ubuntu-18.04, Python 3.8 (https://github.com/ivadomed/ivadomed/pull/1298) to unblock this PR?

I could remove the Windows testing requirement, 

https://github.com/ivadomed/ivadomed/blob/e101ebea632683d67deab3c50dd6b372207de2a9/.github/workflows/run_tests_dummy.yml#L30C76-L31 and also in the admin settings link that's shared in the comment. I think these are the only changes needed.

However as it says they both need to be in sync, changing it in the config file just in the PR might "break" some of the other branches/PR tests (hopefully just temporarily).

Let me know if you'd like me to move ahead and do it if @dyt811 isn't available soon enough to do it.

mathieuboudreau:
@kanishk16 it's been about a week - want me to go ahead and try removing the windows required check myself?

kanishk16:
> @kanishk16 it's been about a week - want me to go ahead and try removing the windows required check myself?
Yup, please go ahead

ivadomed/loader/segmentation_pair.py
```diff
@@ -315,12 +315,12 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
         """
         # For '.png', '.tif', '.tiff', '.jpg' and 'jpeg' extentions
         # Read image as 8 bit grayscale in numpy array (behavior TBD in ivadomed for RGB, RBGA or higher bit depth)
-        if "tif" in extension:
-            img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil'), axis=-1).astype(np.uint8)
-            if len(img.shape) > 3:
```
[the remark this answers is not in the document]
kanishk16: I was under the impression that Armand might have a look at this PR.  I did revisit #1107, and the added tests pass on ivadomed side nevertheless, this PR definitely needs to be tested for the edge cases on the ADS side.

mathieuboudreau:
Done and re-ran the tests - tests are failing due to a Pillow - Tensorboard error.

I checked and Pillow moved from 9.x to 10.0 on July 1st, so this is likely the culprit. I'll add a restricted requirement for <10 version and see how it turns out.

kanishk16:
Ubuntu-18.04, Python 3.8 (https://github.com/ivadomed/ivadomed/pull/1298) required check also needs to be removed as GitHub Actions stopped the support for Ubuntu-18.04 LTS quite some time back.

mathieuboudreau:
Ok so tests pass now, but seems like coveralls & the other matrix test won't pass if the windows tests fail. Could we just remove windows testing altogether? I'm a bit experienced with GitHub action CI YML file writing, but I don't know at first glance how to update it so that the coveralls get reported in the case of failing Windows tests, without rewriting a lot of the structure of the file.

mathieuboudreau:
>@dyt811 Since windows tests aren't passing for some time (unrelated to the current PR) and are required before merging, could you remove the windows tests from the required statuses - Python 3.8 and Ubuntu-18.04, Python 3.8 (https://github.com/ivadomed/ivadomed/pull/1298) to unblock this PR?

Or we could look into how to resolve the one test that's failing in Windows - has someone looked at it yet / is there an issue/PR? Because all the other tests pass in windows, and it would be a shame to stop testing on windows just for this edge case.

mathieuboudreau:
If I were to guess, it has to do with this line

Screenshot 2023-07-05 at 2 23 32 PM

which is formatted for unix-based systems only (MacOS/Linux), vs the line above that uses `Path()` and will correctly format the path to whichever system is being used.

kanishk16:
I second the hesitation to remove Windows tests altogether just for a single failing test.

> which is formatted for unix-based systems only (MacOS/Linux), vs the line above that uses Path() and will correctly format the path to whichever system is being used.
I tried fixing it with something along the lines

```
"bids_config": str(path_repo_root.joinpath('ivadomed', 'config', 'config_bids.json')),
``` 
but unfortunately, tests still fail with the same error AFAIK. I don't think there's a separate issue or a PR to resolve the same.

mathieuboudreau:
I think I resolved the bug - essentially pybids changed how they parse regex over the winter, and at some point it started breaking for Windows. Restricting pybids to <15.6.0 resolved the failing test in the draft PR I setup: https://github.com/ivadomed/ivadomed/pull/1299, see: https://github.com/ivadomed/ivadomed/actions/runs/5467608625/jobs/9954141022


Screenshot 2023-07-05 at 3 31 38 PM

Let me make the change here, and if they all pass, I'll reintroduce the required windows checks and we can move ahead. @hermancollin were you intending on reviewing this PR too?

kanishk16:
> Ok so tests pass now, but seems like coveralls & the other matrix test won't pass if the windows tests fail. Could we just remove windows testing altogether? I'm a bit experienced with GitHub action CI YML file writing, but I don't know at first glance how to update it so that the coveralls get reported in the case of failing Windows tests, without rewriting a lot of the structure of the file.

I looked into the `run_tests.yml` and it seems like we'd need a bit of restructuring the file to skip the windows tests from coveralls. Since we're pretty confident about the changes as there's just a windows test that fails, it won't matter much even if the coveralls report isn't generated.

kanishk16:
> I think I resolved the bug - essentially pybids changed how they parse regex over the winter, and at some point it started breaking for Windows. Restricting pybids to <15.6.0 resolved the failing test in the draft PR I setup: #1299, see: https://github.com/ivadomed/ivadomed/actions/runs/5467608625/jobs/9954141022
> 
> Screenshot 2023-07-05 at 3 31 38 PM
> Let me make the change here, and if they all pass, I'll reintroduce the required windows checks and we can move ahead. @hermancollin were you intending on reviewing this PR too?

@mathieuboudreau I'm not in favor of restricting pybids for a windows test as:
1. Blocks future developments from bids to be integrated into ADS which might be much more significant 
2. "PyBIDS currently supports Python 3 on POSIX operating systems (including Mac OS). **Windows is not officially supported**, though most PyBIDS functionality will probably work fine." (ref: https://github.com/bids-standard/pybids)

Dependency pinning is much more complex than it seems - for detailed discussions https://github.com/ivadomed/ivadomed/issues/1191

mathieuboudreau:
```
@mathieuboudreau I'm not in favor of restricting pybids for a windows test as:

Blocks future developments from bids to be integrated into ADS which might be much more significant
"PyBIDS currently supports Python 3 on POSIX operating systems (including Mac OS). Windows is not officially supported, though most PyBIDS functionality will probably work fine." (ref: https://github.com/bids-standard/pybids)
Dependency pinning is much more complex than it seems - for detailed discussions https://github.com/ivadomed/ivadomed/issues/1191
```

That's fair ! I think the decision will then be, does the IVADOMED team want full Windows compatibility, or pybids > 0.15.5.6, because there's nothing we can do on our side to have both! 

I'm agnostic to either, though I think if you choose the pybids route, you'll have to remove all windows tests to trigger the coveralls step (or, write a completely separate actions file just for windows so that coveralls will trigger for mac/linux).

Let me know which direction you want to go - maybe it's something that the IVADOMED team should discuss internally?

mathieuboudreau:
>Blocks future developments from bids to be integrated into ADS which might be much more significant

This isn't a concern btw, since IVADOMED switched to maintenance only, we're exploring a different package to segment our images.

kanishk16:
Since ivadomed is not under active development and has minimal maintenance, this might be one of the last PRs that goes in just to ease the transitioning process for ADS. Hence, it's better to refrain from introducing a lot of changes i.e. we might want to push this PR in and not even think about supporting windows or removing windows tests altogether for coveralls.

hermancollin:
btw I can test this PR but what should I look for exactly? Just inference?

As for the current dilemma, we can discuss this with the team next Monday.

mathieuboudreau:
>Since ivadomed is not under active development and has minimal maintenance, this might be one of the last PRs that goes in just to ease the transitioning process away from ivadomed. Hence, it's better to refrain from introducing a lot of changes i.e. we might want to push this PR in and not even think or be concerned about supporting windows or removing/restructuring tests for coveralls...

Ok! I reverted the requirements file, and later if this PR is ready but we want to ignore the windows tests failing, an admin could bypass branch protections and merge with the failed checks

mathieuboudreau:
>btw I can test this PR but what should I look for exactly? Just inference?

https://github.com/ivadomed/ivadomed/pull/1297#discussion_r1253142245

the changes on this set of lines, I'm not quite convinced will work the same was as before.

kanishk16:
> btw I can test this PR but what should I look for exactly? Just inference?

I gave a thought to test this PR robustly. I would suggest loading and segmenting (assuming it might include inference) all the different types of images from https://github.com/ivadomed/data-testing/tree/master/data_test_png_tif in Napari, just the way the bug was found in https://github.com/axondeepseg/axondeepseg/issues/737.

hermancollin (changes requested):
I did what Kanishk suggested and here are the results:

<details>

#### on master

| type | PNG | TIF |
|:-:|:-:|:-:|
| 8b-gray| x | x |
| 8b-rgb| x | x |
| 8b-rgba| x | x |
| 16b-gray| x | x |
| 16b-rgb| x | x |
| 16b-rgba| x | x |


So everything works on the master branch for me? IIUC, we would have expected it to fail here? 

#### on this branch

| type | PNG | TIF |
|:-:|:-:|:-:|
| 8b-gray| x | x |
| 8b-rgb| **NOPE** | **NOPE** |
| 8b-rgba| **NOPE** | **NOPE** |
| 16b-gray| x | x |
| 16b-rgb| **NOPE** | **NOPE** |
| 16b-rgba| **NOPE** | **NOPE** |

</details>

So altogether this leaves me kinda confused. This PR breaks RGB compatibility, and I could not reproduce the 16bit bug with the test images.

EDIT: This discussion chain was resolved by https://github.com/imageio/imageio/issues/1017.

ivadomed/loader/segmentation_pair.py
```diff
@@ -315,12 +315,12 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
...
-                img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil', as_gray=True), axis=-1).astype(np.uint8)
-        else:
-            img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
+        _img = imageio.v3.imread(filename)
+        
+        if len(_img.shape) < 3:
```
hermancollin, line 320:
```suggestion
        if len(_img.shape) > 3:
```

Could it be that the sign in this condition was reversed by accident?

hermancollin: Ahh nope sorry about that this doesn't work.

kanishk16:
I'm quite surprised... could you let me know your env details once?

ivadomed/loader/segmentation_pair.py
mathieuboudreau: It's the "as_gray" conversion that's missing. It also changed name in v3 of imageio, can recall what it is but it warns about it @kanishk16 @hermancollin

mathieuboudreau: It should be `mode="L"` in the imread options, see my currently open ivadomed PR that was supposed to deal with it, before this PR was opened. https://github.com/ivadomed/ivadomed/pull/1292/files

hermancollin, line 320:
```suggestion
        if len(_img.shape) >= 3:
            if extension == '.tif' or extension == '.tiff':
                # following line does not work !!!
                _img = imageio.v3.imread(filename, plugin='pillow', extension='.tif', mode='L')
            else:
                _img = imageio.v3.imread(filename, plugin='pillow', mode='L')
        _img = np.expand_dims(_img, axis=-1)
```

After extensive testing, I figured the only way to get back the RGB/RGBA compat was to use the previous logic. Now, it works for PNG images, but I wasn't able to make it work for TIFF images. I think the appropriate backend for TIFF would be `plugin='tifffile'` instead of `plugin='pillow'`, but the tifffile API is obscure to me...

hermancollin:
> @hermancollin Actually, this is exactly what I expected to see!


@mathieuboudreau Yes but what do you think of everything already working on the master branch?

ivadomed/loader/segmentation_pair.py
hermancollin: Yes. The problem is that `mode='L'` works fine for RGB/RGBA PNGs, but not for TIFFs. I wasn't able to find the proper way to do it with imageio.v3

mathieuboudreau: Hmm! Would it be bad practice to first convert the TIFFS to RGB/RGBA, then "load" them? Probably a waste of memory and time, I wonder if that could work somehow

hermancollin:
Opened an issue on the imageio repo https://github.com/imageio/imageio/issues/1017
Let's wait for their response maybe they have a nice way to fix this.

kanishk16:
> > @hermancollin Actually, this is exactly what I expected to see!
> 
> @mathieuboudreau Yes but what do you think of everything already working on the master branch?
> 
> i.e. I'm not sure how to reproduce the bug in the first place... all the 16 bit images worked fine it seems

@hermancollin I guess you missed my previous comment but I'd just like to confirm your environment details.

ivadomed/loader/segmentation_pair.py
mathieuboudreau: >Well currently the rgb/rgba TIFFs are the problem. I assume you mean converting the TIFFs to PNGs or something like this? It would probably work in that case.

Yes woops, mind blank on my part.

>Maybe we should open an issue on the imageio side to ask them the best practice. In the end, all we want to do is convert a multichannel TIFF image to grayscale, it shouldn't be that hard

Definitely we should do that - if it hasn't been done by someone else already! I think the only hiccup is I think tifs are dealt with another dependency, TIFFILE? So maybe there's a conflict between the functionality of those two packages when we want to convert to B&W?

kanishk16:
> I'm afraid its not going to be very informative. I installed ivadomed in editable mode inside an axondeepseg venv. ivadomed -v confirms I'm on the right branch.
Just wanted to confirm the imageio version bcz I needed to upgrade even after a fresh install.

The way I was testing the images was - while running the test `testing/unit_tests/test_loader.py::test_read_png_tif` I visualized all the outputs at the end of all the changes i.e. 
```
img = imageio.core.image_as_uint(_img, bitdepth=8)

self.plot_check(img)
```
```
    @staticmethod
    def plot_check(img, filename):
        import matplotlib.pyplot as plt
        
        fig, ax = plt.subplots(nrows=1, ncols=1)
        
        ax.set_title(filename)
        ax.imshow(img, cmap='gray')

```
It seems to work for all the images with the current state of the PR. Idk why but I wasn't convinced too. Hence, requested another round of reviews. I might be missing something. Perhaps, appears `matplotlib` internally converts images to some common format. 🤷‍♂️

hermancollin:
> Just wanted to confirm the imageio version bcz I needed to upgrade even after a fresh install.

Ah that's a good catch @kanishk16 ! I was on 2.27. However, bumping it to 2.28 doesn't change anything.

Also, let me clarify the bug. The segmentation always fail in the same way and look somewhat like this:
Screenshot_20230711_101923

kanishk16:
> > Just wanted to confirm the imageio version bcz I needed to upgrade even after a fresh install.
> 
> Ah that's a good catch @kanishk16 ! I was on 2.27. However, bumping it to 2.28 doesn't change anything.

Thanks, this made me realize we also need to put a constraint in `requirements.txt`. `imageio` made some significant changes in their recent releases. So, I'd like you to bump to `2.31.1`

> Also, let me clarify the bug. The segmentation always fail in the same way and look somewhat like this: Screenshot_20230711_101923
Yes, I do consider this as a fail, but the original bug, AFAIK should be similar to found in https://github.com/axondeepseg/axondeepseg/issues/737. To reproduce it, I'd suggest upgrading `imageio` to the latest version; then, it might raise an error regarding `as_gray` being deprecated. Replace it with `mode='L'`. Now, try loading it in Napari.

ivadomed/inference.py
```diff
@@ -225,7 +225,9 @@ def pred_to_png(pred_list: list, target_list: list, subj_path: str, suffix: str
     for pred, target in zip(pred_list, target_list):
         filename = subj_path + target + suffix
         data = pred.get_fdata()
-        imageio.imwrite(filename, (data*255/max_value).astype(np.uint8), format='png')
+        _img = (data*255/max_value).astype(np.uint8)
+        img = np.dstack((_img, _img, _img))
```
FirefoxMetzger, line 229: You don't need to do a depth stack here unless you want the image to be stored as a RGB image. 

If you pass a grayscale image (shape `(height, width)`) directly then ImageIO will store it as a grayscale image, which (a) is smaller, and (b) will be read as a grayscale image :)

ivadomed/loader/segmentation_pair.py
```diff
@@ -315,12 +315,12 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
         """
         # For '.png', '.tif', '.tiff', '.jpg' and 'jpeg' extentions
         # Read image as 8 bit grayscale in numpy array (behavior TBD in ivadomed for RGB, RBGA or higher bit depth)
-        if "tif" in extension:
-            img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil'), axis=-1).astype(np.uint8)
-            if len(img.shape) > 3:
-                img = np.expand_dims(imageio.v2.imread(filename, format='tiff-pil', as_gray=True), axis=-1).astype(np.uint8)
-        else:
-            img = np.expand_dims(imageio.v2.imread(filename, as_gray=True), axis=-1).astype(np.uint8)
+        _img = imageio.v3.imread(filename)
```
FirefoxMetzger, line 318: If all you want is 8-bit grayscale for PNG, JPEG, and TIFF, why not use

```python
_img = imageio.v3.imread(filename, plugin="pillow", mode="L")
```

Pillow's TIFF support is much less extensive than tifffile's TIFF support, but assuming that you have simple TIFFs pillow should be able to handle this.

(the diff here uses `TIFF-PIL`, which suggests that this is indeed sufficient.)

hermancollin: Essentially what we are trying to do is accomodate both 8-bits and 16-bits for PNG/TIFF files. Using `mode="F"` works for 8bit and 16bit PNG but pillow is not able to read our TIFFs. Hence, we need to switch to tifffile for TIFF's to be loaded properly

```python
 if 'tif' in extension:
   _img = imageio.v3.imread(filename, plugin='tifffile')
else:
   _img = imageio.v3.imread(filename, plugin="pillow", mode="F")
``` 
Now this here _almost_ works. Only pitfall is that we would like to read rgb/rgba images as grayscale images. The simplest fix I can think of at this point is converting the rgb tiff image to grayscale when we see that a tiff image has more than 1 channel.

ivadomed/inference.py
hermancollin: @FirefoxMetzger thank you for this catch. You are right and the depth stack was also causing a bug for some images.

hermancollin:
At last, every combinations of 8-bit/16-bit, PNG/TIFF, gray/rgb/rgba are working as expected. Maybe someone else would like to confirm that it works on their side?
The conversion to grayscale for TIFF files might look a bit weird but that's how pillow does it as well (see https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert) so we know the conversion is handled identically for PNG and TIFF.

https://github.com/ivadomed/ivadomed/blob/a17d8625f79454a38c63e93876dfe3c9f1f63d33/ivadomed/loader/segmentation_pair.py#L321-L322

ivadomed/loader/segmentation_pair.py
FirefoxMetzger: > I get this error trying to read a simple 8bit TIFF

Interesting. Pillow doesn't get the filepointer for the file that you are trying to read. It could be that this isn't TIFF specific, but rather an issue with how ImageIO and pillow interact. What type of ImageResource are you loading? (file, stream, `bytes`, ...)

If you are able to share an offending TIFF I should be able to run it locally and check if it is perhaps a bug in ImageIO v3. It reminds me of a report for a different format I got a while ago; unfortunately I never managed to reproduce the issue, so I could never figure out what the problem was.

> Another problem is that we often have users with 16 bit images [...] Only pitfall is that we would like to read rgb/rgba images as grayscale images.

I see. In this case I would approach the problem differently. Rather than solving the colorspace conversion indirectly during load by constraining the resulting dtype I would separate the loading (decoding and conversion to `np.array`) and processing (colorspace conversion, and/or buffer unification) steps:

```python
import imageio.v3 as iio

props = iio.improps(filename). # new in v3 - standardized metadata
_img = iio.imread(filename)

# TIFF is a "wild" format. A few assumptions greatly simplify the code below:
# 1. the image is interleaved/channel-last (not planar)
# 2. the colorspace is one of: binary, gray, RGB, RGBA (not aliasing ones like YUV or CMYK)
# 3. the image is flat (not a volume or time-series)
# Note: All of these are trivially true for JPEG and PNG due to limitations of these formats.

# make grayscale (treats binary as 1-bit grayscale)
colorspace_idx = 2 + int(props.is_batch)
if _img.ndims <= colorspace_idx:  # binary or gray
    pass  # nothing to do
elif _img.shape[colorspace_idx] == 3:  # RGB
    _img = np.sum(_img * (.299, .587, .114), axis=-1)
else:  # RGBA
    # discards alpha
    _img = np.sum(_img * (.299, .587, .114, 0), axis=-1)

```

This version lets ImageIO decide how to do the loading (via pillow, tifffile, magic, etc.).

hermancollin: > > I get this error trying to read a simple 8bit TIFF
> 
> Interesting. Pillow doesn't get the filepointer for the file that you are trying to read. It could be that this isn't TIFF specific, but rather an issue with how ImageIO and pillow interact. What type of ImageResource are you loading? (file, stream, `bytes`, ...)

It is a file. This is the image I was talking about: https://github.com/ivadomed/data-testing/blob/0f4d5b644397bf4ab2f4b02f1e30065d4aaba61e/data_test_png_tif/sub-tif/micr/sub-tif_sample-8gray_SEM.tif

> In this case I would approach the problem differently. Rather than solving the colorspace conversion indirectly during load by constraining the resulting dtype I would separate the loading (decoding and conversion to `np.array`) and processing (colorspace conversion, and/or buffer unification) steps:
> 
> ```python
> import imageio.v3 as iio
> 
> props = iio.improps(filename). # new in v3 - standardized metadata
> _img = iio.imread(filename)
> 
> # TIFF is a "wild" format. A few assumptions greatly simplify the code below:
> # 1. the image is interleaved/channel-last (not planar)
> # 2. the colorspace is one of: binary, gray, RGB, RGBA (not aliasing ones like YUV or CMYK)
> # 3. the image is flat (not a volume or time-series)
> # Note: All of these are trivially true for JPEG and PNG due to limitations of these formats.
> 
> # make grayscale (treats binary as 1-bit grayscale)
> colorspace_idx = 2 + int(props.is_batch)
> if _img.ndims <= colorspace_idx:  # binary or gray
>     pass  # nothing to do
> elif _img.shape[colorspace_idx] == 3:  # RGB
>     _img = np.sum(_img * (.299, .587, .114), axis=-1)
> else:  # RGBA
>     # discards alpha
>     _img = np.sum(_img * (.299, .587, .114, 0), axis=-1)
> ```
> 
> This way ImageIO can decide how to do the loading (via pillow, tifffile, magic, etc.). You don't have to create different code paths based on the image format nor do you have to worry about the plugin that is used to do the decoding.

I actually love this solution!

hermancollin (changes requested):
I would like someone else to look at the state of this PR. Only thing I'm uncertain of is the use of `imageio.v3.improps()`.

joshuacwnewton:
Hi all! I'd like to help move this PR along, because it would be very helpful if the fixes from this PR were merged to `master`. 

(The reason I ask is because I'd like to perform some additional maintenance work (updating ivadomed's `torch` version), and I'm running into the same `pybids`/`imageio` bugs that you've encountered here. See: #1303/#1304.) 

Is there anything I can help out with here? (Reviewing the PR, updating `pybids`/`imageio` in the `requirements.txt`, etc.)

mathieuboudreau:
> Hi all! I'd like to help move this PR along, because it would be very helpful if the fixes from this PR were merged to `master`.
> 
> (The reason I ask is because I'd like to perform some additional maintenance work (updating ivadomed's `torch` version), and I'm running into the same `pybids`/`imageio` bugs that you've encountered here. See: #1303/#1304.)
> 
> Is there anything I can help out with here? (Reviewing the PR, updating `pybids`/`imageio` in the `requirements.txt`, etc.)

That would be great @joshuacwnewton ! Let me re-read this thread tonight to collect my thoughts, and will respond with what I think is left missing before merging and we can coordinate moving forward after that - is that ok?

joshuacwnewton:
Perfect! :tada:

mathieuboudreau:
@joshuacwnewton after reviewing this thread, I think it may be ready to merge as-is, but requires an admin to force-merge due to the failing windows tests (unrelated to this PR - https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1611848805).

The last comment in this thread was made by me, expressing some concern for (seemingly arbitrary) hardcoded values that were coded to convert RGBA to greyscale, now that we couldn't use `mode=
as_gray"` anymore: https://github.com/ivadomed/ivadomed/pull/1297#discussion_r1297610802


However, I dug a little bit into the old imageio and pillow codebases, and it appears that as_gray used these same hardcoded values (though using an int16 integer set) to do this conversion (https://github.com/python-pillow/Pillow/blob/b723e9e62e4706a85f7e44cb42b3d838dae5e546/src/PIL/ImageColor.py#L152), so all is well here and I'm not satisfied with knowing that there appears to be consistency between how RGBA was converted to grayscale before to now.

So, either you can do re-review if you'd like for your own purposes, or you or I can force-merge this PR now!

kanishk16:
I'm sorry, as this fell through the cracks. 

> So, either you can do re-review if you'd like for your own purposes, or you or I can force-merge this PR now!

While re-reading this thread (not yet complete), I realized some minor changes still need to be incorporated. @mathieuboudreau If you don't mind, I'd suggest going for a re-review after adding the changes as:
a. Upstream developments (PR https://github.com/imageio/imageio/pull/1031) and related issues (https://github.com/imageio/imageio/issues/1032) around our concerns
b. looming uncertainty https://github.com/ivadomed/ivadomed/pull/1297#pullrequestreview-1539970713 with proposed changes

So, how about I take a look once by tomorrow, add the changes, include suggestions from upstream developments, and ping you and @joshuacwnewton as well for reviews?

mathieuboudreau:
>So, how about I take a look once by tomorrow, add the changes, include suggestions from upstream developments, and ping you and @joshuacwnewton as well for reviews?

Works for me!

hermancollin:
Just a reminder that imageio should also be bumped in the `requirements.txt` file. Currently we are restricting it in the `2.19` minor but the most recent version is `2.31`.

kanishk16:
Just a heads-up, it looks like it will take longer than I had expected.

A little update on tracking the origin of issues - `imageio==2.28.0` (along with `pillow~=9.5.0`) exposed v3 via v2 API, leading to stuff breaking. This was first discovered and reported by Mathieu in #1291. I'm digging just to see what difference is introduced in v3 but this time focussing on the grayscale conversion and what are the possible ways to migrate to v3 without many incompatible changes.

hermancollin:
> A little update on tracking the origin of issues - `imageio==2.28.0` (along with `pillow~=9.5.0`) exposed v3 via v2 API, leading to stuff breaking. This was first discovered and reported by Mathieu in #1291. I'm digging just to see what difference is introduced in v3 but this time focussing on the grayscale conversion and what are the possible ways to migrate to v3 without many incompatible changes.

IIUC what you mean, we already adressed this by applying the grayscale conversion ourself. This is one of my comment from ealier:

> might look a bit weird but that's how pillow does it as well (see https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert) so we know the conversion is handled identically for PNG and TIFF.
> https://github.com/ivadomed/ivadomed/blob/a17d8625f79454a38c63e93876dfe3c9f1f63d33/ivadomed/loader/segmentation_pair.py#L321-L322

We even had an imageio maintainer chip-in directly in this PR so I think we're good on this side. More info on this issue: https://github.com/imageio/imageio/issues/1017
 @mathieuboudreau also confirmed that this colorspace conversion constant was the same in the old imageio/PIL backend.

FirefoxMetzger:
FYI @mathieuboudreau the "random" numbers used in RGB to grayscale conversion go back to [Recommendation BT.601-7](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.601-7-201103-I!!PDF-E.pdf), which is a document on standardizing colors for television and was first published in the 80s.

In the linked document you will find said formula in Annex 2 (page 15, equation 1). The annex describes how to convert sRGB into YCrCb, and the equation everybody uses is this conversion's Y (luminance) component. The numbers aren't uniform because the human eye has different sensitivity to each of the primary colors and choosing them this way is "true to human perception".

From there we get RGBA to grayscale by effectively ignoring the alpha component. There is only one image, so no blending is happening, and subsequently alpha doesn't change the result.

mathieuboudreau:
> FYI @mathieuboudreau the "random" numbers used in RGB to grayscale conversion go back to [Recommendation BT.601-7](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.601-7-201103-I!!PDF-E.pdf), which is a document on standardizing colors for television and was first published in the 80s.
> 
> In the linked document you will find said formula in Annex 2 (page 15, equation 1). The annex describes how to convert sRGB into YCrCb, and the equation everybody uses is this conversion's Y (luminance) component. The numbers aren't uniform because the human eye has different sensitivity to each of the primary colors and choosing them this way is "true to human perception".
> 
> From there we get RGBA to grayscale by effectively ignoring the alpha component. There is only one image, so no blending is happening, and subsequently alpha doesn't change the result.

Wow, thank you so much for clarifying where these came from @FirefoxMetzger! My background is in physics and not CS, so this is not something I would have encountered previously :)

kanishk16:
Thank you @FirefoxMetzger, for the insightful discussion, reviews, and helpful suggestions. BTW, this bug helped me explore imageio extensively and v3 simplifies a lot of the internal stuff making it easier to dive into the codebase.

joshuacwnewton:
Sorry for the delay on this PR! I've been pretty swamped with preparing for SCT's newest release (v6.1) as well as the upcoming SCT Course on November 20th. 

But, once everything is resolved on my end, I'll be freed up to help merge this PR (and subsequently create a new release of ivadomed that contains an updated version of torch (https://github.com/ivadomed/ivadomed/pull/1304).

joshuacwnewton:
Hi folks! I'm finally back to help get this PR merged. :tada:

----

In reviewing the most recent updates to this PR (https://github.com/ivadomed/ivadomed/pull/1297/commits/1424f615394952c75b0d4be599637e69e634f437, Oct 2, 2023), I noticed new test failures stemming from an entirely different PyBIDS-related bug: 

```python
            # Filter dataframe to keep subjects files with available derivatives only
            if has_deriv:
                self.df = self.df[self.df['filename'].str.contains('|'.join(has_deriv))
                                  | self.df['filename'].str.contains('|'.join(deriv))]
            else:
                # Raise error and exit if no derivatives are found for any subject files
>               raise RuntimeError("Derivatives not found.")
E               RuntimeError: Derivatives not found.
```

This seems to be due to a newer version of PyBIDS, as pinning PyBIDS to an older version fixes this error. 

We could spend time debugging this new issue and update `ivadomed` such that it is compatible with newer PyBIDS versions. But, given that we were already somewhat open to the idea of pinning PyBIDS due to the previous Windows-related bug (as previously discussed in (https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1622292897, https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1622302833, etc.), I'd like to propose pinning PyBIDS in the short term. 

If we truly need to unpin PyBIDS in the future (e.g. for a dependent project), we can always revisit how we want to update the ivadomed-specific code in the future. But, I think pinning is a good short-term patch?

----

I've also added some new changes to the `.readthedocs.yml` file to ensure it complies with RTD's new config file requirements, and updated the title of this PR accordingly.

requirements.txt
```diff
@@ -7,13 +7,13 @@
 pyparsing<3,>=2.0.2
 csv-diff>=1.0
 loguru~=0.5
-imageio~=2.19
+imageio>=2.31.4
```
joshuacwnewton, line 10: It seems like we have one more error here:

```
E       AttributeError: module 'PIL.Image' has no attribute 'ANTIALIAS'
```

This is a bit strange, because I tested the changes from this branch in https://github.com/ivadomed/ivadomed/pull/1304, but did not encounter this error.

Nonetheless, I see that `Pillow<10.0.0` was pinned previously in this PR to fix this issue (https://github.com/ivadomed/ivadomed/pull/1297/commits/5672e73b11c9212790ebfd99ebae753cfe7e0056).

That said, I think we can fix this by applying the suggestion from https://stackoverflow.com/a/76616129 instead.

joshuacwnewton (approved):
This has my approval, but it seems wise to have one other ivadomed member review my proposed changes re: Pillow/PyBIDS. :)

joshuacwnewton:
Thank you for adding the additional context! I'm glad we have an issue open for further debugging later on if need be. :)

PR #1303, opened 2023-09-27
Test out `torch==2.0.0` in `requirements.txt` for compatibility with MONAI/nnUNet

joshuacwnewton:
This PR is just a test PR to trigger the CI when `torch==2.0.0` is specified.

The previous upper limit (`1.11.0`) was set in https://github.com/ivadomed/ivadomed/pull/1137 (https://github.com/ivadomed/ivadomed/commit/24fe492e0d296f30c19c2beeba2a47505281d4d0). From the context, it seems like this upper bound was chosen purely because of Python 3.10 support.

I'm not sure if versions higher than `1.11.0` have been tested yet, so I will try that out in this PR.

Relevant for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4182.


requirements.txt
```diff
@@ -21,6 +21,6 @@ tensorboard>=1.15.0
 tqdm>=4.30
 scipy
 torchio>=0.18.68
-torch>=1.8.1,<=1.11.0
+torch==2.0.0
```
joshuacwnewton, line 24: Some initial observations:

- Python 3.7 support stops at `torch==1.13.1`. As far as SCT goes, this is okay, but as far as ivadomed maintenance goes, updating to `torch==2.0.0` means removing support for Python 3.7.
- `torchvision` and `torchio` versions are tied to `torch` versions. So, we'll also have to bump the upper bound of those packages, too.

ivadomed/loader/utils.py
```diff
@@ -13,7 +13,6 @@
 import torch
 from loguru import logger
 from sklearn.model_selection import train_test_split
-from torch._six import string_classes
```
joshuacwnewton, line 16: Our first bit of version incompatibility: We were accessing `torch._six`, which is a now-removed internal API.

The fix for this is simple enough, though -- we can follow `torchvision`'s example here: https://github.com/pytorch/vision/pull/7243

ivadomed/loader/utils.py
joshuacwnewton, line 1: After fixing the `torch._six` issues, I'm able to run the tests, and many of them pass (:tada:)! However, I'm running into quite a lot of errors, all with the same message:

```
   File "/Users/runner/work/ivadomed/ivadomed/ivadomed/loader/bids_dataframe.py", line 183, in create_bids_dataframe
    raise RuntimeError("Derivatives not found.")
```

This seems like it's related to the input datasets, rather than the upgrade to `torch`. To test this, if I retrigger the CI on `master`, I should in theory get the same test failures.

joshuacwnewton: Yep! The issues are independent of `torch`. I'll just double-check that the `torch._six` changes aren't responsible, then I'll dig deeper.

joshuacwnewton: Yep, the `torch._six` changes aren't responsible either. So, I'll re-enable the changes then debug the derivatives issues.

joshuacwnewton: So, the issue here is that ivadomed's dataframe loader isn't finding the derivatives in the `tmp/data_testing` folder. There _is_ a derivatives folder, though! 

Without spending too much time digging into the mechanics of `pybids`, I'm going to assume there was an upstream update in the past 2 months that changed the behavior for how datasets are loaded, given that no changes have been made to `master` in 7 months.

joshuacwnewton: Ah: https://github.com/ivadomed/ivadomed/pull/1297#issuecomment-1622272746

> I think I resolved the bug - essentially pybids changed how they parse regex over the winter, and at some point it started breaking for Windows. Restricting pybids to <15.6.0 resolved the failing test in the draft PR I setup: https://github.com/ivadomed/ivadomed/pull/1299, see: https://github.com/ivadomed/ivadomed/actions/runs/5467608625/jobs/9954141022

It doesn't seem like it's the same bug? But it at least points to an upper version of `pybids` that worked for ivadomed. So, I'll adopt this upper limit and see if it fixes the issue.

joshuacwnewton: Alrighty! This resolved many of the errors, but not all. I'll mark this a resolved.

ivadomed/loader/utils.py
joshuacwnewton: Resolved!

joshuacwnewton:
I'm now encountering errors directly related to `imageio==v2`. Given that there is already an open PR for the maintenance task of migrating `imageio` to v3 (https://github.com/ivadomed/ivadomed/pull/1297), I'm going to close this PR, then remake it using https://github.com/ivadomed/ivadomed/pull/1297 as a base branch.

I'll also try to help https://github.com/ivadomed/ivadomed/pull/1297 get merged, too. :)

PR #1304, opened 2023-09-27
Test out newer versions of PyTorch (`torch>=2.0.0`) for compatibility with downstream projects (SCT, ADS)

joshuacwnewton:
Because ivadomed's inference functionality is incorporated downstream in many of NeuroPoly's projects (e.g. SCT, ADS), it must "play nicely" with other packages installed into those downstream projects. 

One issue with this is that ivadomed currently pins a (relatively old) version of PyTorch, while many other deep learning packages use a newer version of PyTorch. Because of ivadomed's restriction, it often can't be installed into the same environments as other deep learning packages.

So, this PR tests out newer versions of torch (`>=2.0.0`), for the purposes of compatibility with libraries such as nnUnet, MONAI, etc.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,6 +26,6 @@ tensorboard>=1.15.0
 tqdm>=4.30
 scipy
 torchio>=0.18.68
-torch==2.0.0
+torch>=1.8.1
 torchvision>=0.9.1
 wandb>=0.12.11
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,5 +26,5 @@ tqdm>=4.30
 scipy
 torchio>=0.18.68
 torch==2.0.0
-torchvision>=0.9.1,<=0.12.0
+torchvision>=0.9.1
 wandb>=0.12.11
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ joblib~=1.0
 matplotlib>=3.3.0
 nibabel~=3.2
 onnx
-onnxruntime~=1.7
+onnxruntime>=1.7.0,<1.16.0
 pandas>=1.1,<1.5.0
 # We pin Pillow here because we're currently using an older version of `torch`, which in turn uses an
 # older version of `tensorboard`, which in turn uses a deprecated feature of Pillow.
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,10 @@ joblib~=1.0
 matplotlib>=3.3.0
 nibabel~=3.2
 onnx
-onnxruntime>=1.7.0,<1.16.0
+# 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
+# So, pin to >=1.7.0 to avoid having to ask users to install libomp.
+# Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
+onnxruntime>=1.7.0,!=1.16.0
 pandas>=1.1,<1.5.0
 pybids>=0.14.0,<0.15.6
 scikit-learn>=0.20.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ imageio>=2.31.4
 joblib~=1.0
 matplotlib>=3.3.0
 nibabel~=3.2
+onnx
 onnxruntime~=1.7
 pandas>=1.1,<1.5.0
 # We pin Pillow here because we're currently using an older version of `torch`, which in turn uses an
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -11,16 +11,18 @@ imageio~=2.19
 joblib~=1.0
 matplotlib>=3.3.0
 nibabel~=3.2
+onnx
 onnxruntime~=1.7
```
joshuacwnewton, line 15: Alright, building this PR off of #1297 has taken care of another big chunk of failing tests. There are only 2 left! 

The error I'm seeing is:

```python
    def _add_onnxscript_fn(
        model_bytes: bytes,
        custom_opsets: Mapping[str, int],
    ) -> bytes:
        """Insert model-included custom onnx-script function into ModelProto"""
        # TODO(titaiwang): remove this when onnx becomes dependency
        try:
            import onnx
        except ImportError as e:
>           raise errors.OnnxExporterError("Module onnx is not installed!") from e
E           torch.onnx.errors.OnnxExporterError: Module onnx is not installed!
```

`onnxruntime` is installed, but `onnx` is not. Is this a new requirement, then? Hrm...

joshuacwnewton: I've set the upperbound for `onnxruntime` to address the same issue we encountered over at SCT:

After fixing this issue, the tests now pass! (We're still getting a failing CI because of ReadTheDocs, which is still being built using Python 3.7. But, we can take care of that in #1297/#1298.)

requirements.txt
```diff
@@ -11,7 +11,8 @@ imageio>=2.31.4
 joblib~=1.0
 matplotlib>=3.3.0
 nibabel~=3.2
-onnxruntime~=1.7
+onnx
+onnxruntime>=1.7.0,<1.16.0
```
kanishk16, line 15: A nitpick but I'd prefer not to have an upper bound 🤷 
```suggestion
onnxruntime~=1.7
```

requirements.txt
```diff
@@ -11,7 +11,8 @@ imageio>=2.31.4
...
 pandas>=1.1,<1.5.0
 # We pin Pillow here because we're currently using an older version of `torch`, which in turn uses an
 # older version of `tensorboard`, which in turn uses a deprecated feature of Pillow.
```
kanishk16, line 18: 👀

kanishk16 (approved):
Although I did a sanity check on my end using a GPU, would recommend running the test suite once on the lab's GPU. Just going to pre-approve 🚀

requirements.txt
joshuacwnewton: Thank you so much for pointing this out!! 

I had set this upper bound because, at the time, `onnxruntime==1.16.0` contained a bug that was encountered on SCT's end: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225

This bug has since been resolved, so it's sufficient to just set `onnxruntime>=1.7.0,!=1.16.0`

joshuacwnewton: Just out of curiosity, why has the `~=` syntax been proposed? 

From what I understand, this syntax will limit `onnxruntime` to only e.g. `1.7.0`, `1.7.1`, `1.7.2`, etc. which limits compatibility with other packages.

mathieuboudreau:
Testing on ADS now

requirements.txt
joshuacwnewton: Oh!! Thank you so much. It seems I've misunderstood what `~=` refers to.

In other words, since you've suggested `~=1.7`, this will include `1.8`, `1.9`, etc. but _not_ `2.*`, `3.*`, etc.

And, the behavior I describe above would only be true if `~=1.7.0` was specified instead.

In that case, I fully support your suggestion over `>=`. :)

mathieuboudreau:
It's going to take me a bit longer than expected to test, because since imageio was updated from version 2->3 in #1297, I'm going to have to update ADS to make it compatible with this new imageio as well (basically converting all our as_gray alls) before I can attempt to run the tests for this branch

mathieuboudreau:
@joshuacwnewton everything looks like it's working on the ADS side - I tested it out in this PR by pointing the installation to the latest commit: https://github.com/axondeepseg/axondeepseg/pull/739/files#diff-9efd195f4e9bfb79ccd456a1d8370fafcc4bcb0b00ea3799222667d2ae818533R29 

Everything looked good using our napari GUI as well.

So it seems it's a green light from the ADS side! Thanks!

joshuacwnewton:
Thank you so much for your thorough testing (with imageio as well!) :hearts:

PR #1194, opened 2022-09-24
chore: add an upper bound version specifier for pandas to prevent breaking the tests

kanishk16:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [ivadomed's internal developer documentation](https://github.com/ivadomed/ivadomed/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/ivadomed/ivadomed/wiki/tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/ivadomed/ivadomed/wiki/documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
The most recent minor release of pandas, v1.5.0 broke our tests. Since the release 1.5.0 supports python 3.8 or higher, the python 3.7 tests passed but the rest others were broken.

## Linked issues
Discovered in #1193 and related to #1191

PR #1308, opened 2024-01-09
Test upgrades for pinned dependencies to improve downstream compatibility

joshuacwnewton:
## Description

This PR stems from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4319. In that issue, we discovered that ivadomed pins the `3.x` series of `nibabel` releases, while the current major version of the `nibabel` package is `5.x`. Disallowing newer versions of `nibabel` caused an incompatibility with the newest version of another package that SCT relies on (`dipy`). 

As a quick fix, SCT has pinned an older version of `dipy`. But, ideally, we would upgrade both `dipy` _and_ `nibabel`, which requires changes to `ivadomed`'s pinned dependencies.

I'm going to start by testing `nibabel==5.x`, and then move on to other dependencies that currently have some sort of upper bound, just to see the limits of what `ivadomed` can support.

## Linked issues

Partially addresses https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4319.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,8 @@ onnx
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
 # Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
 onnxruntime>=1.7.0,!=1.16.0
-pandas>=1.1
+# `pandas==2.0` removed the `DataFrame.append` call, which we currently use in approx. 13 places (per PyCharm search)
+pandas>=1.1,<2.0
 pybids>=0.14.0,<0.15.6
 scikit-learn>=0.20.3
 scikit-image~=0.17
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ onnx
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
 # Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
 onnxruntime>=1.7.0,!=1.16.0
-pandas>=1.1,<1.5.0
+pandas>=1.1
 pybids>=0.14.0,<0.15.6
 scikit-learn>=0.20.3
 scikit-image~=0.17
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,22 +1,15 @@
...
 matplotlib>=3.3.0
-nibabel~=3.2
+nibabel~=5.2
 onnx
 # 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
 # Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
 onnxruntime>=1.7.0,!=1.16.0
-pandas>=1.1,<1.5.0
+pandas>=1.1
```
joshuacwnewton, line 12: It looks like removing the `pandas<1.5.0` pin caused 2 tests to fail due to the usage of a deprecated call to `Dataframe.append` (more info [here](https://stackoverflow.com/a/75956237)).

Notably, `append` was removed from pandas in `v2.0`, which is a different reasoning than the previous `v1.5.0` pin (see #1194). So, at the very least, we can make the upper bound a little more lax. (Though, I imagine it would be very easy to update the `.append` calls, too.)

joshuacwnewton: - PyCharm -> Right click `DataFrame.append` method -> Go to... -> Implementation
- `pandas.core.frame` -> Right click `DataFrame.append` method -> Find Usages
- Click settings cogwheel -> Scope -> Project Files 

image

As there are ~13 results here, I think it might be better to tackle this in its own PR, and leave the upper bound as `2.0` for now.

joshuacwnewton:
I'd like to try and include this PR in the next ivadomed patch release if possible, alongside #1313. :)

joshuacwnewton:
@kanishk16 I see you've updated this PR! Does this mean you guys have run into issues with the `dipy==1.9.0` release today, too? :P 

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4397

mathieuboudreau:
>Would you be able to review, @mathieuboudreau? (Thank you in advance! ❤️)

Ah shoot I missed this, sorry about that @joshuacwnewton ! Should I take a crack at it now or wait for the dipy situation to resolve itself shortly (hopefully)?

joshuacwnewton:
Ah, sorry! I wasn't clear about the dipy situation -- the dipy stuff is actually dependent on this PR to be fixed. 

So, taking a crack at this PR now would be most appreciated! :)

mathieuboudreau (approved):
LGTM! I tested locally IVADOMED, and through github actions ADS using these changes: https://github.com/axondeepseg/axondeepseg/actions/runs/8238093554/job/22528232069

I did encounter an issue with ADS, but it was resolved after we updated our own nibabel dependency. So, we'll just have to updated it on our end the next time we bump the ivadomed version/commit that we use!

Issue #1312, opened 2024-01-29
convert_file_to_nifti bug for ADS SEM test data


mathieuboudreau:
### Discussed in https://github.com/ivadomed/ivadomed/discussions/1310

<div type='discussions-op-text'>

<sup>Originally posted by **chennyyuee** January 25, 2024</sup>
Dear ivadomed team,

I'm trying to use ivadomed Two-class segmentation to train a model for use in AxonDeepSeg. But when I run ivadomed `-c config_microscopy.json`, using the downloaded tutorial dataset.
I got the following error:
```
  File "C:\Users\ct630\ivadomed_env\lib\site-packages\ivadomed\loader\segmentation_pair.py", line 297, in read_file
    img = self.convert_file_to_nifti(filename, extension, is_gt)
  File "C:\Users\ct630\ivadomed_env\lib\site-packages\ivadomed\loader\segmentation_pair.py", line 360, in convert_file_to_nifti
    _img = np.sum(_img * (.299, .587, .114, 0), axis=-1)
ValueError: operands could not be broadcast together with shapes (744,1154,2) (4,)
```
Is there a way to solve this problem ?

Many thanks!
chenyue
</div>

This user's error is reproducible on my end.

It relates to a part of code that was added (possibly by @kanishk16?) in PR #1297,

https://github.com/ivadomed/ivadomed/blob/4c95d013680ccc3cc1d96a88422e023e6027bcb1/ivadomed/loader/segmentation_pair.py#L352-L360

I ran the debugger using the ADS SEM training data, and found the following for the case where there is an error:

```-> if _img.ndim <= colorspace_idx:  # binary or gray
(Pdb) _img.ndim
2
(Pdb) _img.ndim
2
(Pdb) colorspace_idx
2
(Pdb) props.is_batch
False
(Pdb) _img.shape[colorspace_idx]
*** IndexError: tuple index out of range
(Pdb) _img.shape
(954, 1280)
```

Looking at [imageio's documentation](https://github.com/imageio/imageio/blob/65d79140018bb7c64c0692ea72cb4093e8d632a0/imageio/v2.py#L235-L237), it's not clear what `props.is_batch` is being used for here in relation to setting `colorspace_idx`, @kanishk16 do you recall? The case `is_batch=True` I think would mean it's a list/tuple of images, so what does that have to do with the colorspace?

[...]

PR #1313, opened 2024-01-29
Fix edge case bug where grayscale image has an alpha channel, minor rtd fix

mathieuboudreau:
#### GitHub

#### PR contents

## Linked issues

#1312

### DIFF
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -383,9 +383,9 @@ class SegmentationPair(object):
                     assert _img.shape[colorspace_idx] == 3
                     # Convert RGB to grayscale by multiplying RGB by CCIR 601 constants then summing
                     # See also: https://stackoverflow.com/a/689547
-                    _img[:, :, 0, :] = (_img[:, :, 0, :]*.299)
-                    _img[:, :, 1, :] = (_img[:, :, 1, :]*.587)
-                    _img[:, :, 2, :] = (_img[:, :, 2, :]*.114)
+                    _img[:, :, 0, :] = (_img[:, :, 0, :] * .299)
+                    _img[:, :, 1, :] = (_img[:, :, 1, :] * .587)
+                    _img[:, :, 2, :] = (_img[:, :, 2, :] * .114)
                     _img = np.sum(_img, axis=colorspace_idx)
 
             img = imageio.core.image_as_uint(_img, bitdepth=8)
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -383,9 +383,9 @@ class SegmentationPair(object):
                     assert _img.shape[colorspace_idx] == 3
                     # Convert RGB to grayscale by multiplying RGB by CCIR 601 constants then summing
                     # See also: https://stackoverflow.com/a/689547
-                    _img[:, :, 0, :] *= .299
-                    _img[:, :, 1, :] *= .587
-                    _img[:, :, 2, :] *= .114
+                    _img[:, :, 0, :] = (_img[:, :, 0, :]*.299)
+                    _img[:, :, 1, :] = (_img[:, :, 1, :]*.587)
+                    _img[:, :, 2, :] = (_img[:, :, 2, :]*.114)
                     _img = np.sum(_img, axis=colorspace_idx)
 
             img = imageio.core.image_as_uint(_img, bitdepth=8)
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -353,15 +353,10 @@ class SegmentationPair(object):
             
             # If the image is not a batch of images, then add a new axis. We do this to ensure that
             # we won't have to guess whether or not the last axis contains channel information.
-            print('Start')
-            print(props.is_batch)
-            print(_img.ndim)
             if not props.is_batch:
                 _img=np.expand_dims(_img, axis=-1)
             # After this, image should either be [H, W, N] or [H, W, C, N]
-            print(props.is_batch)
-            print(_img.ndim)
-            print('End')
+
             assert _img.ndim in [3, 4]
 
             # Check if the image batch lacks a channel axis, i.e. is binary/grayscale
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -383,10 +383,7 @@ class SegmentationPair(object):
                     assert _img.shape[colorspace_idx] == 3
                     # Convert RGB to grayscale by multiplying RGB by CCIR 601 constants then summing
                     # See also: https://stackoverflow.com/a/689547
-                    _img[:, :, 0, :] = (_img[:, :, 0, :] * .299)
-                    _img[:, :, 1, :] = (_img[:, :, 1, :] * .587)
-                    _img[:, :, 2, :] = (_img[:, :, 2, :] * .114)
-                    _img = np.sum(_img, axis=colorspace_idx)
+                _img = np.sum(_img * (.299, .587, .114), axis=-1)
 
             img = imageio.core.image_as_uint(_img, bitdepth=8)
 
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -350,9 +350,6 @@ class SegmentationPair(object):
             # Note: All of these are trivially true for JPEG and PNG due to limitations of these formats.
 
             # make grayscale (treats binary as 1-bit grayscale)
-
-
-            # make grayscale (treats binary as 1-bit grayscale)
             colorspace_idx = 2
             if _img.ndim <= colorspace_idx:  # binary or gray
                 pass  # nothing to do
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -356,7 +356,7 @@ class SegmentationPair(object):
             colorspace_idx = 2
             if _img.ndim <= colorspace_idx:  # binary or gray
                 pass  # nothing to do
-            elif _img.shape[colorspace_idx]==2: #gray with alpha channel
+            elif _img.shape[colorspace_idx] == 2:  # gray with alpha channel
                 _img = _img[:, :, 0]
             elif _img.shape[colorspace_idx] == 3:  # RGB
                 _img = np.sum(_img * (.299, .587, .114), axis=-1)
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -353,6 +353,8 @@ class SegmentationPair(object):
             colorspace_idx = 2 + int(props.is_batch)
             if _img.ndim <= colorspace_idx:  # binary or gray
                 pass  # nothing to do
+            elif _img.shape[colorspace_idx]==2: #gray with alpha channel
+                _img = _img[:, :, 0]
             elif _img.shape[colorspace_idx] == 3:  # RGB
                 _img = np.sum(_img * (.299, .587, .114), axis=-1)
             else:  # RGBA
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -353,9 +353,15 @@ class SegmentationPair(object):
             
             # If the image is not a batch of images, then add a new axis. We do this to ensure that
             # we won't have to guess whether or not the last axis contains channel information.
+            print('Start')
+            print(props.is_batch)
+            print(_img.ndim)
             if not props.is_batch:
-                np.expand_dims(_img, axis=-1)
+                _img=np.expand_dims(_img, axis=-1)
             # After this, image should either be [H, W, N] or [H, W, C, N]
+            print(props.is_batch)
+            print(_img.ndim)
+            print('End')
             assert _img.ndim in [3, 4]
 
             # Check if the image batch lacks a channel axis, i.e. is binary/grayscale
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -359,7 +359,7 @@ class SegmentationPair(object):
             assert _img.ndim in [3, 4]
 
             # Check if the image batch lacks a channel axis, i.e. is binary/grayscale
-            if _img.ndim <= 3  # ndim == 3 -> [H, W, N]
+            if _img.ndim <= 3:  # ndim == 3 -> [H, W, N]
                 pass  # nothing to do
 
             # Handle images that *do* have channel information
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -354,7 +354,7 @@ class SegmentationPair(object):
             # If the image is not a batch of images, then add a new axis. We do this to ensure that
             # we won't have to guess whether or not the last axis contains channel information.
             if not props.is_batch:
-                _img=np.expand_dims(_img, axis=-1)
+                _img=np.expand_dims(_img, axis = -1)
             # After this, image should either be [H, W, N] or [H, W, C, N]
 
             assert _img.ndim in [3, 4]
--- a/ivadomed/loader/segmentation_pair.py
+++ b/ivadomed/loader/segmentation_pair.py
@@ -350,18 +350,45 @@ class SegmentationPair(object):
             # Note: All of these are trivially true for JPEG and PNG due to limitations of these formats.
 
             # make grayscale (treats binary as 1-bit grayscale)
-            colorspace_idx = 2 + int(props.is_batch)
-            if _img.ndim <= colorspace_idx:  # binary or gray
+            
+            # If the image is not a batch of images, then add a new axis. We do this to ensure that
+            # we won't have to guess whether or not the last axis contains channel information.
+            if not props.is_batch:
+                np.expand_dims(_img, axis=-1)
+            # After this, image should either be [H, W, N] or [H, W, C, N]
+            assert _img.ndim in [3, 4]
+
+            # Check if the image batch lacks a channel axis, i.e. is binary/grayscale
+            if _img.ndim <= 3  # ndim == 3 -> [H, W, N]
                 pass  # nothing to do
-            elif _img.shape[colorspace_idx]==2: #gray with alpha channel
-                _img = _img[:, :, 0]
-            elif _img.shape[colorspace_idx] == 3:  # RGB
-                _img = np.sum(_img * (.299, .587, .114), axis=-1)
-            else:  # RGBA
-                # discards alpha
-                _img = np.sum(_img * (.299, .587, .114, 0), axis=-1)
-            if len(_img.shape) < 3:
-                _img = np.expand_dims(_img, axis=-1)
+
+            # Handle images that *do* have channel information
+            else:
+                # Since we have an image batch ([H, W, C, N]), use axis=2
+                colorspace_idx = 2
+                n_channels = _img.shape[colorspace_idx]
+                # Handle the error case right away, to ensure all remaining cases will be n_channels <= 4
+                if n_channels > 4:
+                    raise ValueError(f"Number of image channels ({n_channels}) is greater than 4 (RGBA).")
+                # Grayscale + Alpha
+                # Or, grayscale without alpha that happens to have an extra C dimension (?)
+                elif n_channels in [1, 2]:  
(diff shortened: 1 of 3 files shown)


ivadomed/loader/segmentation_pair.py
joshuacwnewton, line 1: I think your solution makes sense! 

Separately, though, while working through this problem on my own (prior to this PR being opened), I tried a rewrite from scratch:

```python
# If the image is not a batch of images, then add a new axis. We do this to ensure that
# we won't have to guess whether or not the last axis contains channel information.
if not props.is_batch:
    np.expand_dims(_img, axis=-1)

# Check if the image batch lacks a channel axis, i.e. is binary/grayscale
if _img.ndim <= 3  # ndim == 3 -> [H, W, N]
    pass  # nothing to do

# Handle images that *do* have channel information
else:
    # Since we have an image batch ([H, W, C, N]), use axis=2
    colorspace_idx = 2
    n_channels = _img.shape[colorspace_idx]
    # Handle the error case right away, to ensure all remaining cases will be n_channels <= 4
    if n_channels > 4:
        raise Exception(f"Number of image channels ({n_channels}) is greater than 4 (RGBA).")
    # Grayscale + Alpha
    elif n_channels == 2:  
        _img = _img[:, :, 0]  # Discard the alpha information
    # RGB or RGB+A
    else:
        if n_channels == 4:  
            _img = _img[:, :, :3]  # Discard the alpha information
        # By this point, _img should only have 3 channels, or else something has gone wrong
        assert _img.shape[colorspace_idx] == 3
        # Convert RGB to grayscale by multiplying RGB by CCIR 601 constants then summing
        # See also: https://stackoverflow.com/a/689547
        _img[:, :, 0, :] *= .299
        _img[:, :, 1, :] *= .587
        _img[:, :, 2, :] *= .114
        _img = np.sum(_img, axis=colorspace_idx)

_img = _img.squeeze()  # Remove [1] axes (i.e. single-image batches)              
```

I've tried to explicitly address each case with comments. Also:

- The previous RGB multiplication didn't seem to work for batches of images? For example, take `_img` with shape `[20, 20, 3, 4]` (batch of 4 RGB images):
  ```python
  rgb = np.sum(rgb * (.299, .587, .114), axis=-1)
    File "/opt/pycharm-2023.2.1/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
      coro = func()
    File "<input>", line 1, in <module>
  ValueError: operands could not be broadcast together with shapes (20,20,3,4) (3,) 
  ```
- To make things consistent, I tried to avoid the previous pattern (adding `1` to `colorspace_idx` if `is_batch`) by making sure we're always working with batches.
- I condensed the two RGB operations (RGB, RGBA) into one by explicitly discarding the alpha information in a separate step.

joshuacwnewton (commented):
If I understand correctly, there are the following cases:

1. Grayscale (1-channel) -> `[H, W]`
2. Grayscale + alpha (2-channel) -> `[H, W, 2]`
3. RGB (3-channel) -> `[H, W, 3]`
4. RGB + alpha (4-channel) -> `[H, W, 4]`

(And any of these cases could have an addition batch axis with `N` images.)

The code assumed that we would only have to deal with cases 1, 3, 4:

https://github.com/ivadomed/ivadomed/blob/4c95d013680ccc3cc1d96a88422e023e6027bcb1/ivadomed/loader/segmentation_pair.py#L348

But we neglected to consider case 2 (grayscale + alpha)? (Hence the value error mentioning `(744,1154,2)`, i.e. the last dim denotes a 2-channel image.)

setup.py
```diff
@@ -21,7 +21,7 @@
     'docs': [
         # pin sphinx to match what RTD uses:
         # https://github.com/readthedocs/readthedocs.org/blob/ecac31de54bbb2c100f933e86eb22b0f4389ba84/requirements/pip.txt#L16
-        'sphinx',
+        'sphinx>-5.0',
```
joshuacwnewton, line 24: Re: Docs error, we ran into the exact same issue with SCT, see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4340

In order to get `sphinx==5.x`, we had to loosen some other docs-related dependencies to avoid conflicts. (Leaving it unpinned should be OK unless something else is holding it back?)

mathieuboudreau: I'll try that thanks!

ivadomed/loader/segmentation_pair.py
mathieuboudreau: I think I'll adopt your way, as it's much more clear! The only somewhat annoying part is needing to force it to become a batch if it isn't.,but I thought about it and couldn't find a less messy way around that. Let me push your changes and see what the tests say

mathieuboudreau: >The previous RGB multiplication didn't seem to work for batches of images? For example, take _img with shape [20, 20, 3, 4] (batch of 4 RGB images):
rgb = np.sum(rgb * (.299, .587, .114), axis=-1)
  File "/opt/pycharm-2023.2.1/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
ValueError: operands could not be broadcast together with shapes (20,20,3,4) (3,) 

  
  Yeah I noticed this too and hinted at it in one of my comment of #1312

joshuacwnewton: Ah wait, I guess this causes an issue, doesn't it...

How do we distinguish between a 3D grayscale image with depth=3 (`[H, W, D=3, N]`) from a 2D RGB image (`[H, W, C=3, N]`)?

mathieuboudreau: @joshuacwnewton the error was happening for an image that has ndim = 2.

I think the issue is that

https://github.com/ivadomed/ivadomed/pull/1313/files#diff-e01feaf172663ae40fefceb9e58de1fea9f793d1d7b8fd530677f59ebb964b7cR357


this line (`np.expand_dims(_img, axis=-1)` is not reassigning to `_img`, so we need `_img=np.expand_dims(_img, axis=-1)`. Testing it locally and if it works I'll push it

joshuacwnewton: Oops, my apologies!! Thank you for debugging. <3

joshuacwnewton: Interesting! There are new test failures, but I think these are actually useful failures?

For the test `test_load_dataset_2d_png`, the image `'sub-rat3_ses-01_sample-data9_SEM.png'` gets loaded as a 2D image by my code (as one might expect).

But, the method `self.get_microscopy_pixelsize` returns 3 values, hence the error. This is due to this logic:

https://github.com/ivadomed/ivadomed/blob/4c95d013680ccc3cc1d96a88422e023e6027bcb1/ivadomed/loader/segmentation_pair.py#L433-L437

So, it looks like the existing code assumes that 2D images will be coerced into 3D images? And in fact, when I revert to `master` then check the test, the existing code loads the image as `[H, W, 1]`.

So, I suppose that me adding the `squeeze` ended up changing the existing behavior.

mathieuboudreau: I noticed that too!

Actually, before we had,

```

            if len(_img.shape) < 3:
                _img = np.expand_dims(_img, axis=-1)

```

which you've actually introduced the opposite,


https://github.com/ivadomed/ivadomed/blob/42bca4792dbfb29cbb2026b67404cc6cc2249010/ivadomed/loader/segmentation_pair.py#L391-L392


So maybe reverting the previous lines will work?

mathieuboudreau: Or actually, just removing lines 391+392 alltogether will be sufficient

joshuacwnewton: Ah, wait! I think what I should have done was:

```python
            if not props.is_batch:
                _img = _img.squeeze(axis=-1)  #
```

That way, the image is only squeezed if an axis was added specifically for the purposes of handling unbatched images.

mathieuboudreau: Done - now one last failing test, due to these lines,

```
        _img[:, :, 0, :] *= .299
        _img[:, :, 1, :] *= .587
        _img[:, :, 2, :] *= .114
```

this error:

>                   _img[:, :, 0, :] *= .299
>E                   numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('uint8') with casting rule 'same_kind'

mathieuboudreau:
This PR is mostly ready for review, except for a test (and test image?) that we should add to cover [line 377](https://github.com/ivadomed/ivadomed/blob/13040369b83571a5b727fcebca213f1da87459d5/ivadomed/loader/segmentation_pair.py#L377) which was the original bug source/fix.

setup.py
joshuacwnewton: It looks like unpinning the other dependencies resulted in the following packages getting installed:

```
sphinx-autodoc-typehints-1.25.3 
sphinx-jinja2-compat-0.2.0.post1 
sphinx-jsonschema-1.19.1 
sphinx-prompt-1.8.0 
sphinx-tabs-3.4.5 
sphinx-toolbox-3.5.0 
sphinx_rtd_theme-2.0.0 
sphinxcontrib-jquery-4.1
```

With our previous pins being:

```
'sphinx-tabs==3.2.0',
'sphinx-toolbox==2.15.2',
'sphinx-jsonschema~=1.16',
```

The last one (`'sphinx-jsonschema~=1.16'`) is compatible with `1.19.1`, but the other two were definitely quite out of date. 

I do notice that `sphinx-toolbox` had a major version upgrade, though. So, I would want to make sure that nothing major has changed in the docs before OK'ing this. I'll look over the changelog now. :)

joshuacwnewton: It looks like `sphinx-toolbox` got sphinx 5.x support in [version 3.2.0](https://sphinx-toolbox.readthedocs.io/en/stable/changelog.html#id3), so upgrading major versions was going to be necessary regardless. Likewise, `sphinx-tabs` added support for sphinx 5.x in [version 3.4.0](https://github.com/executablebooks/sphinx-tabs/blob/master/CHANGELOG.md#340---2022-06-26).

There's nothing that jumps out at me in either changelog as anything to worry about?

- `sphinx-tabs`: Tabs on pages such as https://ivado-medical-imaging--1313.org.readthedocs.build/installation.html seem to render OK.
- `sphinx-toolbox`: We only use the `collapse` utility from sphinx-toolbox, and the collapsible text in https://ivado-medical-imaging--1313.org.readthedocs.build/tutorials/cascaded_architecture.html also seems to render OK.

ivadomed/loader/segmentation_pair.py
```diff
@@ -350,17 +350,44 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
...
+                    if n_channels == 4:  
+                        _img = _img[:, :, :3]  # Discard the alpha information
+                    # By this point, _img should only have 3 channels, or else something has gone wrong
+                    assert _img.shape[colorspace_idx] == 3
+                    # Convert RGB to grayscale by multiplying RGB by CCIR 601 constants then summing
+                    # See also: https://stackoverflow.com/a/689547
+                    _img[:, :, 0, :] = (_img[:, :, 0, :] * .299)
+                    _img[:, :, 1, :] = (_img[:, :, 1, :] * .587)
+                    _img[:, :, 2, :] = (_img[:, :, 2, :] * .114)
+                    _img = np.sum(_img, axis=colorspace_idx)
```
kanishk16, line 389:
```suggestion
                _img = np.sum(_img * (.299, .587, .114), axis=-1)
```

kanishk16 (changes requested):
Idk why but it currently feels like a script to me rather than a module. Hence the suggested changes... More or less, the `assert` statements would not be required once we follow a flat hierarchy something along the lines:

```
colorspace_idx = 2

if _img.ndim <= colorspace_idx:
    pass
elif _img.shape[colorspace_idx] == 2:
    # gray + alpha
     ...
elif _img.shape[colorspace_idx] == 3:
    # RGB
     ...

File: ivadomed/loader/segmentation_pair.py [hunk 3]
```diff
@@ -350,17 +350,44 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
...
+
```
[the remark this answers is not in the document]
mathieuboudreau: >At first glance it appears to be too verbose

A counter argument would be that the previous version was not verbose enough, as I spent a lot of time misunderstanding parts of it yesterday.

mathieuboudreau:
> Idk why but it currently feels like a script to me rather than a module. 

@kanishk16 not quite sure what you mean by this, is it because multiple steps aren't combined anymore? Personally, I prefer clarity at the cost of a few more lines, but I would really like this PR merged ASAP to help out our user in #1310 start training their data. 

Are the changes you suggested mostly style preferences, or do you think they'll impact the performance? If it's the first part, can we push this through so the code becomes functional again now, and then open a new issue to discuss these stylistic details for this block of code?

If this isn't  compromise you're willing to agree to, then I'd suggest we revert all the changes and use my single line bug fix for expediency: https://github.com/ivadomed/ivadomed/pull/1313/commits/340295de00a7c7f17ded6217c4640adaf0efcd06

mathieuboudreau:
There's actually a second user currently wanting to do training using these datasets (the other one posted directly to ADS: https://github.com/axondeepseg/axondeepseg/discussions/783), so this PR is fairly high priority to get merged ASAP.

ivadomed/loader/segmentation_pair.py
```diff
@@ -350,17 +350,44 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
...
-            if _img.ndim <= colorspace_idx:  # binary or gray
+            
+            # If the image is not a batch of images, then add a new axis. We do this to ensure that
+            # we won't have to guess whether or not the last axis contains channel information.
+            if not props.is_batch:
+                _img=np.expand_dims(_img, axis = -1)
+            # After this, image should either be [H, W, N] or [H, W, C, N]
+
+            assert _img.ndim in [3, 4]
+
```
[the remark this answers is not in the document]
joshuacwnewton: Ah!! I hadn't realized that there wouldn't be batches of images. Yes, you're right, that one line definitely colored my assumptions about this block of code.

In that case, if we can safely assume that `props.is_batch` will always be False, then we can definitely simplify the code quite a bit.

mathieuboudreau:
@kanishk16 @joshuacwnewton  I just reverted to my initial solution https://github.com/ivadomed/ivadomed/commit/340295de00a7c7f17ded6217c4640adaf0efcd06, because as I was doing the refactoring requested after @kanishk16 's point-by-point suggestions I noticecd that it simply was just converging back to this old code. The only additional change I maddewas the omission of the pointless "is_batch*.

Let me know if this is good for you

ivadomed/loader/segmentation_pair.py
```diff
@@ -350,17 +350,22 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
             # Note: All of these are trivially true for JPEG and PNG due to limitations of these formats.
 
             # make grayscale (treats binary as 1-bit grayscale)
-            colorspace_idx = 2 + int(props.is_batch)
+
+
+            # make grayscale (treats binary as 1-bit grayscale)
```
joshuacwnewton, line 355: Nit: The comment got duplicated as part of the revision

```suggestion
```

ivadomed/loader/segmentation_pair.py
```diff
@@ -350,17 +350,22 @@ def convert_file_to_nifti(self, filename: str, extension: str, is_gt: bool = Fal
...
+            colorspace_idx = 2
             if _img.ndim <= colorspace_idx:  # binary or gray
                 pass  # nothing to do
+            elif _img.shape[colorspace_idx]==2: #gray with alpha channel
```
joshuacwnewton, line 359: Nit: PEP8 styling (spaces around `==` operator, 2 spaces before comment, space after `#` in comment)

```suggestion
            elif _img.shape[colorspace_idx] == 2:  # gray with alpha channel
```

joshuacwnewton (approved):
I'm good with reverting my suggestion! This has the benefit of making the diff of this PR smaller, making it easier to understand. :)

joshuacwnewton:
> More or less, the assert statements would not be required

Side note: I agree that the assertions definitely weren't required.

<details>
<summary> Thoughts on assertions </summary>

----

I like adding assertions as a way of clarifying the internal state of the code to anyone reading it. (Basically, testing the assumptions we're making about internal state. The assertions should be relevant mainly to developers, but generally should never be seen by users.) In other words, the assertions should never fire if our assumptions are correct? But, if our assumptions are incorrect (as they were with grayscale+alpha images), then the code will fail in an unambiguous way to make debugging easier.

As a specific example, over in SCT, we've started using assertions whenever there is a bare `else:` case:

```python
# Anti-pattern: We assume that argument will be either option_1/2, but what if it isn't?
# Code will continue then probably crash at some unintuitive line of code
if argument == "option_1":
   pass
elif argument == "option_2":
   pass

# Slightly better: All cases are handled thanks to 'else:'
# However, the internal state is now ambiguous -- what is the value of 'argument' inside the 'else:'?
if argument == "option_1":
   pass
else:
   pass

# Preferred by SCT: All cases handled *and* assumptions + internal state are both clear
# If assumptions are correct, assertion statement will do nothing
# If assumptions are false, then the code will fail in a clear/unambiguous way
if argument == "option_1":
   pass
else:
   assert argument == "option_2", f"Unexpected value for argument: {argument}"
   pass
```

----

</details>

I think it might just be a personal style thing -- when I write code, much like my writing in general, I tend to be pretty verbose? :sweat_smile:

kanishk16:
Out of all the comments, maintaining a flat hierarchy is significant as it usually improves performance and in the case of training large data, this could become relevant...

kanishk16:
I guess in all this I forgot to ask one important ques, do we want to add the test img to this PR?

mathieuboudreau:
> I guess in all this I forgot to ask one important ques, do we want to add the test img to this PR?

Yup! Just did in that repo, will restart the tests now to see if there is increased coverage

mathieuboudreau:
> We'd have to cut a release for it and add the new release tag here:
> 
> https://github.com/ivadomed/ivadomed/blob/4c95d013680ccc3cc1d96a88422e023e6027bcb1/ivadomed/scripts/download_data.py#L26
> 
> Let me know if you want me to cut a release!

Ah gotcha - right, that's the same workflow we have in ADS. I just did it, let see how the tests go!

mathieuboudreau:
[Line 357](https://github.com/shimming-toolbox/rf-shimming-7t-neurolibre) got hit per the coverage report

Screenshot 2024-01-30 at 12 57 53 PM

Merging - thanks for the help @joshuacwnewton and @kanishk16 !

kanishk16:
Thank you @mathieuboudreau, for all the patience while I stalled the PR & @joshuacwnewton for helping along.

----

I'd learn smthg.

> I like adding assertions as a way of clarifying the internal state of the code to anyone reading it. (Basically, testing the assumptions we're making about internal state. The assertions should be relevant mainly to developers, but generally should never be seen by users.) In other words, the assertions should never fire if our assumptions are correct? But, if our assumptions are incorrect (as they were with grayscale+alpha images), then the code will fail in an unambiguous way to make debugging easier.

I completely agree with the above idea regarding assertions, but ... let's tackle it as per the case:

```
# Anti-pattern: We assume that argument will be either option_1/2, but what if it isn't?
# Code will continue then probably crash at some unintuitive line of code
if argument == "option_1":
   pass
elif argument == "option_2":
   pass
```
On point! ✔️

```
# Slightly better: All cases are handled thanks to 'else:'
# However, the internal state is now ambiguous -- what is the value of 'argument' inside the 'else:'?
if argument == "option_1":
   pass
else:
   pass
```
✔️ 

```
# Preferred by SCT: All cases handled *and* assumptions + internal state are both clear
# If assumptions are correct, assertion statement will do nothing
# If assumptions are false, then the code will fail in a clear/unambiguous way
if argument == "option_1":
   pass
else:
   assert argument == "option_2", f"Unexpected value for argument: {argument}"
   pass
```
This is where I feel assertion isn't the right way to go... WDYT about this:

```
if argument == "option_1":
   pass
elif argument == "option_2"
   pass
else:
  raise ValueError(f"Unexpected value for argument: {argument}")

...
# some in-place operations on argument
...
# now use assert to clarify the internal state before using it for some other operations
```

---
> I think it might just be a personal style thing -- when I write code, much like my writing in general, I tend to be pretty verbose? 😅

Nah, I didn't mean in that sense...  think about it this way... pretty much in ivadomed there were too few comments, and suddenly, there's code along with comments... 

I like your writing style, and I'd like to believe being verbose is your strength that brings your thoughts alive ❤️

PR #1318, opened 2024-10-03
Update dependencies (`pandas`, `numpy`, `onnx`)

### LINKED ISSUE #1317: Update `pandas` and `numpy` to `>=2.0.0` to allow SCT to update as well
I've been tracking upstream progress for `numpy==2.0.0` compatibility, and everything has finally been addressed on the SCT side of things:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4535

The only thing holding SCT back from upgrading is `ivadomed`, which pins a lower version of `pandas` (see https://github.com/ivadomed/ivadomed/pull/1308#discussion_r1446632262).

This issue is meant to fix the syntax issues, update `pandas`, test `numpy==2.0.0`, then release `ivadomed==2.9.11`.

### LINKED ISSUE #1321: `onnx>=1.16.2` produces a DLL error on Windows
Source: 

- https://github.com/onnx/onnx/issues/6267#issuecomment-2264568402

I tested limiting to `<1.16.2` in https://github.com/ivadomed/ivadomed/pull/1319/commits/7365f6a10fed8c6a548113c738e045d728c4c18d and the error went away. So, the fix should be an easy one.

joshuacwnewton:
## Description

- `pandas<2.0`: Previously held back due to deprecated `df.append` calls. I've replaced the calls with `list.append` + `pd.Dataframe(list)` to avoid costly append/concatenation, as per https://stackoverflow.com/a/75956237.
- `numpy>=2.0`: This update is automatic, and comes with unpinning `pandas`. But, it's also the motivation for this PR in the first place, since it allows SCT to upgrade as well:
    - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4535
- `onnx<1.16.2`: This is a Windows specific bug, see #1321.

## Linked issues

Fixes #1317.
Fixes #1321.

### DIFF
--- a/ivadomed/scripts/visualize_and_compare_testing_models.py
+++ b/ivadomed/scripts/visualize_and_compare_testing_models.py
@@ -135,9 +135,7 @@ def visualize_and_compare_models(ofolders, metric="dice_class0", metadata=None):
     if len(ofolders) < 1:
         raise Exception('No folders were selected - Nothing to show')
 
-    columnNames = ["EvaluationModel", metric, 'subject']
-    df = pd.DataFrame([], columns=columnNames)
-
+    np_lst = []
     for folder in ofolders:
         result = pd.read_csv(str(Path(folder, 'results_eval', 'evaluation_3Dmetrics.csv')))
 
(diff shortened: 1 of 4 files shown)


ivadomed/scripts/visualize_and_compare_testing_models.py
```diff
@@ -161,9 +159,12 @@ def visualize_and_compare_models(ofolders, metric="dice_class0", metadata=None):
 
             folders = [Path(folder).resolve().name] * len(scores)
             subject_id = result["image_id"]
-            combined = np.column_stack((folders, scores.astype(np.object, folders), subject_id)).T
-            singleFolderDF = pd.DataFrame(combined, columnNames).T
```
joshuacwnewton, line 165: Note that this double-transpose (`.T.T`) was unnecessary (the two cancelled each other out), so I excluded them from the rewrite here.

mathieuboudreau (approved):
LGTM!

